Issue #31, opened 2026-02-02
Add `aicpu_build_graph` runtime (AICPU builds + AICPU schedules)


uv-xiao:
Ask for advice on the following plan. Thanks in advance.

## Summary

Today, this repo builds the task graph on the **host** (via a host orchestration `.so` executed during `Runtime.initialize()`), then runs an AICPU scheduler with `N` AICPU instances to dispatch tasks to AICore workers.

Requested change: build the task graph on **AICPU** using **1** AICPU instance, then schedule/execute using **3** AICPU instances (total `4`).

This issue proposes a new runtime implementation named `aicpu_build_graph` under `src/runtime/`, plus a copied example, and documents the required design changes and risks.

## Goals

- Add a runtime `src/runtime/aicpu_build_graph/` discoverable by `RuntimeBuilder` (`python/runtime_builder.py:34`).
- Make graph building execute on AICPU via a programmable function.
- Support two build/schedule modes:
  - **Sequential** build→schedule.
  - **Concurrent** build∥schedule.
- Fixed thread split for this feature: `1` builder + `3` scheduler threads (total `4`), temporarily.

## Proposed approach (high level)

1. **New runtime**: `src/runtime/aicpu_build_graph/` with its own `runtime/runtime.h`.
2. **Programmable AICPU builder**: introduce a function like `extern "C" int build_graph_aicpu(Runtime* runtime)` that runs on AICPU and builds tasks/edges.
3. **Host orchestration becomes “prepare + marshal”**:
   - it allocates/copies tensors on host (unchanged),
   - it writes a **generic `uint64_t[]` payload** into `Runtime` (e.g. `orch_args[]`/`orch_argc`) for the AICPU builder to interpret,
   - it does *not* call `add_task()` / `add_successor()`.
4. **Kernel address binding**: host writes `func_id -> function_bin_addr` into a runtime-visible table so AICPU-built tasks can set `Task::function_bin_addr`.

### Challenge A — Kernel addresses for tasks created on AICPU

If tasks are created on AICPU, host cannot pre-fill `task->function_bin_addr` by iterating tasks (because tasks do not exist yet). Without this field, AICore will skip execution (`src/runtime/host_build_graph/aicore/aicore_executor.cpp:37`).

Proposed fix (runtime-local):

- Add `uint64_t kernel_addrs[MAX_FUNC_ID]` to `src/runtime/aicpu_build_graph/runtime/runtime.h`.
[...]

PR #36, opened 2026-02-03
Add `aicpu_build_graph` runtime (AICPU builds graph, AICPU schedules)

uv-xiao:
## Summary

This PR adds a new runtime `aicpu_build_graph` where the task graph is **built on AICPU** (thread 0) and then scheduled/executed by **3 AICPU scheduler threads** (total 4 AICPU threads). The corresponding issue is #31 .

Key idea: the graph-building logic is **example-provided** as a device-side function `build_graph_aicpu(Runtime*)`, compiled into the AICPU binary (no device-side `dlopen`).

## What’s included
- New runtime: `src/runtime/aicpu_build_graph/`
  - Concurrent build∥schedule implementation with a correct-first `graph_mutex_` strategy.
  - AICPU build APIs for examples:
    - `aicpu_runtime_add_task(..., function_bin_addr)`
    - `aicpu_runtime_add_successor_conditional(...)`
    - `aicpu_runtime_publish_task(...)`
  - Mode selection via `Runtime::build_mode`:
    - `PTO_AICPU_BUILD_GRAPH_BUILD_MODE=1` (default concurrent)
    - `PTO_AICPU_BUILD_GRAPH_BUILD_MODE=0` (sequential barrier)
- Example: `examples/aicpu_build_graph_example/`
  - Host orchestration only prepares tensors and writes `runtime->orch_args[]`.
  - AICPU builder in `examples/aicpu_build_graph_example/kernels/aicpu/build_graph_aicpu.cpp` builds/publishes tasks.
  - `RUNTIME_ENV` in `kernel_config.py` sets `PTO_AICPU_BUILD_GRAPH_BUILD_MODE`.
- Runner updates: `examples/scripts/code_runner.py`
  - Exposes `PTO_KERNELS_DIR` for runtime compilation.
  - Applies `RUNTIME_ENV` during both runtime compilation and `Runtime.initialize()`.

## Perhaps DANGEROUS!

I did some minor updates on [src/platform/a2a3/host/device_runner.cpp](https://github.com/ChaoWao/simpler/compare/main...uv-xiao:pto-runtime:uv/aicpu_build_graph?expand=1#diff-21f08da688161ac42c58a054fabcf0f5e11efa2b76fc81852eab0310661a4db0) and  [src/platform/a2a3sim/host/device_runner.cpp](https://github.com/ChaoWao/simpler/compare/main...uv-xiao:pto-runtime:uv/aicpu_build_graph?expand=1#diff-f33b8a83aa5eaa891aebcd18ee216a395ed70ff6058fe40662bb0af1e20490a6). The purpose is to provide a runtime-visible table of func_id -> kernel code address so the builder can set `Task::function_bin_addr`.

## How it differs from `host_build_graph`
- `host_build_graph`: host orchestration builds tasks/edges and host fills `Task::function_bin_addr` before launch.
- `aicpu_build_graph`: orchestration only marshals inputs; AICPU builds tasks/edges and publishes them while schedulers may already be dispatching tasks (concurrent build∥schedule).

## Validation

```bash

# AICPU build graph (default concurrent)
python examples/scripts/run_example.py \
  -k examples/aicpu_build_graph_example/kernels \
  -g examples/aicpu_build_graph_example/golden.py \
  -p a2a3sim \
  -r aicpu_build_graph
```

## Questions

I've added some logging about AICPU's building graph, but I'm still not sure about the functionality correctness since I failed to find and use the swimlane figure generation API. This needs more assistance.  Besides, for on-board testing, I need more help from guys with A2/A3 platform access.

### DIFF
--- a/examples/scripts/code_runner.py
+++ b/examples/scripts/code_runner.py
@@ -35,8 +35,6 @@ Golden.py interface:
 import importlib.util
 import logging
 import os
-import shutil
-import subprocess
 import sys
 import tempfile
 from contextlib import contextmanager
@@ -442,8 +440,7 @@ class CodeRunner:
         platform_include_dir = str(self.project_root / "src" / "platform" / "include")
         include_dirs = [runtime_include_dir, platform_include_dir]
 
-        # Secure, unique output path in /tmp. This prevents other users/processes from
-        # hijacking a predictable filename (e.g., via symlink races).
+        # Secure, unique output path in /tmp.
         fd, out_s = tempfile.mkstemp(prefix="aicpu_orch_", suffix=".so", dir="/tmp")
         os.close(fd)
         out_path = Path(out_s)
@@ -638,13 +635,11 @@ class CodeRunner:
         extra_aicpu_include_dirs = None
 
         try:
-            runtime_env = _kernel_config_runtime_env(self._kernel_config, self.kernels_dir)
-            with _temporary_env(runtime_env):
-                host_binary, aicpu_binary, aicore_binary = builder.build(
-                    self.runtime_name,
-                    extra_aicpu_source_dirs=extra_aicpu_source_dirs,
-                    extra_aicpu_include_dirs=extra_aicpu_include_dirs,
-                )
+            host_binary, aicpu_binary, aicore_binary = builder.build(
+                self.runtime_name,
+                extra_aicpu_source_dirs=extra_aicpu_source_dirs,
+                extra_aicpu_include_dirs=extra_aicpu_include_dirs,
+            )
         except Exception as e:
             raise RuntimeError(
                 f"Failed to build runtime '{self.runtime_name}' for platform '{self.platform}'.\n"
@@ -683,6 +678,7 @@ class CodeRunner:
             )
             logger.debug(f"Compiled orchestration: {len(orch_so_binary)} bytes")
 
+        # Compile AICPU orchestration plugin if needed (for aicpu_build_graph runtime)
         aicpu_orch_plugin_path: Optional[Path] = None
         if self.runtime_name == "aicpu_build_graph":
             if self.aicpu_orchestration is not None:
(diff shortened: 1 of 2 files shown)


uv-xiao:
According to comments from https://github.com/ChaoWao/simpler/issues/31#issuecomment-3844798847, I've make some changes in the last commits.

Now, this PR enables programming the AICPU-side graph builder as a small .so plugin loaded via dlopen+dlsym on AICPU, instead of linking graph-build logic into the runtime. It also makes kernel address binding for AICPU-created tasks work via a **runtime**-populated `Runtime::kernel_addrs[]` table. Thus, it has no influence on `src/platform` anymore.

What changed (since a9406cb):

- AICPU builder as .so
    - Runner compiles an AICPU orchestration plugin .so from example kernel_config.py and injects it via env.
    - Host orchestration uploads the plugin bytes to device and fills Runtime::
      {aicpu_orch_so_dev_addr,aicpu_orch_so_size,aicpu_orch_func_name}.
    - AICPU builder thread materializes bytes to /tmp/*.so, then dlopen(RTLD_NOW|RTLD_LOCAL) + dlsym().
- Kernel address binding for AICPU-built tasks
    - Runtime::kernel_addrs[func_id] is populated in aicpu_build_graph (runtime) host init (after register_kernel()).
    - Platform-side conditional population was removed.

uv-xiao:
The new commits harden the aicpu_build_graph runtime for the a2a3 device path, and clarifies/solidifies how orchestration code is compiled and executed.

1. Split host orchestration vs AICPU orchestration compilation

  - Host orchestration remains a host-compiled .so (dlopen’d by host runtime during initialize()).
  - AICPU orchestration “builder plugin” is compiled as a separate .so for dlopen on AICPU:
      - a2a3sim: host g++
      - a2a3: aarch64-target-linux-gnu-g++ toolchain selection in python/pto_compiler.py

2. Avoid relying on undefined symbol resolution at plugin load time

  - On some a2a3 deployments, dlopen’d plugins may not resolve undefined symbols from the main AICPU runtime binary.
  - To avoid this, the runtime now provides Runtime::aicpu_build_api (function pointer table) initialized on-device, and the example builder uses this table instead of directly calling aicpu_runtime_* symbols.
  - Additionally, the a2a3 plugin build adds -static-libstdc++ -static-libgcc to reduce runtime deps on device.

3. `feat-rt2`-style plugin byte storage

  - Biggest a2a3 risk: the plugin bytes stored at a device_malloc() address may not be CPU-readable by AICPU syscalls like write(fd, so_data, size).
  - We now store the orchestration plugin .so bytes embedded inside Runtime (fixed-size storage) and the AICPU builder reads from that embedded buffer (same pattern as `feat-rt2`).

uv-xiao:
@ChaoWao `ci.sh` can pass locally. Requesting CI trigger.

PR #72, opened 2026-02-11
Handle partial last block in paged attention via TFILLPAD -inf masking

ChaoWao:
When cur_seq is not a multiple of block_size, the last KV cache block is only partially filled. Previously, the softmax kernel operated on garbage scores for the invalid positions. This adds masking in the softmax kernel using TFILLPAD_INPLACE with PadValue::Min (-inf) so that exp(-inf)=0 gives zero attention weight to invalid positions, propagating correctness through PV matmul and online update without needing changes to those kernels.


ChaoWao (commented):
ok

PR #93, opened 2026-02-14
Refactor: unify register access with generic read_reg/write_reg API

ChaoWao:
Replace purpose-specific register functions (read_task_id_by_reg,
set_aicore_busy/idle, enable/disable_aicore_register, etc.) with a
generic read_reg/write_reg interface parameterized by RegId enum.

Platform layer:
- Add RegId enum and reg_offset() mapping in platform_config.h
- Consolidate AICoreStatus enum into platform_config.h
- Implement read_reg/write_reg for a2a3 (MMIO) and a2a3sim (host memory)
- Rewrite inner_kernel.h for both platforms to use the new interface
- Introduce set_platform_regs()/get_platform_regs() to decouple
  runtime from KernelArgs
- Move shared aicpu sources (platform_regs, unified_log_device) into
  common src/aicpu/ directory

Runtime layer:
- Replace all register helper calls with read_reg/write_reg
- Add physical_core_id to Handshake struct for simulation support
- Pre-set COND=BUSY before task dispatch to prevent false completion
  detection from stale IDLE state

Simulation runner:
- Allocate per-core simulated register blocks in DeviceRunner::run()
- Pass physical_core_id through aicore_execute_wrapper for thread-local
  register base setup


src/platform/a2a3/aicore/inner_kernel.h
```diff
@@ -17,4 +20,43 @@
 // dcci (Data Cache Clean and Invalidate) is provided by CANN headers
 // No need to define it here - it's a hardware instruction
 
+// AICoreStatus is defined in aicpu/aicpu_regs.h for AICPU side.
+// Redeclare here for AICore side (avoids pulling in AICPU headers).
+enum class AICoreStatus : uint32_t {
```
ChaoZheng109: AICoreStatus already in src/platform/include/aicpu/aicpu_regs.h is defined

src/platform/a2a3/aicpu/kernel.cpp
```diff
@@ -4,6 +4,18 @@
 #include "common/unified_log.h"
 #include "common/kernel_args.h"
 #include "aicpu/device_log.h"
+#include "aicpu/platform_regs.h"
+
+// Platform register storage (shared within this .so)
+static uint64_t g_platform_regs = 0;
```
ChaoZheng109: g_platform_regs is only set, but not used

PR #249, opened 2026-03-11
Refactor: cluster-based mixed-task dispatch for AICPU executor

jvjhfhg:
- Replace per-worker-type dispatch with cluster-based model where each cluster groups 1 AIC + 2 AIV cores as a single scheduling unit
- Introduce MixedKernels submit API and 5 resource shapes with dedicated ready queues, probing widest shapes first
- Fix completed_tasks_ counting per-core instead of per-task, which could cause early scheduler exit on multi-core mixed tasks
- Add mixed_example test covering all 5 shapes with golden verification

### DIFF
--- a/src/a2a3/runtime/tensormap_and_ringbuffer/aicpu/aicpu_executor.cpp
+++ b/src/a2a3/runtime/tensormap_and_ringbuffer/aicpu/aicpu_executor.cpp
@@ -707,6 +707,23 @@ void AicpuExecutor::reassign_cores_for_all_threads() {
         }
     }
 
+    // Restore a single core's running/idle state into its new thread's tracker
+    auto reassign_core = [&](int32_t worker_id, CoreTypeTracker& type_tracker,
+                             CoreStateTracker& tracker, int32_t thread_idx) {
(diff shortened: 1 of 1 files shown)


src/a2a3/runtime/tensormap_and_ringbuffer/runtime/pto_scheduler.h
```diff
@@ -501,18 +503,27 @@ struct PTO2SchedulerState {
...
+    int32_t get_ready_task(PTO2ResourceShape shape) {
+        return ready_queues[static_cast<int32_t>(shape)].pop();
     }
 
 #if PTO2_SCHED_PROFILING
-    template<CoreType CT>
-    int32_t get_ready_task(uint64_t& atomic_count, uint64_t& wait_cycle) {
-        return ready_queues[static_cast<int32_t>(CT)].pop(atomic_count, wait_cycle);
+    int32_t get_ready_task(PTO2ResourceShape shape, uint64_t& atomic_count, uint64_t& wait_cycle) {
+        return ready_queues[static_cast<int32_t>(shape)].pop(atomic_count, wait_cycle);
```
poursoul: shape能换成别的概念表达吗

PR #354, opened 2026-03-24
Perf: Reduce unnecessary atomic operations by adding caching mechanisms to PTO2TaskRing and PTO2HeapRing.

yanghaoran29:
Added a cached_last_alive cache field to PTO2TaskRing and a cached_tail cache field to PTO2HeapRing. When allocating task IDs and heap space, the system first reads the cached fields to determine if allocation is possible. If the judgment is affirmative, allocation proceeds directly, reducing atomic operations for reading last_alive_ptr and tail_ptr.


src/a2a3/runtime/tensormap_and_ringbuffer/runtime/pto_ring_buffer.cpp
zhusy54: 缺少cached_last_alive的更新，类比pto2_heap_ring_init

PR #410, opened 2026-03-30
ci: Add dependabot for requirements

georgebisbas:
Add dependabot for handling future python releases on dependencies

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,17 @@
+version: 2
+updates:
+  - package-ecosystem: "pip"
+    directory: "/requirements"
+    schedule:
+      interval: "daily"
+    open-pull-requests-limit: 10
+    labels:
+      - "dependencies"
+
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "daily"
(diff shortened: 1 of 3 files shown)


ChaoWao (changes requested):
Thanks for the contribution! A few concerns:

1. **`--index-url` scope change is a bug** — The original CI only applies `--index-url https://download.pytorch.org/whl/cpu` to torch. After this change it applies to all packages. `--index-url` replaces the default PyPI source, so `pytest` may fail to install if not found on the PyTorch CPU index.

2. **`requirements/ci.txt` is overkill** — CI only needs `pytest` and `torch` now. `pytest` is already managed via `pip install .[test]`. A flat `pip install` line in the workflow is simpler.

3. **torch can't go in the requirements file** — torch needs a separate `--index-url` to get the CPU-only build. Mixing it into the same `pip install -r` either forces `--extra-index-url` (risking the GPU version from PyPI) or breaks the install.

4. **Daily Dependabot for pinned packages adds noise** — These packages are not version-sensitive. Manual updates when needed are sufficient.

georgebisbas:
Thanks for the review!

PR #464, opened 2026-04-07
Perf: cache hash and prefetch chain in TensorMap lookup/insert

chenshengxin2026:
## Summary
- Cache the `hash(addr)` result from `lookup()` and reuse it in the subsequent `insert()` call for INOUT tensors, eliminating a redundant 64-bit multiply per tensor
- Add software prefetch of `next_in_bucket` during chain traversal to hide memory latency on chains longer than one entry
- Add `lookup`/`insert`/`link_entry` overloads that accept precomputed hash

## Benchmark Results

Ascend910 (device 11, 100 rounds, 3 runs averaged):

| Example | Baseline (us) | Optimized (us) | Delta |
|---------|:---:|:---:|:---:|
| alternating_matmul_add | 977.9 | 971.4 | -0.7% |
| benchmark_bgemm | 747.6 | 719.2 | **-3.8%** |
| paged_attention_unroll Case1 | 1165.0 | 1158.2 | -0.6% |
| paged_attention_unroll Case2 | 555.2 | 554.0 | -0.2% |
| batch_paged_attention | 3259.2 | 3239.0 | -0.6% |

The bgemm improvement is expected: it has the highest lookup+dep percentage (45.8% of orch time) and uses INOUT tensors extensively.


ChaoWao:
Not useful any more.

PR #519, opened 2026-04-10
Fix: assert/LOG paths failing when pto2_current_runtime() is null on AICPU

yanghaoran29:
Problem: assert_impl and related code in orchestration/common.cpp use LOG_*
macros that dereference pto2_current_runtime()->ops. On the device path,
pto2_submit_mixed_task and other runtime code run in libaicpu_kernel.so, but
only the orchestration plugin (libdevice_orch_*.so) had g_pto2_current_runtime
set via dlsym(pto2_framework_bind_runtime). Each .so carries its own copy of
common.cpp, so the AICPU image's g_ stayed nullptr and assertion reporting
could crash or misbehave.

Fix: In AicpuExecutor, call the link-resolved pto2_framework_bind_runtime(rt)
for this DSO before orch_bind_runtime_, and pto2_framework_bind_runtime(nullptr)
before pto2_runtime_destroy, mirroring the orchestration SO bind. Same change
for a5 tensormap_and_ringbuffer.

### DIFF
--- a/src/a2a3/runtime/tensormap_and_ringbuffer/aicpu/aicpu_executor.cpp
+++ b/src/a2a3/runtime/tensormap_and_ringbuffer/aicpu/aicpu_executor.cpp
@@ -73,6 +73,10 @@ typedef void (*DeviceOrchestrationBindRuntimeFunc)(PTO2Runtime *rt);
 // Config function exported by orchestration .so
 typedef PTO2OrchestrationConfig (*DeviceOrchestrationConfigFunc)(const ChipStorageTaskArgs &orch_args);
 
+// From orchestration/common.cpp linked into this DSO — updates g_pto2_current_runtime here (distinct from
+// pto2_framework_bind_runtime in the dlopen'd libdevice_orch_*.so).
+extern "C" void pto2_framework_bind_runtime(PTO2Runtime *rt);
+
 constexpr int32_t MAX_AICPU_THREADS = PLATFORM_MAX_AICPU_THREADS;
 constexpr int32_t MAX_CORES_PER_THREAD = PLATFORM_MAX_CORES_PER_THREAD;
 
@@ -2112,6 +2116,7 @@ int32_t AicpuExecutor::run(Runtime *runtime) {
 #if PTO2_PROFILING
             orch_cycle_start = get_sys_cnt_aicpu();
 #endif
+            pto2_framework_bind_runtime(rt);
             if (orch_bind_runtime_ != nullptr) {
                 orch_bind_runtime_(rt);
             }
@@ -2353,8 +2358,8 @@ int32_t AicpuExecutor::run(Runtime *runtime) {
         finished_.store(true, std::memory_order_release);
         // Destroy PTO2 runtime and close orchestration SO (moved from orchestrator path)
         if (!runtime->get_orch_built_on_host() && orch_so_handle_ != nullptr) {
-            // Clear the borrowed pointer in the orchestration SO before destroying
-            // rt, so g_pto2_current_runtime never points to freed memory.
+            // Clear g_pto2_current_runtime in this DSO and in the orchestration SO before destroying rt.
+            pto2_framework_bind_runtime(nullptr);
             if (orch_bind_runtime_ != nullptr) {
                 orch_bind_runtime_(nullptr);
             }
--- a/src/a5/runtime/tensormap_and_ringbuffer/aicpu/aicpu_executor.cpp
+++ b/src/a5/runtime/tensormap_and_ringbuffer/aicpu/aicpu_executor.cpp
@@ -73,6 +73,10 @@ typedef void (*DeviceOrchestrationBindRuntimeFunc)(PTO2Runtime *rt);
 // Config function exported by orchestration .so
 typedef PTO2OrchestrationConfig (*DeviceOrchestrationConfigFunc)(const ChipStorageTaskArgs &orch_args);
 
+// From orchestration/common.cpp linked into this DSO — updates g_pto2_current_runtime here (distinct from
+// pto2_framework_bind_runtime in the dlopen'd libdevice_orch_*.so).
+extern "C" void pto2_framework_bind_runtime(PTO2Runtime *rt);
+
 constexpr int32_t MAX_AICPU_THREADS = PLATFORM_MAX_AICPU_THREADS;
 constexpr int32_t MAX_CORES_PER_THREAD = PLATFORM_MAX_CORES_PER_THREAD;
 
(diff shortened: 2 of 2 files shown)


yanghaoran29:
# Binding Relationships and Fix Instructions for `pto2_current_runtime()` and Various `.so` Files

This document explains: why `pto2_current_runtime()` returns `nullptr` in the AICPU image; the responsibilities and binding differences between `libhost_runtime.so`, `libaicpu_kernel.so`, and `libdevice_orch_*.so`; and the fix approach to synchronously bind `pto2_framework_bind_runtime(rt)` within the current DSO via `AicpuExecutor`.

---

## 1. Phenomenon: `always_assert` and `debug_assert` Fail to Run Normally
When executing `always_assert` and `debug_assert`, the expected assert information is not printed, and the program exits with a segmentation fault instead.

## 2. Root Cause Analysis
`always_assert` and `debug_assert` need to call `pto2_current_runtime()->ops->log_error(__FUNCTION__, fmt, ##__VA_ARGS__)`. When this code is executed, `pto2_current_runtime()` is `nullptr`, causing the error.

## 3. Problem Analysis
### 3.1 Introduction to Each `.so` File (Device Orchestration / Simulation Context)

| Shared Library | Function |
|----------------|----------|
| **`libhost_runtime.so`** | Host-side simulation runtime: loads device threads, `dlopen`s AICPU/AICore, etc.; linked with `runtime/`, `orchestration/common.cpp`, etc. together with `BUILD_CONFIG` of `tensormap_and_ringbuffer`, and **contains one copy** of `g_pto2_current_runtime` (usually not concerned when there is no direct coupling with the hot path of device orchestration). |
| **`libaicpu_kernel.so`** | AICPU device image: implements `aicpu_execute`, `AicpuExecutor`, `pto_orchestrator`, `pto_runtime2`, and **`orchestration/common.cpp`**. When functions such as `pto2_submit_mixed_task` and `assert_impl` execute within this `.so`, they read the **in-image** `g_pto2_current_runtime`. |
| **`libdevice_orch_*.so`** | Compilation output of user orchestration (e.g., the `*_orch.cpp` pointed to by `ORCHESTRATION["source"]` in `kernel_config.py`), plus all `.cpp` files under the `orchestration` directory in `build_config` (including **`common.cpp`**). It is `dlopen`ed by `AicpuExecutor` and then executes `aicpu_orchestration_entry`; `pto2_current_runtime()` in **user orchestration code** reads the `g_` variable **within this `.so`**. |
| **`libaicore_kernel.so`, etc.** | AICore simulation kernels for each `func_id`; generally not linked to the current-runtime path of `pto_orchestration_api`, and has no direct relation to the `g_` variable issue discussed in this document. |

### 3.2 What the "Initialization" of `pto2_current_runtime()` Actually Means and Where It Occurs
- **`pto2_current_runtime()`** is defined in **`pto_orchestration_api.h`** as **`static inline`**, with **no** separate `.so` exported symbol; each call expands to **`pto2_framework_current_runtime()`** within the **current compilation unit**.
- The variable that **actually holds the state** is **`static PTO2Runtime *g_pto2_current_runtime`** (initial value: **`nullptr`**) in **`orchestration/common.cpp`**. **Each `.so` that compiles `common.cpp` into its linking unit has its own copy** of the `g_` variable.
- The **only entry to write to the `g_` variable** (for the given `common.cpp` copy) is **`extern "C" void pto2_framework_bind_runtime(PTO2Runtime *rt)`**: `g_pto2_current_runtime = rt`; unbinding is done via **`pto2_framework_bind_runtime(nullptr)`**.
- **On the device orchestration path**:
  - **Orchestration `.so`**: `AicpuExecutor` obtains the function within the orchestration image via **`dlsym(handle, "pto2_framework_bind_runtime")`**, and executes **`orch_bind_runtime_(rt)`** before calling **`aicpu_orchestration_entry`** → only initializes the `g_` variable in **`libdevice_orch_*.so`**.
  - **`libaicpu_kernel.so`** (after fix): In the same thread and before `orch_bind_runtime_(rt)`, directly call **`pto2_framework_bind_runtime(rt)` linked to the current DSO** to initialize the `g_` variable **within the AICPU image**.

**Note**: `PTO2Runtime *rt` itself is allocated and has **`rt->ops`** set in **`pto2_runtime_create_from_sm`**; this is the creation of the **runtime object**, which is a separate operation from assigning the **`g_pto2_current_runtime` pointer** — `pto2_current_runtime()` returns the **`g_` variable**, not a value that "automatically equals the newly created `rt`".

### 3.3 Reason Why `pto2_current_runtime()` Is `nullptr` (Before Fix)
1. **DSO Isolation**: `g_pto2_current_runtime` **has one copy per `.so`** instead of being a single global variable in the process.
2. **Only the orchestration `.so` is bound**: The **`pto2_framework_bind_runtime`** obtained via `dlsym` modifies the `g_` variable in **`libdevice_orch_*.so`**; the copy of the `g_` variable from `common.cpp` in **`libaicpu_kernel.so`** is **never written** by this call.
3. **Reading the wrong copy**: Code such as **`assert_impl`** executing in **`libaicpu_kernel.so`** expands `LOG_*` macros to **`pto2_current_runtime()->ops->...`**, which invokes `pto2_framework_current_runtime()` **within the AICPU image** → reads the **local DSO `g_` variable that remains at the default `nullptr`**, **regardless of whether `rt` has been created by `pto2_runtime_create_from_sm`**.

**Conclusion (Before Fix)**: The issue is not that "no `PTO2Runtime` exists", but that **"`g_pto2_current_runtime` is unbound in the `.so` where the current code is executing`"`. The user orchestration code has been bound in `libdevice_orch_*.so`, and this copy is **completely isolated** from the one in `libaicpu_kernel.so`.

---

## 4. Fix Solution
### 4.1 Objective
Enable paths in **`libaicpu_kernel.so`** that rely on **`pto2_current_runtime()`** (e.g., `LOG_*` macros in **`assert_impl`**) to obtain the **`rt`** consistent with the current orchestration, matching the `s_runtime_ops` within the current DSO pointed to by **`rt->ops`**.

### 4.2 Implementation
In **`AicpuExecutor`**, **before calling `orch_bind_runtime_(rt)` obtained via `dlsym`**, add a call to the symbol resolved at **link time** for the current DSO:
```cpp
pto2_framework_bind_runtime(rt);
```
This symbol comes from the **same `orchestration/common.cpp`** and modifies the internal `g_pto2_current_runtime` of **`libaicpu_kernel.so`**. This binding **coexists** with the binding in **`libdevice_orch_*.so`** and both point to the **same `rt` pointer**.

**Before `pto2_runtime_destroy(rt)`** (symmetric to cleaning up the orchestration `.so` binding), call:
```cpp
pto2_framework_bind_runtime(nullptr);
```
to avoid **dangling `g_` pointer**.

### 4.3 Relationship with "Orchestration `.so` Binding"
- **Orchestration `.so`**: `orch_bind_runtime_(rt)` / `orch_bind_runtime_(nullptr)` — used by **user orchestration code** (running in the `dlopen`ed image) to call `pto2_current_runtime()`.
- **AICPU `.so`**: `pto2_framework_bind_runtime(rt)` / `nullptr` — used by paths such as the linked `common.cpp` and `pto_orchestrator` **within `libaicpu_kernel.so`**.

The two bindings **do not replace each other**; each binds the `g_` variable **within its own DSO**.

---

poursoul (approved):

PR #554, opened 2026-04-14
Fix: Honor ChipCallable orchestration symbols in runtime dlsym

yanghaoran29:
Resolve device orchestration entry/config symbols from ChipCallable metadata with safe defaults and keep config lookup optional, while plumbing config_name through callable/runtime/bindings and Python build paths for backward-compatible behavior.

### DIFF
--- a/ci.py
+++ b/ci.py
@@ -354,6 +354,7 @@ def compile_task(
         func_name=orchestration["function_name"],
         binary=orch_binary,
         children=kernel_binaries,
+        config_name=orchestration.get("config_name", ""),
     )
 
     all_cases = getattr(golden, "ALL_CASES", {"Default": {}})
--- a/python/bindings/task_interface.cpp
+++ b/python/bindings/task_interface.cpp
@@ -448,7 +448,7 @@ NB_MODULE(_task_interface, m) {
         .def_static(
             "build",
             [](std::vector<ArgDirection> signature, std::string func_name, nb::bytes binary,
-               std::vector<std::tuple<int32_t, PyCoreCallable>> children) -> PyChipCallable {
+               std::vector<std::tuple<int32_t, PyCoreCallable>> children, std::string config_name) -> PyChipCallable {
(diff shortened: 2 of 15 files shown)


yanghaoran29:
# Issue #549 Analysis: Runtime Hardcodes Orchestration Symbol Names

## Issue Description
In the `tensormap_and_ringbuffer` runtime path, the AICPU executor performs `dlsym` using hardcoded symbol names:
- `aicpu_orchestration_entry`
- `aicpu_orchestration_config`

However, the function names in the orchestration configuration are written to `ChipCallable` (`func_name` / `config_name`) and passed to the runtime.
This results in an inconsistency of "configurable in settings but unused in behavior": when the exported symbols differ from the hardcoded names, the runtime will fail during the `dlsym` phase.

## Problems with the Previous Code Implementation
1. `aicpu_executor.cpp` for `a2a3/a5` uses fixed string constants for `dlsym` and ignores `ChipCallable` metadata.
2. `function_name` does not take effect in this path, which easily misleads developers and test cases.
3. The `config` symbol lacks configurability and cannot be customized per scenario.
4. The contracts for entry names and config names are inconsistent, posing high risks for future evolution (renaming symbols).

## Modification Solution
1. Extend `ChipCallable` to add the `config_name` field and its getter.
2. Write `func_name/config_name` to `Runtime` during the `init_runtime_impl(...)` phase.
3. Modify the AICPU executor to use dynamic resolution:
   - Entry symbol: `entry_name = runtime->get_device_orch_func_name()`; fall back to `aicpu_orchestration_entry` if empty
   - Config symbol: `config_name = runtime->get_device_orch_config_name()`; fall back to `aicpu_orchestration_config` if empty
4. Maintain optional semantics for the `config` symbol: failure of `dlsym(config_name)` only triggers a warning and continues execution.
5. The Python build entry passes through the optional `config_name` without breaking old configuration files (default empty string).

ChaoWao (approved):

PR #670, opened 2026-04-24
Add: support deferred completion in a2a3/a5 PTO runtime

jvjhfhg:
- Add kernel and orchestration APIs for tasks that complete after their kernel returns
- Track deferred counter waits in task payloads and scheduler wait state, then complete tasks once registered conditions are ready
- Allocate runtime-owned completion ingress state for future async completion producers
- Add sim and onboard demos that exercise comm-triggered deferred completion through ChipBootstrapConfig and ChipContext
- Use the mainline comm/bootstrap path instead of restoring the old branch-local distributed bootstrap surface


jvjhfhg:
Performance hurts, needs optimization

Update: fixed

poursoul (approved):

Issue #686, opened 2026-04-27
Infer deferred completion from kernel-registered wait conditions instead of submit-time flag


PKUZHOU:
## Background

Current PR #670 requires orchestration to explicitly mark a task as deferred completion by calling a special submit API such as:

```cpp
pto2_rt_submit_aiv_task_deferred(...)
```

This sets `complete_in_future` in the task payload. The scheduler then prepares a deferred-completion context only for those tasks.

However, whether a task actually needs deferred completion is often only known inside the AICore task implementation: if the kernel calls an async operation and registers a completion counter, the task should complete in the future; otherwise it should complete normally.

This suggests that deferred completion could be triggered by the core task's actual use of async APIs, rather than by orchestration knowing this ahead of time.

## Current Behavior

Today the flow is roughly:

```text
orchestration
  -> submit task with complete_in_future = true
scheduler dispatch
  -> prepare task_token + deferred_ingress only for marked tasks
kernel
  -> calls pto2_defer_counter(...)
scheduler completion
  -> register deferred wait conditions
  -> wait for counters before final completion
```

This requires orchestration to know whether the kernel will use async completion.

## Proposed Direction

Make deferred completion auto-detected from the per-dispatch deferred ingress buffer.

Instead of only preparing deferred context for `complete_in_future` tasks, the scheduler can provide every dispatched task with:

```text
task_token
deferred_ingress
deferred_completion_capacity
```

Then the kernel decides whether the task is deferred:

```cpp
PTO2AsyncCtx ctx = pto2_async_ctx(args);

// Only async kernels call this.
pto2_defer_counter(ctx, event_counter, 1, engine);
pto2_defer_flush(ctx);
```

On completion, the scheduler checks the ingress buffer:

```text
if deferred_ingress.count > 0 or deferred_ingress.error_code != NONE:
    register wait conditions
    complete task only after normal_done && all counters ready
else:
    complete task normally
```

This removes the need for orchestration to predict whether the task uses async completion.

## Benefits

- Simpler orchestration API: normal submit and async submit can use the same entry point.
- Avoids mismatches where orchestration marks a task deferred but the kernel does not register any condition.
- Avoids the opposite mismatch where a kernel uses async completion but orchestration forgot to call the deferred submit API.
- More natural ownership: the kernel that launches the async operation also registers the wait condition.
- Works with counter-based async mechanisms such as `TPUT_ASYNC` event counters.

## Design Notes

[...]

PR #700, opened 2026-04-28
Update: simplify deferred completion context

jvjhfhg:
- Always pass async context through dispatch payload on a2a3 and a5
- Remove submit-time deferred flag and redundant AICPU cache maintenance
- Update notify demos and fatal tests for the explicit async context API

Fixes issue #686


jvjhfhg:
Reply to gemini bot review:

1. Prior implement was based on incorrect cache-coherence fact, which leads to redundant cache-flushes/cache-invalidations.
2. False-sharing is no longer possible due to the dual-buffers corresponding to dual-launch on aicore tasks.

poursoul (approved):

PR #711, opened 2026-05-06
Add Python distributed L4 to L3 dispatch

PKUZHOU:
## Summary
- add Python-first gRPC/protobuf distributed dispatch package for L4 -> remote L3
- integrate Worker.add_remote_worker() through a local PROCESS mailbox shim without C++/nanobind changes
- add callable catalog, L3 daemon backend process, heartbeat, tensor-pool surface, examples, and docs

## Tests
- python -m pytest tests/ut/py/test_distributed tests/ut/py/test_worker/test_l4_recursive.py -q
- python -m compileall -q python/simpler/distributed tests/ut/py/test_distributed examples/distributed/l4_l3_remote
- git diff --check

## Notes
- current e2e remote dispatch covers scalar TaskArgs and callable execution
- full remote tensor materialization/output write-back remains future work


jvjhfhg:
`examples/distributed/l4_l3_remote/l4_master.py` seems to imply that remote execution can mutate L4-local Python state:

```python
counter = Counter()

def l3_sub(task_args):
    counter.add(task_args.scalar(0))

...
print(f"remote counter={counter.value}")
return 0 if counter.value == 7 else 1
```

But `l3_sub` is pushed to the L3 daemon through the callable catalog, so the captured counter is serialized with the closure. The remote backend/sub-worker mutates its own deserialized/forked copy, not the original counter in the L4 process.

Meanwhile, the current dispatch response only reports success/failure:

```python
inner.run(orch_fn, args, cfg)
return dispatch_pb2.DispatchResp(task_id=req.task_id, error_code=0), inner
```

and `RemoteWorkerProxy.dispatch()` only checks `error_code`; it does not read or materialize `DispatchResp.output_tensors`.

The tests (`test_l4_l3_remote.py`) avoid this by using externally visible state such as a file/shared-memory counter, which verifies that remote execution happened, but it does not demonstrate a real distributed result-return path. For a cross-host example, this feels misleading.

The example should either avoid expecting L4-local closure state to change, or explicitly use/document an external side effect until `DispatchResp.output_tensors` or another result-return mechanism is implemented.

PKUZHOU (commented):
Retracted: this review was posted from the wrong GitHub identity and will be reposted from the intended project-local account with a self-contained summary.

python/simpler/distributed/catalog.py
```diff
@@ -0,0 +1,131 @@
...
+        self._catalog = catalog
+
+    def PullCallable(self, request, context):  # noqa: N802, ANN001
+        try:
+            payload = self._catalog.export_payload(request.callable_id, request.version)
+        except Exception as e:  # noqa: BLE001
+            context.abort(grpc.StatusCode.NOT_FOUND, str(e))
+        return dispatch_pb2.CallablePayload(
+            callable_id=request.callable_id,
+            version=request.version,
```
PKUZHOU: `version=0` is treated as "latest" by `export_payload()`, but this response returns the literal request version. That can return valid payload bytes with `version=0`, and `install_from_payload()` validates `version == hash(payload)`, so a caller installing the response can fail. Please return the resolved payload version, or remove latest-version semantics from this RPC.

examples/distributed/l4_l3_remote/l4_master.py
```diff
@@ -0,0 +1,51 @@
...
+
+    l3_cid = w4.register(l3_orch)
+    for endpoint in endpoints:
+        w4.add_remote_worker(endpoint)
+    w4.init()
+    try:
+        def l4_orch(orch, task_args, config):
+            for value in (2, 5):
+                sub_args = TaskArgs()
+                sub_args.add_scalar(value)
```
PKUZHOU: This example calls `add_scalar()` before `add_tensor()`, but `TaskArgs` rejects adding tensors after scalars. Running the documented example fails with `RuntimeError: TaskArgs: cannot add tensor after scalar`. Please add tensors before scalars, or change the `TaskArgs` contract/tests if interleaving is intended.

python/simpler/distributed/proto/dispatch.proto
```diff
@@ -0,0 +1,118 @@
...
+  string output_prefix = 6;
+}
+
+message DispatchReq {
+  uint64 task_id = 1;
+  uint64 callable_id = 2;
+  uint64 callable_version = 3;
+  bytes config_blob = 4;
+  repeated uint64 scalar_args = 5;
+  repeated ContinuousTensorRef tensor_args = 6;
```
PKUZHOU: `tensor_args` carries `ContinuousTensorRef.data`, an address-like value from the sender process. That is not a valid execution pointer on remote L3. Since `tensor_refs` is the real remote data-plane schema, please clarify whether `tensor_args` remains supported; if not, reject non-empty `tensor_args` in the daemon or remove it from the active dispatch path.

python/simpler/distributed/remote_proxy.py
```diff
@@ -0,0 +1,472 @@
...
+                        ref, region = self._stage_local_output_tensor(tensor, nbytes, shape, dtype, tag_value)
+                    except (TransportBackendError, TransportUnavailable):
+                        if self._tensor_transport != "auto":
+                            raise
+                    else:
+                        refs.append(ref)
+                        local_output_regions.append(region)
+                        continue
+                data = ctypes.string_at(int(tensor.data), nbytes) if nbytes else b""
+                if nbytes <= self._tensor_inline_threshold:
```
PKUZHOU: This makes inline-vs-TensorPool staging an implicit byte-threshold decision. I think the intended L4 programming model should be explicit: the L4 program allocates/registers remote tensor storage, gets a `TensorHandle`/remote ref, and passes that as the tensor argument. Inline bytes can remain as an explicit small-message/test path, but the normal distributed tensor path should not silently switch based only on size.

python/simpler/distributed/remote_proxy.py
```diff
@@ -0,0 +1,472 @@
...
+                    try:
+                        ref, region = self._stage_local_output_tensor(tensor, nbytes, shape, dtype, tag_value)
+                    except (TransportBackendError, TransportUnavailable):
+                        if self._tensor_transport != "auto":
+                            raise
+                    else:
+                        refs.append(ref)
+                        local_output_regions.append(region)
+                        continue
+                data = ctypes.string_at(int(tensor.data), nbytes) if nbytes else b""
```
PKUZHOU: This reads bytes for every tensor, including `OUTPUT`/`OUTPUT_EXISTING` unless the large RXE output path is selected. Old output-buffer contents are not semantic inputs, so small output tensors and fallback output paths send irrelevant bytes to L3. Please separate input staging from output allocation/writeback.

python/simpler/distributed/remote_proxy.py
```diff
@@ -0,0 +1,472 @@
...
+        except Exception:
+            self._free_remote_handles(remote_handles)
+            self._close_local_output_regions(local_output_regions)
+            raise
+        return refs, remote_handles, local_output_regions
+
+    def _should_stage_local_output(self, tag, nbytes: int) -> bool:  # noqa: ANN001
+        return (
+            self._tensor_transport in {"rxe", "auto"}
+            and getattr(tag, "name", "") in {"OUTPUT", "OUTPUT_EXISTING"}
```
PKUZHOU: The local-output RXE fast path excludes `INOUT`, so large `INOUT` is copy-in/copy-out rather than bidirectional RXE. That is acceptable as an MVP if documented, but it should not be described as shared or in-place remote memory.

python/simpler/distributed/l3_daemon.py
```diff
@@ -0,0 +1,399 @@
...
+        if req.tensor_refs:
+            args, keepalive, writebacks = decode_task_args_with_tensor_refs_and_writebacks(
+                req.tensor_refs,
+                req.scalar_args,
+                tensor_pool,
+            )
+        else:
+            args = decode_task_args(req.tensor_args, req.scalar_args)
+        if req.tensor_refs:
+            run_inner = worker_factory()
```
PKUZHOU: Tensor-ref dispatch creates a fresh backend `Worker(level=3)` per request, initializes it, runs it, and closes it. This is likely needed so newly materialized mmap buffers exist before L3 children fork, but it is a major lifecycle/performance difference from a persistent L3 worker. Please document this tradeoff and the plan for persistent worker + child-visible tensor storage.

python/simpler/distributed/l3_daemon.py
```diff
@@ -0,0 +1,399 @@
...
+        )
+        proc.start()
+        child_conn.close()
+        self._backend_conn = parent_conn
+        self._backend_proc = proc
+
+    def _backend_call(self, msg):
+        if self._backend_conn is None:
+            raise RuntimeError("L3 daemon backend is not running")
+        with self._backend_lock:
```
PKUZHOU: Every foreground RPC is serialized through `_backend_lock` and a blocking Pipe send/recv before backend execution. If this process split is required for fork-safety with gRPC threads, please document it and provide overhead numbers; otherwise consider a direct backend RPC/event loop.

python/simpler/distributed/l3_daemon.py
```diff
@@ -0,0 +1,399 @@
...
+            tb = traceback.format_exc()
+            return dispatch_pb2.DispatchResp(
+                task_id=request.task_id,
+                error_code=1,
+                error_msg=f"{type(e).__name__}: {e}",
+                remote_traceback=[tb],
+            )
+
+    def Heartbeat(self, request, context):  # noqa: N802, ANN001
+        return dispatch_pb2.Health(ok=True, message="ok")
```
PKUZHOU: This heartbeat only proves foreground gRPC liveness. It does not check backend process health, worker initialization, runtime/device readiness, TensorPool capacity, or selected transport. Please either rename/document it as liveness only, or add a deeper readiness RPC.

python/simpler/distributed/catalog.py
```diff
@@ -0,0 +1,131 @@
...
+
+    def __init__(self, *, allowed_modules: Optional[Tuple[str, ...]] = None) -> None:
+        self._functions: dict[tuple[int, int], Callable] = {}
+        self._payloads: dict[tuple[int, int], bytes] = {}
+        self._latest: dict[int, int] = {}
+        self._next_id = 0
+        self._allowed_modules = allowed_modules
+
+    def register(self, fn: Callable, callable_id: Optional[int] = None) -> tuple[int, int]:
+        payload = _pickle_impl.dumps(fn)
```
PKUZHOU: Remote callables are serialized with cloudpickle/pickle, which is not equivalent to local fork/COW callable inheritance. Captured mutable state, raw pointers, file descriptors, sockets, locks, device contexts, imports, and side effects expected to be visible at L4 can behave differently. Please document the remote callable contract and the trusted-cluster assumption.

python/simpler/distributed/catalog.py
uv-xiao, line 122: `version=0` is treated as "latest" by `export_payload()`, but this response returns the literal request version. That can return valid payload bytes with `version=0`, and `install_from_payload()` validates `version == hash(payload)`, so a caller installing the response can fail. Please return the resolved payload version, or remove latest-version semantics from this RPC.

examples/distributed/l4_l3_remote/l4_master.py
uv-xiao, line 35: This example calls `add_scalar()` before `add_tensor()`, but `TaskArgs` rejects adding tensors after scalars. Running the documented example fails with `RuntimeError: TaskArgs: cannot add tensor after scalar`. Please add tensors before scalars, or change the `TaskArgs` contract/tests if interleaving is intended.

python/simpler/distributed/proto/dispatch.proto
uv-xiao, line 79: `tensor_args` carries `ContinuousTensorRef.data`, an address-like value from the sender process. That is not a valid execution pointer on remote L3. Since `tensor_refs` is the real remote data-plane schema, please clarify whether `tensor_args` remains supported; if not, reject non-empty `tensor_args` in the daemon or remove it from the active dispatch path.

python/simpler/distributed/remote_proxy.py
uv-xiao, line 199: This makes inline-vs-TensorPool staging an implicit byte-threshold decision. I think the intended L4 programming model should be explicit: the L4 program allocates/registers remote tensor storage, gets a `TensorHandle`/remote ref, and passes that as the tensor argument. Inline bytes can remain as an explicit small-message/test path, but the normal distributed tensor path should not silently switch based only on size.

python/simpler/distributed/remote_proxy.py
uv-xiao, line 198: This reads bytes for every tensor, including `OUTPUT`/`OUTPUT_EXISTING` unless the large RXE output path is selected. Old output-buffer contents are not semantic inputs, so small output tensors and fallback output paths send irrelevant bytes to L3. Please separate input staging from output allocation/writeback.

python/simpler/distributed/remote_proxy.py
uv-xiao, line 229: The local-output RXE fast path excludes `INOUT`, so large `INOUT` is copy-in/copy-out rather than bidirectional RXE. That is acceptable as an MVP if documented, but it should not be described as shared or in-place remote memory.

python/simpler/distributed/l3_daemon.py
uv-xiao, line 329: Tensor-ref dispatch creates a fresh backend `Worker(level=3)` per request, initializes it, runs it, and closes it. This is likely needed so newly materialized mmap buffers exist before L3 children fork, but it is a major lifecycle/performance difference from a persistent L3 worker. Please document this tradeoff and the plan for persistent worker + child-visible tensor storage.

python/simpler/distributed/l3_daemon.py
uv-xiao, line 130: Every foreground RPC is serialized through `_backend_lock` and a blocking Pipe send/recv before backend execution. If this process split is required for fork-safety with gRPC threads, please document it and provide overhead numbers; otherwise consider a direct backend RPC/event loop.

python/simpler/distributed/l3_daemon.py
uv-xiao, line 96: This heartbeat only proves foreground gRPC liveness. It does not check backend process health, worker initialization, runtime/device readiness, TensorPool capacity, or selected transport. Please either rename/document it as liveness only, or add a deeper readiness RPC.

python/simpler/distributed/catalog.py
uv-xiao, line 36: Remote callables are serialized with cloudpickle/pickle, which is not equivalent to local fork/COW callable inheritance. Captured mutable state, raw pointers, file descriptors, sockets, locks, device contexts, imports, and side effects expected to be visible at L4 can behave differently. Please document the remote callable contract and the trusted-cluster assumption.

uv-xiao (commented):
Implementation review summary for the L4-L3 distributed dispatch PR. I focused on behavioral and semantic issues rather than CI/style.

Main concerns:

1. Callable catalog versioning is internally inconsistent: `PullCallable(version=0)` resolves latest payload bytes but returns version 0, which can fail install-time version validation.
2. The documented L4/L3 example constructs `TaskArgs` in an invalid scalar-before-tensor order.
3. The dispatch proto exposes both legacy address-like `tensor_args` and remote data-plane `tensor_refs`; the contract should be narrowed or validated so raw L4 addresses are not accidentally treated as remote execution pointers.
4. Tensor input staging is currently selected by an internal byte threshold. The distributed L4 programming model should prefer an explicit handle/remote-storage path chosen by the program, with inline bytes only as an explicit small-message/test path.
5. `OUTPUT` and `OUTPUT_EXISTING` staging reads and transfers old local buffer bytes even though output-only tensors do not semantically consume prior contents.
6. `INOUT` currently has copy-in/copy-out semantics and is excluded from the RXE local-output fast path, so it should not be described as shared or in-place remote memory.
7. Tensor-ref dispatch creates an ephemeral backend `Worker(level=3)` per request, which differs materially from persistent local L3 worker reuse and needs rationale/benchmarking or a plan for persistent child-visible tensor storage.
8. L3 daemon dispatch goes through serialized blocking foreground-to-backend IPC after the gRPC call; the process split and overhead need to be justified or simplified.
9. Heartbeat checks only foreground gRPC liveness, not backend/runtime/device/TensorPool readiness.
10. Remote callables use cloudpickle/pickle, whose semantics differ from local fork/COW callable inheritance; the remote callable contract and trusted-cluster assumption should be explicit.

examples/distributed/l4_l3_remote/l4_master.py
```diff
@@ -0,0 +1,51 @@
...
+def main() -> int:
+    parser = argparse.ArgumentParser()
+    parser.add_argument("--remotes", default="127.0.0.1:5050")
+    args = parser.parse_args()
+
+    result = ctypes.c_int64(0)
+    endpoints = [item.strip() for item in args.remotes.split(",") if item.strip()]
+
+    def l3_sub(task_args):
+        output = task_args.tensor(1)
```
uv-xiao, line 17: `task_args` contains one tensor in this example, added below at line 36, so this should be `task_args.tensor(0)`. As written, the subworker tries to read tensor index 1 and the example cannot demonstrate the intended remote L3 writeback behavior.

examples/distributed/l4_l3_remote/l4_master.py
```diff
@@ -0,0 +1,51 @@
...
+        current = ctypes.c_int64.from_address(int(output.data))
+        current.value += int(task_args.scalar(0))
+
+    w4 = Worker(level=4, num_sub_workers=0)
+    sub_cid = w4.register(l3_sub)
+
+    def l3_orch(orch, task_args, config):
+        orch.submit_sub(sub_cid, task_args)
```
uv-xiao, line 25: This makes the example hard to interpret. `submit_sub()` targets an L3 SubWorker, not the next-level remote L3 worker itself. The actual path is L4 `submit_next_level(l3_cid)` -> remote L3 runs `l3_orch` -> remote L3 `submit_sub(sub_cid)` -> SubWorker mutates the tensor. If this example is meant to explain L4-to-remote-L3 dispatch, please either do the mutation directly in `l3_orch`, or make the example explicitly about "L4 dispatch to remote L3, then remote L3 dispatch to its SubWorker".

examples/distributed/l4_l3_remote/l4_master.py
```diff
@@ -0,0 +1,51 @@
...
+    result = ctypes.c_int64(0)
+    endpoints = [item.strip() for item in args.remotes.split(",") if item.strip()]
+
+    def l3_sub(task_args):
+        output = task_args.tensor(1)
+        current = ctypes.c_int64.from_address(int(output.data))
+        current.value += int(task_args.scalar(0))
+
+    w4 = Worker(level=4, num_sub_workers=0)
+    sub_cid = w4.register(l3_sub)
```
uv-xiao, line 22: This is semantically confusing: `l3_sub` is not a W4 SubWorker callable because `w4` has `num_sub_workers=0`; it is intended to be shipped through the L4 catalog and later used by the remote L3 daemon as an L3 SubWorker callable. That works only because `Worker.register()` is an untyped global callable-id table. Please make the callable kind explicit, for example `register_sub(...)` vs `register_orch(...)`, or otherwise document and validate that this id is meant for the remote child's SUB namespace.

examples/distributed/l4_l3_remote/l4_master.py
uv-xiao, line 35: This adds a scalar before adding a tensor, but `TaskArgs` requires tensors before scalars. The example should add the output tensor first and then the scalar. This concrete bug also obscures the higher-level callable-semantics issue because the example can fail before reaching the remote dispatch path.

python/simpler/worker.py
```diff
@@ -616,6 +665,8 @@ def register(self, fn: Callable) -> int:
             raise RuntimeError("Worker.register() must be called before init()")
         cid = len(self._callable_registry)
         self._callable_registry[cid] = fn
+        if self._distributed_catalog is not None:
```
uv-xiao, line 668: `register()` stores every Python callable in one `_callable_registry`, but the callable contracts are different depending on how the id is later used. `submit_sub(cid, ...)` expects `fn(task_args)`, while L4 `submit_next_level(cid, ...)` expects a child orchestration callable shaped like `fn(orch, task_args, config)`. The registry should not be untyped here. Please split this into explicit APIs such as `register_sub()` and `register_orch()`, or return typed handles so the wrong callable kind cannot be submitted to the wrong worker type.

uv-xiao (commented):
Second review round: callable semantics and the L4/L3 remote example.

The main issue is that the public API is ABI-uniform but not semantically uniform. `submit_next_level(callable, ...)` accepts a `uint64`, but the semantic meaning changes by level/target: L3-to-L2 expects a chip callable handle, L4-to-L3 expects a Python orchestration callable id, and remote L3 treats the value as a catalog id. `Worker.register(fn)` also stores both SubWorker callables (`fn(task_args)`) and orchestration callables (`fn(orch, task_args, config)`) in one untyped namespace. The current example makes this hard to understand because `w4.register(l3_sub)` registers a callable intended for a remote L3 SubWorker, while `w4.register(l3_orch)` registers the remote L3 orchestration function.

I think the minimum direction should be typed callable registration/handles, for example `register_sub(...)`, `register_orch(...)`, and a distinct chip callable handle. The internal slot can still carry a compact integer, but the public API should prevent passing a subworker callable id where a next-level orchestration callable is expected.

python/simpler/worker.py
```diff
@@ -633,6 +684,24 @@ def add_worker(self, worker: "Worker") -> None:
...
+            raise RuntimeError("Worker.add_remote_worker() must be called before init()")
+        self._ensure_distributed_catalog()
+        self._remote_worker_specs.append((str(endpoint), dict(options)))
+
+    def _ensure_distributed_catalog(self):
+        if self._distributed_catalog is None:
+            from .distributed.catalog import Catalog  # noqa: PLC0415
+
+            self._distributed_catalog = Catalog()
+            for cid, fn in self._callable_registry.items():
```
uv-xiao, line 701: The distributed catalog mirrors every entry from `_callable_registry`, but it does not preserve whether an entry is a SubWorker callable or a next-level orchestration callable. That is why the L4 example can register both `l3_sub` and `l3_orch` on `w4` and rely on later submit paths to decide their meaning. For remote dispatch this should be explicit in the catalog payload or handle type; otherwise a wrong id can cross the network and fail only later as a signature/runtime mismatch.

python/simpler/worker.py
```diff
@@ -544,6 +558,36 @@ def _child_worker_loop(
...
+    while True:
+        state = _mailbox_load_i32(state_addr)
+        if state == _TASK_READY:
+            cid = struct.unpack_from("Q", buf, _OFF_CALLABLE)[0]
+            code = 0
+            msg = ""
+            try:
+                args = _read_args_from_mailbox(buf)
+                cfg = _read_config_from_mailbox(buf)
+                proxy.dispatch(int(cid), args, cfg)
```
uv-xiao, line 578: This forwards the mailbox `callable` field as a remote catalog id. That is the core semantic mismatch: the same `submit_next_level(callable, ...)` slot can mean a chip callable handle for L3-to-L2, but a Python orchestration/catalog id for L4-to-L3. The public API should expose a typed next-level orchestration handle here, even if the mailbox representation remains a `uint64` internally.

ChaoWao:
Closing this PR — the distributed L4→remote-L3 dispatch feature it adds was meanwhile delivered by the merged `remote L3 worker` stack #1006–#1011 (extracted from #866, merged June 2026, after this branch). `main` now ships `remote_l3_worker.py`, `remote_l3_session.py`, `remote_l3_protocol.py`, and exposes `add_remote_worker` / `RemoteL3`.

Most of this PR's surface has a merged equivalent under a different (non-gRPC) design:

- `proto/dispatch.proto` + gRPC `dispatch_pb2_grpc.py` → `remote_l3_protocol.py` binary frame codec over raw TCP sockets.
- `catalog.py` → callable manifest + `CallableKind` / `RegisterCallableCommand` in `remote_l3_session.py`.
- `l3_daemon.py` (backend process + heartbeat) → `remote_l3_worker.py` subprocess session runner + `_health_loop`.
- `tensor_pool.py` → `_RemoteBufferEntry` + `Export/Import/Release` buffer commands (shm-name based, not raw pointers).
- `serialization.py` → `RemoteTensorDesc` / `RemoteTaskArgsWire` wire codec.
- `remote_proxy.py` (`add_remote_worker` shim) → session runtime + C++ endpoint facade (#1010) + scheduler eligibility (#1009).

The merged design also specifically avoids the issues raised in review here: it doesn't ship raw `tensor.data` pointers across the network (shm names + structured buffer descriptors instead), uses `ChipCallable` manifest/digest rather than cloudpickling closures (cloudpickle ≠ fork/COW callable inheritance), and separates dispatcher vs inner callable identity with explicit `CallableKind` / `RemoteRegistryTarget`.

Two further notes:
- This PR adds `grpcio`, `protobuf`, and `cloudpickle` as **core** runtime dependencies (and `grpcio-tools` to test); the merged stack introduces none of them. Per the project's minimal-dependency stance that's a real cost for a path that's already shipped another way.
- The one piece with no merged equivalent is the **RDMA/RXE data plane** (`rxe_verbs_helper.c`, the RXE paths in `transport_backend.py`). It's incomplete (the PR notes remote tensor materialization / output write-back is future work) and built on the gRPC base rather than the merged wire codec. If a real RDMA requirement emerges, the cleaner path is a transport plugin under #1008's transport-neutral codec rather than reviving this parallel gRPC+RDMA stack.

Happy to reopen or discuss if the RDMA data plane is something you want to pursue on top of the merged stack. Thanks for the work.

PR #782, opened 2026-05-15
Add: tiered --enable-l2-swimlane perf_level (0-4) for L2 swimlane

indigo1973:
Convert --enable-l2-swimlane from a boolean to an integer perf_level
(0-4); bare flag stays backward-compatible by mapping to level 4
(full collection). Introduce a typed L2PerfLevel enum
(DISABLED/AICORE_TIMING/AICPU_TIMING/SCHED_PHASES/ORCH_PHASES) so
gate sites read "if (level >= SCHED_PHASES)" instead of magic ints.

Transport mirrors the PMU pattern with two channels: binary on/off
via KernelArgs::enable_profiling_flag bit1, and granular level via
L2PerfDataHeader::l2_perf_level in shared memory. AICPU promotes the
level from the header in l2_perf_aicpu_init and exposes it through
get_l2_perf_level(). Host-side L2PerfCollector caches the level to
gate JSON sections and stamps the JSON "version" field directly from
perf_level (swimlane_converter and tests now accept versions 1-4).

Apply level gates throughout AICPU code paths: skip dispatch/finish
timestamps and fanout copies below AICPU_TIMING, scheduler phase
records below SCHED_PHASES, and orchestrator phase records below
ORCH_PHASES. Update docs (l2-swimlane-profiling.md,
profiling_levels.md, testing.md) and tests. Note: the a5 backend is
not yet level-aware and still treats the flag as boolean.

PR #841, opened 2026-05-21
Add: tiered perf_level (0-4) for --enable-l2-swimlane on a5

indigo1973:
Add: tiered perf_level (0-4) for --enable-l2-swimlane on a5

Port the tiered L2 swimlane perf_level feature from [#782](https://github.com/hw-native-sys/simpler/issues/782) (a2a3-only)
to the a5 platform, so a5 onboard and a5sim now honor the integer
perf_level (0-4) instead of treating --enable-l2-swimlane as a plain
boolean.

Mirror the a2a3 wiring on a5:
- L2PerfDataHeader::l2_perf_level carries the level into shared memory;
  AICPU promotes it in l2_perf_aicpu_init and exposes it via
  get_l2_perf_level().
- Host-side L2PerfCollector caches the level to gate JSON sections and
  stamps the JSON "version" field directly from perf_level.
- Apply level gates throughout AICPU code paths: skip dispatch/finish
  timestamps and fanout copies below AICPU_TIMING, scheduler phase
  records below SCHED_PHASES, and orchestrator phase records below
  ORCH_PHASES.
- Plumb perf_level through DeviceRunner / pto_runtime_c_api on both
  onboard and sim backends.
- Move l2_perf_aicpu_init out of the dispatch one-time-init block into
  SchedulerContext::init() in scheduler_cold_path.cpp, matching a2a3
  so the orchestrator thread reads a promoted level when caching
  rt->orchestrator.l2_perf_level.
- Align l2_perf_aicpu_record_phase to a2a3 byte-for-byte: remove the
  end-of-function wmb() and the 3 dropped-path wmbs (all introduced
  by [#777](https://github.com/hw-native-sys/simpler/issues/777), none present in a2a3), and unify the accounting comment
  + brace style. Measured ~1.1 ms reduction in L4 orch_cost on
  paged_attention_unroll Case1.
- Align l2_perf_aicpu_complete_record with a2a3: add thread_idx
  parameter (routed from both host_build_graph and tensormap_and_ringbuffer
  callers), introduce an AICPU-private s_perf_records_buffers[] cache as
  the records-buffer SoT, rename switch_buffer -> switch_records_buffer
  and rotate after the write so the just-committed record is preserved,
  and surface ring/task_id mismatch as a dedicated LOG_ERROR
  (completion-before-dispatch invariant violation) separate from
  capacity drops. init / flush_buffers maintain s_perf_records_buffers[]
  in lockstep with state->current_buf_ptr so flush deterministically
  halts subsequent commits.

Update docs (l2-swimlane-profiling.md, profiling_levels.md,
testing.md) to drop the "a5 is boolean-only" caveat and document the
unified integer interface across a2a3 and a5.

### DIFF
--- a/docs/dfx/l2-swimlane-profiling.md
+++ b/docs/dfx/l2-swimlane-profiling.md
@@ -43,7 +43,7 @@ available.
 Enable in one line:
 
 ```bash
-python tests/st/<case>/test_<name>.py -p a2a3 -d 0 --enable-l2-swimlane
+python tests/st/<case>/test_<name>.py -p <platform> -d 0 --enable-l2-swimlane
 ```
 
 ## 3. How to Use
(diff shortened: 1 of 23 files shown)


src/a5/platform/src/aicpu/l2_perf_collector_aicpu.cpp
```diff
@@ -566,7 +587,6 @@ void l2_perf_aicpu_record_phase(
     record->extra2 = extra2;
 
     buf->count = idx + 1;
```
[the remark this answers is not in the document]
indigo1973: The original wmb() was placed AFTER `buf->count = idx + 1`, so it could not function as a release fence regardless of weak memory ordering. Beyond that, `buf->count` is not the host-visible publish
point — host only sees this buffer after `enqueue_ready_buffer` updates `header->queue_tails[]` in switch/flush paths, where any release semantics belong.

Empirical check: a2a3 has run with the same code shape (no barrier at this position) since #654 (Apr 2026). I re-ran paged_attention_unroll Case1 on a2a3 with full L4 swimlane output and verified the resulting JSON against the patterns that would surface out-of-order observation across 1024 tasks + 9298 phase
records: loop_iter / submit_idx / start_time strict monotonicity per thread, zero time-window overlap between consecutive phases within the same submit_idx, and all field bounds. Zero anomalies.

Keeping the wmb() here introduced significant runtime overhead. Happy to revisit if there's a specific failure mode I'm missing.

docs/dfx/l2-swimlane-profiling.md
```diff
@@ -82,9 +75,9 @@ python tests/st/<case>/test_<name>.py -p a2a3 -d 0 --enable-l2-swimlane 2
 # pytest — scheduler phases (level 3)
 pytest tests/st/<case> --platform a2a3 -d 0 --enable-l2-swimlane 3
 
-# a5 onboard / a5sim — boolean only (perf_level integer not honored on a5 yet)
-python tests/st/<case>/test_<name>.py -p a5 -d 0 --enable-l2-swimlane
-python tests/st/<case>/test_<name>.py -p a5sim --enable-l2-swimlane
+# a5 onboard / a5sim — same integer perf_level interface as a2a3
```
ChaoZheng109: 文档描述之前是因为行为不一致单独描述了a5，现在这里a2a3已经和a5的行为统一了，描述一个整体就行，不用分开描述又说行为相同

indigo1973: 已经更新文档。同时更新了docs/task-flow.md中struct CallConfig的内容。

ChaoZheng109 (approved):

PR #863, opened 2026-05-27
Refactor: drop fanout from L2PerfRecord hot path

ChaoWao:
## Summary

The L2 swimlane per-task commit on AICPU was copying up to 128*8B = 1 KB of fanout edges plus walking the producer's fanout linked list, **every task**, on the scheduler completion critical path. The fanout edges are already the static DAG and are reconstructed offline by dep_gen replay into `deps.json` — so the device-side hot path was paying GM-bandwidth and cache-miss cost to duplicate information host tooling already has.

Scope is **a2a3 only**; a5 is untouched.

### Device side

- `L2PerfRecord` drops `fanout[128]` / `fanout_count` (~1088 B → 64 B per record).
- `l2_perf_aicpu_complete_record` drops the trailing `fanout` / `fanout_count` parameters; the impl no longer touches them.
- `scheduler_completion` drops the `fanout_arr` build + linked-list walk; `host_build_graph/aicpu_executor` drops the same pattern at all four call sites.

### Host side

- `l2_perf_collector::export_swimlane_json` emits `"fanout": []` and `"fanout_count": 0` per task to keep the JSON schema shape stable, and drops the top-level `"version"` field, which had drifted into a duplicate of `L2PerfLevel` (see in-flight PR #856 for the misaligned guard cleanup on the consumer side).

### Downstream tools

- `swimlane_converter` already preferred `deps.json` over `task["fanout"]`; it now reads the version-free schema and treats empty fanout as the expected steady state.
- `sched_overhead_analysis` no longer gates phase parsing on the dropped `"version"` field — it gates on presence of `aicpu_scheduler_phases`, which is the right key.

### Tests and comments

- dep_gen tests drop the now-vacuous "fanout ⊆ deps" gate and the auto-add of `--enable-l2-swimlane` that only existed to feed that gate.
- `_swimlane_validate` drops the version assertion.
- `profiling_levels.md`, `dep_gen.h`, `dep_gen_replay.h`, `pto_orchestrator` comments updated to reflect `deps.json` as the sole source of truth for fanout.

## Test plan

- [x] `pip install --no-build-isolation -e .` clean on macOS sim
- [x] `tests/st/a2a3/.../dfx/l2_swimlane/test_l2_swimlane.py` passes (`--platform a2a3sim --enable-l2-swimlane`)
- [x] `tests/st/a2a3/.../dfx/l2_swimlane/test_l2_swimlane_mixed.py` passes (`--platform a2a3sim --enable-l2-swimlane`)
- [x] `tests/st/a2a3/.../dfx/dep_gen/test_dep_gen.py` passes (`--platform a2a3sim --enable-l2-swimlane --enable-dep-gen`)
- [x] `tests/st/a2a3/.../dfx/dep_gen/test_dep_gen_chain.py` passes (`--platform a2a3sim --enable-l2-swimlane --enable-dep-gen`)
- [ ] Hardware run on a2a3 to confirm the AICPU-cycle savings translate to wall-clock improvement on a fanin-tail-bound case (e.g. `paged_attention_unroll_manual_scope` Case1)

## Notes for reviewers

- JSON schema shape is preserved (`fanout: []` / `fanout_count: 0` are still emitted per task), so any consumer that just shape-checks keeps working.
- The `"version"` field removal overlaps with PR #856 in direction. Whichever lands first leaves a trivial conflict for the other.
- Phase-record cost (orchestrator 9-per-submit + scheduler per-iter) is intentionally untouched — that's a separate follow-up if measurement after this still shows AICPU pressure at higher perf levels.

### DIFF
--- a/src/a2a3/platform/include/common/l2_perf_profiling.h
+++ b/src/a2a3/platform/include/common/l2_perf_profiling.h
@@ -60,11 +60,6 @@
 #include "common/core_type.h"
 #include "common/platform_config.h"
 
-// Maximum number of successor tasks per L2PerfRecord (matches Task::fanout)
-#ifndef RUNTIME_MAX_FANOUT
-#define RUNTIME_MAX_FANOUT 128
-#endif
-
 // =============================================================================
 // L2 perf_level — granularity ladder for the L2 swimlane profiler.
 //
@@ -82,7 +77,7 @@
 enum class L2PerfLevel : uint32_t {
     DISABLED = 0,       // No collection at all
     AICORE_TIMING = 1,  // AICore per-task start/end timestamps + task record buffer
-    AICPU_TIMING = 2,   // + AICPU dispatch/finish timestamps + fanout dependency list
+    AICPU_TIMING = 2,   // + AICPU dispatch/finish timestamps
     SCHED_PHASES = 3,   // + scheduler main-loop phase records (SCHED_COMPLETE/DISPATCH/IDLE_WAIT)
     ORCH_PHASES = 4,    // + orchestrator phase records
 };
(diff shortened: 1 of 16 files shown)


src/a2a3/platform/include/common/l2_perf_profiling.h
```diff
@@ -82,7 +77,7 @@
 enum class L2PerfLevel : uint32_t {
     DISABLED = 0,       // No collection at all
     AICORE_TIMING = 1,  // AICore per-task start/end timestamps + task record buffer
-    AICPU_TIMING = 2,   // + AICPU dispatch/finish timestamps + fanout dependency list
```
poursoul, line 85: 这里可以干掉这个注释，让人误以为level 3就会有fanout了

PR #872, opened 2026-05-27
Add scheduler stall snapshot before emergency shutdown

### LINKED ISSUE #868: [Feature] Dump all scheduler thread stall state before emergency shutdown
### Summary

Add a shutdown-time scheduler diagnostic snapshot before
`tensormap_and_ringbuffer` calls `emergency_shutdown()` on scheduler timeout.

The snapshot should reuse the existing `log_stall_diagnostics()` machinery to
dump all active scheduler threads, not only the thread that first reaches
`MAX_IDLE_ITERATIONS`.

### Motivation / Use Case

When one scheduler thread times out, the current log only guarantees a final
stall dump for the triggering thread. Other scheduler threads may have printed
older half-time stall diagnostics, but their exact state immediately before
emergency shutdown is missing.

This makes timeout diagnosis ambiguous. For example, in a paged-attention
unroll debug run, `thread1` triggered:

```text
[STALL thread=1 idle_iterations=800000] TIMEOUT_EXIT after_idle_iterations=800000
Emergency shutdown: sending exit signal to all initialized cores
```

The log showed `thread1` cores stuck on `kernel=2` with AIC COND state `ack`,
while AIV cores were idle. However, the shutdown-time state of `thread0` and
`thread2` was not printed; only older `idle_iterations=400000` diagnostics were
available. A final all-thread snapshot would make it clear whether other
threads were still making progress, also stalled, or blocked behind the same
dependency chain.

### Proposed API / Behavior

On the first scheduler timeout path, before calling `emergency_shutdown()`,
print a one-time shutdown snapshot:

```text
[SHUTDOWN_SNAPSHOT trigger_thread=1 reason=scheduler_timeout idle_iterations=800000] dumping all scheduler threads before emergency shutdown
[STALL thread=0 idle_iterations=800000] TASK ...
[STALL thread=0 idle_iterations=800000] SUMMARY ...
[STALL thread=0 idle_iterations=800000] CLUSTER ...
[STALL thread=1 idle_iterations=800000] CLUSTER ...
[STALL thread=2 idle_iterations=800000] CLUSTER ...
Emergency shutdown: sending exit signal to all initialized cores
```

Suggested implementation:

- Add a cold-path wrapper such as `log_shutdown_stall_snapshot(...)`.
- The wrapper loops over `active_sched_threads_` and calls the existing
  `log_stall_diagnostics(t, total_tasks_, trigger_idle_iterations,
  trigger_last_progress_count)`.
- Call the wrapper from `handle_timeout_exit()` after `completed_.exchange(...)`
  succeeds and before `emergency_shutdown(runtime)`.
- Keep the snapshot best-effort; do not add a cross-thread barrier or wait for
  other scheduler threads to stop.
- Preserve the existing `log_stall_diagnostics()` format so existing grep
  workflows continue to work.

The only semantic compromise in the minimal version is that the
`idle_iterations` printed for every thread would be the trigger thread's count,
not each thread's private local counter. That is acceptable for the initial
diagnostic enhancement because the important core/task/COND state is read live
at snapshot time. If exact per-thread idle counts become necessary later, that
can be added via separate per-thread diagnostic counters.

### Alternatives Considered

- Add per-thread diagnostic counters first. This would make the printed idle
  counts exact, but it touches the scheduler loop and is not necessary for the
  first version.
- Print diagnostics after `emergency_shutdown()`. This is less useful because
  shutdown writes exit signals and deinitializes AICore register blocks, so COND
  state no longer represents the failure site.
- Add synchronization before printing. This risks making the fatal diagnostic
  path more fragile and should be avoided.

### Additional Context

Observed while analyzing:

```bash
task-submit --device auto --run "python tests/st/a2a3/tensormap_and_ringbuffer/paged_attention_unroll/test_paged_attention_unroll.py --device {} --platform a2a3 --enable-l2-swimlane --log-level v0"
```

Relevant files:

- `src/a2a3/runtime/tensormap_and_ringbuffer/runtime/scheduler/scheduler_cold_path.cpp`
- `src/a2a3/runtime/tensormap_and_ringbuffer/runtime/scheduler/scheduler_dispatch.cpp`
- `src/a2a3/runtime/tensormap_and_ringbuffer/runtime/scheduler/scheduler_context.h`

Related: #831, #860

zmnobug:
## Summary
- Add a cold-path shutdown snapshot wrapper for a2a3 and a5 schedulers
- Reuse existing stall diagnostics to dump all active scheduler threads
- Pass last progress count through timeout handling for summary logs
- Clamp snapshot thread count before indexing scheduler trackers

## Testing
- CI

Fixes #868

### DIFF
--- a/src/a2a3/runtime/tensormap_and_ringbuffer/runtime/scheduler/scheduler_cold_path.cpp
+++ b/src/a2a3/runtime/tensormap_and_ringbuffer/runtime/scheduler/scheduler_cold_path.cpp
@@ -327,8 +327,30 @@ void SchedulerContext::log_stall_diagnostics(
     }
 }
 
+void SchedulerContext::log_shutdown_stall_snapshot(
+    int32_t trigger_thread_idx, int32_t trigger_idle_iterations, int32_t trigger_last_progress_count
+) {
+    LOG_WARN(
+        "[SHUTDOWN_SNAPSHOT trigger_thread=%d reason=scheduler_timeout idle_iterations=%d] "
+        "dumping all scheduler threads before emergency shutdown",
+        trigger_thread_idx, trigger_idle_iterations
+    );
+    int32_t thread_count = active_sched_threads_ > 0 ? active_sched_threads_ : aicpu_thread_num_;
+    if (thread_count < 0 || thread_count > MAX_AICPU_THREADS) {
+        LOG_ERROR(
+            "[SHUTDOWN_SNAPSHOT trigger_thread=%d] invalid thread_count=%d, clamping to [0,%d]", trigger_thread_idx,
+            thread_count, MAX_AICPU_THREADS
+        );
+        thread_count = thread_count < 0 ? 0 : MAX_AICPU_THREADS;
+    }
+    for (int32_t t = 0; t < thread_count; t++) {
+        log_stall_diagnostics(t, total_tasks_, trigger_idle_iterations, trigger_last_progress_count);
+    }
+}
+
 int32_t SchedulerContext::handle_timeout_exit(
-    int32_t thread_idx, PTO2SharedMemoryHeader *header, Runtime *runtime, int32_t idle_iterations
+    int32_t thread_idx, PTO2SharedMemoryHeader *header, Runtime *runtime, int32_t idle_iterations,
+    int32_t last_progress_count
 #if PTO2_PROFILING
     ,
     uint64_t sched_start_ts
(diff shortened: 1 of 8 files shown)


ChaoZheng109:
`scheduler_types.h` 里 `STALL_LOG_INTERVAL` 两个 arch 的乘除顺序写反了：

- a2a3：`MAX_IDLE_ITERATIONS / 6 * 10` —— 先整除再乘 ≈ 1.67 × MAX（onboard 下 = 800000/6*10 = 1,333,330，已经超过 `MAX_IDLE_ITERATIONS` 本身），periodic stall 诊断永远不会在超时前触发，且整除在先还丢精度。
- a5：`MAX_IDLE_ITERATIONS * 6 / 10` —— = 0.6 × MAX（= 480,000）。

两边注释都写 "halfway to timeout"，但 a2a3 既和 a5 不一致、值又大于超时阈值。建议把 a2a3 统一成 `MAX_IDLE_ITERATIONS * 6 / 10`。

ChaoZheng109 (approved):

PR #906, opened 2026-05-30
[WIP] A5 performance opt

poursoul:


yanghaoran29:
Follow-up on the M5-derived ideas from this PR — we prototyped and measured two of them on A5 hardware (separate from this PR). Full detail in [#1511](https://github.com/hw-native-sys/simpler/pull/1511) and its investigation doc (`docs/investigations/2026-07-a5-last-task-alive-cross-cluster.md`).

**Scheme B — orch co-located with scheds (placement): measured negative, dropped.**
Moved orch into the sched-majority die so the `last_task_alive` handoff stays intra-die — on Scenario A, `ALLOWED_CPUS {5,6,7,8,3}` → `{3,6,7,8,5}` (orch die0 → die1). Measured on `spmd_multiblock_mix`, 100 rounds × 2 repeats: orch_cost went **up ~13%** (main 70.5/78.1 µs vs B 81.6/86.4 µs, ranges non-overlapping). Co-locating orch with scheds saved the cross-die `last_task_alive` read but cost more in L3/cache contention (orch's tensormap lookup + dep wiring competing with scheds in die1). Takeaway: the default cross-die orch isolation looks deliberate, not an oversight — at least at the workload size we could fit on this box.

**Scheme C — single-writer lazy publish: deferred (premise challenged).**
Rests on the same premise as M5 (cross-die `last_task_alive` traffic is the bottleneck). B's result suggests that isn't the dominant cost on A5, so C is doubtful here; our box's SNR (orch ~75 µs, ~10% run-to-run) also can't resolve the expected few-% effect. Settling it needs an isolated cross-die micro-benchmark + a full-SKU box (paged_attention Case1, block_dim=36, which our partial-good cube=28 devices can't run).

**Side find — real bug in `cleanup_retired` (a2a3 + a5), fixed in #1511.**
Surfaced while stress-testing the stale-SM tolerance C would need: `cleanup_retired` freed a slot's whole task chain assuming it belonged to one task (guarded only by a `debug_assert`). Under cleanup lag + slot reuse the chain mixes in a still-live task's entries and gets freed → silent tensormap corruption in release builds. Fixed by selective free (only entries whose `producer_task_id` matches the retiring `local_id`). Worth noting for this PR: the **"range cap" approach in the M5 design doc would not have been sufficient** — the live entry is freed on the first visit to a mixed chain regardless of the cap; the per-entry match is what actually fixes it.

yanghaoran29:
## Follow-up: PR #906 feature split and A5 measurements

I split the main performance ideas in this PR into independently measurable features and tested the viable subsets against `main` on A5 hardware. The primary workload was `paged_attention_unroll` Case1 on the same device, using STRACE timings.

| Feature | Change | Measurement vs main | Result / status |
|---|---|---|---|
| F2 | Poll AICore mailbox only from the selected scheduler threads | Orch **-0.57%**, Sched **-0.66%** | Noise-level; kept only on private branch [`exp/pr906-F2`](https://github.com/yanghaoran29/simpler/tree/exp/pr906-F2) |
| F3 | Publish `last_task_alive` to SM every K=16 local advances | Orch **-5.46%**, Sched **-4.77%** (100 rounds x 3) | Significant and repeatable; follow-up draft PR [#1575](https://github.com/hw-native-sys/simpler/pull/1575) |
| F7 | `always_inline` for `prepare_task` / `submit_task_common` | After rebasing onto `main@e80eaaf6`: Device **-0.00%**, Effective **-0.01%**, Orch **+0.04%**, Sched **-0.05%** (100 rounds x 3, alternating on the same card) | No credible benefit; the earlier ~2.6% result did not reproduce. [Detailed commit comment](https://github.com/yanghaoran29/simpler/commit/066599fcb8f0355ab084210dac1a9110eca76f8f#commitcomment-194302400) |
| F9 | Two-pass orchestration: emit all QKs before SF/PV/UP | Orch **+7.87%**, Sched **+1.68%** | Clear orchestration regression; do not land alone. Private branch [`exp/pr906-F9`](https://github.com/yanghaoran29/simpler/tree/exp/pr906-F9) |

Additional validation:

- F3 passed the full A5 tensormap ST suite: **19/19 PASS** (20 rounds per sample). On the same-card full-suite comparison, Case1 remained directionally consistent: Orch **-4.22%**, Sched **-3.82%**.
- F7 passed the A5 full-scene sweep and DFX-specific runs after rebasing, but remained performance-neutral versus the matching latest-main baseline.
- F1/F5/F8 were skipped because main already had equivalent behavior in the measured path or the original mechanism no longer applied cleanly.
- F4 (dedicated wiring thread) and F6 (FaninPool in the SM ring header) were not isolated: both require large structural/ABI-sensitive ports across substantial mainline drift.

Bottom line: among the isolated PR #906 ideas, **F3 is the only demonstrated standalone performance win**. F2 and F7 do not show credible benefit, while F9 regresses.

PR #1006, opened 2026-06-08
docs: add remote l3 worker contract

puddingfjz:
## Summary

This is PR 1 of the remote L3 worker split stack. It adds the document
contract for the remote L3 worker design and the audit artifacts used to split
the original large remote L3 branch into smaller, reviewable PRs.

The intent of this PR is to make the contract reviewable before the code that
implements it lands. Later PRs in the stack should be reviewed against these
documents.

  ## Relationship to PR #866 and split stack

  This PR is part of the split stack extracted from #866.

  Stack:
  - #1006: docs: add remote L3 worker contract
  - #1007: feat: add remote import callable descriptors
  - #1008: feat: add remote L3 wire codec
  - #1009: feat: add remote endpoint eligibility scheduling
  - #1010: feat: add remote L3 endpoint facade
  - #1011: feat: add remote L3 Python session runtime

  The top of the stack was verified to match the corrected `remote-l3-worker-design`
  branch content from #866, excluding empty CI retrigger commits.

## Stack position

- Base: `main`
- Head: `remote-l3-split-docs`
- Next PR: `remote-l3-split-callable-identity`

## Scope

This PR documents:

- The remote L3 worker model and the host/session/worker responsibilities.
- The remote L3 protocol surface, including frames, task submission,
  completion, control messages, health, and reserved fields.
- Buffer and transport responsibilities, including ownership, import,
  release, and future transport work.
- Updates to scheduler, orchestrator, worker-manager, task-flow, and
  hierarchical runtime documents so remote endpoints fit the existing runtime
  model.
- The split and audit plan used to divide the original remote L3 branch.
- Audit artifacts: compliance matrix, split map, risk register,
  verification checklist, and future-work list.

## Requirements covered

This PR establishes the source of truth for the rest of the stack:

- Separates required behavior from explicitly unsupported behavior.
- Marks hardware HCOMM profiles as future work for this split.
- Keeps Remote CommDomain support reserved for a later PR instead of exposing a
  partial implementation.
- Requires unsupported or reserved protocol paths to fail explicitly.
- Records the expected review boundaries for callable identity, wire codec,
  scheduler eligibility, C++ remote endpoint, and Python session runtime work.

## Important exclusions

This PR intentionally does not implement runtime behavior. It only defines the
contract and review artifacts. The implementation arrives in later stacked PRs.

The following remain future work unless a later dedicated PR changes the
contract:

- A2 RoCE hardware profile support.
- A3 HCCS hardware profile support.
[...]

PR #866, opened 2026-05-27
Add remote L3 worker

puddingfjz:
## Summary
  - Add a Remote L3 worker design for scheduling remote `Worker(level=3)` instances as L4 NEXT_LEVEL children while preserving the existing Orchestrator/Scheduler DAG model
  - Introduce a `WorkerEndpoint` boundary so local fork/shm children and remote L3 sessions share the same run/control/shutdown contract
  - Define fork-safe remote daemon/session startup, including runner prestart before `HELLO READY` and before any transport or health threads handle task traffic
  - Specify remote callable routing with separate outer remote-orch and inner L3 cid namespaces, transactional register/commit/abort, unregister tombstones, and unified
  `register(callable, workers=...)` API semantics
  - Define remote TaskArgs wire representation with canonical frame encoding, remote tensor sidecars, tag consumption before dispatch, explicit remote buffer handles, and first-cut
  rejection of null remote OUTPUT tensors
  - Define failure semantics for explicit success/failure completion, endpoint failure, downstream dependency poisoning, timeout handling, and cleanup
  - Specify HCOMM-backed steady-state control/data communication after bootstrap, including HCOMM RPC command lanes, HCOMM data adapter profiles for A2 RoCE/A3 HCCS/A5 UB, and
  independent health signaling
  - Keep Remote CommDomain and possible `CommContext` ABI changes as later work while reserving domain-related controls for future extension
  - Provide a staged implementation plan and test matrix covering local adapter compatibility, frame validation, remote simulation, callable registration, remote buffers, health, and
  hardware-gated HCOMM profiles

  ## Testing
  - Docs only; no tests run
  - `git diff --check` passed

PR #1011, opened 2026-06-08
feat: add remote l3 python session runtime

puddingfjz:
## Summary

This is PR 6 of the remote L3 worker split stack. It adds the Python remote L3
session runtime and simulation-facing remote worker integration on top of the
documents, callable identity, wire codec, scheduler abstraction, and C++
endpoint facade introduced by the earlier PRs.

This is the top implementation PR in the split stack. At this point the stacked
branch content matches the corrected remote L3 feature branch, excluding
drop-only empty CI retrigger commits.

  ## Relationship to PR #866 and split stack

  This PR is part of the split stack extracted from #866.

  Stack:
  - #1006: docs: add remote L3 worker contract
  - #1007: feat: add remote import callable descriptors
  - #1008: feat: add remote L3 wire codec
  - #1009: feat: add remote endpoint eligibility scheduling
  - #1010: feat: add remote L3 endpoint facade
  - #1011: feat: add remote L3 Python session runtime

  The top of the stack was verified to match the corrected `remote-l3-worker-design`
  branch content from #866, excluding empty CI retrigger commits.

## Stack position

- Base: `remote-l3-split-cpp-remote-endpoint`
- Head: `remote-l3-split-python-runtime`
- Previous PR: `remote-l3-split-cpp-remote-endpoint`
- Top of split stack: yes

## Scope

This PR adds or updates:

- `simpler-remote-worker`.
- `simpler-remote-l3-session`.
- Python remote session lifecycle management.
- Python remote worker daemon behavior.
- `RemoteCallable` integration.
- `RemoteWorkerSpec` integration.
- `RemoteBufferHandle`, `RemoteBufferExport`, `RemoteTensorRef`, and hidden remote tensor sidecar integration.
- Simulation remote buffer allocation, copy, export, import, release-import,
  and owner-free behavior.
- Remote dispatcher and inner worker registry controls.
- Health lane and session teardown handling.
- Python tests covering task-args materialization, callable identity, session
  behavior, remote buffer lifecycle, and CLI smoke paths.

## Requirements covered

This PR covers the Python runtime and simulation requirements from the remote
L3 contract:

- Remote session startup preserves the fork-before-thread invariant.
- `HELLO READY` is used as the scheduling barrier before remote work is made
  eligible.
- Remote callables are registered through stable callable identity descriptors.
- Remote task arguments materialize through explicit remote tensor sidecars.
- Remote buffer owner/import/release-import lifetimes are tracked.
- Owner frees wait for active slot references and imports to drain.
- Partial register and partial import failures clean up visible state.
- Session shutdown tears down health and command-lane state deterministically.
- Reserved Remote CommDomain controls fail explicitly in this split.
- Hardware transport profiles remain future work and are not partially exposed.

## Why this is separate

This PR contains the highest-level process and lifecycle behavior. Keeping it
at the top of the stack means earlier PRs can be reviewed for their contracts
and lower-level invariants first, while this PR focuses on Python runtime
integration, process lifecycle, health, cleanup, and simulation behavior.

## Important exclusions

This PR intentionally does not add:

- A2 RoCE hardware profile support.
- A3 HCCS hardware profile support.
- A5 UB HCOMM hardware profile support.
- Remote CommDomain support.

Those remain future work per the contract documents.

## Verification

Targeted verification already run during the split:

- Remote task args sidecar tests: passed.
- Remote L3 session task-args materialization tests: passed.
- Callable identity Python unit tests in sandbox: passed, with socket-denied
  cases skipped by the sandbox.
- Callable identity Python unit tests outside the sandbox: passed.
- Editable build/install smoke: passed.
- `simpler-remote-worker --help`: passed.
- `simpler-remote-l3-session --help`: passed.

Top-of-stack verification:

- The top split branch was compared against the corrected
  `remote-l3-worker-design` branch.
- Tracked file contents matched exactly at the time of split verification.
- Empty CI retrigger commits were intentionally dropped.

## Reviewer notes

The main review focus should be lifecycle correctness:

- No Python session should fork after network or C++ worker threads have
  started.
- Health threads must not outlive the transport/session objects they reference.
- A command-lane wait must not hide a health failure indefinitely.
- Owner buffer frees must not race active imports or task slot references.
- Partial register/import failure paths should leave no visible committed
  remote state.


puddingfjz:
# PR 1011 Remote L3 Review Fix Plan

Local planning note for PR #1011. This file is not intended to be committed
and is not a public API contract.

## Recommended API Direction

Use the ordinary public `TaskArgs` as the user-facing argument container. Put
remote semantics on first-class handles such as `RemoteTensorRef` and
`RemoteBufferHandle`.

The public API should be:

```python
args = TaskArgs()
args.add_tensor(remote_ref, TensorArgType.INPUT)
orch.submit_next_level(remote_callable, args)
```

Do not add a public `RemoteTaskArgs` wrapper in PR #1011. Keep
`RemoteTaskArgsWire`, sidecars, transport descriptors, and remote tensor
descriptors internal.

The model is closer to Ray/PyTorch RPC in public API shape: ordinary argument
containers can carry opaque remote references. It should not copy Ray's
top-level value-dereference semantics. A `RemoteTensorRef` means remote
placement plus access capability, not "fetch this into a local
`ContinuousTensor` before submit".

Do not add public `remote_materialize` in PR #1011. If local materialization is
needed later, design it as a separate explicit API with copy, synchronization,
error, and lifecycle semantics.

## Finding 1: Access Flags Are Not Enforced

### Current Problem

Imported handle access bits are documented but not enforced during submit.
`_sidecar_from_ref()` builds the remote descriptor without preserving the
handle's access contract in a place that validation uses, endpoint filtering
only intersects endpoint ids, and C++ sidecar validation checks shape and
endpoint form rather than the handle access mask.

As a result, a read-only import can be submitted as `OUTPUT` or `INOUT`, and a
write-only import can be submitted as `INPUT`.

### Design Decision

Keep `access_flags`. Do not remove them and do not turn every remote handle
into implicit read-write. Remote buffer import is a capability/lease, not
ownership transfer of the underlying allocation.

Do not add another public read/write flag to `RemoteTensorRef`. The two
existing pieces are enough:

- `RemoteBufferHandle.access_flags`: granted capability.
- `TensorArgType`: use intent for this submission and dependency tracking.

Submit validation should check that the tag's minimum required access is a
subset of the handle's granted capability. It should not mutate or downgrade
the handle.

Recommended mapping:

```text
INPUT            requires READ
OUTPUT           requires WRITE
OUTPUT_EXISTING  requires WRITE
INOUT            requires READ | WRITE
NO_DEP           requires READ | WRITE
```

Remote `NO_DEP` is allowed. Its semantics are:

```text
NO_DEP skips dependency tracking only. It does not declare read-only or
write-only use. Because it carries no direction, a remote NO_DEP tensor must
use a handle with READ | WRITE capability.
```

This keeps behavior close to local `NO_DEP`, where the tag does not participate
in dependency inference and the user callable/kernel is responsible for its
actual memory behavior. The remote path still preserves the capability
contract by requiring a handle that covers all possible access directions.

### Fix Plan

- Add a Python helper that maps `TensorArgType` to required remote access.
- Validate remote sidecars against `args.tag(i)` before C++ submit.
- Keep validation in submit path, not in wire `RemoteTensorDesc.flags`.
- Do not use `RemoteTensorDesc.flags` for access; protocol v1 reserves it and
  rejects non-zero values.
- Optionally add defensive C++ validation when enough handle capability
  metadata is available there, but the first user-facing error should happen in
  Python submit validation.

Suggested tests:

- Read-only handle as `OUTPUT` is rejected.
- Read-only handle as `INOUT` is rejected.
- Write-only handle as `INPUT` is rejected.
- Read-write handle as `INPUT`, `OUTPUT`, `OUTPUT_EXISTING`, `INOUT`, and
  `NO_DEP` is accepted.
- Read-only or write-only handle as remote `NO_DEP` is rejected.

## Finding 2: Public API Exposes Raw Transport Details

### Current Problem

`RemoteBufferExport` is a public dataclass with fields such as `remote_addr`,
`rkey_or_token`, `ub_ldst_va`, and `transport_descriptor`. The simulation
session writes `entry.addr + offset` into `remote_addr` and places the shared
memory name into the transport descriptor. `Worker.remote_export()` returns
this object directly to user code.

That conflicts with the documented opaque-handle model. It also leaks simulator
host process details through public Python API and freezes transport internals
as user-visible API.

### Design Decision

Keep remote handles/export descriptors opaque and controlled. Users may pass
them back to runtime APIs, but they must not inspect raw transport keys,
addresses, HCOMM descriptors, UB addresses, shared-memory names, or simulator
host pointers.

### Fix Plan

- Replace the public dataclass shape with a controlled opaque class.
- Use `__slots__` and an internal construction token, or an equivalent
  provenance mechanism, so users cannot forge valid exports.
- Store raw wire payload in private fields or an internal payload object.
- Expose only safe metadata through read-only properties, such as size, owner
  endpoint, address-space class, access mask, transport profile name, and
  release state if those are documented as stable.
- Ensure `Worker.remote_export()` is the only public way to obtain a valid
  export descriptor.
- Ensure `Worker.remote_import()` rejects forged, stale, or wrong-session
  descriptors.
- Keep simulator `remote_addr` and shared-memory names internal to the session
  and binding path. They must not appear as public attributes or in `repr()`.

Suggested tests:

- Public construction of a raw export fails or cannot produce an importable
  descriptor.
- `RemoteBufferExport` has no public `remote_addr`, `rkey_or_token`,
  `ub_ldst_va`, or `transport_descriptor` attributes.
- `repr(exported)` does not include transport keys, host addresses, or shared
  memory names.
- `remote_import()` rejects forged export-like objects.

## Finding 3: Public RemoteTaskArgs Is Documented But Not Implemented

### Current Problem

Docs and the PR body claim a public `RemoteTaskArgs` wrapper and examples such
as `RemoteTaskArgs.add_tensor(...)`, but the code exposes ordinary `TaskArgs`
plus `RemoteTensorRef` support. `__all__` does not include `RemoteTaskArgs`.

### Design Decision

Do not add a public `RemoteTaskArgs` wrapper. The better public API is ordinary
`TaskArgs` carrying `RemoteTensorRef`, with remote sidecars and
`RemoteTaskArgsWire` kept internal.

This is both easier for user code and closer to established distributed APIs
where ordinary argument containers can carry opaque references.

### Fix Plan

- Update docs and PR body to remove public `RemoteTaskArgs` claims.
- Document `TaskArgs.add_tensor(RemoteTensorRef(...), tag=...)` as the public
  remote tensor argument path.
- Keep `RemoteTaskArgsWire`, sidecars, and remote descriptor structs internal.
- Update examples to use `TaskArgs`.
- Keep public docs clear that `TaskArgs` can contain local tensors and remote
  tensor refs only under validated, documented endpoint and capability rules.

Suggested tests:

- Public import surface does not expose `RemoteTaskArgs`.
- `TaskArgs.add_tensor(RemoteTensorRef(...), tag)` remains covered.
- Docs examples match runnable API.

## Finding 4: Bootstrap Manifest Hashid Is Not Revalidated

### Current Problem

Dynamic CONTROL registration validates import targets and canonical hash ids,
but bootstrap manifest `remote_task_dispatcher` entries are directly loaded
with `bytes.fromhex(entry["hashid"])` and `_load_import_target(entry["target"])`.

That creates two different identity paths. A bad manifest can bind hashid A to
target B and still let the session reach `HELLO READY`.

### Severity And Scope

This is a reasonable `Should fix`, not necessarily `Must fix`, unless the
manifest is treated as untrusted input. It is not over-validation if it stays
within canonical callable identity checks already required by the docs and
dynamic registration path.

### Fix Plan

- Extract a shared validator for remote import callable descriptors.
- Validate bootstrap manifest entries before sending or before accepting
  `HELLO READY`.
- Check hashid format and digest length.
- Check target import syntax and loadability.
- Recompute the canonical descriptor hash from target and kind.
- Reject hashid/target/kind mismatches.
- Reject duplicate hashids bound to different targets.
- Do not add unrelated import allowlists or environment policy checks in this
  PR unless the docs already require them.

Suggested tests:

- Bootstrap manifest with malformed hashid is rejected.
- Bootstrap manifest with hashid/target mismatch is rejected before READY.
- Bootstrap manifest and dynamic CONTROL path compute the same hashid for the
  same target.
- Duplicate hashid with different target is rejected.

## Finding 5: Partial Init Failure Can Leak Remote Sessions

### Current Problem

`_init_hierarchical()` can open remote sessions and register remote sockets
before `Worker.init()` sets `_initialized = True`. If a later session
connection, HELLO, or health setup fails, `init()` raises while `_initialized`
is still false. A later `close()` returns immediately, so already-created
resources may be orphaned.

### Severity And Scope

This is a lifecycle cleanup issue, not over-validation. It is a reasonable
`Should fix`. Do not fix it by setting `_initialized = True` early; that would
blur object state.

### Fix Plan

- Wrap `init()` in a failure cleanup path:

```python
try:
    self._init_hierarchical()
except BaseException:
    self._cleanup_partial_init()
    raise
```

- Make `_cleanup_partial_init()` best-effort and idempotent.
- Preserve the original init exception. Cleanup errors may be logged or
  attached, but must not replace the primary failure.
- Clean already-opened remote sessions.
- Tear down any created `_Worker` endpoint state.
- Close and unlink created shared-memory mailboxes.
- Reap or shut down child processes if the failure point can occur after fork.
- Clear partial lists so a second cleanup does not double free.

Suggested tests:

- Inject failure after the first remote session opens and verify session close.
- Inject failure after `add_remote_l3_socket()` and verify endpoint teardown.
- Calling `close()` after failed `init()` is safe and idempotent.

## Hidden Sidecar Robustness

The current hidden sidecar strategy can remain internal, but it should not be a
validation bypass.

Fix plan:

- Attach sidecar creation to `TaskArgs.add_tensor(RemoteTensorRef(...))`.
- Validate sidecar contents again at submit time using `TaskArgs` tags.
- Keep sidecar ownership and lifetime deterministic.
- Avoid relying on `id(args)` global state unless cleanup handles object reuse,
  exceptions, and `TaskArgs.clear()`.
- Transfer or clear sidecar refs after submit/error paths in a predictable way.

## Pending Remote Free Error Handling

Keep this as a cleanup follow-up from the review even though it was not one of
the five main findings.

Fix plan:

- Do not remove pending release/free records before the control command
  succeeds.
- If release/free fails, keep the record pending or move it to an explicit
  uncertain state.
- Report uncertain cleanup during teardown without losing the original error.

## Acceptance Criteria

- Public docs no longer claim a public `RemoteTaskArgs`.
- Public docs show `TaskArgs.add_tensor(RemoteTensorRef(...), tag=...)`.
- `RemoteTensorRef` in `TaskArgs` means remote capability consumption, not
  implicit local materialization.
- Remote `NO_DEP` is supported and documented as dependency-only; it requires a
  read-write-capable remote handle.
- Submit-time validation enforces endpoint, tensor metadata, byte range,
  release state, generation, and tag-vs-access compatibility.
- No public Python object exposes simulator host pointers, shared-memory names,
  raw rkeys, transport descriptors, or UB LD/ST addresses.
- Users cannot forge valid imports or export descriptors with public
  constructors alone.
- Bootstrap manifest hashid/target mismatches fail before `HELLO READY`.
- Partial remote-session initialization failure is cleaned up deterministically.
- PR 1011 does not add public local materialization or hardware HCOMM/RDMA
  implementation; those remain separate future work.

puddingfjz:
**First**, we should replace endpoint_id with worker_id. The identity exposed here is not a transport endpoint identity; it is the user-facing identity of a NEXT_LEVEL worker, regardless of whether that worker is backed by a local mailbox child or a remote L3 endpoint. Using endpoint_id makes the local Worker child case awkward and suggests the id is remote/transport-specific. worker_id better matches the API contract:` add_worker() `and `add_remote_worker()` both return stable worker ids, and those ids are used for affinity, eligibility, control operations, and remote buffer ownership.

  **Second**, once NEXT_LEVEL uses worker_id, we need to keep it distinct from the C++ internal vector index. We want to allow NEXT_LEVEL workers to mix local Worker children and remote L3 workers. For example, a user may call `add_remote_worker()` before `add_worker()`, or the other way around, while the C++ WorkerThread vector is ordered by the runtime startup flow rather than by the user-facing add order. Therefore APIs, affinity, eligible worker sets, control paths, and remote buffer handles/results should use and resolve by stable worker_id; only direct internal vector lookup should use worker_index.

puddingfjz:
While reviewing this PR, we found a broader scheduler contract issue that already exists on main and should be tracked separately from the worker-id cleanup here. Current group dispatch is all-or-nothing: the scheduler first selects workers for all group members, and only dispatches after every member has a selected worker. For automatically selected members, `pick_idle_excluding_eligible()` excludes workers already selected for earlier members. As a result, a shape like `eligible_endpoint_ids = {{0}, {0}}` can pass validation when endpoint 0 exists, but the scheduler cannot select the second member, requeues the whole group slot, and may retry forever. This PR should not force a distinct-worker contract for that case; it should only fix the narrower PR-local issue of rejecting unknown eligible worker ids at submit time. The broader question of whether group members may reuse the same worker, or should be rejected up front, is tracked separately in the new issue https://github.com/hw-native-sys/simpler/issues/1105.

puddingfjz:
# Remote L3 AI-Generated Review Comment Fix Summary

This update addresses the high-priority items from the AI-generated review
comments on the Remote L3 PR.

## Fixed Review Comment Items

1. Remote session bind safety

   Wildcard session binds such as `0.0.0.0` now require explicit opt-in, and
   the default session bind host is derived from the remote endpoint host.

2. Remote session lifecycle cleanup

   Session startup and post-READY accept now have bounded timeouts. Session
   runner processes are reaped after successful startup, and active health
   connections are closed during shutdown.

3. `RemoteTensorRef` sidecars on SUB submits

   `submit_sub()` and `submit_sub_group()` now reject `RemoteTensorRef`
   sidecars because SUB submits use the local Python worker path and do not
   deliver remote sidecar metadata.

4. `HOST_INLINE` payload size validation

   `RemoteTensorRef.host_inline()` now requires the inline payload length to
   match the declared `shape * dtype` size.

5. `Worker.remote_import()` rollback safety

   `Worker.remote_import()` now pins the owner buffer before issuing the remote
   import control command, rolls back that pin on failure, and best-effort
   releases the remote import mapping if Python handle construction fails after
   the remote import succeeds.

6. Remote sidecar slot-ref capture rollback

   Remote sidecar slot-ref capture now rolls back previously acquired refs if a
   later handle acquisition fails.

7. Remote callable register/unregister control exceptions

   Remote callable register/unregister now handles direct control exceptions
   consistently with `ok=false` replies. Register prepare/commit exceptions
   enter the partial-failure cleanup path; prepared workers are aborted and
   committed workers are unregistered where needed; hashids are marked
   uncertain when cleanup state cannot be trusted; remote unregister remains
   best-effort and records uncertain cleanup instead of propagating transport
   exceptions.

## Validation

Added focused regression coverage for each fixed path.

The affected unit suites pass:

```text
tests/ut/py/test_callable_identity.py
tests/ut/py/test_task_interface.py
tests/ut/py/test_remote_l3_lifecycle.py

180 passed, 13 skipped
```

jvjhfhg (approved):

PR #1023, opened 2026-06-10
fix(runtime): allow zero-size view at any offset (Tensor::view bounds-check)

csy0225:
## Summary

`Tensor::view()` asserts `view_offsets[i] + view_shapes[i] <= shapes[i]` for every dim — but the orchestration codegen emits guards of the form `(off >= shape ? 0 : min(...))` to clamp the **shape** dim to 0 when a dynamic offset runs past the parent, leaving the **offset** as the raw (possibly out-of-range) value and expecting the runtime to honour the empty-view contract. With the current assertion, that codegen pattern fails on board.

This PR loosens `view()` to detect any zero-shape dim up front and skip both the per-dim offset assertion and the `start_offset` advancement (no bytes captured, so offset bounds are irrelevant) and the trailing `assert_in_buffer_bounds()`. All previously valid views still pass through the same code path; only previously-asserting-but-now-OK views (zero shape + out-of-range offset) now correctly return an empty Tensor.

## Repro signature on board (a2a3)

The current assertion surfaces as a generic AICPU exception `errorCode=0x2a` → host-side `aclrtSynchronizeStreamWithTimeout (AICPU) failed: 507018`. With `ASCEND_GLOBAL_LOG_LEVEL=1` and `ASCEND_PROCESS_LOG_PATH=<dir>`, the device-side plog shows:

```
[ERROR] AICPU: Assertion failed:
        view_offsets[i] + view_shapes[i] <= shapes[i]
[ERROR] AICPU: Location:
        .../runtime/tensormap_and_ringbuffer/runtime/tensor.h:356
```

The trigger is the codegen lowering of dynamic offsets like `static_cast<uint32_t>(pos)` where `pos = ctx_len - 1`. When `ctx_len` is 0 (the very first prefill step, or a dummy-weight bring-up where seq_lens is not pre-populated), `pos = -1` and the cast underflows to `0xFFFFFFFF`. The codegen guard correctly sets `cos_row_shapes[0] = 0` for this offset, but the OFFSET passed to `view()` is the raw `0xFFFFFFFF`, which fails the assertion.

A typical generated chip-side site (paraphrased from `aicpu_orchestration_entry`):

```cpp
int32_t ctx_len = get_tensor_data<int32_t>(ext_seq_lens, 1, indices_ctx_len);
int64_t pos = (static_cast<int64_t>(ctx_len) - 1);
uint32_t cos_row_offsets[2] = {static_cast<uint32_t>(pos), 0};
uint32_t cos_row_shapes[2] = {
    (cos_row_offsets[0] >= ext_rope_cos.shapes[0]
        ? 0u
        : std::min<uint32_t>(1, ext_rope_cos.shapes[0] - cos_row_offsets[0])),
    ...
};
Tensor cos_row = ext_rope_cos.view(cos_row_shapes, cos_row_offsets);  // <-- asserts
```

## Why fix it in the runtime, not the codegen

The codegen's intent is unambiguous: `(off >= shape ? 0 : ...)` is exactly *"if offset is out of range, return an empty view"*. Hardening every codegen call site to also clamp the offset would touch many emit paths and is prone to regression; honoring the empty-view contract once in the runtime is a strict improvement and matches what the codegen already documents in its emitted comments.

## Diff shape

- Two byte-identical sibling headers (a2a3 and a5) updated symmetrically.
- ABI unchanged: `sizeof(Tensor)` is untouched and the `static_asserts` in `comm_context.h` continue to hold.
- No call-site changes anywhere — purely a one-method semantic widening.

## Test plan

- [x] Existing `tests/ut/cpp/{a2a3,a5}/test_tensormap.cpp` view-based overlap tests continue to pass (no preexisting test exercises the zero-shape-view + out-of-range-offset path).
- [ ] Suggest adding a regression test: `base.view([0, ...], [0xFFFFFFFF, 0])` returns an empty Tensor with `numel()==0` and does not abort.
- [x] Validated locally on Ascend a2a3 single-rank decode: with this fix the assertion never fires; the chip-side run progresses past the rope-view sites it previously aborted at.

## References

The originating model-side investigation is documented in this companion issue / mirror in the consumer project: pypto-side root-cause walkthrough exists at `docs/upstream-issues/pypto-1702-followup.md` (separate codepath from PR `hw-native-sys/pypto#1718`).

### DIFF
--- a/simpler_setup/runtime_compiler.py
+++ b/simpler_setup/runtime_compiler.py
@@ -284,7 +284,11 @@ class RuntimeCompiler:
                     dest_dir.mkdir(parents=True, exist_ok=True)
                     dest_dispatcher = dest_dir / dispatcher_name
                     shutil.copy2(dispatcher_so, dest_dispatcher)
-                    subprocess.run(["strip", "-s", str(dest_dispatcher)], check=True)
+                    # Cross-arch strip: aicpu .so is aarch64 even on x86 host;
+                    # GNU strip 2.38 on Ubuntu 22.04 cannot read it. Prefer
+                    # llvm-strip (multi-arch) when available.
+                    strip_bin = shutil.which("llvm-strip") or "strip"
+                    subprocess.run([strip_bin, "-s", str(dest_dispatcher)], check=True)
             if output_dir is not None:
                 od = Path(output_dir)
                 od.mkdir(parents=True, exist_ok=True)
--- a/src/a2a3/platform/onboard/host/CMakeLists.txt
+++ b/src/a2a3/platform/onboard/host/CMakeLists.txt
@@ -36,10 +36,17 @@ endif()
 # issue. The SDMA workspace init in comm_hccl.cpp pulls in pto-isa
 # headers and CANN-9.0-only aclnn symbols (aclnnShmemSdmaStarsQuery*),
 # even though it is logically orthogonal to HCCL comm bootstrap and only
-# needed by the sdma_async_completion_demo. Until that coupling is
-# refactored away, the macro is forced ON: PTO_ISA_ROOT and CANN 9.0+ are
-# hard build/runtime preconditions for a2a3 onboard.
-set(SIMPLER_ENABLE_PTO_SDMA_WORKSPACE ON)
+# needed by the sdma_async_completion_demo.
+#
+# Forced OFF: on this driver/CANN the ShmemSdmaStarsQuery AICPU kernel
+# (libcpu_kernels.so::RunCpuKernel) raises an AICPU exception (errorCode
+# 0x2a) during comm bootstrap, which faults the whole chip and cascades
+# every subsequent op to 507899/507018 (notably aclrtIpcMemImportByKey in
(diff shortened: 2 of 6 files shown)


ChaoZheng109:
本 PR 的全部实质改动已通过 #1158(squash 合入 `main`,merge commit 47a411c1)落地:
- `Tensor::view()` 零尺寸视图修复(并补齐了 `is_contiguous`/`extent_elem_cache` 重置,修掉本 PR 版本会导致的空视图 overlap 误判);
- `runtime_compiler.py` 跨架构 strip fallback;
- a2a3/a5 HCCL `--no-as-needed` 链接修复;
- a2a3/a5 comm_hccl P2P/IPC best-effort + `ACL_RT_IPC_MEM_IMPORT_FLAG_ENABLE_PEER_ACCESS`。

唯一差异:本 PR 把 `SIMPLER_ENABLE_PTO_SDMA_WORKSPACE` 改为 `OFF`,经确认是误操作——#1158 保持 `ON`。

#1158 已合入,本 PR 已过时(且会与 main 冲突),关闭。

PR #1028, opened 2026-06-11
Add: selective scalar tensor dump

### LINKED ISSUE #1026: [Feature] Support selective scalar dump via Arg::dump(...)
### Summary

Support selecting scalar arguments in the existing `Arg::dump(...)` partial-dump API, so users can request mixed tensor/scalar dumps such as:

```cpp
params.dump(tensor1, tensor2, scalar1, scalar2);
```

This should dump only the selected tensor and scalar arguments instead of requiring all scalar arguments on the task to be dumped.

### Motivation / Use Case

Partial tensor dump already supports selecting specific tensor arguments with `params.dump(tensor1, tensor2)`. Scalar dump should follow the same user-facing model so users can focus debug output on the scalar values that matter for a given task.

Without selective scalar dump, `params.dump()` dumps every tensor and scalar argument on the `Arg`, which is noisy for kernels with many scalar parameters. Users also need mixed selection to debug real orchestration code where one tensor output and one scalar control value are the relevant pair.

### Proposed API / Behavior

```cpp
params.dump();                          // dump all tensor + scalar args
params.dump(tensor1, tensor2);          // dump selected tensor args
params.dump(tensor1, scalar1, scalar2); // dump selected tensor/scalar args
```

Expected behavior:

- no-arg `dump()` keeps task-level behavior and dumps all tensor and scalar args on the `Arg`;
- variadic `dump(...)` can select tensors and scalar lvalues in one call;
- selected scalar args are recorded using the same task arg-index selection mechanism as selected tensors;
- scalar records are emitted at `before_dispatch`;
- scalar payload belongs to the task, so mixed-subtask tasks should record each selected scalar once per task, while tensor records remain tied to callable signature traversal;
- if the same scalar lvalue is added more than once, selection can use the first matching scalar arg and log a warning so users know to use distinct local variables when they need a later duplicate.

### Alternatives Considered

A separate scalar handle API such as `add_scalar_arg(...)` was considered, but it would introduce a new user-facing concept that tensor selection does not need. Reusing `params.dump(scale)` keeps the scalar and tensor partial-dump interfaces consistent.

Value-based scalar matching alone is not enough, because repeated scalar values, equivalent bit patterns, and type conversion to `uint64_t` can be ambiguous. The API should treat scalar selection as selecting registered scalar lvalues rather than every scalar with the same value.

### Additional Context

Related: #965

This tracks the selective scalar dump enhancement for tensor dump DFX. It is separate from scalar dtype decoding: this issue is about which scalar arguments are selected and recorded, not how scalar values are decoded in the manifest.

zmnobug:
## Summary
- Extend `Arg::dump(...)` so partial dump can select scalar lvalues together with tensors.
- Use one tensor/scalar arg mask for AICPU filtering and dump scalars once per task.
- Update tensor dump docs and ST coverage for task, scalar-only, and mixed selection.

## Testing
- `git diff --check`
- `git diff --cached --check`
- `g++ -std=c++17 -fsyntax-only ... src/a2a3/.../pto_orchestrator.cpp`
- `g++ -std=c++17 -fsyntax-only ... src/a5/.../pto_orchestrator.cpp`
- `python3 -m py_compile tests/st/a2a3/tensormap_and_ringbuffer/dfx/tensor_dump/test_tensor_dump.py`
- Not run: full runtime rebuild / ST. `npu-smi info` timed out locally, and the known local a2a3sim libstdc++ ABI issue blocks the tensor-dump ST.

Fixes #1026

### DIFF
--- a/src/a2a3/runtime/tensormap_and_ringbuffer/runtime/pto_types.h
+++ b/src/a2a3/runtime/tensormap_and_ringbuffer/runtime/pto_types.h
@@ -28,6 +28,9 @@
 #include <stdint.h>
 #include <string.h>
 
+#include <type_traits>
+#include <utility>
+
 #if defined(__aarch64__)
 #include <arm_neon.h>
 #endif
@@ -37,6 +40,10 @@
 #include "task_args.h"
 #include "tensor.h"
 #include "tensor_arg.h"
+
+#ifndef PTO2_PROFILING
+#define PTO2_PROFILING 1
+#endif
 
 // Task arguments — alias the common CORE_MAX_* constants (single source of
 // truth in src/common/task_interface/arg_direction.h, transitively included
@@ -177,14 +184,22 @@ struct Arg : TaskArgsTpl<TensorRef, uint64_t, MAX_TENSOR_ARGS, MAX_SCALAR_ARGS, 
     const char *error_msg{nullptr};
     PTO2LaunchSpec launch_spec;  // SPMD launch parameters (block_num, etc.)
 
+    void clear() {
+        TaskArgsTpl<TensorRef, uint64_t, MAX_TENSOR_ARGS, MAX_SCALAR_ARGS, TensorArgType>::clear();
+#if PTO2_PROFILING
+        dump_arg_mask_ = 0;
+        dump_arg_index_ambiguous_mask_ = 0;
+        clear_scalar_sources();
+        memset(scalar_dtypes_, 0, sizeof(scalar_dtypes_));
+#endif
+        explicit_deps_ = nullptr;
+        explicit_dep_count_ = 0;
+    }
+
     void reset() {
         clear();
         has_error = false;
         error_msg = nullptr;
-        tensor_dump_arg_mask_ = 0;
-        explicit_deps_ = nullptr;
-        explicit_dep_count_ = 0;
-        memset(scalar_dtypes_, 0, sizeof(scalar_dtypes_));
     }
 
     void set_error(const char *msg) {
@@ -196,23 +211,32 @@ struct Arg : TaskArgsTpl<TensorRef, uint64_t, MAX_TENSOR_ARGS, MAX_SCALAR_ARGS, 
 
     template <typename... Args>
     void dump(Args &&...args) {
+#if PTO2_PROFILING
         static_assert(
             (std::is_lvalue_reference_v<Args> && ...),
-            "dump: temporaries are not allowed — pass tensors already added to this Arg"
+            "dump: temporaries are not allowed — pass tensors/scalars already added to this Arg"
         );
         static_assert(
-            ((std::is_same_v<std::decay_t<Args>, Tensor> || std::is_same_v<std::decay_t<Args>, TensorCreateInfo>) &&
-             ...),
-            "dump: all arguments must be Tensor or TensorCreateInfo"
+            (is_supported_dump_arg_v<Args> && ...),
+            "dump: all arguments must be Tensor, TensorCreateInfo, or scalar lvalues"
         );
         if constexpr (sizeof...(Args) == 0) {
-            mark_all_tensor_dump_arg();
+            mark_all_dump_args();
         } else {
-            (mark_tensor_dump_arg(args), ...);
+            (mark_dump_arg(args), ...);
         }
+#else
+        ((void)args, ...);
+#endif
     }
 
-    uint64_t tensor_dump_arg_mask() const { return tensor_dump_arg_mask_; }
+#if PTO2_PROFILING
+    uint64_t tensor_dump_arg_mask() const { return dump_arg_mask_; }
+    uint64_t tensor_dump_arg_index_ambiguous_mask() const { return dump_arg_index_ambiguous_mask_; }
+#else
+    uint64_t tensor_dump_arg_mask() const { return 0; }
+    uint64_t tensor_dump_arg_index_ambiguous_mask() const { return 0; }
+#endif
 
     template <typename... Args>
     void add_input(Args &&...args) {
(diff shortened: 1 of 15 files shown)


zmnobug:
@gemini-code-assist Addressed the review feedback in `71d95b80`.

- `dump_tensors_for_task` now uses `tensor_index` when querying the dump arg mask, and records the emitted tensor slot index consistently. This avoids shifted filtering when callable args mix tensors and scalars.
- Removed the function-local static `logged_arg_warning`. The warning-once state is now stored in `PTO2OrchestratorState` for both a2a3 and a5, so it is explicitly initialized with each orchestrator state instead of leaking across AICPU runs.

Validation run:
- `git diff --check`
- `g++ -std=c++17 -fsyntax-only` for a2a3 `pto_orchestrator.cpp`
- `g++ -std=c++17 -fsyntax-only` for a5 `pto_orchestrator.cpp`
- `python3 -m py_compile tests/st/a2a3/tensormap_and_ringbuffer/dfx/tensor_dump/test_tensor_dump.py`

ChaoZheng109:
Two changes requested before merge — both apply to **both** the a2a3 and a5 copies of `pto_orchestrator.cpp` (and the matching `pto_types.h`).

### 1. Gate the `log_arg_warning_once` call behind `#if PTO2_PROFILING`

The new `log_arg_warning_once(orch, task_id.raw, args.warning_msg())` call sits in `submit_task_common` right after `result.set_task_id(task_id)` — **outside any profiling gate**, so it runs on every task submission whether or not the dump/profiling feature is compiled in.

Its functional neighbor a few lines below — `set_dump_tensor_task_mask(...)`, which propagates the dump arg mask — is already wrapped in `#if PTO2_PROFILING`. The warning is the only piece of dump machinery that leaks onto the always-on submit path. That cuts against our rule that profiling overhead stays behind the macro + runtime switch, and it's inconsistent with the immediately adjacent dump code.

It also means the warning fires even when dump is **disabled at runtime**, where the `dump()` marker is a no-op — i.e. we warn about something that has no effect.

**Minimum fix:** move the `log_arg_warning_once(...)` call into the existing `#if PTO2_PROFILING` block next to `set_dump_tensor_task_mask`. The `warning_msg_` field on `Arg` and the `logged_arg_warning` flag on the orch state should follow the same gating so they cost nothing when profiling is compiled out.

### 2. Surface the ambiguity through the dump JSON, not (only) a log line

Rather than building a new log path for this, record the ambiguity as **structured data** and let it ride the channel that is already gated:

- In `mark_dump_arg(scalar)`, when `match_count > 1`, record the selected scalar arg **index** (or an ambiguous bitmask) on the `Arg` instead of a free-text `warning_msg`.
- Ship it alongside the dump mask inside the gated block (next to `set_dump_tensor_task_mask`).
- Emit it in `tensor_dump.json` (host writer in `src/common/platform/shared/host/tensor_dump_collector.cpp`) as a per-task / per-scalar field, e.g. `"dump_ambiguous_scalar_idx": 2`.

Benefits: zero cost on the non-dump path (rides the already-gated dump channel), a deterministic index for the consumer instead of a string, reaches exactly the audience that enabled dump, and correctly stays silent when dump is off.

If we want console immediacy too, keep a single `LOG_WARN` — but **inside** the `#if PTO2_PROFILING` gate, per point 1.

zmnobug:
@ChaoZheng109 addressed in `169485dd`.

- Removed the console warning path entirely: no more `log_arg_warning_once`,
  `warning_msg_`, or `logged_arg_warning` state in either a2a3 or a5. This
  keeps the always-on submit path free of dump warning work.
- The dump metadata propagation now stays on the existing gated path: inside
  `#if PTO2_PROFILING` and under `is_dump_tensor_enabled()`, next to the
  existing dump mask propagation.
- Replaced the free-text warning with structured metadata.
  `Arg::mark_dump_arg(scalar)` records duplicate scalar-lvalue ambiguity in
  `dump_arg_index_ambiguous_mask_`; `set_dump_tensor_task_mask(task_id, mask,
  flags)` ships it with the dump mask; the host JSON emits
  `arg_index_ambiguous: true` on the selected scalar entry.
- Applied the same changes to both a2a3 and a5 copies, and kept scalar records
  emitted once per task outside the subtask loop.
- Also moved scalar `value` earlier in the JSON entry so it appears next to the
  kind/dtype fields.

Validation:
- `g++ -fsyntax-only` for a2a3/a5 `pto_orchestrator.cpp`
- `g++ -fsyntax-only` for a2a3/a5 `tensor_dump_aicpu.cpp`
- `python -m py_compile tests/st/a2a3/tensormap_and_ringbuffer/dfx/tensor_dump/test_tensor_dump.py`
- `conda run -n zm_pypto pre-commit run --from-ref upstream/main --to-ref HEAD`

src/a2a3/runtime/tensormap_and_ringbuffer/runtime/pto_types.h
```diff
@@ -38,6 +41,10 @@
 #include "tensor.h"
 #include "tensor_arg.h"
 
+#ifndef PTO2_PROFILING
+#define PTO2_PROFILING 1
```
ChaoZheng109, line 45: 直接声明src/a2a3/runtime/tensormap_and_ringbuffer/runtime/pto_runtime2_types.h可以吗

zmnobug: 不建议在 pto_types.h 里直接 include pto_runtime2_types.h。

  原因是现在依赖方向已经是反过来的：

  // pto_runtime2_types.h
  #include "pto_types.h"

  如果 pto_types.h 再 include pto_runtime2_types.h，会形成循环 include。因为 include guard，结果很容易变成 pto_runtime2_types.h 看到一个“还没真正定义完 Arg/TaskOutputTensors 的 pto_types.h”，后面类型解析会变脆。

  而且 pto_types.h 设计上是更轻的公共 Arg header，orchestration 侧也直接 include 它。让它依赖 pto_runtime2_types.h 会把 runtime 内部的大量类型/常量带进 orchestration header，耦合会变重。

  更合适的做法：

  #ifndef PTO2_PROFILING
  #define PTO2_PROFILING 1
  #endif

  这两行可以保留在 pto_types.h。因为 pto_types.h 自己用了 #if PTO2_PROFILING，它需要在本 header 内有默认值，不能假设 include 方一定先 include 了 pto_runtime2_types.h。

  如果觉得重复，可以后续抽一个很小的公共配置头，比如 pto_profiling_config.h，只放这些 profiling macro 默认值，然后 pto_types.h 和 pto_runtime2_types.h 都 include 它。这个比让 pto_types.h include pto_runtime2_types.h 干净。

ChaoZheng109 (approved):

PR #1031, opened 2026-06-11
Fix: monotonic scope_stats heap accounting for multi-wrap scopes

### LINKED ISSUE #996: [Code Health] scope_stats heap-ring wrap: scope_high_water / scope_alloc miscompute when heap_top < heap_tail
### Category

### Component

Other — DFX / `scope_stats` (heap ring usage accounting)

### Description

`scope_stats` snapshots each ring's head/tail at every `PTO2_SCOPE` boundary and reports usage
as `end - start` deltas. This is correct for the **task_window** ring (`task_head`/`task_tail`
are monotonic sequence numbers, so `head >= tail` always), but not for the **heap** ring, whose
`heap_top_`/`heap_tail_` are **wrapping byte offsets** in `[0, heap_size)`. Two consequences:

1. The on-device contract comment ("`heap_end - heap_start` is bytes in use") is wrong once the
   heap has wrapped (`heap_end < heap_start`), so any consumer that subtracts the raw fields gets
   a garbage/negative value. The Python plotter compensates with a single-fold wrap correction,
   which only happens to be enough for instantaneous occupancy (always `< capacity`).

2. The span/cumulative metrics are not recoverable from two wrapped snapshots. `scope_alloc` and
   `scope_high_water` are not bounded by capacity: a scope whose cumulative heap throughput
   exceeds `heap_size` wraps more than once, and the wrapped end/begin offsets can no longer
   reconstruct the true value. (Backpressure bounds *instantaneous* occupancy, not per-scope
   throughput.) Separately, `scope_high_water = end.top - begin.tail` is not even a true peak in
   the no-wrap case — it is the total address span touched, an upper bound, not the realized peak.

### Location

- `src/common/platform/include/common/scope_stats.h:72-89` — contract comment + raw heap fields
- `src/a2a3/runtime/tensormap_and_ringbuffer/runtime/pto_ring_buffer.h:183-206` — monotonic task
  head/tail vs wrapping heap top/tail
- `src/a2a3/runtime/tensormap_and_ringbuffer/runtime/pto_orchestrator.cpp:443-448,466-471` —
  capture site
- `simpler_setup/tools/scope_stats_plot.py:65-67,88-94` — metrics + single-fold wrap correction
- a5 mirror: `src/a5/runtime/tensormap_and_ringbuffer/runtime/{pto_ring_buffer.h,pto_orchestrator.cpp}`

### Proposed Fix

Record **monotonic (non-wrapping) heap accounting** at the scope boundary instead of the raw
wrapping `heap_top_`/`heap_tail_`, so every metric becomes an exact subtraction and the Python
wrap correction can be removed. Relabel `scope_high_water` to reflect what it actually is (an
upper bound on occupancy, not an observed peak). The task_window ring needs no change. Keep a5
in sync and add a regression test for a scope that wraps the heap more than once.

### Priority

---

Tracked under #995. Distinct from #991 (HTML readability / dep_pool channel) and #902 (per-task
granularity).

doraemonmj:
heap_start/heap_end were wrapping byte offsets in [0, heap_cap), so subtracting two boundary snapshots lost every wrap in between — per-scope alloc / high-water metrics were wrong once a scope's heap throughput exceeded heap_cap, and unrecoverable from two wrapped samples.

The allocator is the only place each individual wrap is observable (each allocation is < heap_cap), so it now reports every wrap to the collector via scope_stats_note_heap_wrap(side). The collector accumulates per-ring wrap counts, attributes each wrap to the current scope's ring (derived from depth, so the runtime stores no scope_stats state), and unrolls the sampled offsets into monotonic cumulative-byte values. Every host-side delta is now an exact, non-negative subtraction for any wrap count.

A debug-only invariant assert (unrolled occupancy == modular occupancy) guards against wrap-count drift; it is skipped for rings that report no wraps, so a5's pre-migration path is unaffected.

Applied to both a2a3 and a5. dep_pool and task_window are already monotonic linear counters and need no change. Docs updated for the new non-wrapping semantics.

fixes: #996

### DIFF
--- a/docs/dfx/scope-stats.md
+++ b/docs/dfx/scope-stats.md
@@ -92,16 +92,18 @@ and common patterns.
 
 | Metric | Formula | What it tells you |
 | ------ | ------- | ----------------- |
-| `scope_high_water` | `end.head − begin.tail` | Highest occupancy the scope held — residual not yet released on entry, plus what this scope added. The true peak. |
-| `real_occupancy` | `end.head − end.tail` | What is still occupied at scope exit (the live level on leaving). |
-| `scope_alloc` | `end.head − begin.head` | How far the allocation frontier advanced — this scope's own net (unreleased) allocation. |
+| `scope_high_water` | `end.head − begin.tail` | Upper bound on the occupancy this scope reached — entry backlog plus everything this scope allocated, *without* subtracting what was released mid-scope. Not a realized peak, and not bounded by capacity: a scope that streams more than `cap` reports more than `cap`. |
+| `real_occupancy` | `end.head − end.tail` | What is still occupied at scope exit (the live level on leaving). Always in `[0, cap]`. |
+| `scope_alloc` | `end.head − begin.head` | How far the allocation frontier advanced over the scope — this scope's total allocation throughput (can exceed `cap`). |
 
 For `task_window`, `heap`, and `dep_pool`, `head` / `tail` mean that
 resource's allocation frontier and released boundary at the sampled scope
-boundary. Ring-buffer occupancy is non-negative, so a negative delta is folded
-back by one buffer length (`(head + cap − tail) % cap`) — a wrap past the
-buffer end. `tensormap` is reported as a single in-use value (no head/tail), so
-it shows only the `real_occupancy` curve.
+boundary. All three are reported as **monotonic, non-wrapping** counters
+(byte totals for `heap`, entry counters for the others), so every delta above
+is an exact, non-negative subtraction regardless of how many times the
+underlying ring wrapped during the scope — no wrap correction is applied.
+`tensormap` is reported as a single in-use value (no head/tail), so it shows
+only the `real_occupancy` curve.
 
 ## 2. What gets captured
 
@@ -127,10 +129,13 @@ it shows only the `real_occupancy` curve.
 ## 3. Output: `scope_stats.jsonl`
 
 NDJSON. Line 1 is run metadata; each subsequent line is one scope
-sample (`begin` or `end`). Schema version 5:
+sample (`begin` or `end`). Schema version 6 (bumped from 5 when
+`heap_start`/`heap_end` changed from wrapping ring offsets to monotonic
+cumulative bytes — a `version == 5` file's heap fields wrap, a `version == 6`
+file's do not):
 
 ```json
-{"version": 5, "fatal": false, "dropped": 0, "total": 4, "task_window_max": [8, 4], "heap_max": [268435456, 268435456], "dep_pool_max": [1024, 1024], "tensormap_max": 65536}
+{"version": 6, "fatal": false, "dropped": 0, "total": 4, "task_window_max": [8, 4], "heap_max": [268435456, 268435456], "dep_pool_max": [1024, 1024], "tensormap_max": 65536}
 {"site": "example_orchestration.cpp:77", "phase": "begin", "depth": 1, "ring": 1, "task_window_start": 0, "task_window_end": 0, "heap_start": 0, "heap_end": 0, "dep_pool_start": 1, "dep_pool_end": 1, "tensormap": 0}
 {"site": "example_orchestration.cpp:77", "phase": "end", "depth": 1, "ring": 1, "task_window_start": 0, "task_window_end": 4, "heap_start": 0, "heap_end": 8192, "dep_pool_start": 1, "dep_pool_end": 6, "tensormap": 5}
 ```
@@ -139,7 +144,7 @@ Metadata line (line 1):
 
 | Field | Type | Meaning |
 | ----- | ---- | ------- |
-| `version` | int | Schema version (`5`) |
+| `version` | int | Schema version (`6`) |
 | `fatal` | bool | `true` iff a fatal was latched during the run; records past it are diagnostic-only |
 | `dropped` | uint | Records dropped on device (free_queue empty / ready_queue full); `0` on a healthy run |
 | `total` | uint | Total records the device attempted (collected + dropped) |
@@ -158,8 +163,8 @@ Per-sample lines, oldest-first:
 | `ring` | int | Ring this scope used; indexes `task_window_max` / `heap_max` / `dep_pool_max` |
 | `task_window_start` | int | Task-window ring tail at this boundary |
 | `task_window_end` | int | Task-window ring head at this boundary |
-| `heap_start` | uint | Heap ring tail (released boundary) in bytes |
-| `heap_end` | uint | Heap ring head (allocation frontier) in bytes |
+| `heap_start` | uint | Heap reclaim boundary — monotonic cumulative bytes reclaimed (non-wrapping) |
+| `heap_end` | uint | Heap allocation frontier — monotonic cumulative bytes allocated (non-wrapping) |
 | `dep_pool_start` | int | Scheduler-published dependency-list pool tail at this boundary |
 | `dep_pool_end` | int | Scheduler-published dependency-list pool top at this boundary |
 | `tensormap` | int | Tensormap entries in use |
--- a/src/common/platform/include/common/scope_stats.h
+++ b/src/common/platform/include/common/scope_stats.h
@@ -83,8 +83,8 @@ struct ScopeStatsRecord {
                                   // human-readable path without dereferencing a
                                   // device pointer (the string table lives in the
                                   // orchestration .so, not in shared memory).
-    uint64_t heap_start;          // Heap ring reclaim pointer (heap_tail_).
-    uint64_t heap_end;            // Heap ring allocation pointer (heap_top_).
+    uint64_t heap_start;          // Heap reclaim pointer, unrolled to monotonic cumulative bytes.
+    uint64_t heap_end;            // Heap allocation pointer, unrolled to monotonic cumulative bytes.
     int32_t site_line;
     int32_t task_start;      // Task ring tail (last_task_alive).
     int32_t task_end;        // Task ring head (next task id).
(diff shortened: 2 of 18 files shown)


ChaoZheng109 (commented):
对照 issue #996 的 Proposed Fix 复核本 PR。

**核心修复正确**:设备侧改为单调(非回绕)堆记账,wrap 归属(ring==scope 深度、scope 内恒定、与 collector depth 锁步)与时机(kernel 入口先启用、堆指针从 0 起算)均成立,a5 也已同步。未发现正确性缺陷。

**相对 #996 完整诉求仍有缺口**(逐条标在对应位置):
1. 回归测试:#996 明确要求"a scope that wraps the heap more than once",本 PR 未加 → 见 test_stubs.cpp 评论。
2. 删除 plotter 的单折 wrap 修正 + scope_high_water 重新措辞:只改了 docs,plotter 未同步 → 见 scope-stats.md 评论。
3. 过时注释(a5 已迁移却写"still feeds raw offsets") → 见 collector 评论。
4. heap_start/heap_end 语义变更但 schema version 未 bump → 见 scope_stats.h 评论。

src/common/platform/shared/aicpu/scope_stats_collector_aicpu.cpp
```diff
@@ -164,6 +174,44 @@ void switch_buffer() {
...
+    if (s_scope_stats_header == nullptr) return offset;
+    if (ring_id < 0 || ring_id >= PTO2_SCOPE_STATS_MAX_RING_DEPTH) return offset;
+    return offset + s_heap_wraps[ring_id][side] * s_scope_stats_header->heap_cap[ring_id];
+}
+
+// Debug-only: the unrolled occupancy (mono_end - mono_start) must equal the true
+// modular occupancy derived from the raw wrapping offsets. A drift means a wrap
+// was missed or misattributed (scope_stats_note_heap_wrap), which always shifts
+// the result by a multiple of heap_cap. Guarded on wraps>0 so it never fires for
+// an arch/ring that doesn't report wraps (e.g. a5 today still feeds raw wrapping
```
ChaoZheng109: **过时注释 (doc-consistency)**:这里写 `(e.g. a5 today still feeds raw wrapping offsets)`,但本 PR 已让 a5 的 AICPU kernel 链接公共 collector(a5 `compile_commands.json` 把 `scope_stats_collector_aicpu.cpp` 编进 `aicpu_kernel.dir`),且 a5 的 `pto_ring_buffer.h` 也在 `#if PTO2_PROFILING` 下调用了 `scope_stats_note_heap_wrap` —— a5 实际已迁移、会上报并展开,不再 feed raw offsets。这与 commit message "Applied to both a2a3 and a5" 矛盾。

`wraps>0` 这个 guard 本身有价值(保护任何尚未 wrap 过的 ring),建议把 a5 的举例删掉或改成"例如某 scope 吞吐还没到一次 wrap 的 ring"。

tests/ut/cpp/stubs/test_stubs.cpp
```diff
@@ -67,6 +67,20 @@ void unified_log_info_v(const char *func, int v, const char *fmt, ...) {
...
+//
+// pto_ring_buffer.h (PTO2_PROFILING build) calls these from the inlined
+// allocator paths to report heap-ring wraps. Unit tests of the allocator do not
+// link the platform collector, so provide weak no-op fallbacks here: the real
+// collector's strong definitions win whenever it is linked, and the allocator
+// test stays self-contained (scope_stats is disabled in tests, so the gated
+// calls never fire anyway).
+
+extern "C" __attribute__((weak)) bool is_scope_stats_enabled() { return false; }
+extern "C" __attribute__((weak)) void scope_stats_note_heap_wrap(int /* side */) {}
```
ChaoZheng109: **缺少回归测试(本 PR 相对 #996 最关键的缺口)**。#996 的 Proposed Fix 明确要求 "add a regression test for a scope that wraps the heap more than once";本 PR 只加了让 allocator UT 能链接的 weak 桩,没有任何真正触发**多圈 wrap** 的用例。

现有 ST smoke `test_scope_stats.py` 用 `vector_example`(64KB,远不到一圈)且只校验字段存在/未 drop,**不校验 `heap_start/heap_end` 数值** → 修复前后都会通过,不覆盖被修的 bug。

建议补:`a2a3_rt_objs` 已链接真实 collector(见 `tests/ut/cpp/CMakeLists.txt`),可直接写一个 C++ UT:`set_scope_stats_enabled(true)` + `set_platform_scope_stats_base(...)` 后灌入若干 `scope_stats_note_heap_wrap` 与带 wrap 偏移的 begin/end 快照,断言展开后 `scope_alloc`/`high_water` delta 精确(且可 > cap)。

src/common/platform/include/common/scope_stats.h
```diff
@@ -83,8 +83,8 @@ struct ScopeStatsRecord {
                                   // human-readable path without dereferencing a
                                   // device pointer (the string table lives in the
                                   // orchestration .so, not in shared memory).
-    uint64_t heap_start;          // Heap ring reclaim pointer (heap_tail_).
-    uint64_t heap_end;            // Heap ring allocation pointer (heap_top_).
+    uint64_t heap_start;          // Heap reclaim pointer, unrolled to monotonic cumulative bytes.
+    uint64_t heap_end;            // Heap allocation pointer, unrolled to monotonic cumulative bytes.
```
ChaoZheng109, line 87: **稳定边界 / 版本故事**:`heap_start`/`heap_end` 的语义从"环形偏移"变成了"单调累计字节",但 scope_stats jsonl 的 schema version 未 bump(ST 测试里仍断言 `version == 5`)。

同一 `version=5` 下,旧文件(回绕偏移)与新文件(单调字节)对同一字段语义不同,下游消费者(plotter / 任何手算 `end-start` 的人)无法区分。建议同步 bump schema version,并在 `docs/dfx/scope-stats.md` 注明版本切换点。

docs/dfx/scope-stats.md
```diff
@@ -92,16 +92,18 @@ and common patterns.
...
 For `task_window`, `heap`, and `dep_pool`, `head` / `tail` mean that
 resource's allocation frontier and released boundary at the sampled scope
-boundary. Ring-buffer occupancy is non-negative, so a negative delta is folded
-back by one buffer length (`(head + cap − tail) % cap`) — a wrap past the
-buffer end. `tensormap` is reported as a single in-use value (no head/tail), so
-it shows only the `real_occupancy` curve.
+boundary. All three are reported as **monotonic, non-wrapping** counters
+(byte totals for `heap`, entry counters for the others), so every delta above
+is an exact, non-negative subtraction regardless of how many times the
+underlying ring wrapped during the scope — no wrap correction is applied.
```
ChaoZheng109, line 104: **文档/代码不一致**:这里(及上方 `scope_high_water` 改为"上界非峰值")只改了文档,但实际的 plotter `simpler_setup/tools/scope_stats_plot.py` 未同步:
1. 第 118 行 `_wrap()` 单折修正仍在。对单调值它是无害空操作(三个 metric 的 delta 恒非负,不会触发折叠),但 #996 明确要求删除它,且此处文档已声称 "no wrap correction is applied" —— doc 与 code 互相矛盾。
2. 第 75 行 `scope_high_water` 的描述仍是 "Peak pressure seen by this scope",未按本次重新措辞改为"占用上界"。

建议在本 PR 内一并更新 plotter,保持文档/代码一致。

src/common/platform/shared/aicpu/scope_stats_collector_aicpu.cpp
```diff
@@ -164,6 +174,44 @@ void switch_buffer() {
...
+    if (cap == 0) return;
+    uint64_t wraps_alloc = s_heap_wraps[ring_id][SCOPE_STATS_HEAP_SIDE_ALLOC];
+    uint64_t wraps_reclaim = s_heap_wraps[ring_id][SCOPE_STATS_HEAP_SIDE_RECLAIM];
+    if (wraps_alloc == 0 && wraps_reclaim == 0) return;  // wrap accounting inactive
+    uint64_t mono_occ = (raw_end + wraps_alloc * cap) - (raw_start + wraps_reclaim * cap);
+    uint64_t modular_occ = (raw_end + cap - raw_start) % cap;
+    // When the raw offsets coincide, modular_occ folds to 0 but the ring is
+    // either empty (occ 0) or exactly full (occ cap) — indistinguishable from
+    // the offsets alone, so accept both. Any other mismatch is a real drift.
+    assert((mono_occ == modular_occ || (modular_occ == 0 && mono_occ == cap)) && "scope_stats heap wrap-count drift");
```
ChaoZheng109: Suggest removing `debug_assert_heap_invariant` (and this call site).

The unroll correctness is already pinned down by the new UT —
`test_scope_stats_collector.cpp` asserts the exact monotonic
`heap_start`/`heap_end` for the multi-wrap and full-ring cases — so this
in-situ invariant is redundant for catching wrap-count drift.

It also carries a device-side risk the comment above doesn't reflect. The
comment says "Compiled out under NDEBUG", but the AICPU onboard build
(`src/{arch}/platform/onboard/aicpu/CMakeLists.txt`) sets no `CMAKE_BUILD_TYPE`
and passes `-O3 -g` with no `-DNDEBUG`, so `NDEBUG` is *not* defined there — the
`assert()` is compiled into the device `.so`, and a drift would `abort()` on the
AICPU. That's too aggressive for an opt-in DFX path and inconsistent with this
collector's own degrade-and-continue style (`LOG_ERROR` + drop at the flush
site, ~line 314).

Since the UT covers the invariant, dropping the assert (rather than converting
it to `LOG_ERROR`) keeps the hot path clean. If you'd rather keep a runtime
guard, a `LOG_ERROR` soft-fail would match the rest of the file — but it isn't
needed for coverage.

doraemonmj: 已删除

src/common/platform/include/common/scope_stats.h
doraemonmj: 已修改

docs/dfx/scope-stats.md
doraemonmj: _wrap()已删除，注释已修改

ChaoZheng109 (approved):

Issue #1045, opened 2026-06-12
[Performance] Wrong AICPU affinity assignment causes significant performance decrease


noabauma:
### Platform

a2a3 (Ascend 910B/C hardware)

### Runtime Variant

tensormap_and_ringbuffer

### Summary

Simpler is occasionally choosing AICPU threads across NUMA (AICPU Package) domains to work as scheduler and orchestrator threads. This severely affects AICPU performance during the execution of a kernel (up to 7%, given our measurements). This is in line with what we have recently documented in our recent [Report](https://jx.huawei.com/redirect/qe8vkejFB94) (Huawei-employee only), where we concluded that having the scheduler and orchestrator, all in the same NUMA domain, is a necessary condition for scheduling performance.

The cause is the `platform_aicpu_affinity_gate()` function (see: https://github.com/hw-native-sys/simpler/blob/2668be5a7627c2ae56ecb0012ec2be1181dc9455/src/common/platform/onboard/aicpu/platform_aicpu_affinity.cpp#L39) whose logic does not seem to constrain affinity assignment to a single AICPU package.

To solve this issue, one should make `platform_aicpu_affinity_gate()` enforce that all threads lie in the same NUMA domain — ideally making placement 100% deterministic, except when `runtime->aicpu_thread_num` exceeds the size of a single NUMA domain.

A pull request with a proposed solution can be found here: #1046 

### Git Commit ID

196bcc677b759f8298900871059af5a64fc846fb

### CANN Version

9.0.0

### Driver Version

26.0.rc1

### Host Platform

Linux (aarch64)

### Reproduction

```bash
Let Claude run src/simpler/.claude/skills/benchmark/SKILL.md /benchmark
```

You can use the following snippet to check the assignment is correct or not:

```C++

    #ifdef __linux__
    #include <sched.h>
   #endif

    // Get current CPU ID
    int cpu = sched_getcpu();

    LOG_INFO_V0("Thread[%d] is mapped to cpu %d", cpu);
```

Then check that all cores are set to AICPU package 1 (assigned to 4-7)

### Expected Performance

Example                                          Base (main)   HEAD (us)   Delta (us)   Change (%)
-------------------------------               ---------      ---------   ----------   ----------
benchmark_bgemm (Case0)                    665.3         619.3           -46.0        **-6.92% ✓**
  (host)                                               107086.4    113684.5      +6598.1        +6.16%
  (device)                                                2230.5        1957.1         -273.4       -12.26%
[...]

PR #1046, opened 2026-06-12
[Fix] [Performance] Fix AICPU thread affinity assignment to fit in a single AICPU package (a.k.a NUMA domain)

noabauma:
This PR proposes a way to fix issue #1045

This algorithm gives a 100% guarantee of selecting the right threads to serve as the scheduler and orchestrator by choosing those in the same NUMA domain. As described in the associated issue, this fix improved average AICPU scheduling/orchestration performance by up to 7% on the tests we performed. This was only tested in A3.


ChaoWao:
Duplicate #1138 .

PR #1081, opened 2026-06-17
Fix: atomic acquire/release for sim register handshake gate (a2a3 + a5)

ChaoZheng109:
## Problem

The nightly **tsan / a2a3sim** sanitizer job intermittently fails on
`test_dedup_shared_so_independent_unregister` with a scheduler stall:

```
[STALL thread=N] TIMEOUT_EXIT after_idle_iterations=...
PTO2 runtime failed: orch_error_code=0 sched_error_code=100 runtime_status=-100
RuntimeError: run_prepared failed with code -100
```

Root cause is **not** a too-small timeout and is unrelated to the
sync_start drain fix (#1074). The AICPU<->AICore handshake registers
(`COND`, `DATA_MAIN_BASE`) are accessed in the **sim** platform as plain
`volatile` loads/stores ordered only by standalone fences
(`OUT_OF_ORDER_*_BARRIER` / `rmb` / `wmb`). In sim those "registers" are
plain host memory shared across the AICore and AICPU host threads, so a
bare fence beside a non-atomic access establishes no inter-thread
happens-before in the C++ memory model — ThreadSanitizer correctly flags
every COND/dispatch access as a data race, and the unsynchronized COND
poll can miss the AICore's FIN publish, leaving a task stuck `RUNNING`
(`cond_reg_state=ack`) until the idle watchdog fires.

## Fix

Make the register cell itself the synchronizing object on the sim path:

- **AICore sim** `read_reg`/`write_reg` → `__atomic_load_n(ACQUIRE)` /
  `__atomic_store_n(RELEASE)` (sim-only `inner_kernel.h`).
- **AICPU** side: add `reg_load_acquire` / `reg_store_release` in
  `platform_regs.h`, gated on `__CPU_SIM` (now defined on the sim aicpu
  target). `write_reg`, the shared `read_reg` (a2a3) /
  `inner_platform_regs.cpp` (a5), and the hot-path `*core.cond_ptr`
  completion poll route through them.

**Onboard is unchanged**: `__CPU_SIM` is undefined there, so the
accessors stay plain Device-MMIO `volatile` + the existing explicit
`rmb()`/`wmb()` (atomics on Device-nGnRnE memory are not valid). The
atomic release/acquire subsumes the now-removed per-access fences on the
sim path only.

## Validation

Built with `SIMPLER_SANITIZER=tsan` and ran the previously-failing test
**40x on a2a3sim under TSAN**:

- 40/40 pass, **0 stalls**, **0 ThreadSanitizer data races**
  (previously dozens at these exact lines: `aicore_executor.cpp`,
  `handshake_all_cores`, `check_running_cores_for_completion`,
  `platform_regs`).
- The built `libaicpu_kernel.so` now contains `__tsan_atomic_*` calls,
  confirming the accesses became instrumented atomics.

## Notes

- Independent of #1074; touches only `src/{a2a3,a5}/platform/sim` +
  the sim register-cell accessor and one completion-poll line.
- Startup/teardown races on `bind_runtime` / `on_orchestration_done`
  seen in the original failing log are a separate pre-existing issue
  (scheduler init/teardown, not the register handshake) and are not
  addressed here.

### DIFF
--- a/src/a2a3/platform/include/aicpu/platform_regs.h
+++ b/src/a2a3/platform/include/aicpu/platform_regs.h
@@ -84,6 +84,20 @@ uint64_t get_platform_pmu_reg_addrs();
 volatile uint32_t *get_reg_ptr(uint64_t reg_base_addr, RegId reg);
 
 /**
+ * Register-cell synchronizing accessors for the AICPU<->AICore handshake gate
+ * (COND / DATA_MAIN_BASE). Implemented per-variant:
+ *   sim/aicpu/inner_platform_regs.cpp     -- atomic acquire/release. The
+ *       simulated registers are plain host memory shared across host threads,
+ *       so the access itself must carry inter-thread happens-before (and be
+ *       visible to ThreadSanitizer) against the AICore side.
+ *   onboard/aicpu/inner_platform_regs.cpp -- plain Device-nGnRnE MMIO load/store
+ *       (atomics are not valid on Device memory); ordering is the caller's
+ *       explicit rmb()/wmb(), so the hardware path is the legacy behavior.
+ */
+uint32_t reg_load_acquire(const volatile uint32_t *p);
+void reg_store_release(volatile uint32_t *p, uint32_t v);
+
+/**
  * Read a register value from an AICore's register block
  *
  * No memory barrier is emitted. Callers that read a hand-off bit
@@ -109,7 +123,9 @@ uint64_t read_reg(uint64_t reg_base_addr, RegId reg);
  * @param value          Value to write (truncated to register width)
  */
 inline void write_reg(uint64_t reg_base_addr, RegId reg, uint64_t value) {
-    *reinterpret_cast<volatile uint32_t *>(reg_base_addr + reg_offset(reg)) = static_cast<uint32_t>(value);
+    reg_store_release(
+        reinterpret_cast<volatile uint32_t *>(reg_base_addr + reg_offset(reg)), static_cast<uint32_t>(value)
+    );
 }
 
 /**
--- a/src/a2a3/platform/onboard/aicpu/inner_platform_regs.cpp
+++ b/src/a2a3/platform/onboard/aicpu/inner_platform_regs.cpp
@@ -13,15 +13,22 @@
  * @brief Variant-specific platform_regs hooks for real hardware (a2a3)
  *
  * a2a3 sim and onboard share the same register layout, so read_reg /
- * write_reg live in the shared src/aicpu/platform_regs.cpp. This file
- * exists only for the deinit-timeout split where onboard keeps the
- * legacy 1 s budget while sim widens it — see platform_regs.h for the
- * rationale.
+ * write_reg live in the shared src/aicpu/platform_regs.cpp. This file holds
+ * the variant-specific hooks: the reg_load_acquire / reg_store_release
+ * handshake-gate accessors (plain MMIO here) and the deinit-timeout budget —
+ * see platform_regs.h for the rationale of each.
  */
 
 #include <cstdint>
 #include "aicpu/platform_regs.h"
 #include "common/platform_config.h"
+
+// Plain Device-nGnRnE MMIO load/store (atomics are not valid on Device
+// memory); cross Device<->Normal ordering is the caller's rmb()/wmb(). See
+// platform_regs.h.
+uint32_t reg_load_acquire(const volatile uint32_t *p) { return *p; }
+
+void reg_store_release(volatile uint32_t *p, uint32_t v) { *p = v; }
 
 /**
  * @brief Deinit ACK-wait budget on real hardware: 1 s.
--- a/src/a2a3/platform/shared/aicpu/platform_regs.cpp
+++ b/src/a2a3/platform/shared/aicpu/platform_regs.cpp
@@ -50,7 +50,9 @@ volatile uint32_t *get_reg_ptr(uint64_t reg_base_addr, RegId reg) {
     return reinterpret_cast<volatile uint32_t *>(reg_base_addr + reg_offset(reg));
 }
 
-uint64_t read_reg(uint64_t reg_base_addr, RegId reg) { return static_cast<uint64_t>(*get_reg_ptr(reg_base_addr, reg)); }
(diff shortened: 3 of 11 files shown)


ChaoZheng109:
### Sanitizer validation (temporary `pull_request` trigger, now removed)

The COND/dispatch data race only reproduces under the CI runner's 4-vCPU
oversubscription — a high-core local box never trips the window — so this was
validated by temporarily enabling the Sanitizers workflow on this PR. That
temp trigger has been dropped; the branch is back to just the fix.

**A/B in x86 CI (`tsan, a2a3sim`):**

| | main (pre-fix) | this PR (fixed) |
| --- | --- | --- |
| ThreadSanitizer data races | 28 (5 at the COND/dispatch handshake lines) | **0** |
| stall (`TIMEOUT_EXIT` / `run_prepared -> -100`) | intermittently present | **none** |
| test result | flaky (nightly red) | 2 passed |

**Flakiness probe — Sanitizers re-run 5x on this PR:**

| run | tsan/a2a3sim | tsan/a5sim |
| --- | --- | --- |
| 1 | ✅ 0 races | ✅ 0 races |
| 2 | ✅ 0 races | ✅ 0 races |
| 3 | ✅ 0 races | ✅ 0 races |
| 4 | ✅ 0 races | ✅ 0 races |
| 5 | ✅ 0 races | ✅ 0 races |

All four sanitizer jobs (`asan`/`tsan` × `a2a3sim`/`a5sim`) green every time.
Workflow run: https://github.com/hw-native-sys/simpler/actions/runs/27681815742

This is consistent by construction: once the register cell is an atomic
release/acquire pair on the same address, TSAN cannot report a race between the
AICore store and the AICPU load, and the missed-FIN stall (the downstream
failure mode) cannot occur.

ChaoZheng109:
### Update: implementation refactored to the per-variant split

Following review, `reg_load_acquire` / `reg_store_release` no longer use
`#if defined(__CPU_SIM)` macro isolation in the shared header. They now follow
the established pattern in this tree — interface **declared** in
`aicpu/platform_regs.h`, **implemented per-variant**:

- `sim/aicpu/inner_platform_regs.cpp` → `__atomic_load_n/store_n` (acquire/release)
- `onboard/aicpu/inner_platform_regs.cpp` → plain Device-MMIO load/store

(the same sim/onboard split already used for `read_reg`/`write_reg`). The
`__CPU_SIM` compile-define gate is removed entirely, so the final diff carries
**no build-system change and no macro** — 11 files, platform/runtime only.

Re-validated on x86 CI with the refactored code (sanitizers temporarily enabled
on the PR, now removed):

| | tsan/a2a3sim | tsan/a5sim |
| --- | --- | --- |
| data races | **0** | **0** |
| result | 2 passed | 1 passed |

All four sanitizer jobs green. Run:
https://github.com/hw-native-sys/simpler/actions/runs/27685708946

ChaoZheng109:
## Onboard benchmark: no measurable performance impact (within noise)

Since this is a **sim-only** change (the onboard accessor `reg_load_acquire`/`reg_store_release` compiles to the same plain `volatile` Device-MMIO load/store — see `onboard/aicpu/inner_platform_regs.cpp`), the only onboard-observable diff is the completion-poll routing through a possibly-non-inlined wrapper. Benchmarked to confirm it doesn't regress the scheduler hot path.

**Setup:** a2a3, `tensormap_and_ringbuffer`, rebased onto current `main`. Baseline `a756969c` (pre-fix) vs HEAD (with fix), **same device (NPU:3), sequential**, 100 rounds each, via `task-submit`. Both runs exit 0; qwen3 stress case completed with no stall.

**Per-window breakdown, Base → HEAD (us), with Δ%.** The changed line lives in the completion poll → **Sched** window; **Orch** (graph-build) does *not* run the poll and is byte-identical between builds — it's the control column.

| example / case | Orch (base→HEAD) | Orch Δ% | Sched (base→HEAD) | Sched Δ% | Effective Δ% |
| --- | ---: | ---: | ---: | ---: | ---: |
| qwen3_14b_decode / StressBatch16Seq3500 | 899.4 → 892.1 | −0.81% | 2144.9 → 2136.3 | −0.40% | −0.40% |
| batch_paged_attention / Case1 | 3807.1 → 3814.9 | +0.20% | 4530.3 → 4514.8 | −0.34% | −0.34% |
| paged_attention_unroll / Case1 | 1110.1 → 1106.6 | −0.32% | 1288.2 → 1287.0 | −0.09% | −0.10% |
| alternating_matmul_add / Case1 | 933.7 → 925.6 | −0.87% | 933.8 → 925.7 | −0.87% | −0.87% |
| paged_attention_unroll / Case2 | 494.8 → 500.5 | +1.15% | 666.0 → 672.2 | +0.93% | +0.93% |
| paged_attention_unroll_manual_scope / Case2 | 449.5 → 461.4 | +2.65% | 634.7 → 638.7 | +0.63% | +0.63% |
| paged_attention_unroll_manual_scope / Case1 | 999.5 → 1038.9 | +3.94% | 1206.8 → 1243.4 | +3.03% | +3.05% |
| benchmark_bgemm / Case0 | 856.4 → 932.4 | +8.87% | 900.4 → 974.2 | +8.20% | +8.19% |

**The two >2% cases are contention noise, not the change — now visible directly in the table:**

1. **Orch and Sched move together by the same magnitude** on the outliers (bgemm: Orch +8.87% / Sched +8.20%; manual_scope C1: Orch +3.94% / Sched +3.03%). Orch's producing code is unchanged between the two builds, so its +8.87% is purely environmental — and Sched riding the same common-mode swing means Sched's move isn't from the wrapper either.
2. The cases that run the poll the *most* (qwen3 stress, batch_paged — long and stable) are **−0.40% / −0.34%** on both Sched and Effective. A real per-poll cost would surface there first, not vanish.
3. Host swung ±50% run-to-run (alternating −57%, bgemm +20%); the box had 3–4 concurrent pypto/qwen NPU jobs during the run, and `bgemm` (~900 us) is the shortest/most jitter-sensitive kernel.

**Verdict:** stable cases sit in the −0.4% … +0.1% noise band on Sched/Effective; the short-kernel scatter is common-mode contention (Orch self-evidently moves with it). Consistent with the change being bit-identical onboard — **no performance regression.**

Issue #900, opened 2026-05-29
[Bug] A2A3 `spmd_paged_attention_highperf` hardware run times out or produces partial zero output while `a2a3sim` passes


MirkoDeVita98:
### Platform

a2a3 (Ascend 910B/C hardware)

### Runtime Variant

tensormap_and_ringbuffer

### Description

`spmd_paged_attention_highperf` does not behave correctly on A2A3 hardware, even though the same test passes on `a2a3sim` simulation. The original hardware run timed out in the runtime. After diagnostic changes that avoided unresolved relocations in the dynamically loaded AICore payload, the kernel completed but produced a golden mismatch: a subset of attention heads were computed accurately, while many heads remained all zero.

This points to a hardware-only dynamic payload loading / SPMD context issue rather than a numerical error in the attention math itself. The relevant loader currently extracts only raw `.text` bytes from AIC/AIV ELF objects and does not apply `.rela.text` relocations. The PA kernel object contains relocations for out-of-line calls and block-local/global symbols, so those references can be invalid when the raw payload is called directly on device.

### Steps to Reproduce

From PR https://github.com/hw-native-sys/simpler/pull/899 :

```bash
cd tests/st/a2a3/tensormap_and_ringbuffer/spmd_paged_attention_highperf
python test_spmd_paged_attention_highperf.py -p a2a3
```

Control comparison:

```bash
python test_spmd_paged_attention_highperf.py -p a2a3sim
```

### Expected Behavior

The A2A3 hardware run should match the golden output within the scene test tolerance, as the `a2a3sim` run does.

### Actual Behavior

The original hardware run failed with an AICPU/runtime timeout similar to:

```text
aclrtSynchronizeStreamWithTimeout (AICPU) failed: 507018
PTO2 runtime failed: orch_error_code=0 sched_error_code=100 runtime_status=-100
```

After diagnostic changes to avoid some payload relocations, the run completed but failed validation:

```text
FAILED: Golden mismatch on 'out': max_diff=0.60693359375, rtol=0.005, atol=0.02
```

A manual device-output dump showed that the device is computing some heads correctly and leaving many heads unwritten/all zero. Example stats from the diagnostic run:

```text
Device shape: (1, 32, 128), dtype: torch.float16
Device nonzero: 1280 / 4096
Golden nonzero: 4096 / 4096
Device min/max: -0.515625 / 0.41748046875
Golden min/max: -0.5966796875 / 0.60693359375
Max diff: 0.60693359375
[...]

PR #1088, opened 2026-06-18
Replace qwen3_14b_decode with a 2-layer (decode_fwd_layers N=2) SceneTestCase

lwDavid:
## What

Replaces the single-layer `qwen3_14b_decode` example with a **fused 2-layer
decode chunk** — pypto-lib `models/qwen3/14b/decode_layer.py` entry
`decode_fwd_layers` with `_CHUNK_NLAYERS == 2` (FP32 inter-layer residual carry,
hidden → hidden, no LM head), harvested from pypto codegen so simpler devs run
it directly (no lib/JIT descent).

**Parameter regime matches `stress_profile.py`** (the vLLM serving stress run):

| Param | Value | Source |
| ----- | ----- | ------ |
| `BATCH` | 16 | `CONCURRENCY` (decode kernel BATCH=16) |
| `MAX_SEQ` | 5500 | `max_model_len` |
| decode `seq_len` | 3500 | the ~3500-token prompt |
| layers | 2 | `decode_fwd_layers` N=2 |

- `examples/a2a3/tensormap_and_ringbuffer/qwen3_14b_decode/` — test + README + 40
  sources (8 AIC + 31 AIV + orchestration); `CALLABLE` from the codegen
  `kernel_config.py`. Weights + paged KV pool stacked ×2 (one slice per layer;
  layers reuse layer-0 weights per the lib const-layer-0 stacked-fwd reference).
- `simpler_setup/goldens/qwen3_14b_decode.py` — 2-layer torch golden + fixture
  (per-layer KV pools, FP32 carry, single bf16 round at the chunk tail).
- `fa_fused_aiv` uses the native PTO-ISA `get_subblockid()` instead of the
  codegen default `[[block_local]]` static (avoids the `.text` relocation the
  AICore loader can't apply — issue #900).

The earlier single-layer `qwen3_14b_decode` and the WIP `qwen3_14b_decode_layer`
are removed; this is the one canonical qwen decode example.

## seq=3500 result

Measured on a2a3 (device): **4/5 runs XPASS** (clean) — finite output lanes match
the torch reference at `max_abs_diff ≈ 0.016`, both layers' KV writes match. The
remaining ~20% hit **one output lane = NaN**, the known non-deterministic
**SceneTestCase-L2 run-path race** (framework-level, independent of layer
count / seq_len; pypto's `execute_compiled` runs the identical compiled artifacts
deterministically clean). The case is `xfail(strict=False)` so CI stays green
whether a run XPASSes or XFAILs, and flips to a hard signal once the framework
run-path is fixed. Details in the example `README.md`.


lwDavid:
Addressed the review feedback and squashed the branch to a single commit.

- **clang-format**: all harvested kernels + orchestration formatted to ≤120 cols.
- **CodeRabbit**: removed the unused `Tensor out = ext_out;` in the orchestration; removed the duplicate `pipe_barrier`/`dcci`/`dsb` cache-flush in `fa_work_build.cpp`; added `seqlen_max` range validation in the golden; added a language to the README fenced block (MD040).
- **Gemini**: replaced the `<n>` device placeholders with `${DEVICE}` in the README commands.
- **pyright**: added `from __future__ import annotations` to the golden (repo targets py3.9, so the `int | None` union needs lazy annotations).

The two generated-C++ cleanups are documented in the example README's provenance section. All review threads resolved.

For context: the `st-onboard-a2a3` failure on the earlier run was an unrelated CI-runner device-poison cascade (`507018` on `dev=11`) that also failed several unrelated tests (`all_to_all_distributed`, `ffn_tp_parallel`, `reduce_scatter_distributed`, `domain_rank_map`, `paged_attention_manual_scope`); this PR touches none of those paths. `Batch16Varied` itself is `xfail(strict=False)` for the isolated SceneTestCase-L2 run-path issue described in the README.

ChaoZheng109 (dismissed):

examples/a2a3/tensormap_and_ringbuffer/qwen3_14b_decode_layer/test_qwen3_14b_decode_layer.py
```diff
@@ -0,0 +1,355 @@
...
+    }
+
+    CASES = [
+        {
+            "name": "Batch16Varied",
+            "platforms": ["a2a3"],
+            # block_dim=0 → auto (DeviceRunner resolves to stream max capacity),
+            # matching the lib RunConfig default for qwen3_decode_mpmd.
+            "config": {"aicpu_thread_num": 4, "block_dim": 0},
+            "params": {"seed": 1234, "full_seq": False},
```
ChaoZheng109: “full_seq": False是什么意思，现在的seqlen实际是多少

lwDavid: `full_seq=False` means the per-lane sequence lengths are **not** pinned to `MAX_SEQ`; instead they are sampled deterministically (from `seed=1234`) as varied random lengths in `[1, MAX_SEQ] = [1, 4096]`. This is the default regime of the lib's `decode_layer.py`, and it exercises multi-block paged attention and the cross-block `online_softmax` reduction. `full_seq=True` instead sets every lane to `MAX_SEQ=4096` (maximum KV-cache load, for a stable full-load perf run).

For the current case (`seed=1234, BATCH=16, full_seq=False`) the actual seq_lens are:

```
[2864, 724, 1319, 3126, 3277, 3445, 665, 2042, 3728, 2738, 280, 1258, 3227, 31, 3756, 1183]
```

i.e. `ceil(seq_len / 128)` = 1–30 paged blocks per lane (`BLOCK_SIZE=128`). The `full_seq` semantics are also documented in the golden's `generate_inputs` docstring.

ChaoZheng109 (approved):

Issue #1022, opened 2026-06-10
[Bug] SPMD Paged Attention Highperf Fails For >= 8192 sequences


MirkoDeVita98:
### Platform

a2a3

### Runtime Variant

tensormap_and_ringbuffer

### Description

The `spmd_paged_attention_highperf` scene test works for shorter decode
sequence lengths, but fails when extended to longer paged-attention sequences.
The failure is reproducible for the 8192-token case added in
[hw-native-sys/simpler#986](https://github.com/hw-native-sys/simpler/pull/986).

The compared PyPTO SPMD implementation can run the same logical GQA paged
attention problem for 8192 and 16384 tokens, so the long-sequence failure
appears specific to the simpler highperf kernel/runtime path rather than the
input shape itself.

### Steps to Reproduce

1. Check out the code from
   [hw-native-sys/simpler#986](https://github.com/hw-native-sys/simpler/pull/986).
2. Build/install the project as usual for onboard a2a3 testing.
3. Run the highperf SPMD paged-attention long sequence case on an a2a3 device:

```bash
   python tests/st/a2a3/tensormap_and_ringbuffer/spmd_paged_attention_highperf/test_spmd_paged_attention_highperf.py \
     -p a2a3 -d 6 --case b1_h32_kv8_s8192_bs128_fp16
```

### Expected Behavior

The highperf SPMD paged-attention kernel should complete successfully for the
8192-token GQA decode case and produce output matching the golden reference.

The expected shape is:

```text
batch = 1
num_heads = 32
num_kv_heads = 8
head_dim = 128
kv_seq = 8192
block_size = 128
dtype = fp16
```

### Actual Behavior

The 8192-token highperf case fails with an AICore timeout. The observed error
code is:

```text
507018
```

Shorter cases such as 128, 512, and 4096 tokens complete successfully. The
failure has been reproduced for the 8192-token highperf case.

### Git Commit ID

afb5c5a95cf05d5bb346eaef83a318c6f3164971

### CANN Version

9.0.0

### Driver Version

25.5.1

### Host Platform

Linux (aarch64)

MirkoDeVita98:
Fix: https://github.com/hw-native-sys/simpler/pull/1091

PR #1076, opened 2026-06-17
test(#1070): enable highperf s128 + s8192 onboard a2a3 in CI

ChaoZheng109:
Closes #1070. Supersedes #1094 (folds its s8192 enable in here).

## Background

#1091 (merged) fixed the #1070 onboard `out` golden mismatch with a
**producer-side** cross-core DDR fence (`DdrBarrierBeforeFfts`) before each
decode FFTS signal — but it left `b1_h32_kv8_s128` **a2a3sim-only**, so CI never
guarded the regression onboard, and its long-sequence cases were all
`manual:True` (CI-excluded).

## A/B validation

I ran a controlled onboard experiment on the #1070 `s128` case:

| arm | kernel | st-onboard-a2a3 |
|-----|--------|-----------------|
| producer-side only (#1091 as merged) | `DdrBarrierBeforeFfts` | **10/10 green** |
| producer + consumer-side invalidate | + `dcci` invalidate after each `wait_flag_dev` | 9/9 green (+1 infra flake) |

→ **The producer-side fence alone closes the race** (19/19 onboard rounds clean
total); the consumer-side invalidate added no measurable benefit, so it is **not
included**.

## Change (test-only, no kernel change)

- Re-enable `b1_h32_kv8_s128` on **a2a3** (was a2a3sim-only) so CI guards the
  #1070 regression onboard.
- Drop `manual:True` from `b1_h32_kv8_s8192` so CI also guards the long-sequence
  fix onboard (was PR #1094).

Issue #1070, opened 2026-06-16
[Bug] spmd_paged_attention_highperf b1_h32_kv8_s128_bs128_fp16 regressed: sim scheduler stall (-100) + onboard golden mismatch


ChaoZheng109:
### Platform

a2a3 (Ascend 910B/C hardware) **and** a2a3sim (Ascend 910B/C simulation) — fails on both.

### Runtime Variant

tensormap_and_ringbuffer

### Description

`tests/st/a2a3/tensormap_and_ringbuffer/spmd_paged_attention_highperf/test_spmd_paged_attention_highperf.py::TestSpmdPagedAttentionHighPerf::test_run` fails on the only non-manual case, `b1_h32_kv8_s128_bs128_fp16` (batch=1, num_heads=32, kv_heads=8, head_dim=128, kv_seq=128, block_size=128, fp16).

This is a **regression on `main`**: the `st-sim-a2a3` and `st-onboard-a2a3` jobs both passed at commit `17fa04aa` (2026-06-11, run 27324277512) and fail at the current HEAD `19b2c0be`. The failure reproduces deterministically on simulation, so it is not hardware flakiness.

The failure presents differently on the two backends:

- **a2a3sim**: scheduler stall — the workload never completes and the runtime times out.
- **a2a3 (onboard)**: numerical golden mismatch on `out`, and the resulting hang/fault poisons the device (cascading 507018 into the next test in the same worker).

### Steps to Reproduce

```
# simulation (deterministic)
pytest tests/st/a2a3/tensormap_and_ringbuffer/spmd_paged_attention_highperf/test_spmd_paged_attention_highperf.py \
    --platform a2a3sim --device 0-15 -v

# onboard
python -m pytest tests/st/a2a3/tensormap_and_ringbuffer/spmd_paged_attention_highperf/test_spmd_paged_attention_highperf.py \
    --platform a2a3 --device <range> -v
```

Observed in CI run 27612538258:
- sim (ubuntu): https://github.com/hw-native-sys/simpler/actions/runs/27612538258/job/81640333256
- onboard: https://github.com/hw-native-sys/simpler/actions/runs/27612538258/job/81640333245

### Expected Behavior

`test_run` passes for `b1_h32_kv8_s128_bs128_fp16` on both a2a3sim and a2a3, as it did at `17fa04aa`.

### Actual Behavior

a2a3sim:

```
E       RuntimeError: run_prepared failed with code -100
[...]

PR #1094, opened 2026-06-22
test: enable spmd_paged_attention_highperf s8192 case in CI

ChaoZheng109:
## Why

Issue #1022 (SPMD paged attention highperf fails for >= 8192 sequences) was
closed by PR #1091. However, #1091 added the new long-sequence cases
(s4096 / s6144 / s8192 / s16384 / b2 variants) with `manual: True`, so CI
never actually exercised the `s8192` case that the bug report targeted — the
fix went in unverified by the pipeline.

## What

Drop `manual: True` on `b1_h32_kv8_s8192_bs128_fp16` so the case that
reproduced #1022 now runs in the a2a3 onboard CI and guards against
regression. The other newly-added long-sequence cases remain `manual` to keep
CI runtime bounded.

## Test

CI (a2a3 onboard) will run the s8192 highperf paged-attention case.


ChaoZheng109:
Folded into #1076, which enables both b1_h32_kv8_s128 (onboard #1070 regression guard) and b1_h32_kv8_s8192 (this PR's case) in one change. Closing in favor of #1076.

PR #1107, opened 2026-06-22
[Performance] Wireless (Polling-based) task readiness detection brings 15~40% speedup

SergioMartin86:
I am posting this as a draft PR because it has diverged so much from main, it'd be too disruptive to propose it as normal PR.

Nevertheless, this is proof there is a high potential for >>10% speedups if we change the current "wiring" strategy to keep track of dependencies, and instead we use a polling-based approach. The gist:

- On adding a task, only remember the indexes of the tasks upon which the task depends
- To detect whether the task is ready to go, the scheduler "constantly" polls a common shared array (initially full of zeros). If all its dependencies are 1, then the task is ready to go
- If the task only has one dependency left, assign it as a "notify" task to that producer, so polling is no longer needed
- Consumers do not notify producers, whose memory are freed up regularly (not on notification).

Here are my experimental result, across several tests, comparing this PR (branch: [wireless2](https://github.com/huawei-csl/simpler/tree/wireless2)) with upstream/main @ c4b0aac2 (Fix: monotonic scope_stats heap accounting for multi-wrap scopes (#996) (#1031) Date: 2026-06-17)

image

Wireless2 is faster on every case, by 16% to 48%. Geo-mean speedup of 30%. These are all speedups on Device Time (host time is not considered, as it's outside the scope of our investigation)

Noah Baumann (@noabauma) is currently working on a minimal patch to apply these optimizations to main without too many disruptive changes. Nevertheless, I thought you'd like to have access to this PR to test it yourselves.

Adding AI-generated summary of the optimizations, which may help replicating it:

[wireless-architecture.md](https://github.com/user-attachments/files/29202845/wireless-architecture.md)


SergioMartin86:
Closing, as this change has been refined into a much less invasive PR: 
https://github.com/hw-native-sys/simpler/pull/1137

Issue #1111, opened 2026-06-23
[Code Health] In-place card recovery (force_reset_device) silently fails / is a2a3-unverified — make it observable and self-checked


ChaoZheng109:
### Category

Robustness (potential edge-case failure)

### Component

Platform (a2a3 / a2a3sim)

### Description

The in-place card-recovery path
(`DeviceRunner::force_reset_device()` → `aclrtResetDeviceForce`, called from
`DeviceRunner::finalize()` on the `device_unusable_` poison path) is the intended
fix for the AICore op-timeout cascade: it heals the poisoned card in the same
process so the next `Worker.init()` succeeds and the L2 cases that would
otherwise be skipped actually run (see
`docs/investigations/2026-06-a5-aicore-op-timeout-cascade.md`).

This recovery is **hardware-verified on a5 only**. The a2a3 mirror is
**CI-verified only** — the dev box has no a2a3 silicon, and the investigation
explicitly flags that "the CANN force-reset semantics may differ from a5"
(see its "When to reconsider" section). In practice we still see L2 cases fall
back to the poison-`skip` path, which means force-reset is **not reliably
recovering the card** (at least on a2a3, possibly on rare a5 force-reset
failures too).

Worse, when force-reset fails the failure is **invisible to the runtime**, so
the system cannot tell "recovered" from "still poisoned":

1. **The reset rc is swallowed.** `force_reset_device()` only `LOG_ERROR`s a
   non-zero `aclrtResetDeviceForce` rc and returns void — no caller can act on
   it (`device_runner.cpp:568-573`).
2. **`device_unusable_` is cleared unconditionally** at the end of `finalize()`
   regardless of whether the reset succeeded (`device_runner.cpp:649`), so a
   reused/next runner starts with a clean flag even if the card is still
   poisoned.
3. **No post-reset self-check.** Whether the card is actually clean is only
   discovered later, in Python, when the next `Worker.init()` fails — at which
   point the only remaining option is the fixture's poison-`skip` fallback.

Net effect: when the in-place reset silently fails, the would-be-skipped tests
lose coverage in that run, and we have no signal at the C++ layer that recovery
failed.

### Location

- `src/a2a3/platform/onboard/host/device_runner.cpp:553-574` — `force_reset_device()` (rc swallowed)
- `src/a2a3/platform/onboard/host/device_runner.cpp:642-649` — `finalize()` poison path + unconditional `device_unusable_ = false`
[...]

Issue #1110, opened 2026-06-23
[Code Health] L2 device-poison skips leave tests uncovered — add a dispatcher-level fresh-process retry pass


ChaoZheng109:
### Category

Robustness (potential edge-case failure)

### Component

Tests

### Description

When an AICore op-timeout poisons a gw xdist worker's L2 device context, the
triggering case FAILs and every subsequent same-runtime L2 case on that worker
process is marked `_l2_poisoned` and `pytest.skip`ped (see
`docs/investigations/2026-06-a5-aicore-op-timeout-cascade.md`). The skip
correctly stops the misleading 507899 cascade, but it means **those cases get
no coverage at all** in that run — they are neither verified on the poisoned
card nor anywhere else.

The intended recovery is the in-place `aclrtResetDeviceForce` path in
`DeviceRunner::finalize()`, which heals the card so the next `Worker.init()`
succeeds and the would-be-skipped cases actually run. That path is
hardware-verified on **a5** but the a2a3 mirror is **CI-verified only** (no a2a3
silicon on the dev box). When force-reset does not recover the card (a2a3, or a
rare a5 force-reset failure), the fixture falls back to the skip — and those
cases silently lose coverage.

This issue tracks **option B**: a dispatcher-level retry pass that re-runs the
poison-skipped cases in a fresh subprocess (= a clean card, which the
investigation confirms always recovers the poison), independent of whether the
in-process force-reset worked. Option A (making force-reset itself reliable on
a2a3) is the better end-state but is an empirical hardware fix gated on a2a3
silicon; B is a deterministic, hardware-semantics-independent safety net that
restores coverage today.

**Critical correctness constraint (design note):** the retry pass must collect
cases by a marker the poison guard *actively registers*, NOT by
`outcome == "skipped"`. Several legitimate skips would otherwise be wrongly
retried:

- `@pytest.mark.skip` / `skipif` (author-intentional)
- platform-required skips (`conftest.py:554/573/1104`)
- `st_worker requires SceneTestCase` (`conftest.py:1149`)
- `No cases matched ...` (`simpler_setup/scene_test.py:1418`)

(Platform-mismatch cases are *deselected*, not skipped — `conftest.py:515` — so
they never appear as skipped, but the others above do.)

Proposed mechanism: at the two poison-skip points, register
`(nodeid, runtime)` into a structured sink (per-worker JSONL file, or xdist
`workeroutput`); the dispatcher reads exactly that sink after the L2 xdist
[...]

PR #1121, opened 2026-06-23
Add: dispatcher fresh-process retry for L2 poison-skips (#1110)

doraemonmj:
## What

Option B for #1110: restore coverage for L2 classes that get
`pytest.skip`ped after an AICore op-timeout poisons a gw xdist worker's
device context. The dispatcher re-runs exactly those classes in a fresh
subprocess (= clean card) so they aren't silently dropped.

Real CI evidence of the gap: run 27742754024 `st-onboard-a2a3` gw3 —
`TestPagedAttentionManualScope` FAILED (507018) then **16** tensormap L2
classes SKIPPED, all losing coverage.

## How

- Each of the two `st_worker` poison guards registers the skipped class
  `(selector, nodeid)` into a per-runtime JSONL sink (`O_APPEND`, crosses
  the xdist-worker → L2-subprocess → dispatcher boundaries).
- The dispatcher reads exactly that sink after the L2 subprocess exits and
  re-runs only those classes via `--case ClassName::` in a fresh subprocess
  with the sink env stripped (single bounded pass).
- Collection is strictly sink-based, never `outcome == "skipped"`, so
  legitimate skips (`@pytest.mark.skip`, platform-required, `No cases
  matched`) are never retried.
- Independent of whether the in-place force-reset recovery fired — this is
  the hardware-independent coverage floor. Complementary to #1111 (make the
  in-place recovery decision observable / self-checked).

## Verification

- Unit + end-to-end tests drive the real `_dispatch_test_phases` loop with a
  stateful fake subprocess that reproduces the cascade → fresh-process
  retry; a mutation check confirms the e2e test fails if the retry is removed.
  (Tests kept local for now, not in this PR.)
- a2a3 hardware (Ascend910, exclusive task-submit lock): a fresh process
  recovers a poisoned card 10/10; force-reset itself recovers 10/10 when
  triggered. Investigation doc updated with these measurements and the
  drain-success false-negative that explains why the cascade still happens.

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -17,12 +17,14 @@ subprocess per runtime so each gets a clean CANN context. See docs/testing.md.
 from __future__ import annotations
 
 import faulthandler
+import json
 import logging
 import os
 import re
 import signal
 import subprocess
 import sys
+import tempfile
 import time
 import typing
 
@@ -899,6 +901,15 @@ def _dispatch_test_phases(session, resource_specs):  # noqa: PLR0912
         cmd = base_args + ["--runtime", rt, "--level", "2"]
         if xdist_available:
             cmd += ["-n", str(max_parallel), "--dist", "loadfile"]
+        # Per-runtime sink for the in-process poison guards (issue #1110). Each
+        # xdist worker appends the classes it poison-skips; we re-run them in a
+        # fresh subprocess after this one exits so they don't silently lose
+        # coverage. mkstemp gives an unpredictable, exclusively-created path
+        # (this dev box is shared — a predictable /tmp name is exposed to symlink
+        # / TOCTOU attacks and pid-recycle collisions); the children just append.
+        sink_fd, sink_path = tempfile.mkstemp(prefix="simpler_l2_poison_", suffix=f"_{rt}.jsonl")
+        os.close(sink_fd)
+        run_env = {**os.environ, _L2_POISON_SINK_ENV: sink_path}
         # L2 subprocesses run serially (one runtime at a time) so we don't
         # need to buffer their stdout — we can stream it directly through
         # the group markers. ``::group::`` on its own line before the run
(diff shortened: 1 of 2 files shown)


doraemonmj:
## Real-hardware end-to-end validation (a2a3, Ascend910)

Confirmed option B fires and restores coverage on real a2a3 silicon
(exclusive `task-submit` device lock). To force the poison-skip path —
which this box's in-process `force_reset_device` recovery otherwise
prevents — `force_reset_device()` was temporarily disabled (reverted +
rebuilt afterward; not in this PR) and a throwaway hang L2 SceneTestCase
reused `tests/st/aicore_op_timeout`'s STARS-reaped hang kernel.

Run: `pytest <hang_dir> <spmd_basic> <multi_round_paged_attention> <hbg/matmul> --platform a2a3 --device <1 id>` through the real dispatcher.

```
L2 tensormap_and_ringbuffer subprocess (force-reset disabled):
  FAILED   TestHangPoison::test_run      # real op-timeout 507018 poisons the card
  1 failed, 2 skipped                    # spmd_basic + multi_round poison-SKIPPED
  --- L2 tensormap_and_ringbuffer: FAIL rc=1 ---

dispatcher retry (option B):
  ::group::L2 tensormap_and_ringbuffer poison-retry (2 class(es), fresh process)
    multi_round_paged_attention  PASSED  # recovered on the clean fresh-process card
    spmd_basic                   PASSED
  --- poison-retry: PASS ---             # coverage restored
```

Behaviour is exactly as designed:
- The triggering FAIL (`TestHangPoison`) is **not** retried — it falls out
  of the retry's `--case` selectors as a legitimate "No cases matched" skip.
- The retry re-ran **only** the two sink-registered poison-skips, on a fresh
  process (= clean card; a fresh process recovers even with force-reset
  disabled, since poison does not survive process exit).
- Job still exits non-zero: the real failure is reported, not masked.

This closes the last gap — the full `skip → sink → fresh-process retry →
recover` chain now verified on a2a3 hardware, complementing the unit +
end-to-end dispatcher tests and the 10/10 force-reset/fresh-process
recovery probe recorded in the investigation doc.

ChaoZheng109 (dismissed):

ChaoZheng109 (approved):

PR #1124, opened 2026-06-23
Fix: trigger + verify + retry + observe in-place card recovery (#1111)

doraemonmj:
## What

Closes the gaps in the in-place card recovery (`force_reset_device` →
`aclrtResetDeviceForce`) called from `DeviceRunner::finalize()` on the
`device_unusable_` poison path. Mirrored across a2a3 and a5.

The root cause my hardware testing isolated is **not** that force-reset
fails (it recovers a poisoned a2a3 card 10/10 when triggered) — it's that
force-reset was **silently not triggered**: the trigger gate keyed on the
bounded drain's rc, which false-negatives on a still-poisoned card. Real
impact: CI run 27742754024 (`st-onboard-a2a3` gw3) — 1 FAILED + **16
SKIPPED** after the drain reported success on a poisoned card.

## How (maps to #1111's proposed fixes)

1. **Trigger gate (root cause, not in #1111's list).**
   `recover_device_or_mark_unusable()` now always sets `device_unusable_`
   on an AICore error; the bounded drain stays a best-effort settle but no
   longer gates recovery. The op-timeout sticky-error is only cleared by a
   force reset (per-card safe under the exclusive `task-submit` lock).
2. **Post-reset self-check (#1111 fix 2).** After the reset,
   `force_reset_device()` re-binds and does a trivial `rtMalloc`/`rtFree`
   (the a2a3 poison's surfacing call, 507899) and returns 0 only when that
   confirms the card is clean.
3. **Bounded retry (#1111 fix 3).** `finalize()` retries the
   drain-then-reset cycle up to 3 attempts until the probe confirms clean;
   if all fail the card stays flagged for the layer above to recover. No
   unbounded loop.
4. **Observability (#1111 fix 1).** `force_reset_device()` returns its rc
   (was `void`) and logs the previously silent early-return failures;
   `finalize()` clears `device_unusable_` only when recovery actually
   succeeded.

## Verification (a2a3 hardware, Ascend910, exclusive lock)

- poison → reset → **probe confirmed clean** → in-process `Worker.init`
  recovers, across 3/3 and 5/5 runs; the probe never false-fails a
  recovered card (the key regression risk).
- Retry loop exercised by injecting two probe failures (temp, reverted):
  attempts 1-2 logged `did not confirm clean`, attempt 3 recovered.

**Caveat:** the a5 mirror is symmetric but unvalidated locally (no a5
silicon). Its `rtMalloc` probe is a weaker signal on a5 since a5 poison
surfaces earlier (register-map / stream) than malloc — it never false-fails
a clean card, but a still-poisoned a5 card could pass it. Flagged for
reviewers; `st-onboard-a5` CI is the a5 channel.

## Relationship to #1110

Complementary, independent. This (#1111, option A) makes the in-place
recovery reliable + observable, reducing how often the #1110 dispatcher
retry (option B, hardware-independent coverage backstop) has to fire.

### DIFF
--- a/src/a2a3/platform/onboard/host/device_runner.cpp
+++ b/src/a2a3/platform/onboard/host/device_runner.cpp
@@ -462,26 +462,29 @@ void DeviceRunner::recover_device_or_mark_unusable(int aicore_rc) {
     // st_worker pool hands one ChipWorker to every test class on a device),
     // that one error poisons every later test in the xdist worker process.
     //
-    // Try to drain the device so a subsequent run can recover in place. Use
-    // the BOUNDED aclrtSynchronizeDeviceWithTimeout, NOT an unbounded
-    // aclrtSynchronizeStream* on the error-state stream — the latter wedges
-    // subsequent tests (see DeviceRunnerBase::finalize_common's comment on
-    // why finalize deliberately skips a pre-destroy sync). If the bounded
-    // device drain itself errors, the context is unrecoverable without a full
-    // device reset, so mark the runner unusable: run() then fails fast with a
-    // clear message instead of cascading, and finalize() force-resets the card.
+    // Best-effort BOUNDED drain (aclrtSynchronizeDeviceWithTimeout, NOT an
+    // unbounded aclrtSynchronizeStream* on the error-state stream — the latter
+    // wedges subsequent tests, see DeviceRunnerBase::finalize_common). But DO
+    // NOT gate recovery on the drain's rc: the bounded drain can return success
+    // on a still-poisoned card — a false-negative that left force_reset_device()
+    // untriggered and cascaded into 16 skipped L2 cases on CI run 27742754024
+    // (st-onboard-a2a3 gw3). The op-timeout sticky-error is only cleared by a
+    // force reset (a soft reset/drain does not), so always mark the runner
+    // unusable here: run() fails fast and finalize() force-resets the card, so
+    // the next Worker.init lands clean regardless of what the drain reported.
     int sync_rc = aclrtSynchronizeDeviceWithTimeout(PLATFORM_STREAM_SYNC_TIMEOUT_MS);
     if (sync_rc != ACL_SUCCESS) {
         LOG_ERROR(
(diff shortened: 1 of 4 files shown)


doraemonmj:
Addressed @gemini-code-assist's device-reference-leak finding (976a244b): the post-reset probe's manual `aclrtSetDevice` is replaced with a fresh `DeviceBindGuard`, so the reference is released on every exit path — restoring `force_reset_device`'s own no-leak invariant. Applied to both a2a3 and a5.

Re-verified on a2a3 hardware (Ascend910, exclusive task-submit lock): poison → reset → probe confirmed clean → in-process `Worker.init` recovers, 5/5, with no spurious probe failure.

doraemonmj:
Addressed @coderabbitai's review (force-pushed, squashed to one commit):

- **Reference balance + rtFree check** — `force_reset_device` now scopes the initial bind/drain/reset in its own block (released before the probe), the probe uses a fresh `DeviceBindGuard`, and every probe rc is checked including `rtFree`/`aclrtDestroyStream` (any failure → non-zero, card stays flagged).
- **Drain in valid ACL context** — the best-effort `aclrtSynchronizeDeviceWithTimeout` moved out of `finalize()`'s retry loop and into `force_reset_device()` after the `AclInitGuard`/`DeviceBindGuard`, so it no longer runs after `aclFinalize`/`rtDeviceReset` teardown.
- **a5 probe strength** — the probe now exercises both poison surfaces: a stream create/destroy (a5's `rtStreamCreate`/`halResMap` path) and an HBM alloc/free (a2a3's `rtMalloc` 507899). The a5 mirror is still unvalidated locally (no a5 silicon); `st-onboard-a5` CI is the channel.

Re-verified on a2a3 hardware (Ascend910, exclusive lock): poison → reset → probe confirmed clean → in-process `Worker.init` recovers, 5/5, no spurious probe failure; the retry loop was exercised via injected probe failures (attempts 1-2 failed, attempt 3 recovered).

ChaoZheng109:
closes #1111

ChaoZheng109 (approved):

PR #1134, opened 2026-06-24
ci: read the pinned pto-isa commit from a committed pto_isa.pin

lyfne123:
## Why

The pinned pto-isa revision lives only in workflow `env: PTO_ISA_COMMIT` literals (`ci.yml`, `sanitizers.yml`). Downstream consumers that pin this repo as a submodule — notably **pypto** — have no stable, machine-readable file to derive it from, so each re-declares its own pto-isa pin and they drift out of sync.

## What

- **`pto_isa.pin`** (new, repo root): the committed source of truth — a single 40-char pto-isa commit.
- **`.github/actions/read-pto-isa`** (new composite action): reads `pto_isa.pin`, validates it is a 40-char hex commit (fails early otherwise), and exports `PTO_ISA_COMMIT` for the calling job.
- Every job that needs the pin adds one line after checkout: `- uses: ./.github/actions/read-pto-isa` (six jobs in `ci.yml`, the sanitizer job in `sanitizers.yml`).
- The workflow-level `PTO_ISA_COMMIT` literals are removed — `pto_isa.pin` is now the only place the revision is declared, and the read+validate logic lives once in the composite action.

**No version change** — `pto_isa.pin` equals the previous pin (`ddafa8da…`). All existing `${{ env.PTO_ISA_COMMIT }}` / `${PTO_ISA_COMMIT}` references are unchanged; only their source moved from a literal to the file.

## Context

Upstream half of a cross-repo effort to keep pypto / pypto-lib / runtime pinning the same pto-isa. pypto's side (`toolchain/versions.env` SSOT) is in hw-native-sys/pypto#1851; a follow-up there derives pypto's pto-isa from `runtime/pto_isa.pin` once this lands.

### DIFF
--- a/.github/actions/read-pto-isa/action.yml
+++ b/.github/actions/read-pto-isa/action.yml
@@ -1,0 +1,18 @@
+name: Read pinned pto-isa commit
+description: >-
+  Read the committed pto_isa.pin (the repo-root source of truth for the pinned
+  pto-isa revision) and export it as PTO_ISA_COMMIT for subsequent steps in the
+  calling job. Requires the repository to be checked out first.
+runs:
+  using: composite
+  steps:
+    - name: Read and validate pto_isa.pin
+      shell: bash
+      run: |
+        commit="$(tr -d '[:space:]' < pto_isa.pin)"
+        if ! printf '%s' "$commit" | grep -qE '^[0-9a-f]{40}$'; then
+          echo "::error file=pto_isa.pin::expected a 40-char hex commit, got '$commit'"
+          exit 1
+        fi
+        echo "PTO_ISA_COMMIT=$commit" >> "$GITHUB_ENV"
+        echo "pinned pto-isa: $commit"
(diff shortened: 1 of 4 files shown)


lyfne123:
Reworked per review: CI now **reads** `pto_isa.pin` directly (a `read-pto-isa` lead job feeds the six jobs via `needs.read-pto-isa.outputs.commit`; sanitizers loads it into `$GITHUB_ENV`). Dropped the duplicated workflow literal and the verify job — `pto_isa.pin` is now the single declaration.

lyfne123:
Reworked again to address the review:

- **Per-job `env` removed** (the concern raised separately): each consuming job now adds a single `- uses: ./.github/actions/read-pto-isa` step after checkout. No global `env` is possible here because workflow-level `env:` can't reference job outputs or read a file — so the read happens in a composite action instead, centralizing the logic in one place.
- **CodeRabbit – validate `pto_isa.pin`**: the composite action validates the value is a 40-char hex commit and fails early with a clear message otherwise. ✅
- **CodeRabbit – pin/harden the `read-pto-isa` job checkout**: that lead job is gone entirely; the composite action performs no checkout (it reads the file the job already checked out), so there is no new `actions/checkout` to pin. ✅

ChaoWao (approved):

PR #1152, opened 2026-06-25
Right-size DFX buffers per cross-subsystem capacity audit

### LINKED ISSUE #977: [Code Health] DFX buffer sizing audit — dep_gen drops on realistic workloads, cross-subsystem capacity review
### Category

### Component

Other (please specify in description) — DFX subsystems (dep_gen, l2_swimlane, pmu, tensor_dump, scope_stats) under `src/a2a3/platform/include/common/platform_config.h` and the AICPU collector implementations.

### Description

`PLATFORM_DEP_GEN_BUFFERS_PER_INSTANCE = 4` × `PLATFORM_DEP_GEN_RECORDS_PER_BUFFER = 32` gives only **128 in-flight records** before the device-side rotation in `dep_gen_collector_aicpu.cpp` hits `free_queue` empty and silently drops 32 records at a time (overwrite current buffer + bump `dropped_record_count`). On the realistic `paged_attention_unroll_manual_scope` Case1 workload (1024 `submit_task` calls, ~16 ms total run, burst window ~1 ms), this drops **512/1024 records = 50%**, and `dep_gen_collector.cpp:reconcile_counters()` (correctly) refuses to emit `deps.json`. The user-visible symptom is "`--enable-dep-gen` alone produces no deps.json" while paired `--enable-dep-gen --enable-l2-swimlane` works (because l2_swimlane slows the AICPU dispatch loop enough for the host mgmt thread to drain in time — an accidental coupling, not a reliable workaround).

This issue is broader than just dep_gen. The same SPSC + ProfilerBase pattern is used by all five DFX subsystems, but the sizing budgets across them are wildly asymmetric, and dep_gen is the smallest by **2000×** measured in in-flight record count, only partly explained by its larger record size. Below is the audit data + a design discussion.

#### 1. The drop path (all DFX subsystems share this contract)

\`dep_gen_collector_aicpu.cpp:97-104\`:

\`\`\`cpp
if (head == tail) {  // free_queue empty
    LOG_WARN(\"dep_gen: no free buffer, overwriting current (dropped %u records)\", full_buf->count);
    s_dep_gen_state->dropped_record_count += full_buf->count;
    full_buf->count = 0;
    return;
}
\`\`\`

Identical pattern in `l2_swimlane_collector_aicpu.cpp:switch_records_buffer()` and `pmu_collector_aicpu.cpp` (with the matching \"avoids blocking the AICPU dispatch loop\" header comment). The design philosophy is consistent and explicit: **DFX records are best-effort; orch / dispatch hot paths must never block on diagnostics**. Backpressure (spin-wait until host drains) is structurally forbidden because it would change the observed system's timing — defeating the purpose of profiling.

This is the right invariant. The bug is **sizing under this invariant**, not the invariant itself.

#### 2. Cross-subsystem sizing comparison

| Subsystem | Per-unit | BUFFERS_PER_? | RECORDS_PER_BUFFER | Record size | In-flight records (per unit) | Bytes per unit |
|---|---|---|---|---|---|---|
| l2_swimlane AICpu task | core | 8 | 1000 | 64 B | 8000 | ~500 KB |
| l2_swimlane AICore task | core | 4 | 1024 | 32 B | 4096 | ~128 KB |
| l2_swimlane phase (sched/orch) | thread | 16 | **16384** | 40 B / 32 B | **262 144** | ~10 MB |
| tensor_dump | thread | 8 | 256 | 128 B | 2048 | 256 KB |
| pmu | core | 4 | 512 | ~64 B | 2048 | ~128 KB |
| scope_stats | instance | 8 | 512 | small | 4096 | small |
| **dep_gen (current)** | **instance** | **4** | **32** | **2624 B** | **128** | **336 KB** |

dep_gen sits 16–2000× below every other subsystem on in-flight record count. The justification given in the existing comment (`platform_config.h:266`) is:

> Each DepGenRecord is ~2.3 KB ... sized to fit a typical example's submit count (~100-200) in a few buffers.

That \"typical example\" assumption (100–200 submits) is **stale**. Realistic workloads in the repo today:

- `paged_attention_unroll_manual_scope` Case1: 1024 submits / launch — already 5–10× the assumption
- `paged_attention_manual_scope` CaseSmall1: ~50 submits — within assumption
- A `qwen3_14b` decode step per-layer: substantially more

#### 3. Empirical verification (commit `20ba6a2f`, onboard a2a3 via task-submit)

| `PLATFORM_DEP_GEN_BUFFERS_PER_INSTANCE` | Dropped records | deps.json emitted? |
|---|---|---|
| 4 (default) | 512 / 1024 | ❌ |
| 16 | 384 / 1024 | ❌ |
| 64 | 0 / 1024 | ✅ (865 KB) |

So even at 16 the workload still drops 38%. The fix has to be substantial, not incremental.

#### 4. Why \"few big buffers\" beats \"many small buffers\" for dep_gen

Within the same total byte budget there are two ways to scale:

| Sizing | RECORDS_PER_BUFFER | BUFFERS_PER_INSTANCE | In-flight | Total memory | Rotations for 1024 records |
|---|---|---|---|---|---|
| Many small | 256 | 16 | 4096 | 10.7 MB | 4 |
| **Few big (proposed)** | **1024** | 4 | **4096** | **10.7 MB** | **1** |

Few-big matches the l2_swimlane phase pool philosophy (16384 records per buffer, 16 buffers). For dep_gen specifically, few-big is the better trade:

1. **Per-rotation host cost is non-trivial.** `mirror_shm_from_device` + `try_pop_aicpu_entry` + `proactive_replenish` runs every mgmt loop iteration regardless of how many records were drained. Fewer rotations → fewer mgmt loop wakeups during a burst.
2. **AICPU rotation cost scales with rotation count, not buffer size.** A single SPSC pop+push is the same work whether the buffer holds 32 or 1024 records.
3. **Failure-mode granularity is acceptable**: when a drop *does* occur, losing 1024 records is no worse than losing 32 — the reconcile path already rejects the whole deps.json (it's an all-or-nothing graph artifact, unlike per-task swimlane records that degrade gracefully).

### Location

- Sizing constants: `src/a2a3/platform/include/common/platform_config.h:266-282`
- Device-side drop path: `src/a2a3/platform/shared/aicpu/dep_gen_collector_aicpu.cpp:97-104`
- Host reconcile gate: `src/a2a3/platform/shared/host/dep_gen_collector.cpp:155-215` (`reconcile_counters`)
- Cross-subsystem reference: `src/a2a3/platform/include/common/l2_swimlane_profiling.h:496` (phase = 16384/buffer), `src/a2a3/platform/include/common/pmu_profiling.h` (pmu = 512/buffer), `src/a2a3/platform/include/common/tensor_dump.h:124` (dump = 256/buffer), `src/a2a3/platform/include/common/scope_stats.h` (scope_stats = 512/buffer)

### Proposed Fix

**Primary**: bump dep_gen's per-buffer records to align with other DFX subsystems' headroom.

\`\`\`cpp
// platform_config.h
constexpr int PLATFORM_DEP_GEN_RECORDS_PER_BUFFER = 1024;  // was 32
constexpr int PLATFORM_DEP_GEN_BUFFERS_PER_INSTANCE = 4;   // unchanged
// In-flight: 4 * 1024 = 4096 records (32× current).
// Memory: 4 buffers * 1024 records * 2624 B/record ≈ 10.7 MB per instance.
// Aligns with l2_swimlane Aicpu task pool (8000 records/core, ~500 KB)
// in capacity philosophy. Verified to eliminate drops on
// paged_attention_unroll Case1 (1024 submits).
\`\`\`

Update the stale `\"typical example's submit count (~100-200)\"` comment with measured numbers + the new headroom rationale.

**Secondary**: as part of the same PR or a follow-up, audit the other DFX subsystems against realistic workloads to confirm they have adequate headroom. The l2_swimlane phase pool already does (262K in-flight); pmu and tensor_dump may be worth re-measuring.

**Not appropriate** (explicitly considered and rejected):

- **Spin-wait backpressure on the AICPU side**: violates the \"DFX never blocks dispatch\" invariant that all five subsystems share. Would change the observed system's timing → swimlane measurements stop reflecting real workload behavior. The current \"overwrite + drop + accept count\" is the right contract.
- **Faster host drain (smaller poll sleep)**: tested via reasoning — at BUFFERS=16 the workload still drops 38%, meaning the bottleneck isn't drain frequency but pool depth.

### Priority

Medium (minor risk, should fix in next few releases) — dep_gen is opt-in (`--enable-dep-gen`), and the broken case is \"solo dep_gen on workloads >200 submits\". Users who notice silently get steered toward paired mode (which works for the wrong reason). But the silent failure mode + the broader sizing-philosophy question across DFX subsystems make it worth fixing before more subsystems hit the same wall.

Related: #860 (tensor dump host drain backpressure — same SPSC pattern), #908 (dep_gen fanout overhead), #959 (scheduler wiring deadlock — different mechanism, same hot-path-must-not-block principle).

doraemonmj:
Audited all 5 DFX collectors (l2_swimlane / dep_gen / pmu / scope_stats /
tensor_dump) for buffer over/under-provisioning, using Little's Law
(L = produce-rate × recycle-latency) cross-checked against onboard
worst-case flood runs (paged_attention 65536-submit + manual_scope variants).

### Per-subsystem result

| DFX | Verdict | Change | Lever | Memory |
| --- | ------- | ------ | ----- | ------ |
| **l2_swimlane** | phase pools 2–16× over (orch uses ~5/16, sched <1/16) | `BUFFERS_PER_THREAD` split → `SCHED 16→6` / `ORCH 16→8` | count | phase 72 → **28 MB** |
| **pmu** | over (free margin 75%, L≈1) | `BUFFERS_PER_CORE 4→2` | count | 9.2 → **4.6 MB** |
| **scope_stats** | over (free margin 87.5%, L≈1) | `BUFFERS_PER_INSTANCE 8→4` | count | 0.21 → **0.11 MB** |
| **dep_gen** | rate-limited drops, not capacity | `RECORDS_PER_BUFFER → 1024` | size | corrects 2048 overshoot, frees ~17.5 MB |
| **tensor_dump** | adequate (0 overwrite across 6 cases) | unchanged | — | — |

### Logic

- **Two distinct over-provision modes.** pmu/scope/l2 waste **buffer count**
  (Little's L≈1–5 but configured 4–16×); dep_gen's issue was per-buffer
  **size**. We cut the dimension that's actually over-provisioned and leave
  the other alone (e.g. l2 phase buffer *size* is 89%+ used, so untouched).
- **Flood drops are rate-bound, not capacity-bound.** dep_gen drops only
  under bare no-dependency flood; doubling any buffer knob doesn't move them,
  and real dependency-paced workloads never drop — so buffer sizing isn't the
  fix there (the 32→1024 change addresses the unroll 50%-drop under-size, not
  the flood).
- **l2_swimlane zero-ABI split.** sched/orch have asymmetric throughput, so
  the shared `BUFFERS_PER_THREAD` is split per kind. The device-visible
  ready_queue is decoupled (held at 16), so the shm header layout is unchanged.

### Validation

Onboard zero-drop on a2a3 under the 65536-submit flood for every changed pool
(per-pool reconcile `dropped=0`; free-queue watermarks confirm touch-bottom is
recycle-latency-bound, not capacity). Mirrored to a5 by the same
arch-independent theory; a5-silicon validation pending.

**Net: ~66 MB freed, zero ABI.**

### Docs

Full per-subsystem analysis (theory + measurements) in
`docs/dfx/dfx-buffer-capacity-audit.md` (added here). Also corrects the
dep_gen truncation-tail example in `docs/dfx/dep_gen.md` to the shipped 1024
slots (addresses review feedback).

close #977

### DIFF
--- a/docs/dfx/dep_gen.md
+++ b/docs/dfx/dep_gen.md
@@ -340,8 +340,8 @@ within one buffer — replay scans linearly and ties them back to the
 base via `task_id`.
 
 **Truncation tail.** A submit whose chain exceeds the buffer's slot
-budget (`PLATFORM_DEP_GEN_RECORDS_PER_BUFFER = 32` slots → roughly
-`64 + 31 × 582 = 18106` deps max in the best case) is logged via
+budget (`PLATFORM_DEP_GEN_RECORDS_PER_BUFFER = 1024` slots → roughly
+`64 + 1023 × 582 = 595450` deps max in the best case) is logged via
 `LOG_ERROR` and truncated to the largest dc that fits. Runtime
 correctness is unaffected — `Arg::set_dependencies` keeps the full dep
 list; only the dep_gen replay graph loses the tail.
--- a/src/a2a3/platform/include/common/platform_config.h
+++ b/src/a2a3/platform/include/common/platform_config.h
@@ -133,11 +133,18 @@ constexpr int PLATFORM_PROF_BUFFERS_PER_CORE = 8;
 constexpr int PLATFORM_AICORE_BUFFERS_PER_CORE = 4;
 
 /**
- * L2SwimlaneAicpuSchedPhaseBuffer / L2SwimlaneAicpuOrchPhaseBuffer
- * pre-allocation count per AICPU thread (applies independently to both pool
- * kinds). 1 goes into the free_queue at init, the rest into the recycled pool.
+ * Host preallocation count per AICPU thread for the two phase pools, split per
+ * kind (sched vs orch) because their throughput is asymmetric — a single shared
+ * value over-provisions the lighter one. 1 buffer seeds the free_queue at init,
+ * the rest the recycled pool.
+ *
+ * Floor for both: SLOT_COUNT(4) + 1 = 5 (free_queue fillable + 1 active buffer).
+ * Pure host preallocation — zero ABI (the device-visible ready_queue is decoupled
+ * below, held at the original 16). Sizing rationale + measurements:
+ * docs/dfx/dfx-buffer-capacity-audit.md.
  */
-constexpr int PLATFORM_PROF_BUFFERS_PER_THREAD = 16;
+constexpr int PLATFORM_PROF_SCHED_BUFFERS_PER_THREAD = 6;
+constexpr int PLATFORM_PROF_ORCH_BUFFERS_PER_THREAD = 8;
 
 /**
  * Ready queue capacity for performance data collection.
@@ -146,10 +153,18 @@ constexpr int PLATFORM_PROF_BUFFERS_PER_THREAD = 16;
  * over the four buffer kinds (L2SwimlaneAicpuTaskBuffer per core,
  * L2SwimlaneAicpuSchedPhaseBuffer + L2SwimlaneAicpuOrchPhaseBuffer per
  * AICPU thread, L2SwimlaneAicoreTaskBuffer per core).
+ *
+ * The phase term is sized to the ORIGINAL per-thread count (16), decoupled from
+ * PLATFORM_PROF_{SCHED,ORCH}_BUFFERS_PER_THREAD, on purpose: this array sizes the
+ * device-visible L2SwimlaneDataHeader::queues[], so shrinking the host
+ * preallocation must NOT shrink it (that would change the shm layout = ABI).
+ * 992 is far above the observed peak backlog (~40); the slack is harmless.
  */
-constexpr int PLATFORM_PROF_READYQUEUE_SIZE = PLATFORM_MAX_CORES * PLATFORM_PROF_BUFFERS_PER_CORE +
-                                              2 * PLATFORM_MAX_AICPU_THREADS * PLATFORM_PROF_BUFFERS_PER_THREAD +
-                                              PLATFORM_MAX_CORES * PLATFORM_AICORE_BUFFERS_PER_CORE;
+constexpr int PLATFORM_PROF_READYQUEUE_BUFFERS_PER_THREAD = 16;
+constexpr int PLATFORM_PROF_READYQUEUE_SIZE =
+    PLATFORM_MAX_CORES * PLATFORM_PROF_BUFFERS_PER_CORE +
+    2 * PLATFORM_MAX_AICPU_THREADS * PLATFORM_PROF_READYQUEUE_BUFFERS_PER_THREAD +
+    PLATFORM_MAX_CORES * PLATFORM_AICORE_BUFFERS_PER_CORE;
 
 /**
  * System counter frequency (get_sys_cnt)
@@ -240,8 +255,13 @@ constexpr int PLATFORM_PMU_SLOT_COUNT = 4;
 
 /**
  * Pre-allocated PmuBuffer count per AICore.
+ * Sized at 2 = 2× the measured peak in-flight (Little's Law L≈1: PMU records
+ * are produced at the slow fixed sampling rate, decoupled from submit, and the
+ * 64 B record drains instantly, so the host never lets more than 1 buffer be
+ * borrowed). Was 4 (a 4× over-provision). See
+ * docs/dfx/dfx-buffer-capacity-audit.md.
  */
-constexpr int PLATFORM_PMU_BUFFERS_PER_CORE = 4;
+constexpr int PLATFORM_PMU_BUFFERS_PER_CORE = 2;
 
 /**
  * Ready queue capacity for PMU data collection.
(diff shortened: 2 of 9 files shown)


docs/dfx/dep_gen.md
```diff
@@ -340,8 +340,8 @@ within one buffer — replay scans linearly and ties them back to the
 base via `task_id`.
 
 **Truncation tail.** A submit whose chain exceeds the buffer's slot
-budget (`PLATFORM_DEP_GEN_RECORDS_PER_BUFFER = 32` slots → roughly
-`64 + 31 × 582 = 18106` deps max in the best case) is logged via
+budget (`PLATFORM_DEP_GEN_RECORDS_PER_BUFFER = 2048` slots → roughly
```
ChaoZheng109: **[必须修复] 文档与本 PR 自己下发的代码常量矛盾。** 这里写 `PLATFORM_DEP_GEN_RECORDS_PER_BUFFER = 2048`（及 `64 + 2047 × 582 = 1191418`），但 `platform_config.h` 实际设的是 **1024**。2048 恰是本 PR 声称要消除的 “#977 过冲值”。应改为 `1024` → `64 + 1023 × 582 = 595450`。

src/a2a3/platform/include/common/platform_config.h
```diff
@@ -240,8 +259,13 @@ constexpr int PLATFORM_PMU_SLOT_COUNT = 4;
 
 /**
  * Pre-allocated PmuBuffer count per AICore.
+ * Sized at 2 = 2× the measured peak in-flight (Little's Law L≈1: PMU records
+ * are produced at the slow fixed sampling rate, decoupled from submit, and the
+ * 64 B record drains instantly, so the host never lets more than 1 buffer be
+ * borrowed). Was 4 (a 4× over-provision). See docs/my/profiling-config-audit.md
```
ChaoZheng109: **[应修复] 注释依据指向仓库里不存在的文件。** 本 PR 的 a2a3/a5 两个 `platform_config.h` 共 6 处引用 `docs/my/profiling-config-audit.md`（§4.0/§3.1/§4.3 等），但 `docs/my/` 不在工作树中、本 PR 未新增、也未被 gitignore，是个人草稿路径。所有魔数改动的数值依据都指向评审者和未来维护者都打不开的文件。建议把审计文档提交到 `docs/dfx/`（或 `docs/investigations/`）并改引该处，或把关键推理内联进注释（`.claude/rules/doc-consistency.md` §5）。

src/a5/platform/include/common/platform_config.h
```diff
@@ -281,8 +299,14 @@ constexpr int PLATFORM_PMU_SLOT_COUNT = 4;
 
 /**
  * Number of PmuBuffers pre-allocated per core (initial pool size).
+ * Sized at 2 = 2× the Little's Law need (L≈1: PMU records are produced at the
+ * slow fixed sampling rate, decoupled from submit, and the 64 B record drains
+ * instantly, so peak in-flight is 1). Was 4 (a 4× over-provision). Mirrors the
+ * a2a3 change (validated zero-drop on a2a3 silicon); the L≈1 argument is
+ * arch-independent, but **a5-silicon onboard validation is still pending**.
+ * See docs/my/profiling-config-audit.md §4.0.
```
ChaoZheng109: **[应修复] 同 a2a3：注释引用了仓库里不存在的 `docs/my/profiling-config-audit.md`。** a5 侧此处及 §3/§4.1 等多处同样指向该未提交文件。请与 a2a3 一并修正（提交审计文档或内联推理）。

src/a2a3/platform/include/common/platform_config.h
```diff
@@ -146,10 +157,18 @@ constexpr int PLATFORM_PROF_BUFFERS_PER_THREAD = 16;
  * over the four buffer kinds (L2SwimlaneAicpuTaskBuffer per core,
  * L2SwimlaneAicpuSchedPhaseBuffer + L2SwimlaneAicpuOrchPhaseBuffer per
  * AICPU thread, L2SwimlaneAicoreTaskBuffer per core).
+ *
+ * The phase term is sized to the ORIGINAL per-thread count (16), decoupled from
+ * PLATFORM_PROF_BUFFERS_PER_THREAD, on purpose: this array sizes the
```
ChaoZheng109: **[应修复] 注释引用了本 PR 已删除的符号。** 这里写 “decoupled from `PLATFORM_PROF_BUFFERS_PER_THREAD`”，而 a5 的对应注释已正确写成 `PLATFORM_PROF_{SCHED,ORCH}_BUFFERS_PER_THREAD`。请把 a2a3 改成一致。

ChaoZheng109 (changes requested):
本次审计与重构总体可靠：零-ABI 推理成立（`PLATFORM_PROF_READYQUEUE_SIZE` 数值未变）、`BUFFERS<SLOT` 的边界被 `batch_size()` 的 `kBatch < 1 ? 1` 钳住、两处结构体尺寸注释（64B/4672B）是真修对了。但有几处文档/注释与代码不一致需要在同一提交里修掉，详见行内评论。

**另一条无法做成行内评论（文件不在本 PR diff 内），在此说明：**

**[应修复] `docs/dfx/l2-swimlane-profiling.md:844, 904` 引用了被本 PR 删除的 `PLATFORM_PROF_BUFFERS_PER_THREAD`。** 两处都是调优建议（“raise `PLATFORM_PROF_BUFFERS_PER_CORE` / `PLATFORM_PROF_BUFFERS_PER_THREAD`”），该符号已被拆成 `PLATFORM_PROF_{SCHED,ORCH}_BUFFERS_PER_THREAD`，建议同步更新（参见 `.claude/rules/doc-consistency.md` §1：重命名符号需在同一提交里 grep `docs/`）。

src/a5/platform/include/common/platform_config.h
doraemonmj: 提交文档，修改了注释

src/a2a3/platform/include/common/platform_config.h
doraemonmj: 已修改

docs/dfx/dep_gen.md
doraemonmj: 已修改

src/a2a3/platform/include/common/platform_config.h
doraemonmj: 已解决

ChaoZheng109 (approved):

PR #1154, opened 2026-06-25
Gate residual orch/scheduler PMU & dep_gen enable() calls under PTO2_PROFILING (a2a3 + a5)

### LINKED ISSUE #1146: [Code Health] Two profiling enable() gates leak onto orch/scheduler hot path (not covered by PTO2_PROFILING)
### Category

### Component

AICPU Scheduler

### Description

The orchestrator/scheduler hot path contains `if (xxx_enable())` profiling gates. Each gate resolves to an O(1) global-bool read (`return g_enable_pmu;` etc.), but the gate functions are `extern "C"` / `weak` symbols defined in a **separate translation unit** from the scheduler loop, so the compiler **cannot inline them and cannot hoist them out of the loop** — every call is a real `bl`/`ret` on the hot path. This is already acknowledged in-tree (a2a3 `scheduler_dispatch.cpp` ~L810: *"is_pmu_enabled() is extern "C" and the compiler cannot hoist it across the dispatch loop on its own"*).

The existing `PTO2_PROFILING` compile macro (`src/common/task_interface/profiling_config.h`, default `1`) already wraps **most** of these gates — `is_dump_args_enabled()`, `is_scope_stats_enabled()`, etc. are inside `#if PTO2_PROFILING` and compile out when `PTO2_PROFILING=0`. Two residual sites are **not** gated and remain as non-inlinable calls even when profiling is compiled out:

1. **`is_pmu_enabled()`** — called **once per scheduler main-loop iteration** (the tightest spin loop in the system). Note this one is **load-bearing**: `pmu_active` is passed into `dispatch_ready_tasks(...)` and forces single-issue dispatch, so the fix must hard-set `pmu_active = false` under `#if !PTO2_PROFILING`, not just delete the call (when profiling is compiled out, PMU is definitionally off).
2. **`is_dep_gen_enabled()`** — called **once per task submission** in `submit_task_common`.

**Magnitude is unmeasured.** No profile was taken; the original report ("much slower") is a mechanism-true but quantitatively unverified claim. The cost is one non-inlinable function call per scheduler iteration / per task submission. The in-tree function-scope caching of `is_pmu_enabled()` is the strongest existing evidence the team already treats the call cost as real and worth avoiding.

Full analysis: `docs/investigations/2026-06-orch-profiling-enable-gates-hot-path.md`.

### Location

- `src/a5/runtime/tensormap_and_ringbuffer/runtime/scheduler/scheduler_dispatch.cpp:813` — `is_pmu_enabled()`
- `src/a2a3/runtime/tensormap_and_ringbuffer/runtime/scheduler/scheduler_dispatch.cpp:812` — `is_pmu_enabled()`
- `src/a5/runtime/tensormap_and_ringbuffer/runtime/pto_orchestrator.cpp:610` — `is_dep_gen_enabled()`
- `src/a2a3/runtime/tensormap_and_ringbuffer/runtime/pto_orchestrator.cpp` — `is_dep_gen_enabled()` in `submit_task_common`

### Proposed Fix

Gate the two residual sites under the **existing** `PTO2_PROFILING` macro — do **not** add a new `PTO_PROFILING` macro (it would duplicate `PTO2_PROFILING`). For the load-bearing PMU site:

```cpp
#if PTO2_PROFILING
    const bool pmu_active = is_pmu_enabled();
#else
    constexpr bool pmu_active = false;
#endif
```

and the analogous wrap around `is_dep_gen_enabled()` in `submit_task_common`. Recommend taking an AICPU scheduler profile first to confirm the cost is a non-trivial slice before/after, since the magnitude is currently unverified.

### Priority

---

Related: #1103

doraemonmj:
## What

Closes #1146.

Two profiling `enable()` gates leaked onto the orchestrator/scheduler hot path uncovered by `PTO2_PROFILING`:

1. **`is_pmu_enabled()`** — once per scheduler main-loop iteration (`scheduler_dispatch.cpp`).
2. **`is_dep_gen_enabled()`** — once per task submission in `submit_task_common` (`pto_orchestrator.cpp`).

Both predicates are `extern "C"`/weak symbols defined in a separate translation unit from the scheduler loop, so the compiler can neither inline nor hoist them — each is a real `bl`/`ret` on the hot path even when profiling is off. Most other gates were already inside `#if PTO2_PROFILING`; these two were the residual.

## How

Gate both sites, both arches (a2a3 + a5), under the **existing** `PTO2_PROFILING` macro — **no new macro** (would duplicate `PTO2_PROFILING`).

- The PMU site is **load-bearing** (`pmu_active` feeds `dispatch_ready_tasks` and forces single-issue dispatch), so it is hard-set to `constexpr bool pmu_active = false` under `#if !PTO2_PROFILING` rather than dropped — PMU is definitionally off when profiling is compiled out, which keeps the overlapping fast path. The a5 read was additionally **hoisted to function scope** (it was re-called every scheduler loop iteration) to match a2a3; the value is loop-invariant (PMU latched once at kernel entry).
- **dep_gen is gated as a whole subsystem.** The per-task capture (`record_submit`) *and* the three cold sites — `dep_gen_aicpu_init` (`scheduler_cold_path.cpp`) and `set_orch_thread_idx` / `flush` (`aicpu_executor.cpp`) — are all folded under `#if PTO2_PROFILING`. Gating only the per-task capture would leave a half-on state at `=0` (buffer init + empty flush, no records); gating the whole subsystem compiles it out cleanly. The prior in-tree "dep_gen gated independently of PTO2_PROFILING" comment was updated accordingly.

## Scope notes

- Default build (`PTO2_PROFILING=1`) behavior is **unchanged**; a `=1` build still toggles dep_gen at runtime via the host SubmitTrace flag exactly as before.
- The `PTO2_PROFILING=0` + dep_gen-enabled combo is given up on purpose: nothing exercises it today (the only `=0` CI leg `profiling-flags-smoke / pto2-off` runs `vector_example`, which never enables dep_gen), and `=0` is the production-inference build where dep_gen (a diagnostic) is unwanted.
- **Magnitude unmeasured** — the cost is mechanism-true (one non-inlinable call per iteration/submit) but no AICPU scheduler profile was taken. Full analysis in `docs/investigations/2026-06-orch-profiling-enable-gates-hot-path.md`.

## Test

All touched translation units (`scheduler_dispatch.cpp`, `pto_orchestrator.cpp`, `scheduler_cold_path.cpp`, `aicpu_executor.cpp`) `-fsyntax-only` clean under `PTO2_PROFILING=1` **and** `PTO2_PROFILING=0` (a2a3 + a5).

### DIFF
--- a/src/a2a3/runtime/tensormap_and_ringbuffer/aicpu/aicpu_executor.cpp
+++ b/src/a2a3/runtime/tensormap_and_ringbuffer/aicpu/aicpu_executor.cpp
@@ -539,6 +539,7 @@ int32_t AicpuExecutor::run(Runtime *runtime) {
             }
 #endif
 
+#if PTO2_PROFILING
             // dep_gen plugs into the orchestrator thread (single-instance subsystem):
             // record the per-thread ready_queue index before any submit_task fires
             // inside orch_func_.
@@ -546,7 +547,6 @@ int32_t AicpuExecutor::run(Runtime *runtime) {
                 dep_gen_aicpu_set_orch_thread_idx(thread_idx);
             }
 
-#if PTO2_PROFILING
             // scope_stats streams scope_end records off the orchestrator thread:
             // record the per-thread ready_queue index. No-op (writer shared
             // state null) when scope_stats is disabled; the current buffer is
@@ -565,12 +565,12 @@ int32_t AicpuExecutor::run(Runtime *runtime) {
             (*p_func)(orch_args_cached_);
             rt_scope_end(rt);
 
+#if PTO2_PROFILING
             // Flush the (potentially partially-filled) DepGenBuffer so the host
             // collector can pick it up before this orchestrator thread joins.
             if (is_dep_gen_enabled()) {
                 dep_gen_aicpu_flush();
             }
-#if PTO2_PROFILING
             // Push the partially-filled scope_stats buffer so the host gets the
             // final scope_end records. Idempotent / no-op when disabled.
             scope_stats_aicpu_flush_buffers();
--- a/docs/investigations/2026-06-orch-profiling-enable-gates-hot-path.md
+++ b/docs/investigations/2026-06-orch-profiling-enable-gates-hot-path.md
@@ -1,0 +1,133 @@
+# Gating the two residual profiling `enable()` calls left on the orch/scheduler hot path
+
+**Date**: 2026-06-25
+**Verdict**: applied (gated under the **existing** `PTO2_PROFILING`); magnitude
+deliberately **not** measured; a new `PTO_PROFILING` macro was rejected.
+
+## Question
+
+Profiling must not slow the orchestration/scheduler flow it instruments. The
+runtime off switch (`is_*_enabled()` returning `false`) is not free: even when
+off, each gate is a live `if (xxx_enable())` on the hot path. Proposal: let
+production inference and perf-measurement runs shed that residual cost.
+
+Why the residual cost is non-zero: each gate is an O(1) global-bool read
+(`return g_enable_pmu;`), but `is_pmu_enabled()`, `is_dep_gen_enabled()`, … are
+`extern "C"` / weak symbols defined in a **different translation unit** from the
+scheduler loop. The compiler **cannot inline or hoist them**, so each "off"
+check is still a real `bl`/`ret` — once per scheduler iteration, once per task
+submit.
+
+This is already acknowledged in-tree: a2a3 `scheduler_dispatch.cpp` caches the
+PMU gate at function scope with the comment *"is_pmu_enabled() is extern "C" and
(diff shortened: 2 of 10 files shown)


src/a2a3/runtime/tensormap_and_ringbuffer/aicpu/aicpu_executor.cpp
```diff
@@ -539,14 +539,14 @@ int32_t AicpuExecutor::run(Runtime *runtime) {
             }
 #endif
 
+#if PTO2_PROFILING
```
ChaoZheng109, line 542: 这里和上一个endif可以合并

ChaoZheng109: 算了之后专门开一个issue统一进行规范

ChaoZheng109 (approved):

PR #1169, opened 2026-06-26
fix(a5/profiling): defer collector device-buffer member publish past rollback guard (#1085)

### LINKED ISSUE #1085: [Code Health] a5 profiling collectors leave device-buffer member pointers dangling after a failed initialize()
### Category

### Component

Host Runtime

### Description

The a5 host-side profiling collectors assign their device-buffer **member pointers** mid-`initialize()`, before the success commit, while the `InitRollbackGuard` is still armed. If `initialize()` fails at any return point after that assignment, the guard frees all framework-tracked device buffers, leaving the member pointing at freed memory.

Surfaced by this review comment (on `l2_swimlane_collector.cpp:254`):
> If initialize fails after rotation_table_dev is allocated and assigned to aicore_ring_addr_table_dev_ (line 254), aicore_ring_addr_table_dev_ will be left pointing to the freed memory (since the rollback guard will free it). This can lead to undefined behavior if get_aicore_ring_addr_table_device_ptr() is called or if finalize is subsequently invoked.

**The dangling itself is confirmed real** (full chain verified on `main`):
- `alloc_paired_buffer` registers `rotation_table_dev` via `BufferPoolManager::register_mapping` (adds it to `dev_to_host_`).
- There are real fail-return points after line 254 and before `guard.commit()` (the `init_phase_pools` alloc failures).
- On those, the un-committed `InitRollbackGuard` destructor calls `release_all_owned`, which iterates `dev_to_host_` and frees every dev_ptr — including `rotation_table_dev`. The member is then dangling.

**It does NOT cause UB in the current call flow** (hence low severity) — both UB triggers the comment names are currently unreachable:
- `finalize()` early-returns on `shm_host_ == nullptr`, and `shm_host_` is set only at the very end of `initialize()` (after all fail-return points), so a failed init never reaches the `release_one_buffer(<member>)` path → no double-free.
- `get_..._device_ptr()` is only called by `device_runner` on init **success**.

So it is a latent footgun: a member left pointing to freed memory after a failed init. A future refactor (relocating the accessor, or changing the `shm_host_`-gated `finalize` early-return) could turn it into UB / double-free.

Not isolated to one collector — the same mid-init-assign pattern is shared across the a5 collectors, so it should be addressed uniformly. a2a3 is unaffected: its collectors do not use `InitRollbackGuard`, so there is no rollback-free → no dangling member.

Found during code review; not a runtime failure.

### Location

Both on `main`:
- `src/a5/platform/shared/host/l2_swimlane_collector.cpp:254` — `aicore_ring_addr_table_dev_` assigned mid-`initialize` (`shm_host_` set ~line 352 via `set_memory_context`, `guard.commit()` ~line 363). Introduced by #1058.
- `src/a5/platform/shared/host/pmu_collector.cpp:104` — `aicore_ring_addrs_dev_` assigned mid-`init`, fail-returns after, `finalize` guarded by the same `shm_host_ == nullptr` early-return.

### Proposed Fix

Either:
1. Defer the device-buffer member assignments to the post-`guard.commit()` region (next to `perf_shared_mem_dev_ = perf_dev_ptr` / `shm_dev_ = ...`), so a failed init leaves them at their default `nullptr`; or
2. Have `InitRollbackGuard` reset the registered member pointers to `nullptr` on rollback.

Apply uniformly across the a5 collectors that share this mid-init assignment pattern (l2_swimlane, pmu, and any sibling).

### Priority

---

Related (other profiling/collector resource-lifecycle code-health items, different root cause): #1082, #977

doraemonmj:
## Summary

Fixes #1085.

The a5 `L2SwimlaneCollector`, `PmuCollector`, and `DepGenCollector`
published device-buffer **members**, the memory context, and/or the
`initialized_` flag mid-`init()` while the `InitRollbackGuard` was still
armed. A later init failure frees those buffers on unwind (via
`release_all_owned` / the guard's `direct_ptrs_`), leaving the collector
pointing at freed memory with its "initialized" gate already set — so
`finalize()` skips its `!initialized_` early-return and double-frees /
UAFs what the guard released.

The failure is low-probability (the throwing call is a `std::bad_alloc`
from a `std::vector::assign` or from `set_memory_context` copy-assigning
its `std::function` members), but it is a real exception-safety hole, not
just a latent footgun.

## Fix (issue option 1, scoped to `src/a5/`)

Route the device buffers through locals and publish them, the memory
context, and the init gate to their members only **after**
`guard.commit()`, with the init gate published **last**. A failed init
then leaves the members at their defaults and the gate false, so
`finalize()` is a no-op.

- **l2_swimlane** — rotation table → `aicore_ring_addr_table_dev_`,
  `perf_shared_mem_dev_`, and `set_memory_context`.
- **pmu** — ring-address table dev ptr + paired host shadow + per-core
  ring vector (`aicore_ring_addrs_dev_` / `aicore_ring_addrs_host_` /
  `aicore_rings_dev_`), `shm_dev_`, `set_memory_context`, then
  `initialized_` last.
- **dep_gen** — `shm_dev_` / `shm_size_`, `set_memory_context`, then
  `initialized_` last. dep_gen has no dangling-member problem (its only
  buffer member is set after the last fail-return), but it published
  `initialized_` before commit with a throwing `set_memory_context` in
  between — the same `finalize()` double-free class — so it is hardened
  the same way to satisfy #1085's "apply uniformly across the a5
  collectors" ask.

The `guard.commit()` → publish window contains no return or throwing
call, so the success path is behaviorally identical.

## Scope check

- **a2a3** is unaffected — its collectors use no `InitRollbackGuard`, so
  a failed init leaks rather than dangles (separate, pre-existing,
  lower-severity concern).
- The shared **`src/common/`** collectors that *do* use the guard
  (`tensor_dump_collector`, `scope_stats_collector`) are already safe by
  construction: their buffer member / init gate is published after the
  last fail-return or with no rollback-able call between the publish and
  `guard.commit()`.

## Testing

Builds clean; CI (incl. `st-onboard-a5`) green. No regression test
added: the UB is reached only via an allocation-failure unwind that is
impractical to trigger deterministically (per discipline.md's "test
impractical" exception), so the fix is verified by inspection as an
equivalent transform on the success path.

### DIFF
--- a/src/a5/platform/shared/host/dep_gen_collector.cpp
+++ b/src/a5/platform/shared/host/dep_gen_collector.cpp
@@ -115,23 +115,26 @@ int DepGenCollector::init(
     // free_queue contents) to device.
     profiling_copy_to_device(shm_dev_local, shm_host_local, shm_size);
 
+    LOG_INFO_V0(
+        "DepGen collector initialized: %d threads, SHM=0x%lx (records held in memory until replay)", num_threads,
+        reinterpret_cast<unsigned long>(shm_dev_local)
+    );
+    guard.commit();
+    // Publish members + memory context only after the rollback guard is
+    // disarmed; initialized_ is published last. set_memory_context copy-assigns
+    // std::function members and can throw std::bad_alloc, so keeping it (and the
+    // initialized_ store) before commit would let the guard free the shm/buffers
+    // on unwind while initialized_ is already true — finalize() would then skip
+    // its !initialized_ early-return and double-free what the guard released.
+    // start(tf) gates on shm_host_ (published by set_memory_context), so this is
+    // the moment the collector becomes startable.
     shm_dev_ = shm_dev_local;
     shm_size_ = shm_size;
-    initialized_ = true;
-
-    // Re-set_memory_context now that the shm region is ready. start(tf) gates
-    // on shm_host_ being non-null, so this is the moment the collector
-    // becomes startable.
     set_memory_context(
         alloc_cb, register_cb, free_cb, profiling_copy_to_device_for_ops, profiling_copy_from_device_for_ops,
         shm_dev_local, shm_host_local, shm_size, device_id
     );
-
-    LOG_INFO_V0(
-        "DepGen collector initialized: %d threads, SHM=0x%lx (records held in memory until replay)", num_threads,
-        reinterpret_cast<unsigned long>(shm_dev_)
-    );
-    guard.commit();
+    initialized_ = true;
     return 0;
 }
 
--- a/src/a5/platform/shared/host/l2_swimlane_collector.cpp
+++ b/src/a5/platform/shared/host/l2_swimlane_collector.cpp
@@ -243,15 +243,19 @@ int L2SwimlaneCollector::initialize(
     // direct access to `&ac_state->head` device addresses, no
     // host-to-device translation needed). AICore reads
     // rotation_table[block_idx] at kernel entry.
+    // Held in a local and published to aicore_ring_addr_table_dev_ only after
+    // guard.commit() (see end of this function). The alloc registers the buffer
+    // in the rollback guard, so a later init failure frees it via
+    // release_all_owned; assigning the member here would leave it dangling.
+    void *rotation_table_dev = nullptr;
     {
         size_t table_bytes = static_cast<size_t>(num_aicore) * sizeof(uint64_t);
         void *rotation_table_host = nullptr;
-        void *rotation_table_dev = alloc_paired_buffer(table_bytes, &rotation_table_host);
+        rotation_table_dev = alloc_paired_buffer(table_bytes, &rotation_table_host);
         if (rotation_table_dev == nullptr) {
             LOG_ERROR("Failed to allocate l2_swimlane_aicore_rotation_table (rotation) table (%zu bytes)", table_bytes);
             return -1;
         }
-        aicore_ring_addr_table_dev_ = rotation_table_dev;
     }
 
     // Step 6: Initialize per-thread phase pools — both sched and orch. Each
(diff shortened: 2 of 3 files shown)


ChaoZheng109:
Nice fix — the l2/pmu reordering is correct and CI (incl. `st-onboard-a5`) is green.

One follow-on within this PR's scope: please also unify `dep_gen_collector.cpp` so all three a5 collectors follow the same hardened "publish after commit" pattern. Issue #1085 explicitly asked to "apply uniformly across the a5 collectors (l2_swimlane, pmu, and any sibling)", and dep_gen is the lone outlier left.

`DepGenCollector::is_initialized()` returns `initialized_`, and init publishes it **before** `guard.commit()`:

```cpp
// src/a5/platform/shared/host/dep_gen_collector.cpp
shm_dev_  = shm_dev_local;     // 118
shm_size_ = shm_size;          // 119
initialized_ = true;           // 120  <-- published while guard still armed
set_memory_context(...);       // 125  <-- copy-assigns std::function members
guard.commit();                // 134
```

`set_memory_context` copy-assigns several `std::function` members, which can throw `std::bad_alloc`. If that fires after line 120, the guard (un-committed) frees the shm + every `DepGenBuffer` on unwind, but `initialized_` is already `true`. A subsequent `finalize()` does `if (!initialized_) return;`, so it does **not** early-return and double-frees / UAFs the buffers the guard already released. This is exactly the failure mode the l2/pmu reorder in this PR eliminates by publishing `initialized_` last.

dep_gen does *not* have the dangling-member problem (its only buffer member `shm_dev_` is set after the last fail-return), so this is the only residual exposure — but it's the same class, and leaving it makes dep_gen inconsistent with the now-hardened siblings.

Minimal change (mirror the pmu pattern — log up, publish after commit, `initialized_` last):

```cpp
    profiling_copy_to_device(shm_dev_local, shm_host_local, shm_size);

    LOG_INFO_V0(
        "DepGen collector initialized: %d threads, SHM=0x%lx (records held in memory until replay)",
        num_threads, reinterpret_cast<unsigned long>(shm_dev_local)
    );
    guard.commit();
    // Publish members + memory context after the guard is disarmed; initialized_ last.
    shm_dev_  = shm_dev_local;
    shm_size_ = shm_size;
    set_memory_context(
        alloc_cb, register_cb, free_cb, profiling_copy_to_device_for_ops, profiling_copy_from_device_for_ops,
        shm_dev_local, shm_host_local, shm_size, device_id
    );
    initialized_ = true;
    return 0;
```

Behaviorally identical on the success path; closes the issue's "uniformly" ask in one go.

---

Scope check while here (confirms the PR's "a2a3 unaffected" claim, with the precise reason):

- **a2a3's own collectors** (l2_swimlane / pmu / dep_gen) use no `InitRollbackGuard`, so there is no rollback-free path → no dangling member / double-free (a failed init leaks rather than dangles — separate, pre-existing, lower-severity concern). Confirmed unaffected.
- **The shared `src/common/` collectors that *do* use the guard** (used by both arches) are already safe by construction:
  - `tensor_dump_collector` (gate `shm_host_`): `dump_shared_mem_dev_` is published at L153, after the last fail-return (L132) — guard can't roll back with the member already published.
  - `scope_stats_collector` (gate `initialized_`): `initialized_ = true` at L123 is before `guard.commit()` (L125), but there is no return/throwing call between them, and `set_memory_context` runs *after* commit (L130) — so no rollback can occur with `initialized_` already true. Safe, though cosmetically it's the same "publish before commit" shape; optional to tidy for consistency.

So `dep_gen` is the only collector with a genuine (if extremely-low-probability) residual exposure, because its `set_memory_context` sits *between* `initialized_ = true` and `guard.commit()`.

doraemonmj:
Thanks for the precise catch — confirmed and fixed. `dep_gen` now mirrors the pmu pattern: log moved up to use the local, members + `set_memory_context` published after `guard.commit()`, and `initialized_` published last, so an unwind that frees the shm/buffers can no longer leave `initialized_ == true` for `finalize()` to double-free. Your scope check on a2a3 (no `InitRollbackGuard`) and the shared `src/common/` collectors matches what I see, so no further changes there. All three commits are squashed into one and the PR description is updated to reflect the dep_gen change.

ChaoZheng109 (approved):

PR #1172, opened 2026-06-26
Qwen3 test shapes for spmd paged attention highperf

MirkoDeVita98:
```bash
python tests/st/a2a3/tensormap_and_ringbuffer/spmd_paged_attention_highperf/test_spmd_paged_attention_highperf.py \
  -p a2a3 \
  --manual only \
  --case b16_h40_kv8_s16384_bs128_fp16
```

### DIFF
--- a/tests/st/a2a3/tensormap_and_ringbuffer/spmd_paged_attention_highperf/test_spmd_paged_attention_highperf.py
+++ b/tests/st/a2a3/tensormap_and_ringbuffer/spmd_paged_attention_highperf/test_spmd_paged_attention_highperf.py
@@ -298,6 +298,38 @@ class TestSpmdPagedAttentionHighPerf(SceneTestCase):
                 "dtype": "float16",
             },
         },
+        {
+            "name": "b16_h40_kv8_s8192_bs128_fp16",
+            "manual": True,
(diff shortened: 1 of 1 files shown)


ChaoWao:
Closing — onboard results on the Qwen3 batch-16 shape don't meet the perf bar.

**Tested (a2a3, tensormap_and_ringbuffer):** `b16_h40_kv8_s3584_bs128_fp16`
(batch 16, num_heads 40, kv_heads 8, head_dim 128, kv_seq 3584 ≈ 3.5k,
block_size 128, fp16), 10 rounds. Golden **PASSED** — the shape is correct.

**Effective (device) wall time ≈ 1.40 ms/step steady-state:**

| Metric | Device (effective) |
| ------ | ------------------ |
| steady-state avg (rounds 1–9) | **1404 µs** |
| median | 1.41 ms |
| min / max | 1.26 ms / 1.59 ms |
| round 0 (JIT + first-launch warmup, excluded) | 13.3 ms |

That is **~5.2× slower** than the Qwen3 target of ~270 µs/step, so the
high-perf SPMD paged-attention kernel isn't ready to carry these Qwen3
shapes as tracked test cases yet. Closing until the kernel closes the gap;
the shapes can be re-added once perf lands near the 270 µs target.

PR #986, opened 2026-06-04
Paged Attention benchmark vs pypto

MirkoDeVita98:
# Paged Attention Benchmark

```bash
python3 paged_attention_benchmark_vs_pypto/benchmark_spmd_paged_attention_compare.py \
  -p a2a3 \
  -d 6 \
  --rounds 50 \
  --warmup 5 \
  --shape b1_h32_kv8_s4096_bs128_fp16
```

```text
== summary ==
highperf host e2e avg:        127563.3 us
highperf runtime device avg:  1495.3 us
highperf device-log total:    60.2 us
highperf device-log orch:     7.3 us
highperf device-log sched:    60.2 us
PyPTO host e2e avg:           813021.3 us
PyPTO device-log total:       1483.8 us
PyPTO device-log orch:        719.8 us
PyPTO device-log sched:       1483.8 us
PyPTO / highperf host e2e:    6.37x
```

PR #1173, opened 2026-06-26
Integrate TraCR as the runtime/kernel profiler for Simpler

noabauma:
Upstreams **[TraCR](https://github.com/huawei-csl/TracR)** into Simpler as a
first-class profiler, replacing the separate development branch. No more
out-of-tree branch to rebase — TraCR lives in `tools/tracr/` and is wired into
the `tensormap_and_ringbuffer` runtime build behind `BUILD_TRACR=ON`.

## What is TraCR

A lightweight, nanoscale C++ instrumentation library. Markers store fixed
16-byte records into a **lock-free per-thread ring buffer**; on device the
timestamps come from the **hardware counter**, and the buffers are downloaded
after a run to `~/ascend/tracr_<N>/` and post-processed (`tracr_process`) into
**Perfetto** / Paraver timelines. It supports **multi-device** profiling today.

It captures the full on-device picture: the **AICPU scheduler threads**
(orchestration / scheduling / phase / drain markers) **and** per-core task
execution on the **AICube / AIVector** units.

Perfetto view:
Image

## The selling point: overhead lands *outside* the region you measure

For fine-grained branch-to-branch tuning (does optimization X shave a few % off
the scheduler / kernel?) the profiler's *own* cost must be smaller than the
effect you're chasing. It isn't, with the current L2 swimlane collector.

Measured on `Case1` paged attention (a2a3 onboard, device-clock timing of
`aicpu_execute`, vanilla baseline **~39.8 ms**):

| Profiler | `aicpu_execute` overhead | Where the cost lands |
| --- | --- | --- |
| **L2 swimlane** (level 4) | **+3.7 ms (+9%)** | **inside** the scheduling loop |
| **TraCR** | **~0 ms in-loop** (+0.5–0.9 ms one-shot dump) | **outside** the scheduling loop |

**When you're hunting a 2–5% improvement between branches, swimlane's own +9%
perturbation is larger than the effect you're trying to measure — and it lands
directly on the region under test. TraCR leaves that region within noise of
vanilla.** That is the whole reason TraCR is needed.

## Why the current profiler (L2 swimlane) isn't enough

- **It perturbs the very region it measures.** Full swimlane (level 4) raised
  `aicpu_execute` from **~39.8 ms → ~43.5 ms (+3.7 ms, ~9%)**, and that cost is
  *inside* the scheduling loop. It's front-loaded (heaviest until the buffers
  fill), so it distorts *relative* comparisons rather than adding a constant
  offset.
- **It is lossy by design.** Swimlane records per-task timings into bounded,
  rotating device ring buffers that **cap at 2¹⁶ records per stream and silently
  drop the rest**. On `Case1` that ceiling is ≈ *one round's* worth of tasks:
  the report contained exactly `65,536` aicore-task + `65,536` aicpu-task records
  (both pinned at the cap). You cannot attribute a small delta to a branch change
  on data that's partially dropped.
- **The "zero-overhead" appearance is the data loss, not efficiency.** Over many
  iterations swimlane looks free only because it hits the cap and stops
  collecting — the extra rounds contribute nothing because they're dropped.
- **The always-on part is coarse.** What's genuinely always-on (independent of
  the flag) is just ~5 phase spans (`device_wall` / `orch` / `sched`) — a
  bird's-eye split, not fine-grained kernel/scheduler timing.

### TraCR on the same run

- **Overhead sits *outside* the measured region.** Timing between `TRACR_START()`
  and `TRACR_FINALIZE()`, the scheduling region is **within noise of vanilla**
  (~40.4 ms vs ~39.7 ms). TraCR's entire cost is a single deferred **~0.5–0.9 ms
  trace-buffer dump at finalize** — it does *not* inflate the hot path it's
  measuring.
- **It collects more, at higher fidelity, with no drops.** The same `Case1` round
  produced **330,833 TraCR records** (orchestrator thread `263`; the three
  scheduler threads `~107k–117k` each) — ~2.5× swimlane's *capped* volume — at
  16 B / nanosecond HW-counter resolution, into a 1 M-per-thread buffer (no
  truncation). Cheap append on the hot path (`_traces[idx++] = payload`), one
  bulk ~5 MB copy to HBM at the end.

## Why TraCR

- **Multi-device.** Profiles multi-device runs today!
- **Extremely low, hot-path-clean overhead.** Compile-time gated (zero-cost stubs
  when off), no locks, fixed-size records, HW-counter timestamps. It does **not**
  write to the AICPU `device_log` on the hot path, so it avoids the
  op-serialization / op-execute-timeout pitfalls of log-based profiling
  (`PTO2_PROFILING`). Measured cost is confined to a one-shot post-run buffer
  dump *outside* the scheduling region (see above), vs. the L2 swimlane
  collector's ~9% in-loop perturbation.
- **Right level of detail.** Profiles at both the **Simpler runtime** (scheduler
  orchestration) and **kernel** level — the visibility needed to profile PyPTO /
  Simpler kernels.
- **Easy to use.** One build flag, run as usual, one post-process command → a
  Perfetto trace you open in `ui.perfetto.dev`.
- **Demand.** Many users are asking to profile their kernels; this gives them a
  supported, in-tree tool instead of a side branch.

## Where it's already been used

- **qwen3-14B decode attention** (`fa_fused` + `online_softmax`) in pypto-lib —
  captured the full decode-layer timeline (AICPU + 24 AICube + 48 AIVector) (see [pypto-lib #607](https://github.com/hw-native-sys/pypto-lib/issues/607).
- **Optimized pypto vs CCE paged attention** — the follow-up to the
  6x in #986 , comparing the
  optimized attention against the CCE highperf kernel
  #899 at matched shapes.
- DeepSeek V4 CSA decode-attention kernel [pypto-lib #590](https://github.com/hw-native-sys/pypto-lib/issues/590)

- General PyPTO / Simpler kernel profiling.

## Future work

- **A5** platform support — pending A5 hardware access (coming).

Repo: <https://github.com/huawei-csl/TracR>


ChaoWao:
Thanks for the substantial work here, @noabauma — the profiler design is sound and the write-up/measurements are genuinely helpful. After review, we've decided **not to land TraCR as a separate, parallel profiler**. We'd rather **improve the profiling we already have (the L2 swimlane collector)** than carry a second, independent profiling stack in-tree — one path to maintain, document, and wire through CI.

Two hard constraints shape this, and they're why a big-buffer / dump-once design can't simply replace the swimlane:

**1. Production runs must rotate — no large hold-everything buffer.**
TraCR's headline win (near-zero hot-path cost) comes from appending to a large per-thread buffer (1M records ≈ 16 MB/thread) and copying it out once at finalize. That only works because the *whole run* fits in the buffer. In production — long-running, multi-stream, memory-constrained — we can't reserve that; the profiler **must rotate bounded buffers** and drain them during the run. That is exactly what the swimlane does, and it's the source of the per-record ordering cost. So the goal isn't to adopt the unbounded-buffer model — it's to make **rotation itself cheaper**: e.g. the redundant AICPU per-record `wmb()` that #1301 removes, batching publication, and attacking the real cost, the AICore-side per-record write-back.

**2. It must not capture less than the swimlane does.**
The 16-byte marker payload is strictly less structured than the swimlane records. As instrumented here it drops:
- per-task **unique identity** (`task_token_raw`, the full ring+local id) — the marker's 32-bit `extraId` only carries `func_id`, i.e. *which kernel*, not *which task instance*;
- the **AICPU↔AICore 1:1 join** (`reg_task_id`) needed to stitch a task's dispatch side and execute side under SPMD / MIX / dual-issue;
- per-phase scheduler **telemetry** (`tasks_processed`, `pop_hit`/`pop_miss`, per-shape ready-queue depths);
- **true on-core** start/end timestamps — the current markers are AICPU-observed dispatch/complete, not AICore-measured execution.

A replacement that profiles *less* than the swimlane isn't an option for us.

If you're up for it, we'd welcome PRs that improve the swimlane **in place** — cheaper rotation/publication, and wider (but still bounded, rotating) buffers to relieve the 2¹⁶ truncation you rightly flagged — without dropping any of the fields above. Happy to help scope that; #1301 is a first concrete step in that direction. Closing this in favor of it.

PR #1222, opened 2026-06-30
Fix: release _ChipWorker handle in Worker.close() L2 branch

### LINKED ISSUE #1221: [Code Health] Worker.close() L2 branch leaks _ChipWorker nanobind instance at interpreter shutdown
### Category

### Component

Host Runtime

### Description

`Worker.close()` is contracted to release every resource the Worker holds. The
L2 branch violates this: it calls `self._chip_worker.finalize()` but never drops
the Python reference (`self._chip_worker = None`). The `_ChipWorker` nanobind
instance therefore stays alive on the closed `Worker` object.

This is inconsistent with the two sibling teardown paths, both of which *do*
drop the handle:

- the L>=3 branch sets `self._worker = None` right after `self._worker.close()`;
- the error/abort path already does `self._chip_worker = None` after `finalize()`.

**Observed symptom (CI, intermittent):** nanobind prints a reference-leak dump at
interpreter shutdown, e.g.

```
nanobind: leaked 1 instances!
 - leaked instance 0x... of type "_task_interface._ChipWorker"
nanobind: leaked 15 types!
nanobind: leaked 165 functions!
nanobind: this is likely caused by a reference counting issue in the binding code.
```

(The full types/functions list is dumped because nanobind cannot cleanly unload
the module while any one of its instances is still live.)

**Why it is intermittent / "sometimes":** when a pytest case *fails or errors*,
pytest retains that case's traceback for reporting, and the traceback strongly
references the failing frame's locals — including the `worker` object, which in
turn pins `_ChipWorker`. Those references survive until interpreter exit, where
nanobind's leak check runs and reports them. Passing runs release the locals
normally, so no dump appears. `tests/st/aicore_op_timeout/test_aicore_op_timeout.py`
is a frequent trigger because it asserts on a timing-sensitive 507xxx code and an
`elapsed < 10` bound that can fail on a busy/shared box.

This is a benign teardown-ordering artifact (not a runtime C++ leak), but it is
noisy in CI logs and masks any future *real* nanobind refcount regression.

### Location

- \`python/simpler/worker.py\` — \`Worker.close()\`, L2 branch (\`if self.level == 2:\`), the \`self._chip_worker.finalize()\` line.

For reference, the consistent siblings:
- \`python/simpler/worker.py\` — L>=3 branch: \`self._worker = None\` after \`self._worker.close()\`.
- \`python/simpler/worker.py\` — error/abort path: \`self._chip_worker = None\` after \`finalize()\`.

### Proposed Fix

Drop the handle in the L2 branch immediately after finalizing, mirroring the
other two paths:

```python
if self.level == 2:
    if self._chip_worker:
        self._chip_worker.finalize()
        self._chip_worker = None
```

This releases the `_ChipWorker` instance as soon as the Worker is closed, so it
no longer outlives the module even when a failing test's traceback pins the
`Worker` object. One line; aligns `close()` with the L>=3 branch and the error
path.

### Priority

### Environment

- Git commit: `11d03d9b81e1d29162eb30b7b39386842328559d`
- Host platform: Linux (aarch64)

Related: #1082, #980, #1018, #824 (Worker lifecycle / cleanup — distinct root causes).

ChaoZheng109:
## Summary

- `Worker.close()` L2 branch called `self._chip_worker.finalize()` but never dropped the Python reference, leaving the `_ChipWorker` nanobind instance alive on the closed Worker.
- Set `self._chip_worker = None` after `finalize()`, mirroring the L>=3 branch (`self._worker = None`) and the error path (which already does this).

## Why

When a pytest case fails, pytest retains its traceback, which strongly references the failing frame's `worker` local and through it the `_ChipWorker` instance — surviving until interpreter exit, where nanobind's leak check fires:

```
nanobind: leaked instance ... of type "_task_interface._ChipWorker"
nanobind: leaked 15 types! / leaked 165 functions!
```

(nanobind dumps the whole types/functions list because it cannot cleanly unload the module while any one instance is live.) This is a benign teardown-ordering artifact, not a runtime C++ leak, but it is noisy in CI and masks future real refcount regressions. Dropping the handle on close releases the instance so it no longer outlives the module.

## Testing

- [x] Simulation tests pass — `examples/workers/l2/vector_add` on `a2a3sim` (exercises L2 register/init/run/close), no leak dump.
- [ ] Hardware tests pass (not run; teardown-only change, sim-covered).

Closes #1221

### DIFF
--- a/python/simpler/worker.py
+++ b/python/simpler/worker.py
@@ -3915,6 +3915,7 @@ class Worker:
         if self.level == 2:
             if self._chip_worker:
                 self._chip_worker.finalize()
+                self._chip_worker = None
         else:
             if self._worker:
                 self._worker.close()


ChaoWao (approved):

yanghaoran29:
*How to Trigger a Genuine Local Dump*
nanobind registers the leak checker routine `internals_cleanup` via **`Py_AtExit`** (`nb_internals.cpp:562`). This routine runs at the **very end** of interpreter finalization — after module dictionary cleanup and the mandatory garbage collection pass enforced by CPython during shutdown. This leads to the following behavior:
- Instances referenced solely by ordinary global variables are released during module cleanup before the leak checker executes, so no leak dump is generated.
- Self-referential cycles combined with `gc.disable()` are also ineffective, as the forced GC triggered at finalization ignores the `gc.disable()` flag.

To reproduce a real leak dump, the target instance must **survive the entire finalization workflow**. The tracebacks generated by failed test cases can trigger this leak because they pin stack frames and create persistent references that CPython cannot collect automatically. I use an equivalent method — `ctypes.pythonapi.Py_IncRef(worker)` — to "immortalize" the worker object: its reference count will never drop to zero, making it unreclaimable by CPython. This faithfully replicates the edge case where the worker outlives module teardown logic. The test script can be found at `/tmp/leaktest/nb_dump.py`.

*Before the Fix (Buggy Version — `self._chip_worker = None` Removed)*
```
[nb_dump] worker immortalized (Py_IncRef); exiting -> nanobind Py_AtExit leak check follows
nanobind: leaked 2 instances!
 - leaked instance 0xfffdaeed9f98 of type "_task_interface.ChipCallable"
 - leaked instance 0xfffdaf0fb648 of type "_task_interface._ChipWorker"
nanobind: leaked 15 types!
 - leaked type "_task_interface._Worker"
 - leaked type "ArgDirection"
 - leaked type "_task_interface.RuntimeEnv"
 - leaked type "_task_interface._ChipWorker"
 - leaked type "_task_interface.ChipCallable"
 - ... skipped remainder
nanobind: leaked 165 functions!
 - ... skipped remainder
nanobind: this is likely caused by a reference counting issue in the binding code.
See https://nanobind.readthedocs.io/en/latest/refleaks.html
```
This output exactly matches the leak dump attached to the issue ticket (`leaked 15 types! / leaked 165 functions!`), and explicitly includes the target instance `_task_interface._ChipWorker`.

*After the Fix (Corrected Version — Exact Code from PR HEAD)*
```
nanobind: leaked 1 instances!
 - leaked instance 0xfffd9b1cd128 of type "_task_interface.ChipCallable"
nanobind: leaked 15 types!
 - leaked type "_task_interface._ChipWorker"
nanobind: leaked 165 functions!
nanobind: this is likely caused by a reference counting issue in the binding code.
```
The **`_ChipWorker` instance is removed entirely from the leak list** (instance count reduced from `2 instances` to `1 instances`) — this is the exact intended outcome of the PR.

> Note: Both states are functionally valid. The residual leaks are benign teardown artifacts and do not impact computational correctness:
> `[vector_add] golden check PASSED`, `max |host_out - expected| = 0.000e+00`.

*Further Experimental Test: Can Clearing Registries Eliminate the Dump Completely?*
Building on the current PR, I ran an experimental test by adding extra logic to clear nanobind handle references held at the Worker layer within the L2 branch of `close()`:
```python
if self.level == 2:
    if self._chip_worker:
        self._chip_worker.finalize()
        self._chip_worker = None
        self._callable_registry.clear()   # Newly added line
        self._identity_registry.clear()   # Newly added line
        self._live_handles.clear()        # Newly added line
```
When re-running the identical `nb_dump.py` script (with the worker still immortalized via `Py_IncRef`), **zero `nanobind: leaked` logs appear in stderr**, while the golden result check `golden check PASSED` remains unaffected.

## Three-Way Comparison (Official nanobind Exit Leak Check)
| State | Operations Performed in `close()` L2 Branch | nanobind Exit Leak Check Result |
| --- | --- | --- |
| Pre-fix baseline | Only call `finalize()` | `leaked 2 instances!` → `ChipCallable` + **`_ChipWorker`**, plus 15 types / 165 functions leaked |
| **Current PR implementation** | `finalize()` + set `_chip_worker=None` | `leaked 1 instances!` → only `ChipCallable` remains; 15 types / 165 functions still leak (the `_ChipWorker` leak is resolved, residual noise persists) |
| Current PR + registry clearing | Add `_callable_registry/_identity_registry/_live_handles.clear()` on top of PR logic | **Zero leak logs printed; the dump is fully eliminated 100%** |

In short: The module can only be unloaded cleanly with all leak noise eliminated if **all nanobind handles retained by the Worker after `close()` are fully released**. This PR addresses just one link in that chain: the `_ChipWorker` reference.

> Merge Tradeoff Note: The semantic contract of `close()` is "release all held resources". The routines `_cleanup_l3_l2_regions()` and `finalize()` already execute before `close()` reaches this branch, so clearing these dictionary registries is logically consistent. For rigor, prior to merging this extended cleanup logic, we must verify there are no code paths that access `_callable_registry` / `_identity_registry` / `_live_handles` after `close()` completes — this can be captured as a dedicated regression test requirement for this modification.

Issue #1249, opened 2026-07-02
[Code Health] Harden a5 DFX scene tests: PMU artifact validation gaps + missing st-sim-a5 smoke coverage


doraemonmj:
### Category

Robustness (potential edge-case failure)

### Component

Tests

### Description

During review of the commit that mirrors the a2a3 DFX scene tests to a5, four
follow-up hardening points surfaced. None block the mirror (it is a faithful
copy), but each weakens the coverage the DFX tests are supposed to provide.

**A. PMU artifact validation is too lenient** (issues 1–3 below live in
`test_pmu.py::_validate_pmu_artifact`; the a2a3 and a5 files are byte-identical
apart from arch strings, so all three exist in **both** and should be fixed in
both to avoid divergence):

1. **Silent skip on capture failure.** `if not matches: return` — when PMU
   capture fails so badly that no output directory is produced, the glob
   matches nothing and the method returns without any assertion, so the test
   *passes* on a real capture regression instead of failing.

2. **Stale-directory risk.** `matches[-1]` picks the most-recent directory by
   mtime but never confirms it belongs to the current test invocation. A
   directory left over from a prior run/session with the same case label
   satisfies the glob and gets validated in place of (or in addition to) the
   current run's real output — masking a current-run capture regression.

3. **Header ordering not verified.** The docstring claims the smoke asserts the
   header "starts with the documented prefix" (leading, ordered), but the loop
   only checks `col in header_cols` (membership). A header with the same columns
   in a different order passes incorrectly. Either tighten the check to a
   leading ordered-prefix comparison, or correct the docstring to match the
   membership-only intent.

**B. CI coverage gap — `st-sim-a5` runs no DFX per-feature smoke steps.**
`.github/workflows/ci.yml`'s `st-sim-a5` job runs only the default
`pytest examples tests/st --platform a5sim ...` with no `--enable-*` flags.
The a5 DFX cases are no-ops without their flag, so they are **collected but not
validated** in sim CI. By contrast `st-sim-a2a3` mounts four dedicated steps
(`dep_gen`, `l2_swimlane`, `PMU`, `args_dump`). Note the onboard job already
mirrors all four a5 smokes — the gap is **only on the sim side**.

### Location

```markdown
- `tests/st/a5/tensormap_and_ringbuffer/dfx/pmu/test_pmu.py:94-96` (silent skip + stale dir)
[...]

PR #1254, opened 2026-07-02
Harden DFX artifact validation: bind to the current run

doraemonmj:
## What

The DFX scene tests locate their output directory by globbing
`outputs/<label>_*` and taking the newest by mtime, without confirming it
belongs to the current test invocation. Two problems, both from issue #1249's
section A:

- **Stale-directory (A.2):** a leftover `outputs/<label>_<old-ts>/` from a
  prior run/session with the same case label satisfies the glob and gets
  validated in place of (or instead of) the current run's real output —
  masking a current-run capture regression.
- **Silent-skip (A.1):** `_swimlane_validate.py` and `dep_gen/test_dep_gen.py`
  returned without asserting when the glob matched nothing, so a total capture
  failure (no output dir produced) passed as green.

## How

Capture a `time.time()` marker before `super().test_run(...)`, thread it into
the validators, and keep only directories whose `mtime >= marker`; then
`assert` at least one such directory exists. Fixed in lockstep across the
a2a3 and a5 mirrors so they stay byte-identical modulo arch strings:

- `pmu`, `dep_gen`, `dep_gen_chain`, `args_dump`, `scope_stats`,
  `l2_swimlane` (+ `mixed`)
- shared `_swimlane_validate.validate_perf_artifact` now takes a `since=` marker

## PMU header check (A.3)

Kept membership-only. Confirmed nothing in the repo reads `pmu.csv`
positionally (the only consumer is the test itself), so enforcing column
order would add brittleness without catching a real break. Instead the
docstring is corrected to state column order is not enforced — resolving the
doc/impl mismatch without tightening behavior.

## Testing

All six DFX smokes pass on `a2a3sim` (silicon-agnostic) with their
`--enable-*` flags. The marker-filter + assert paths are exercised by real
captures.

## Scope

Refs #1249. Section B (adding the DFX smoke steps to the `st-sim-a5` CI job)
is intentionally left out of this PR.

### DIFF
--- a/tests/st/a2a3/tensormap_and_ringbuffer/dfx/args_dump/test_args_dump.py
+++ b/tests/st/a2a3/tensormap_and_ringbuffer/dfx/args_dump/test_args_dump.py
@@ -20,6 +20,7 @@ use the unified schema, and no legacy args-only manifest is emitted.
 import json
 import subprocess
 import sys
+import time
 
 import torch
 from simpler.task_interface import ArgDirection as D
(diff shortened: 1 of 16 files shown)


doraemonmj:
close #1249

ChaoZheng109 (approved):

PR #1259, opened 2026-07-02
Refactor: remove orphaned WorkerManager idle-picker methods

ChaoWao:
## Summary

`#1009` (remote endpoint eligibility scheduling) migrated the scheduler's worker selection from `pick_idle_excluding` to `pick_idle_excluding_eligible`, but left the old entry points behind. This removes the migration residue. Git history — not just a zero-caller grep — confirms each is dead, not in-progress scaffolding:

- **`pick_idle_excluding`** — `#1009`'s diff rewrote the scheduler's only call site to `pick_idle_excluding_eligible`, then kept the old method as a thin forwarding shim (`return pick_idle_excluding_eligible(type, exclude, unconstrained)`). Nothing calls the shim.
- **`pick_idle` / `pick_n_idle`** — base variants with **no call site anywhere in the repo's history** (pickaxe on `pick_idle(`/`pick_n_idle(` finds only declaration/definition commits since `#587`).
- **`shutdown_children`** — `WorkerManager` method with zero call sites, distinct from the live `WorkerThread::shutdown_child` it wrapped.

All four are zero-call, host-only, non-virtual, and not part of any wire protocol or serialized struct.

The `scheduler.h` data-flow comment that named `pick_n_idle` is updated to reference the actual dispatch path (`pick_idle_excluding_eligible`).

**Deliberately kept:** `pick_idle_excluding_eligible` (the live selector, called from `scheduler.cpp:375`) and `WorkerThread::shutdown_child` (8 live call sites).

Net: **−42 lines**, 3 files, no behavior change.

## Testing

- [x] Both runtimes build + link clean under `-Wall -Wextra`, no warnings (a2a3/a5 × host_build_graph/tensormap_and_ringbuffer)
- [x] `WorkerManager`-referencing Python test passes (`test_mailbox_atomics` 6/6)
- [ ] cpput `hierarchical` suite — a pre-existing gtest ABI link error in the local tree (reproduces on clean `upstream/main`, unrelated to this change) blocks local run; CI builds cpput in a clean container and covers it.


ChaoWao:
Heads-up for reviewers: the red \`ut (ubuntu)\` / \`ut (macos)\` checks are a **pre-existing CI flake, not caused by this PR** — they \`cancel\` because \`test_scheduler\` (cpput test #5) hangs into the job's 15-min cap.

Evidence it's unrelated to this change (a behavior-preserving rename + zero-caller deletions):
- The unrelated \`l3-l2-orch-message-queue\` PR hit the *identical* hang at the same test (run 28584354306).
- Both this PR's commits (deletion-only, and rename) hung at the same test, same timestamp pattern.
- The rename (\`pick_idle_excluding_eligible\` → \`pick_idle\`) has no default args, so no silent overload-resolution shift; everything that compiles has byte-identical runtime behavior.

Tracked in #1260. All other 14 checks (incl. \`ut-a2a3\`, \`ut-a5\`, both \`st-onboard\`, \`st-sim\`, \`pre-commit\`, \`packaging\`) are green.

PR #1264, opened 2026-07-03
Refactor: move fanout wiring to orchestrator, drop wiring queue

ChaoWao:
## Summary

Moves all task fanout wiring from the scheduler's deferred wiring queue onto the **orchestrator**, and deletes the wiring-queue subsystem entirely.

- The orchestrator now wires each task inline at submit time: `dep_pool.ensure_space` (block / reclaim / overflow-detect) → `wire_task`. This mirrors the `fanin_pool.ensure_space` already on the submit path.
- `dep_pool` becomes **orchestrator-exclusive (single writer) → no lock needed**. The `dep_pool_lock` added in #1263 was only required by the hybrid O+S design; a single-wirer model is lock-free. The only cross-thread serialization remains the per-producer `fanout_lock`, shared with the scheduler's completion handler.
- **Deleted:** `PTO2SpscQueue`, `WiringState`, `drain_wiring_queue`, `report_wiring_deadlock`, `producer_blocked`, `orch_needs_drain`, `dep_deadlock_reported`, and the Wire swimlane phase.
- Forward-progress safety moves from the S-side structural wiring-deadlock detector to `PTO2DepListPool::ensure_space`'s existing no-progress overflow detector (same `PTO2_ERROR_DEP_POOL_OVERFLOW` fatal).

## Why

A single wirer gives a simpler model — lock-free dep_pool, no SPSC queue, no batch/backoff machinery. The benchmark below tests whether the wiring queue's batching was still pulling its weight.

## Measured impact (a2a3 onboard, device 2, 100 rounds, trimmed mean; vs mainline `c4cfc968`)

| Example | Effective Δ | Device Δ | Orch Δ |
|---|---:|---:|---:|
| qwen3_14b_decode (StressBatch16Seq3500) | **−0.1%** | −0.4% | +11.2% |
| paged_attention_unroll (Case1 / Case2) | −1.1% / −0.5% | ~flat | +14.7% / +2.8% |
| paged_attention_unroll_manual_scope (Case1 / Case2) | −0.8% / +0.8% | ~flat | +11.4% / +6.2% |
| alternating_matmul_add (Case1) | **+4.6%** ⚠️ | +2.7% | +4.7% |
| benchmark_bgemm (Case0) | **+9.4%** ⚠️⚠️ | +5.6% | +13.8% |
| batch_paged_attention (Case1) | **+5.6%** ⚠️ | +4.4% | +9.2% |

**Neutral on S-bound workloads (incl. qwen3); regresses O-bound ones (where `Orch ≈ Effective`) by 4.6–9.4%.** Orch rises everywhere because the wiring work shifts S→O; where the orchestrator was already the bottleneck, that increase lands directly on Effective. The wiring queue's batching was load-bearing for those cases.

This is opened to get CI/review input on the **full-removal** approach versus the **hybrid** in #1263 (keep the queue + add O-side fast paths), which avoids these regressions. It is **not a clear net win as-is.**

## Testing

- [x] a2a3sim `tensormap_and_ringbuffer` suite — 30 passed, 1 skipped.
- [x] Builds clean (a2a3 sim + onboard runtimes); clang-format + markdownlint pass.
- [ ] cpput (hardware-gated) — pending CI.
- [x] Onboard benchmark vs mainline (table above); `test_spsc_queue.cpp` (a2a3) removed, drain cases dropped from `test_wiring.cpp` (wire_task cases retained).

## Known caveat

`PTO2_SERIAL_ORCH_SCHED=1` (default-off, debug-only) now requires a graph's in-flight fanout to fit in dep_pool, since no task completes during orchestration to reclaim it. Documented in `RUNTIME_LOGIC.md`.


ChaoWao:
Closing as superseded by #1263 — it is a strict superset (a2a3 + a5, dep and fanin wiring moved to O, scheduler_timeout sim-test fixed, dep_pool-overflow deferral folded in, rebased on current main, CI-green). This PR was the minimal variant opened to measure the queue-removal design (3-way benchmark confirmed identical perf), so it has served its purpose.

PR #1269, opened 2026-07-05
Early-dispatch: align MIX selection with normal dispatch + staging/trigger cleanups

ChaoWao:
Two early-dispatch (`try_speculative_early_dispatch` / `stage_consumer_blocks`)
changes. **Normal dispatch (`dispatch_shape`) and the Phase-4b early-dispatch
trigger gate are untouched.**

## 1. Align MIX cluster selection with normal dispatch

Early dispatch selected MIX clusters via `get_idle_core_offset_states(MIX)` /
`get_pending_core_offset_states(MIX)`, which inspect all three cores of a
cluster and **ignore the task's `active_mask`**. Normal dispatch uses
`classify_mix_cluster(core_mask)`, which scopes the check to the cores the task
actually uses. So a **partial** MIX (e.g. 1C1V = AIC+AIV0) whose *unused* AIV
was busy was placed by normal dispatch but stranded by early dispatch.

Early dispatch now selects MIX clusters with an **inline one-pass loop over the
same `classify_mix_cluster` predicate** (no new helper — mirrors normal
dispatch's inline style), routing each cluster to the running or pending
bucket. Selection only; placement stays whole-cluster.

## 2. Early-dispatch staging cleanups

- **Batch-pop** the `early_dispatch_queue` (`pop_batch` up to DRAIN_MAX)
  instead of one pop per consumer, mirroring `pop_ready_tasks_batch`; the
  unprocessed remainder is pushed back for peers via `push_batch`.
- **Single `wmb()`**: stage both buckets (idle -> running, pend -> gated
  pending) in one prepare pass + one barrier + one publish loop.

## Explicitly NOT changed

- Normal dispatch (`dispatch_shape`).
- **The Phase-4b early-dispatch trigger gate** (`try_pushed` / `any_ready_work`).
  An earlier revision of this PR relaxed it to `has_any_free_slot()`; that
  changed early-dispatch *behavior* -- qwen3 `q_proj` SPMD blocks went from a
  synchronized start to a staggered one -- so it has been **dropped**, along with
  its now-dead `has_any_free_slot()` helper. This PR is now MIX-selection +
  staging-cost only, with **no** early-dispatch trigger-frequency change.
- Self-ring memory ordering stays `seq_cst` (store-load / Dekker pattern).
- `sync_start` exclusion and the CAS-based block claim.

## Testing

- **Onboard a2a3** (qwen3-14b `decode_fwd` 2 layers, 3 runs each vs merge-base):
  early-dispatch staged volume identical (78 tasks/run), AICPU early-dispatch
  cost **-41%** -- full numbers in the perf comment below.
- The retained changes (MIX selection + staging) are unchanged from the prior
  reviewed revision; this revision only removes the trigger-gate relaxation and
  its dead helper. CI (a2a3sim + onboard) re-validates on this push.

### DIFF
--- a/src/a2a3/runtime/tensormap_and_ringbuffer/runtime/scheduler/scheduler_dispatch.cpp
+++ b/src/a2a3/runtime/tensormap_and_ringbuffer/runtime/scheduler/scheduler_dispatch.cpp
@@ -567,14 +567,16 @@ int32_t SchedulerContext::stage_consumer_blocks(
     uint64_t my_cores[PTO2_SPEC_CORE_MASK_WORDS] = {0};  // cores this thread gated (for self-ring)
     int32_t staged = 0;
     int32_t block = start;
-    auto stage_from = [&](CoreTracker::BitStates &avail, bool to_pending) {
-        // Mirror the normal flush_publish (scheduler_dispatch.cpp wmb()+publish loop):
-        // prepare all claimed blocks' payloads, one wmb(), then publish. The wmb
-        // guarantees the not_ready gate + args are globally visible before any
-        // DATA_MAIN_BASE token — without it a gated core can pick up the token and
-        // dcci a stale payload (the doorbell/release path mirrors normal dispatch).
-        PublishHandle handles[CoreTracker::MAX_CLUSTERS * 3];
-        int n = 0;
+    // Mirror the normal flush_publish (scheduler_dispatch.cpp wmb()+publish loop):
+    // prepare ALL claimed blocks' payloads (idle bucket -> running slot, pend bucket
+    // -> gated pending), then ONE wmb(), then publish. The wmb guarantees the
+    // not_ready gate + args are globally visible before any DATA_MAIN_BASE token —
+    // without it a gated core can pick up the token and dcci a stale payload. The
+    // shared `count` budget bounds total blocks <= free clusters/cores, so both
+    // buckets fit one handles[] buffer.
+    PublishHandle handles[CoreTracker::MAX_CLUSTERS * 3];
+    int n = 0;
+    auto prepare_from = [&](CoreTracker::BitStates &avail, bool to_pending) {
         while (count > 0 && avail.has_value()) {
             int32_t core_offset = avail.pop_first();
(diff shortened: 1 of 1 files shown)


ChaoZheng109:
## Onboard perf — a2a3, qwen3-14b `decode_fwd` (2 layers + LM head), 3 runs each

Measured this PR's early-dispatch staging path against the merge-base (no PR).
Signal = the `early_dispatch` AICPU scheduler-phase bars from the L2 swimlane
(level 4); each run normalized to its own t0. Workload:
[pypto-lib#691](https://github.com/hw-native-sys/pypto-lib/pull/691)
`decode_fwd.py --validate-fwd --fwd-layers 2 --enable-l2-swimlane`, each variant
run 3× under an exclusive `task-submit` device lock.

| metric | merge-base (before) | this PR | Δ |
| --- | --- | --- | --- |
| ED staged tasks / run | 78 (78 / 78 / 78) | 78 (78 / 78 / 78) | **unchanged** |
| ED bars / run | 10.3 | 11.0 | ~same |
| **total ED time / run** | **60.6 µs** (72.6 / 70.1 / 39.1) | **35.6 µs** (35.0 / 37.3 / 34.5) | **−41 %** |
| **cost per staged task** | **0.78 µs** | **0.46 µs** | **−41 %** |
| ED bar duration (median) | 5.48 µs | 2.42 µs | −56 % |
| ED bar duration (max) | 11.6 µs | 8.0 µs | −30 % |

**Read:** the staging cleanups (batch-pop + single `wmb()` + one publish loop)
stage the *exact same* set of tasks — 78/run, same two-wave structure, same
per-wave breakdown — but cut the per-pass AICPU cost ~41 %. So behavior is
unchanged; only the staging cost drops, and its run-to-run variance collapses
(before 39–73 µs → after 34–37 µs).

The dropped trigger-gate relaxation was verified separately to be the piece that
changed early-dispatch *behavior* (`q_proj` synchronized→staggered start); the
MIX-selection + staging parts kept here are behavior-neutral, cost-only.

<sub>Method: ED bars from `l2_swimlane_records.json`, `kind == early_dispatch`,
`tasks_processed` = staged count, 50 MHz clock. 3 runs per variant.</sub>

PR #1272, opened 2026-07-06
feat(a2a3 scheduler): MIX split placement with doorbell-enforced simultaneity

ChaoWao:
## Summary

Allow MIX (cube+AIV) task blocks to stage their AIC/AIV subtasks on **whatever core slot is free** — running slot if the core is idle, pending slot if the core is busy with a free pending slot — independently. This fixes the "stranded AIV cores" problem: when AIV finishes early but its sibling AIC is still running, the cluster is partially-idle and currently no MIX block can use it.

The in-kernel simultaneity contract (FFTS `syncall` + cross-core pipe `tpush`/`tpop`) is upheld by **gating the whole task behind the speculative doorbell** until every subtask has promoted to a running slot, so all subtasks activate at the same instant at doorbell-ring.

## Changes

| File | Change |
|------|--------|
| `scheduler_types.h` | `classify_mix_cluster` → `MixPlacementPlan{SPLIT, pending_mask}` with deadlock guard (pending subtask must NOT land behind a gated-running core) |
| `pto_scheduler.h` | `maybe_ring_split_doorbell`: rings only when fanin complete + **fully staged** (`next_block_idx >= logical_block_num`) + all subtasks promoted. `try_speculative_release`: split tasks defer ring; non-split keep immediate ring |
| `scheduler_dispatch.cpp` | `prepare_block_for_dispatch` per-subtask `pending_mask`; `stage_consumer_blocks` sources split clusters + counts `promoted_subtasks` + calls `maybe_ring`; `try_speculative_early_dispatch` split sourcing with deadlock guard |
| `scheduler_completion.cpp` | Case 3.3 promote: increments `promoted_subtasks` + triggers `maybe_ring` |
| `pto_runtime2_types.h` | `promoted_subtasks` + `requires_split_gate` fields on `PTO2TaskPayload` |
| `scheduler_context.h` | Updated signatures |
| `test_scheduler_state.cpp` | SPLIT classifier UTs |

## Key design decisions

1. **Full-staging gate**: `maybe_ring_split_doorbell` checks `next_block_idx >= logical_block_num` BEFORE ringing — prevents a partial ring that would spin the launched blocks of a full-occupancy syncall task (e.g. `fa_fused`) holding cores the un-staged blocks need.
2. **Deadlock guard**: `classify_mix_cluster` excludes cores whose running task is gated (STAGING) from pending-eligibility — staging a pending subtask behind a gated running task deadlocks.
3. **Non-split tasks unchanged**: only tasks with `requires_split_gate` use the deferred gate; non-split tasks ring immediately at fanin (sticky doorbell token handles pending subtasks).

## Test
- `ctest -R test_scheduler_state` — classifier + gate UTs pass.
- `python decode_fwd.py -p a2a3 -d <dev> --validate-fwd --fwd-layers 2 --enable-l2-swimlane` — golden argmax passes.


Crane-Liu:
Latest perf-only comparison (PR691 pto-lib, Qwen3-14B decode, `fwd_layers=2`, seed=1234, device=1):

| set | pass | device_wall_us | sched_us | 对比 baseline |
|---|---:|---:|---:|---:|
| baseline_x5 | 5/5 | 4132.152 | 3513.652 | 0 |
| current_step4_x3 | 3/3 | 4089.180 | 3475.460 | device -1.040%, sched -1.087% |

相对收益：device 1.040%，sched 1.087%。

PR #1273, opened 2026-07-06
Refactor: shard L2 swimlane collector accumulation

### LINKED ISSUE #1237: [Code Health] L2Swimlane collector: shard per-collector accumulation state to drop hot-path locks/atomics (measure first)
### Category

### Component

Host Runtime

### Description

PR #1162 makes the L2Swimlane collector run N-wide (`kCollectorThreadCount = PLATFORM_MAX_AICPU_THREADS`), so `on_buffer_collected` and its `copy_*_buffer` helpers now execute on multiple collector threads concurrently. To stay race-free, the shared accumulation state was synchronized two ways:

- scalar tallies/flags made `std::atomic` (the running `total_*_collected_` counters and `has_phase_data_`);
- the per-core / per-thread record vectors guarded by per-index `std::mutex` arrays (perf / aicore records keyed by core; sched / orch phase records keyed by AICPU thread).

**Proposed optimization — shard by writer, reduce at the end.** Instead of collectors contending on shared state, give each collector its own slice:

- counters/flags → per-collector arrays (`[collector]`), each collector does a plain (non-atomic) write to its own slot;
- record vectors → per-collector-by-instance (`[collector][core]` / `[collector][thread]`), each collector appends only to its own slice with no lock.

Then reduce once at `reconcile_counters` / export time — this runs after `stop()` has joined every collector, so there is a natural "all writes done" barrier. This removes every lock and atomic from the collector hot path.

**Ordering is safe:** the export path already sorts records by timestamp, so the "collector-0's records, then collector-1's, ..." concatenation order after the merge does not matter.

**Caveats to handle when implementing:**
- **False sharing:** pad each per-collector counter slot to its own cache line (`alignas(64)` or equivalent). Adjacent `uint64_t` slots share a line and would ping-pong across cores — that can be *slower* than the single atomic it replaces.
- Extra memory: N copies of the outer vector structures (small — the inner vectors start empty).
- One-time O(total records) merge at the end (single-threaded, unavoidable work).

**Measure before doing this.** The mutex is taken once per *buffer*, not per record, and is typically uncontended (only when two collectors touch the same core simultaneously). The dominant cost in `copy_*_buffer` is the per-record `push_back` and, on the non-SVM (a5) path, the `copy_buffer_from_device` rtMemcpy that already happened upstream. So the expected speedup may be small and could be eaten by false sharing. Gate this on a profile (`/profile` or the dfx timing tools) that shows the lock/atomic is actually hot. Prior L2Swimlane micro-opt context: `docs/investigations/2026-06-l2-swimlane-defer-wmb.md`; if this is tried and shows no signal, record that in `docs/investigations/` per the working-discipline rule.

**Do this after #1162 merges** — that PR is being trimmed and its line layout is still moving.

### Location

(Symbols only — no line numbers, since #1162 is still being edited.)

- `src/a2a3/platform/include/host/l2_swimlane_collector.h` and `src/a5/platform/include/host/l2_swimlane_collector.h` — members `total_perf_collected_`, `total_sched_phase_collected_`, `total_orch_phase_collected_`, `has_phase_data_`, `perf_record_mutexes_`, `aicore_record_mutexes_`, `sched_phase_record_mutexes_`, `orch_phase_record_mutexes_`, and the `collected_*_records_` vectors.
- `src/a2a3/platform/shared/host/l2_swimlane_collector.cpp` and `src/a5/platform/shared/host/l2_swimlane_collector.cpp` — `copy_perf_buffer`, `copy_sched_phase_buffer`, `copy_orch_phase_buffer`, `copy_aicore_buffer` (writers) and `reconcile_counters` (the reduce site).

### Proposed Fix

Replace the shared atomics + per-index mutexes with per-collector-sharded state (`[collector]` for scalars, `[collector][instance]` for record vectors), written lock-free on the hot path and merged once at reconcile/export after all collector threads join. Cache-line-pad the per-collector scalar slots. Land only with a before/after profile demonstrating a real collector-side speedup.

### Priority

Related: #1161, #1162

zmnobug:
## Summary

- Add an optional shard-aware collector callback in `ProfilerBase`.
- Shard L2Swimlane host accumulation state by collector for a2a3 and a5.
- Merge per-collector records and counters during reconcile/export after
  collector threads have stopped.

## Motivation

Refs #1237.

PR #1162 made the L2Swimlane collector run with multiple collector threads.
The existing implementation keeps correctness by protecting shared record
vectors with per-index mutexes and shared counters with atomics. This refactor
lets each collector write only to its own shard, then reduces those shards once
the collector threads have joined.

## Validation

- `env PATH=/data/zhaomin/.conda/envs/zm_pypto/bin:$PATH /data/zhaomin/.conda/envs/zm_pypto/bin/pre-commit run --files ...`
- Synthetic producer comparison on the same upstream/main baseline:
  - baseline effective receive median: `3.6992 GB/s`
  - current effective receive median: `3.7312 GB/s`
  - baseline effective receive mean: `3.7650 GB/s`
  - current effective receive mean: `3.7029 GB/s`
  - median drop rate was effectively unchanged

The synthetic comparison does not show a stable throughput gain, so this PR is
positioned as a code-health refactor rather than a proven performance fix.

close #1237

### DIFF
--- a/src/a2a3/platform/shared/host/l2_swimlane_collector.cpp
+++ b/src/a2a3/platform/shared/host/l2_swimlane_collector.cpp
@@ -20,12 +20,14 @@
 
 #include "host/l2_swimlane_collector.h"
 
+#include <cassert>
 #include <cinttypes>
 #include <cstdlib>
 #include <ctime>
 #include <filesystem>
 #include <fstream>
 #include <string>
+#include <type_traits>
 #include <vector>
 
 #include "common/memory_barrier.h"
@@ -93,9 +95,11 @@ int L2SwimlaneCollector::initialize(
     aicpu_thread_num_ = aicpu_thread_num;
     l2_swimlane_level_ = l2_swimlane_level;
     output_prefix_ = output_prefix;
-    total_perf_collected_.store(0, std::memory_order_relaxed);
-    total_sched_phase_collected_.store(0, std::memory_order_relaxed);
-    total_orch_phase_collected_.store(0, std::memory_order_relaxed);
+    total_perf_collected_ = 0;
+    total_sched_phase_collected_ = 0;
+    total_orch_phase_collected_ = 0;
+    has_phase_data_ = false;
+    collector_shards_merged_ = false;
 
     // Stash the memory context on the base up-front so alloc_single_buffer
     // sees consistent values during init. shm_host_ stays nullptr until the
@@ -282,10 +286,10 @@ int L2SwimlaneCollector::initialize(
     // allocated only for the first buffer_count pools. For sched the two are
     // equal; orch is a single instance (pool 0), so it zeroes all slots but
     // allocates buffers for just pool 0 — no buffers wasted on unused slots.
-    auto init_phase_pools = [&](auto buffer_tag, L2SwimlaneAicpuTaskPool *(*get_state)(void *, int, int),
+    auto init_phase_pools = [&](auto *buffer_tag, L2SwimlaneAicpuTaskPool *(*get_state)(void *, int, int),
                                 int state_count, int buffer_count, int buffers_per_thread, ProfBufferType recycle_kind,
                                 const char *kind_label) -> int {
-        using Buffer = typename decltype(buffer_tag)::type;
+        using Buffer = std::remove_pointer_t<decltype(buffer_tag)>;
         constexpr size_t buffer_bytes = sizeof(Buffer);
         for (int t = 0; t < state_count; t++) {
             auto *state = get_state(perf_host_ptr, num_aicore, t);
@@ -317,23 +321,13 @@ int L2SwimlaneCollector::initialize(
         return 0;
     };
 
-    // Type tags so the templated lambda can deduce the buffer type without
-    // having to spell out an explicit template argument (not portable on a
-    // generic lambda before C++20 explicit template-parameter syntax).
-    struct SchedTag {
-        using type = L2SwimlaneAicpuSchedPhaseBuffer;
-    };
-    struct OrchTag {
-        using type = L2SwimlaneAicpuOrchPhaseBuffer;
-    };
-
     // Sched: actual scheduler-thread count is unknown at host-alloc time, so
     // size buffers to the platform max. Orch: a single instance (pool 0), so
     // allocate buffers for just one pool while still zeroing all MAX states.
     if (init_phase_pools(
-            SchedTag{}, get_sched_phase_buffer_state, /*state_count=*/num_phase_threads,
-            /*buffer_count=*/num_phase_threads, /*buffers_per_thread=*/PLATFORM_PROF_SCHED_BUFFERS_PER_THREAD,
-            ProfBufferType::AICPU_SCHED_PHASE, "sched"
+            static_cast<L2SwimlaneAicpuSchedPhaseBuffer *>(nullptr), get_sched_phase_buffer_state,
+            /*state_count=*/num_phase_threads, /*buffer_count=*/num_phase_threads,
+            /*buffers_per_thread=*/PLATFORM_PROF_SCHED_BUFFERS_PER_THREAD, ProfBufferType::AICPU_SCHED_PHASE, "sched"
         ) != 0) {
         return -1;
     }
@@ -341,7 +335,8 @@ int L2SwimlaneCollector::initialize(
         return get_orch_phase_buffer_state(base, n_cores, t);
     };
     if (init_phase_pools(
-            OrchTag{}, orch_get_state, /*state_count=*/num_phase_threads, /*buffer_count=*/1,
+            static_cast<L2SwimlaneAicpuOrchPhaseBuffer *>(nullptr), orch_get_state,
+            /*state_count=*/num_phase_threads, /*buffer_count=*/1,
             /*buffers_per_thread=*/PLATFORM_PROF_ORCH_BUFFERS_PER_THREAD, ProfBufferType::AICPU_ORCH_PHASE, "orch"
         ) != 0) {
         return -1;
@@ -367,10 +362,7 @@ int L2SwimlaneCollector::initialize(
         total_size, device_id
     );
 
-    collected_perf_records_.assign(num_aicore_, {});
-    collected_aicore_records_.assign(num_aicore_, {});
-    collected_sched_phase_records_.assign(PLATFORM_MAX_AICPU_THREADS, {});
-    collected_orch_phase_records_.assign(PLATFORM_MAX_AICPU_THREADS, {});
+    reset_collector_shards();
 
     LOG_INFO_V0("Performance profiling initialized (dynamic buffer mode)");
     return 0;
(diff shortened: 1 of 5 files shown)


src/a2a3/platform/shared/host/l2_swimlane_collector.cpp
```diff
@@ -380,59 +371,149 @@ int L2SwimlaneCollector::initialize(
 // ProfilerBase callbacks
 // ---------------------------------------------------------------------------
 
-void L2SwimlaneCollector::copy_perf_buffer(const ReadyBufferInfo &info) {
+size_t L2SwimlaneCollector::normalize_collector_shard(int collector_shard) const {
+    if (collector_counters_.empty() || collector_shard < 0) {
+        return 0;
+    }
+    return static_cast<size_t>(collector_shard) % collector_counters_.size();
```
ChaoZheng109: **取模兜底会把去锁所依赖的核心不变量降级为静默数据竞争。**

本次去掉 per-index mutex 的正确性,完全建立在「每个 collector 线程拿到唯一、互不重叠的 shard,因此每个分片切片单写」之上。而这里是全仓库唯一对 shard 做边界处理的地方,却用了 `%`。

今天 `collector_shard ∈ [0, kCollectorThreadCount)` 且 `collector_counters_.size() == kCollectorThreadCount`(同一个 `L2SwimlaneModule::kCollectorThreadCount`),取模是恒等——纯死代码。但一旦将来某处传入越界的 shard,`%` 会**静默地把两个 collector 线程映射到同一个分片**,而该分片此时已无锁 → 对同一 `vector` 并发 `push_back`,数据撕裂/损坏,不崩溃、只偶发。

取模恰好违反了它本应守护的不变量。建议改成 fail-loud:debug 下 `assert(0 <= collector_shard && (size_t)collector_shard < collector_counters_.size())`;release 下返回一个越界哨兵(如 `collector_counters_.size()`),让四个 `copy_*_buffer` 已有的 `shard < ...size()` 守卫把该 buffer 丢弃,而不是无声地与别的线程共用分片。

src/a5/platform/shared/host/l2_swimlane_collector.cpp
```diff
@@ -382,59 +373,149 @@ int L2SwimlaneCollector::initialize(
 // ProfilerBase callbacks
 // ---------------------------------------------------------------------------
 
-void L2SwimlaneCollector::copy_perf_buffer(const ReadyBufferInfo &info) {
+size_t L2SwimlaneCollector::normalize_collector_shard(int collector_shard) const {
+    if (collector_counters_.empty() || collector_shard < 0) {
+        return 0;
+    }
+    return static_cast<size_t>(collector_shard) % collector_counters_.size();
```
ChaoZheng109: **取模兜底会把去锁所依赖的核心不变量降级为静默数据竞争。**(同 a2a3 一致)

去掉 per-index mutex 的正确性完全依赖「shard 每线程唯一、互不重叠 → 分片切片单写」。这里是唯一对 shard 做边界处理的地方却用 `%`:今天恒等、是死代码,但将来若传入越界 shard,`%` 会静默地把两个 collector 线程折叠到同一无锁分片 → 并发 `push_back` 同一 `vector`,数据撕裂且只偶发。

建议 fail-loud:debug `assert`,release 返回越界哨兵让 `copy_*_buffer` 现有的 `shard < ...size()` 守卫丢弃该 buffer,而非静默别名。

src/a2a3/platform/shared/host/l2_swimlane_collector.cpp
```diff
@@ -490,19 +571,21 @@ void L2SwimlaneCollector::copy_aicore_buffer(const ReadyBufferInfo &info) {
     }
 }
 
-void L2SwimlaneCollector::on_buffer_collected(const ReadyBufferInfo &info) {
+void L2SwimlaneCollector::on_buffer_collected(const ReadyBufferInfo &info) { on_buffer_collected(info, 0); }
```
ChaoZheng109: 这个 1 参 `on_buffer_collected` 现在是**死代码**:`consume()` 里的 `if constexpr (ProfilerDerivedShardAwareCollector<...>::value)` 对 L2 恒为 true,基类只会调 2 参版;全仓库也没有别的调用者。它现在只是转发到 `(info, 0)`。

建议要么删掉,要么在此加一行注释,明确它保留为基类文档契约里的默认签名,以免读者误以为还存在 `shard=0` 的真实调用路径。

src/a5/platform/shared/host/l2_swimlane_collector.cpp
```diff
@@ -492,19 +573,21 @@ void L2SwimlaneCollector::copy_aicore_buffer(const ReadyBufferInfo &info) {
     }
 }
 
-void L2SwimlaneCollector::on_buffer_collected(const ReadyBufferInfo &info) {
+void L2SwimlaneCollector::on_buffer_collected(const ReadyBufferInfo &info) { on_buffer_collected(info, 0); }
```
ChaoZheng109: 同 a2a3:此 1 参 `on_buffer_collected` 现为死代码——`consume()` 的 `if constexpr (ProfilerDerivedShardAwareCollector<...>::value)` 对 L2 恒 true,只调 2 参版,且无其它调用者。建议删除,或加注释说明保留为基类默认签名。

ChaoZheng109 (commented):
两条针对本次去锁重构的实现备注(inline)。均非 blocker;点 1 触及去锁方案赖以成立的核心不变量,建议处理。其余 DFX collector(PMU/DepGen/ScopeStats 同型,TensorDump 因输出为串行流而形态不同)的差异留作后续,不在本 PR 范围。

src/a2a3/platform/shared/host/l2_swimlane_collector.cpp
ChaoZheng109: 删掉

src/a2a3/platform/shared/host/l2_swimlane_collector.cpp
zmnobug: Fixed in 9bb80c64. normalize_collector_shard now asserts on invalid shards in debug builds and returns shard_count as a release sentinel, so the existing shard bounds checks drop invalid buffers instead of aliasing them onto another shard.

src/a5/platform/shared/host/l2_swimlane_collector.cpp
zmnobug: Fixed in 9bb80c64 with the same a2a3 behavior: invalid collector shards assert in debug and return an out-of-range sentinel in release, preserving the single-writer shard invariant.

src/a2a3/platform/shared/host/l2_swimlane_collector.cpp
zmnobug: Fixed in 9bb80c64. Removed the one-argument on_buffer_collected declaration and definition; L2 now keeps only the shard-aware callback path.

src/a5/platform/shared/host/l2_swimlane_collector.cpp
zmnobug: Fixed in 9bb80c64. Removed the dead one-argument on_buffer_collected overload on a5 as well, matching the shard-aware L2 path.

ChaoZheng109 (approved):

PR #1278, opened 2026-07-06
CI: add st-a5 and st-sim-a5 DFX per-feature smoke steps.

### LINKED ISSUE #1249: [Code Health] Harden a5 DFX scene tests: PMU artifact validation gaps + missing st-sim-a5 smoke coverage

yanghaoran29:
Fixes https://github.com/hw-native-sys/simpler/issues/1249

The a2a3/a2a3sim job already runs four dedicated DFX smoke steps (dep_gen, l2_swimlane, PMU, args_dump) with the matching --enable-* flags. The a5/a5sim job only ran the default pytest sweep, so a5 DFX scene tests were collected but their capture pipelines were never validated in sim CI.

Mirror the st-a2a3/st-sim-a2a3 smoke set in st-a5/st-sim-a5 with a5/a5sim paths and flags. graphviz is already installed in the a5/a5sim job for dep_gen.

Items A.1-A.3 (PMU artifact validation hardening in test_pmu.py) were already landed on main; this commit closes the remaining sim CI gap.

Also update a5's DeviceRunner::finalize_collectors() to use finalize() instead of stop() for diagnostics collectors, matching a2a3's implementation. This ensures shared memory is properly released on any exit from run() — success or early error — preventing stale state issues when workers are reused across runs (e.g., pytest session-scoped worker pool). Add scope_stats_collector_ cleanup with args.scope_stats_data_base reset.

### DIFF
--- a/src/a5/platform/onboard/host/device_runner.cpp
+++ b/src/a5/platform/onboard/host/device_runner.cpp
@@ -641,25 +641,9 @@ int DeviceRunner::finalize() {
     }
 
     // Cleanup all profiling subsystems (free shm + per-buffer dev/host
-    // shadows). All four shared collectors use the same alloc/free shape
-    // on a5: no unregister callback (a5 doesn't use halHostRegister) +
-    // prof_free_cb (rtFree directly).
-    if (l2_swimlane_collector_.is_initialized()) {
-        l2_swimlane_collector_.finalize(/*unregister_cb=*/nullptr, prof_free_cb);
-    }
-    if (dump_collector_.is_initialized()) {
-        dump_collector_.finalize(/*unregister_cb=*/nullptr, prof_free_cb);
-    }
-    if (pmu_collector_.is_initialized()) {
-        pmu_collector_.finalize(/*unregister_cb=*/nullptr, prof_free_cb);
-    }
-    if (dep_gen_collector_.is_initialized()) {
-        dep_gen_collector_.finalize(/*unregister_cb=*/nullptr, prof_free_cb);
-    }
-    if (scope_stats_collector_.is_initialized()) {
-        scope_stats_collector_.finalize(/*unregister_cb=*/nullptr, prof_free_cb);
-        kernel_args_.args.scope_stats_data_base = 0;
-    }
+    // shadows). Normally already done by run()'s exit path; this is the
+    // backstop for the no-run-since-init case.
+    finalize_collectors();
 
     // Shared cleanup body — streams, kernel_args, callable/orch maps,
     // chip-callable buffer pool, the three arenas, device_wall,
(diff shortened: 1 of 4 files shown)


src/a5/platform/onboard/host/device_runner.cpp
```diff
@@ -748,17 +748,26 @@ int DeviceRunner::finalize() {
...
+    // otherwise re-enter init_l2_swimlane() with stale state still allocated.
+    // Matches a2a3's finalize_collectors().
     if (l2_swimlane_collector_.is_initialized()) {
-        l2_swimlane_collector_.stop();
+        l2_swimlane_collector_.finalize(/*unregister_cb=*/nullptr, prof_free_cb);
     }
     if (dump_collector_.is_initialized()) {
-        dump_collector_.stop();
+        dump_collector_.finalize(/*unregister_cb=*/nullptr, prof_free_cb);
     }
```
ChaoZheng109, line 745: onboard `finalize_collectors()` 现在做的是完整 finalize,但 `finalize()`(约 647–662 行)仍保留着一份等价的内联 finalize 序列,两处逻辑重复。

建议统一:把 `finalize()` 里 `:647-662` 这段替换成直接调用 `finalize_collectors()`,与
- a2a3 onboard(`finalize()` → `finalize_collectors()`,`device_runner.cpp:703`)
- 本 PR 的 a5 sim(`:643`)

保持一致,消除重复,避免以后两处漂移。功能上等价(两条路径都会 finalize 释放 shm,且双调用安全),纯属结构去重。

yanghaoran29: 已按建议去重:把 `finalize()` 里 `:647-662` 的内联 finalize 序列替换成直接调用 `finalize_collectors()`,与 a2a3 onboard(`device_runner.cpp:703`)和本 PR 的 a5 sim(`:643`)保持一致。功能等价(is_initialized() 守卫保证 run() 退出路径已 finalize 后的二次调用为 no-op)。已提交于 `ff475e7a`。

ChaoZheng109 (approved):

PR #1280, opened 2026-07-07
perf(a2a3 scheduler): align early-dispatch staging + remove second complete poll

ChaoWao:
## Summary

Three scheduler-loop performance changes, no functional change.

### 1. `stage_consumer_blocks`: single prepare pass (align with `flush_publish`)
Merge two separate `stage_from` lambda calls (idle then pending, **each with its own wmb+publish**) into a single prepare pass + **one wmb** + one publish loop — mirrors normal dispatch's `flush_publish`. Saves one `wmb` (memory barrier) per stage call.

### 2. `stage_consumer_blocks`: relax seq\_cst → release/acquire
`staged_core_mask.fetch\_or`: `seq\_cst` → `release`. `spec\_state` DISPATCHED check: `seq\_cst` → `acquire`. The stager writes its bits before the OR (release); the releaser CASes then reads the mask (acquire). The release/acquire pair gives the needed happens-before — seq\_cst was over-conservative.

### 3. Remove second completion poll (Phase 5)
The top-of-loop Phase 1 poll already catches every FIN. The second poll only saved ~7µs detection latency in rare busy windows but **doubled the per-iteration MMIO poll cost** (72-core LDR COND scan at ~95ns/core) — the dominant scheduler overhead on this workload.

## Test
- `ctest -R test\_scheduler\_state` — UTs pass.
- Device golden (`--validate-fwd --fwd-layers 2`) verified during development.

### DIFF
--- a/src/a2a3/runtime/tensormap_and_ringbuffer/runtime/scheduler/scheduler_context.h
+++ b/src/a2a3/runtime/tensormap_and_ringbuffer/runtime/scheduler/scheduler_context.h
@@ -252,7 +252,7 @@ private:
 
     void dispatch_shape(
         int32_t thread_idx, PTO2ResourceShape shape, CoreTracker::DispatchPhase phase, CoreTracker &tracker,
-        bool &entered_drain, bool &made_progress, bool &try_pushed
+        bool &entered_drain, bool &made_progress
     );
 
     // Speculative early-dispatch (Hook 1). After normal dispatch leaves idle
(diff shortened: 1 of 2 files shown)


ChaoWao:
Superseded by #1288 — same Phase 5 removal + try_pushed gate removal, but #1288 adds has_any_free_slot() local pre-check (avoids shared queue access when fully loaded). The seq_cst→release/acquire change in stage_consumer_blocks will be re-proposed separately if needed.

PR #1286, opened 2026-07-07
fix(dfx): keep SPMD dependency arrows on the swimlane Scheduler View

yanghaoran29:
## Problem

The swimlane converter draws each `deps.json` edge twice — on the **Worker View** (pid=4, AICore timestamps) and mirrored on the **Scheduler View** (pid=3, AICPU timestamps). For SPMD tasks the Scheduler View silently dropped arrows the Worker View kept, so dependencies "disappeared" on the AICPU-eye view.

Root cause: the mirror bound each arrow to the literal min-`core_id` anchor and required a captured AICPU `finish` on both endpoints. Two triggers:

- **Successor without a captured completion** (tail/terminal task — dispatched fine, `finish<=0`): lost *every* inbound arrow, even though the arrow lands at the successor's `dispatch`.
- **SPMD pred whose literal min-core subtask lacked a `finish`**: lost its outbound arrow, even when sibling blocks carried valid timing.

The Worker View is unaffected (AICore `end_time` is always present), so the two views disagreed.

## Fix

`_scheduler_anchor_row()` resolves each endpoint to the first subtask the Scheduler View can actually bind to — Worker-View parity, "connect the first (visible) subtask". For an SPMD task it falls back to the first same-`core_type` sibling that carries the needed timestamp; a `finish` is required only on the source (its ts), not on the dispatch-anchored destination. Non-SPMD behavior is unchanged.

## Tests

- Two unit regressions in `test_swimlane_converter.py` reproducing both triggers (verified to **fail pre-fix**, Scheduler=0 vs expected 1).
- A Worker↔Scheduler dependency-arrow **parity check** added to the shared **a2a3 + a5** `l2_swimlane` st validator (`_swimlane_validate.py`), so the mirror contract is exercised in both arch CI steps on real captured data.
- `docs/dfx/l2-swimlane-profiling.md` §3.5 updated to document the fallback.

Reproduced and validated against real qwen AllSpmd-decode and batch-paged-attention swimlane captures on disk (no NPU currently reachable on the dev box; heavy SPMD onboard case is a2a3-only).


yanghaoran29:
Addressed the CodeRabbit nitpick in \`1a31f29\`: extracted \`verify_dependency_view_parity\` into a single shared source of truth at \`tests/st/_l2_swimlane_shared.py\`, loaded by both the a2a3 and a5 \`_swimlane_validate.py\` via a \`__file__\`-relative module load (the two leaf packages have no shared package parent under \`--import-mode=importlib\`). The parity contract now lives in one place instead of being hand-synced across platforms.

PR #1291, opened 2026-07-08
Refactor: move L3-L2 orch comm data path to L3 Host in sim

ccyywwen:
## Summary

  This PR is based on PR #1236.
  That base is intentional: PR #1236 restructures the L3-L2 message queue and
  primitive payload-transfer layering, so the mapped-region data path here is
  implemented against that newer primitive-layer shape rather than the older queue
  staging structure.

  This PR implements the first production slice of the L3-L2 communication data
  path refactor. It adds lifecycle create/release controls for L3-L2 regions and
  switches simulation platforms to an L3 Host mapped-region data path. In sim,
  payload and counter operations now execute from the L3 Host through private
  mapped-region helpers, while the six-scalar descriptor remains unchanged and
  continues to carry L2-side addresses for the L2 AICPU endpoint.

  On onboard platforms, the public API remains supported through the existing
  L2 Host-mediated path. No public primitive or message queue API changes are
  introduced.

## Changes

  - Add region create/release mailbox controls, fixed wire structs, bindings, and
    Python encode/decode support.
  - Allocate one L2 Host-owned sim backing region and partition it into payload
    and counter ranges.
  - Initialize counter bytes to zero before publishing create replies.
  - Add private L3 Host mapped-region helpers for sim import/close, payload
    read/write, and counter notify/test/wait.
  - Store mapped-region metadata privately on the L3 Host side; it is not exposed
    through `descriptor_scalars()`.
  - Route sim `payload_write`, `payload_read`, `SIGNAL_NOTIFY`, `SIGNAL_TEST`,
    and `SIGNAL_WAIT` through the mapped-region path.
  - Preserve live/released/poison semantics, create rollback, and post-drain
    cleanup ordering.
  - Keep queue staging behavior behind primitive payload transfer.
  - Update user-facing docs for the current API and platform behavior.
  - Remove the non-main design/implementation planning docs from this branch.


jvjhfhg:
Closed since a superset #1305 has been merged.

PR #1295, opened 2026-07-08
Fix: retain callable chip buffers by registration

### LINKED ISSUE #1082: [Code Health] AICPU orch .so is double-uploaded (orch_so_dedup_ duplicates the ChipCallable buffer) and chip_callable_buffers_ leaks until finalize
### Category

### Component

Host Runtime

### Description

Two related device-memory hygiene problems in the onboard host runner, surfaced while investigating #1019 (which turned out to be unrelated — an op-timeout window, fixed by #1035).

**1. The AICPU orchestration `.so` is uploaded to the device twice.**
- `upload_chip_callable_buffer()` uploads the **whole** `ChipCallable` (header + `storage_`) into `chip_callable_buffers_`. `storage_` *starts with* the orchestration `.so` — `ChipCallable::binary_data()` **is** the orch `.so` (the host itself reads it that way: `runtime_maker.cpp` does `orch_so_binary = callable->binary_data()`).
- `register_callable()` then uploads **just the orch `.so` bytes again** into a separate pool `orch_so_dedup_`, and points the AICPU at it via `runtime.set_dev_orch_so(addr, size)`.

So the same orch bytes occupy device GM twice. The orch already lives at a known offset inside the chip buffer: `chip_dev + offsetof(ChipCallable, storage_)`, length `binary_size()`. `orch_so_dedup_` exists only because the AICPU reads `dev_orch_so` (the standalone copy) and `dlopen`s it, instead of pointing into the chip buffer.

**2. `chip_callable_buffers_` is never freed on unregister — only at `finalize()`.**
- `unregister_callable()` decrements + frees `orch_so_dedup_` (it has a refcount), but does **not** touch `chip_callable_buffers_`.
- `ChipCallableBuffer` has no `refcount` field (unlike `OrchSoBuffer`), and the comment says "never freed until finalize".
- Result: in a long-lived worker (e.g. the L2 `st_worker` pool reuses one `ChipWorker` per (runtime, device) for a whole xdist session), every distinct chip-callable buffer accumulates on-device until the runner finalizes, even after its callable is unregistered. The header comment at the `chip_callable_buffers_` declaration already *claims* "refcounted by hash", but that was specified and never implemented.

Neither is a crash today (the #1019 repro showed max-live=1 for a single callable), but both are real growth/duplication that scale with callable churn.

### Location

- `src/common/platform/onboard/host/device_runner_base.cpp`
  - `upload_chip_callable_buffer()` ~`385`–`431` (uploads whole ChipCallable incl. orch)
  - `register_callable()` ~`476`–`535` (separate orch `.so` upload → `orch_so_dedup_`)
  - `prepare_orch_so()` ~`437`–`470` (`set_dev_orch_so`)
  - `unregister_callable()` `571`–`594` (frees `orch_so_dedup_`, **not** `chip_callable_buffers_`)
  - finalize-only free of `chip_callable_buffers_` ~`680`–`689`
- `src/common/platform/onboard/host/device_runner_base.h`
  - `struct ChipCallableBuffer` `625`–`628` (no `refcount`), map `629`
  - `struct OrchSoBuffer` `654`–`657` (has `refcount`), map `660`
- `src/common/task_interface/callable.h:117`–`118` — `ChipCallable::binary_data()/binary_size()` = orch `.so`
- `src/{a5,a2a3}/runtime/tensormap_and_ringbuffer/host/runtime_maker.cpp` ~`133`–`134` — `orch_so = callable->binary_data()`
- `src/{a5,a2a3}/runtime/tensormap_and_ringbuffer/aicpu/aicpu_executor.cpp` ~`269`–`318` — AICPU reads `dev_orch_so` → writes temp file → `dlopen`

### Proposed Fix

Unify into a single refcounted device buffer:
1. `prepare_orch_so` (device-orch path): set `dev_orch_so = chip_dev + offsetof(ChipCallable, storage_)`, size `= binary_size()` — point the AICPU at the orch bytes already inside the chip buffer.
2. Delete `register_callable`'s separate orch upload + the `orch_so_dedup_` map/refcount.
3. Add `refcount` to `ChipCallableBuffer` and free-on-unregister in `unregister_callable()` (mirror the existing `orch_so_dedup_` pattern), so the unified buffer (orch + kernels) is released when its last callable unregisters.
4. Start with a failing repro test: register N distinct callables → unregister → assert the chip pool is freed (today it isn't).

**Must verify / preserve (the make-or-break check):**
- **Byte-identity**: `upload_chip_callable_buffer` runs `patch_chip_callable_scratch_for_device` (patches *child* `resolved_addr_`). Confirm that patching does **not** touch the orch region `storage_[0 .. binary_size)`, so the in-buffer orch is byte-identical to what `dlopen` needs.
- **hbg path**: `host_build_graph` doesn't H2D the orch (host `dlopen` via `state.host_dlopen_handle`). This change is for the device-orch (`tensormap_and_ringbuffer`) path only; the hbg branch must stay.

### Priority

puddingfjz:
Keep one refcounted ChipCallable buffer lease per registered callable id.

Use the retained chip storage slice as the device orch SO.

This removes the second orch SO upload for registered callables.

Release the retained chip buffer on unregister and registration failure.

Fixes #1082

### DIFF
--- a/src/common/platform/onboard/host/device_runner_base.cpp
+++ b/src/common/platform/onboard/host/device_runner_base.cpp
@@ -30,6 +30,7 @@
 
 #include <algorithm>
 #include <cassert>
+#include <cstddef>
 #include <cstdint>
 #include <cstring>
 #include <limits>
@@ -625,7 +626,7 @@ void DeviceRunnerBase::print_handshake_results() {
 // =============================================================================
 
 uint64_t DeviceRunnerBase::upload_chip_callable_buffer(const ChipCallable *callable) {
-    if (callable == nullptr || callable->child_count() == 0) {
+    if (callable == nullptr) {
         return 0;
     }
     if (stream_aicpu_ == nullptr) {
@@ -638,9 +639,10 @@ uint64_t DeviceRunnerBase::upload_chip_callable_buffer(const ChipCallable *calla
     // Content-hash dedup: identical bytes → return cached chip_dev.
     auto it = chip_callable_buffers_.find(layout.content_hash);
     if (it != chip_callable_buffers_.end()) {
+        it->second.refcount++;
         LOG_DEBUG(
-            "Chip callable dedup hit: chip_dev=0x%lx, size=%zu, hash=0x%lx", it->second.chip_dev, it->second.total_size,
-            layout.content_hash
+            "Chip callable dedup hit: chip_dev=0x%lx, size=%zu, hash=0x%lx, refcount=%d", it->second.chip_dev,
+            it->second.total_size, layout.content_hash, it->second.refcount
         );
         return it->second.chip_dev;
     }
@@ -668,12 +670,32 @@ uint64_t DeviceRunnerBase::upload_chip_callable_buffer(const ChipCallable *calla
         return 0;
     }
 
-    chip_callable_buffers_.emplace(layout.content_hash, ChipCallableBuffer{chip_dev, layout.total_size});
+    chip_callable_buffers_.emplace(layout.content_hash, ChipCallableBuffer{chip_dev, layout.total_size, 1});
     LOG_DEBUG(
         "Uploaded chip callable: chip_dev=0x%lx, size=%zu, child_count=%d, hash=0x%lx", chip_dev, layout.total_size,
         callable->child_count(), layout.content_hash
     );
     return chip_dev;
+}
+
+int DeviceRunnerBase::release_chip_callable_buffer(uint64_t hash) {
+    if (hash == 0) {
+        return 0;
+    }
+    auto it = chip_callable_buffers_.find(hash);
+    if (it == chip_callable_buffers_.end()) {
+        LOG_WARN("release_chip_callable_buffer: hash=0x%lx not found", hash);
+        return 0;
+    }
+    if (--it->second.refcount <= 0) {
+        mem_alloc_.free(reinterpret_cast<void *>(it->second.chip_dev));
+        LOG_DEBUG(
+            "Freed chip callable buffer: chip_dev=0x%lx, size=%zu, hash=0x%lx", it->second.chip_dev,
+            it->second.total_size, hash
+        );
+        chip_callable_buffers_.erase(it);
+    }
+    return 0;
 }
 
 int DeviceRunnerBase::stamp_orch_so(Runtime &runtime, int32_t cid) {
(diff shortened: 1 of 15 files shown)


src/common/platform/onboard/host/device_runner_base.cpp
```diff
@@ -882,20 +887,13 @@ int DeviceRunnerBase::unregister_callable(int32_t callable_id) {
     CallableState state = std::move(it->second);
     callables_.erase(it);
     aicpu_seen_callable_ids_.erase(callable_id);
+    release_chip_callable_buffer(state.chip_buffer_hash);
```
ChaoWao, line 890: **Should fix — no regression test for this lifetime/refcount fix.**

Both bugs #1082 describes are silent ("Neither is a crash today"), and every existing test passes with or without the fix: the `prepared_callable` ST asserts only functional correctness and the *monotonic* `aicpu_dlopen_count` (never decremented on unregister), so it cannot observe a chip-buffer leak; `test_chip_callable_upload_immutable.cpp` is a buffer-immutability contract test unrelated to refcounting.

If a future edit drops this `release_chip_callable_buffer` call, or the `chip_buffer_guard`/`dismiss()` → `CallableState` ownership handoff regresses, nothing goes red. Per `.claude/rules/discipline.md` §3 a bugfix should leave a barrier behind.

Cheapest lock-in: a cpput test on `DeviceRunnerBase` — register a callable, assert `chip_callable_buffers_.size() == 1`; unregister, assert 0; register identical bytes at two cids, assert the dedup refcount, unregister one → still present, unregister the other → freed. (Alternatively expose a live-buffer counter mirroring `aicpu_dlopen_count` and assert `delta == 0` after register+unregister in the existing ST.)

ChaoWao (commented):
Mechanism is correct and faithfully implements #1082 (both the double-upload and the leak-until-finalize). Verified the layout assumption (`ChipCallable::binary_data()` == `storage_`, orch SO is the leading slice) and traced every register/unregister failure path — the refcount balance holds on all of them. Leaving 3 non-blocking comments inline.

src/common/task_interface/prepare_callable_common.h
```diff
@@ -73,32 +76,34 @@ struct CallableArtifacts {
...
 inline int upload_and_collect_child_addrs(
-    const ChipCallable *callable, uint64_t (*upload_fn)(const void *), std::vector<ChildKernelAddr> *out
+    const ChipCallable *callable, uint64_t (*upload_fn)(const void *), std::vector<ChildKernelAddr> *out,
+    uint64_t *out_chip_dev = nullptr, uint64_t *out_chip_hash = nullptr
 ) {
     if (callable == nullptr || upload_fn == nullptr || out == nullptr) return -1;
     out->clear();
-    if (callable->child_count() <= 0) return 0;
 
+    const ChipCallableLayout layout = compute_chip_callable_layout(callable);
```
ChaoWao, line 98: **Consider (minor).** `compute_chip_callable_layout(callable).content_hash` is computed here, and `upload_chip_callable_buffer` recomputes the same layout/hash internally to key `chip_callable_buffers_`. It's a once-per-registration FNV pass over the whole buffer — negligible, and the two-computation split keeps the `upload_fn` pointer signature clean. Fine to leave; noting only for completeness.

src/common/platform/onboard/host/device_runner_base.cpp
```diff
@@ -793,59 +816,36 @@ int DeviceRunnerBase::record_device_orch_callable(
...
-    } else {
-        buf_it->second.refcount++;
-        dev_addr = reinterpret_cast<uint64_t>(buf_it->second.dev_addr);
-        LOG_INFO_V0(
-            "record_device_orch_callable: hash=0x%lx shared buffer (refcount=%d)", hash, buf_it->second.refcount
-        );
-    }
-
     CallableState state;
     state.hash = hash;
```
ChaoWao, line 831: **Consider (informational).** With `orch_so_dedup_` gone, `state.hash` (the orch ELF build-id) is no longer used for any buffer lookup — its only remaining consumer is the `callable_hash()` getter (unchanged by this PR, still used elsewhere). Not dead code, but the field's role has narrowed to just that accessor; worth a note so a future reader doesn't assume it still drives dedup.

ChaoWao (approved):

PR #1309, opened 2026-07-09
Refactor: remove the user-facing block_dim knob

doraemonmj:
> **Stacked on #1478.** Its commit is currently in this branch and drops out
> once it merges. #1477 (CoreTracker capacity) has landed, which is what lets
> the four cases keeping `aicpu_thread_num: 2` run at the full 24-cluster
> width.

## What

A run now always takes the whole device. `block_dim` was a per-call knob that
every scene test pinned to the platform maximum anyway, and it forced host
orchestration to guess a width before the device had reported one.

## The knob, end to end

- **`CallConfig` loses the field.** The packed wire layout is one int32 shorter,
  so the remote-L3 `PROTOCOL_VERSION` goes **1 → 2** in both codecs and the fork
  mailbox format string drops an `i`. The compile-time layout guard in
  `call_config.h` is updated with it — it is what caught the drift.
- **`DeviceRunner` resolves the width unconditionally**: every cluster the
  AICore stream reports onboard, `SIM_AUTO_BLOCKDIM` on sim.
- The nanobind binding, `scene_test`'s config plumbing, and the `l0_swimlane`
  replay lose their `block_dim` paths.

Two latent problems surfaced while doing it and are fixed here:

- `remote_l3_session.py` hardcoded `protocol_version=1` in the HELLO payload
  instead of referencing the constant — a version bump would have silently
  missed it.
- `spmd_basic`'s `Case2_AutoBlockDim` became an exact duplicate of `Case1` once
  the knob was gone, so it is dropped.

## Scene tests

178 pinned values across 75 files removed. Kernel-side `params["block_dim"]` — a
tiling argument, not a runtime knob — is untouched.

`l0_swimlane` can no longer infer an SPMD replay width from the test file, since
a cohort now sizes itself on device: `--spmd-block-num` is required to replay
one, and the help/doc say so.

## Coverage note

Sim resolves to `SIM_AUTO_BLOCKDIM` (8) and onboard to the device's real width
(24 on a2a3), so **sim no longer exercises wide-device behaviour**. That is a
deliberate trade — sim runs one OS thread per AICore, and 24-36 clusters is
72-108 threads per case — but it is why #1477's bug was invisible until an
onboard run. Wide-device coverage rests on the onboard job.

## Verification

Measured on top of #1478 (#1477 is now in main):

| Suite | Result |
| ----- | ------ |
| `pytest examples tests/st --platform a2a3` (onboard) | 60 passed, 0 failed |
| `pytest examples tests/st --platform a2a3sim` | 51 passed, 0 failed |
| `pytest examples tests/st --platform a5sim` | 39 passed, 0 failed |
| `pytest tests/ut -m "not requires_hardware"` | 813 passed |
| pre-commit (all hooks) | passed |

The four cases that keep `aicpu_thread_num: 2` (`dummy_task`,
`predicated_dispatch` on both runtimes, `dep_gen_chain`) pass at the full
24-cluster width with no thread-count workaround, which is the check that #1477
actually fixed the cause rather than moving it.

## History

This PR started as "zero out the hardcoded `block_dim`". That premise stopped
holding when #1458 landed `SIM_AUTO_BLOCKDIM = 8`: on sim the auto sentinel no
longer resolves to `PLATFORM_MAX_BLOCKDIM`, so zeroing the pins is a behaviour
change rather than a no-op. Removing the knob outright makes "a run takes the
whole device" the only shape there is. @doraemonmj's cohort-width analysis lives
on in #1478.

## Cost: the under-count check from #1472

#1472 added a `Pinned` case to the three `available_aicore_counts` tests that
fixed `block_dim` so the host held an expectation independent of what the device
reported — the only thing able to catch an **under**-reported count. That handle
is exactly the knob this PR removes, so the case goes with it.

What survives: the count is range-checked, `aiv == 2 * clusters` is checked, and
the cohort is actually spent, so an **over**-reported count still trips the
`require_sync_start` deadlock guard on device. What is lost: an under-reported
count is now self-consistent — fewer blocks launch, fewer slots are expected,
the tail is zero on both sides. The test docstrings say so rather than implying
coverage that is not there.

Restoring it needs a host-side handle on the device width that is not a second
source of truth for a device property. There is not one today.


ChaoWao:
Heads-up: #1458 landed and it invalidates this PR's core premise on sim.

This PR's rationale is that `block_dim=0` already resolves to the literal being
removed:

> **Sim**: uses `PLATFORM_MAX_BLOCKDIM` directly (a2a3=24, a5=36).

That is no longer true. #1458 added `SIM_AUTO_BLOCKDIM = 8`
(`src/common/platform/sim/host/device_runner_base.h`): sim now resolves the
auto sentinel to **8**, not 24/36, because sim runs one OS thread per AICore and
taking the whole modelled chip costs 72-108 threads per case. An explicit
`block_dim` is still honoured up to `PLATFORM_MAX_BLOCKDIM`, so this PR's
zeroing is the exact operation that flips behaviour.

So on sim this stops being a no-op cleanup, and at least two cases break
outright rather than merely running narrower:

| Case | Submits | After zeroing (sim) |
| ---- | ------- | ------------------- |
| `spmd_sync_start` (pins 24) | `block_num=12`, `require_sync_start` | 12 > limit 8 → `REQUIRE_SYNC_START_INVALID` |
| `spmd_sync_start_edge` (pins 24) | `block_num=23`, `require_sync_start` | 23 > limit 8 → same |

The fix is the rework this PR's follow-up was always going to need: SPMD
cohorts must size themselves from `rt_available_cluster_count()` (added in
#1458) instead of hardcoding a width, and cases whose block count is load-bearing
need their incore tiling adjusted to match. I'm preparing that change and will
link it here.

Until then, please don't merge this as-is on top of current main — the two
together are a red sim suite. Either hold it for the cohort rework, or update
the body so the sim behaviour change is stated rather than described as a no-op.

ChaoWao:
Heads-up: I've force-pushed onto this branch, replacing `775d2800`. Your work is
not discarded — the commit carries `Co-authored-by: majin0824`, and your semantic
analysis of the cohort widths (stress at `total/6, /2, /3, /6`; early_dispatch in
waves of at most `cluster_count`) is what the new orchestrations implement. Sorry
for landing on top of yours rather than alongside; happy to restructure if you'd
prefer your commit kept separate underneath.

**What changed relative to your version**

Two things, one additive and one a design swap.

*Additive:* the PR now also removes the `block_dim` field itself, not just the
pinned values — `CallConfig`, both remote-L3 wire codecs (`PROTOCOL_VERSION`
1→2, the payload is an int32 shorter), the fork-mailbox format string, the
nanobind binding, `scene_test`, and `l0_swimlane`. Zeroing the values without
removing the knob leaves the sim behaviour change (24/36 → 8) implicit; removing
it makes "a run takes the whole device" the only shape there is.

*Design swap:* `simpler_setup/available_block_dim.py` is gone. Instead each
orchestration writes the geometry it actually used into a `layout` output
tensor, and the test rebuilds its golden from that (via the `compare_outputs`
hook from #1470). The reason is that the helper made the host a second source of
truth for a device property:

- `_SIM_AUTO_BLOCKDIM = 8` and `_PLATFORM_MAX_BLOCKDIM = {a2a3: 24, a5: 36}` are
  hand-mirrored from `device_runner_base.h` / `platform_config.h`, with nothing
  enforcing they stay in sync.
- The onboard path opens a second ACL client inside the pytest process
  (`acl.init()` → `set_device` → `create_stream`) defaulting to `device_id=0`,
  but the resource scheduler hands each case whatever devices are free. On a
  shared box that both asks the wrong device and touches one the run does not
  hold.
- Every failure path falls back to `ceiling`, which is exactly the wrong answer
  on a device that reports fewer cores than the arch maximum.

With the device reporting its own width there is nothing to keep in sync and no
second ACL context.

**State:** in progress — 4 of 6 sync_start cases converted (`spmd_sync_start`,
`_aiv`, `_edge`, `_mix_spill`). `_stress` and `_early_dispatch` are still on the
old literals, so CI is red until those land. Pushing now for visibility rather
than sitting on it.

ChaoWao (approved):

PR #1343, opened 2026-07-13
feat(dfx): scale host drain/collector threads with aicpu_thread_num

yanghaoran29:
Every enabled DFX subsystem started N drain + 1 replenish + N collector threads with N fixed at compile time to PLATFORM_MAX_AICPU_THREADS (a2a3=4, a5=7), regardless of how many AICPU threads the run launched. The host shard arrays and the proactively-replenished device buffers were sized the same way, so a 2-thread run spun drain/collector threads on queues that provably never get a producer and preallocated device buffers into recycled lanes no drain thread owns.

Split the constant's two roles: the std::array capacity stays compile time (no dynamic allocation, no layout change), while the number of threads/shards in use becomes a runtime value latched in Derived::init() via ProfilerBase::set_aicpu_thread_num(). Three quantities, no longer equal:
  queue_count_ = aicpu_thread_num                     device queues to scan
  shard_count_ = min(n, Module::kMaxCollectorThreads) drain/collector threads
  kMaxCollectorThreads                                std::array capacity

Per-DFX thread construction. kMaxCollectorThreads is set per subsystem by who the sole device-side producer is:
  - L2Swimlane  kMaxCollectorThreads = PLATFORM_MAX_AICPU_THREADS. Producers are the scheduler threads (task / sched-phase records) plus the orchestrator (orch-phase records), one queue per AICPU thread. shard_count_ == queue_count_ == aicpu_thread_num: drain i owns queue i and feeds collector shard i, 1:1.
  - ArgsDump    kMaxCollectorThreads = PLATFORM_MAX_AICPU_THREADS. Producers are the scheduler threads, one per AICPU thread. Same 1:1 drain<->queue<->collector shape as L2Swimlane.
  - PMU         kMaxCollectorThreads = PLATFORM_MAX_AICPU_THREADS. Producers are the scheduler threads that own each core, one per AICPU thread. Same 1:1 shape.
  - DepGen      kMaxCollectorThreads = 1. The orchestrator is the sole device-side producer (enqueues into queues[s_orch_thread_idx]). One drain thread scans all queue_count_ queues to find the single producer, so here queue_count_ > shard_count_ == 1; extra shards would only be empty.
  - ScopeStats  kMaxCollectorThreads = 1. Same orchestrator-only shape as DepGen: one drain thread, one collector shard, scanning every live queue.

SPSC still holds everywhere: drain count == collector count == shard_count_, so each ready shard has exactly one drain-thread producer.

The count is injected through a dedicated setter rather than set_memory_context() because PMU seeds its recycled lanes *before* its set_memory_context() call, and every shard index folds modulo the manager's shard count; set_aicpu_thread_num() must precede the first push_recycled().

Two latent defects fixed alongside, same root cause:
- L2Swimlane/PMU derived their recycled watermark from ceil(MAX_CORES / PLATFORM_MAX_AICPU_THREADS). With fewer live shards each owns more cores, so the watermark was under-provisioned by up to 4x (a2a3) / 7x (a5): the recycled pool runs dry, drain cannot refill the device free_queue, and records are dropped. Now computed from the runtime shard count via the two-arg recycled_warm_target(kind, shard_count); aggregate memory is unchanged.
- L2Swimlane allocated sched-phase buffers for all PLATFORM_MAX_AICPU_THREADS pools, pushing the surplus of pools t >= aicpu_thread_num into recycled lane t, which no drain thread owns. Buffers now follow aicpu_thread_num while the shm layout keeps its fixed PLATFORM_MAX_AICPU_THREADS stride (AICPU's pool offsets depend on it).

Also fixes a failure mode the scaling itself introduces: the collector's idle-timeout guard only exempted never-used shards when shard_count > 1, so a single-shard subsystem (DepGen/ScopeStats) that legitimately emits nothing for a whole run would log "collector idle timeout" and abandon its shard. An empty shard is a valid run shape at any shard count; execution_complete_ is the exit path.

Device shm layout is untouched — this is a host-side change only.

Docs: docs/dfx/dynamic-dfx-thread-scaling.md documents the three-quantity model and the per-subsystem shard shapes; profiling-framework.md, pmu-profiling.md and l2-swimlane-profiling.md describe kMaxCollectorThreads as the array capacity with the live count = min(aicpu_thread_num, cap).

Tests: new tests/ut/cpp/common/test_profiler_base.cpp covers the thread fanout, the queue_count_ > shard_count_ shape, and the idle-timeout regression. The DFX scene tests only ran at aicpu_thread_num=4, which on a2a3 is exactly the degenerate shard_count == max case, so scope_stats and l2_swimlane gain an aicpu_thread_num=2 case. (1 is not a valid config: sched_thread_num = n - 1, so it would leave zero schedulers.)

### DIFF
--- a/src/common/platform/include/host/profiler_base.h
+++ b/src/common/platform/include/host/profiler_base.h
@@ -45,10 +45,15 @@
  *   static constexpr uint32_t kHostRecycledQueueSize;
  *   static constexpr uint32_t kSlotCount;      // FreeQueue::buffer_ptrs[] length.
  *   static constexpr const char* kSubsystemName; // "PMU" / "L2Swimlane" / "Dump".
- *   // Optional: number of mgmt drain shards (defaults to 1).
- *   static constexpr int      kMgmtDrainThreadCount;
- *   // Optional: number of collector threads / host ready-queue shards.
- *   static constexpr int      kCollectorThreadCount;
+ *   // Optional: CAPACITY of the drain / collector shard arrays (defaults to
+ *   // 1). Bounds the shard arrays at compile time; the number of threads
+ *   // actually started is the runtime min(aicpu_thread_num,
+ *   // kMaxCollectorThreads), latched by ProfilerBase::set_aicpu_thread_num().
+ *   // Subsystems whose only device-side producer is the orchestrator (DepGen,
+ *   // ScopeStats) set this to 1: one drain thread scans every AICPU ready
+ *   // queue and finds the single producer's, so extra shards would only ever
+ *   // be empty.
+ *   static constexpr int      kMaxCollectorThreads;
  *   // Optional: refresh cached queue metadata before a replenish pass.
  *   template <typename Mgr>
  *   static void refresh_replenish_metadata(Mgr&, DataHeader*);
@@ -58,6 +63,15 @@
  *
  *   // Per-kind alloc batch size for proactive_replenish's free-queue fallback.
  *   static int batch_size(int kind);
+ *
+ *   // Optional: steady-state low-water mark for a shard-local recycled lane.
+ *   // Two forms; the two-arg one wins if both are present:
+ *   //   static int recycled_warm_target(int kind);
+ *   //   static int recycled_warm_target(int kind, int shard_count);
+ *   // Take the two-arg form when the target depends on how many shards share
+ *   // the device's cores — with `shard_count` live shards each one owns
+ *   // ceil(cores / shard_count) cores, so the target must grow as shards
+ *   // shrink. Omitting both means no watermark (0).
  *
  *   // Required only when kBufferKinds > 1: discriminate which recycled bin
  *   // a finished buffer belongs to. Single-kind modules omit this method;
@@ -191,16 +205,6 @@
 #include "host/profiling_copy.h"
 
 namespace profiling_common {
-
-template <typename Module, typename = void>
-struct ProfilerModuleDrainThreadCount {
-    static constexpr int value = 1;
-};
-
-template <typename Module>
-struct ProfilerModuleDrainThreadCount<Module, std::void_t<decltype(Module::kMgmtDrainThreadCount)>> {
-    static constexpr int value = Module::kMgmtDrainThreadCount;
-};
 
 template <typename Derived, typename ReadyBufferInfo, typename = void>
 struct ProfilerDerivedShardAwareCollector {
@@ -468,11 +472,12 @@ struct ProfilerAlgorithms {
             }
         });
 
+        const int shard_count = mgr.shard_count();
         uint64_t pushed = 0;
         for (int kind = 0; kind < Module::kBufferKinds; kind++) {
             if (buffer_sizes[static_cast<size_t>(kind)] == 0) continue;
-            for (int shard = 0; shard < static_cast<int>(Mgr::kCollectorShardCount); shard++) {
-                size_t target = clamped_recycled_warm_target<Mgr>(kind, shard);
+            for (int shard = 0; shard < shard_count; shard++) {
+                size_t target = clamped_recycled_warm_target<Mgr>(kind, shard, shard_count);
                 if (target == 0) continue;
                 size_t current = mgr.recycled_count(kind, shard);
                 if (current >= target) continue;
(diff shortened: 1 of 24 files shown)


src/common/platform/include/host/profiler_base.h
```diff
@@ -503,8 +525,8 @@ struct ProfilerAlgorithms {
     static int recycled_warm_target_impl(int, int, ...) { return 0; }
 
     template <typename Mgr>
-    static size_t clamped_recycled_warm_target(int kind, int shard) {
-        int target = recycled_warm_target(kind, shard);
+    static size_t clamped_recycled_warm_target(int kind, int shard, int shard_count) {
```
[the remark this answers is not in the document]
yanghaoran29: Not applied — this is a false positive. `shard` is not unused: it is passed to the `LOG_WARN` on the clamp path (`"...for shard=%d kind=%d..."`, `Module::kSubsystemName, shard, kind, ...`) to report which shard's watermark exceeded capacity. Applying the suggested `int /*shard*/` would make that `shard` reference undeclared and break the build under `-Werror`. The parameter that genuinely became index-independent is the watermark target, which is why `recycled_warm_target` now takes `shard_count` (a count) instead of `shard` (an index); the per-shard `shard` arg is retained solely for the diagnostic. Resolving as-is.

src/common/platform/shared/host/args_dump_collector.cpp
```diff
@@ -55,6 +55,17 @@ int ArgsDumpCollector::initialize(
         LOG_ERROR("ArgsDumpCollector already initialized");
         return -1;
     }
+    if (num_dump_threads <= 0 || num_dump_threads > PLATFORM_MAX_AICPU_THREADS) {
+        LOG_ERROR(
+            "TensorDumpCollector::initialize: invalid num_dump_threads=%d (valid range: 1-%d)", num_dump_threads,
```
ChaoZheng109: 日志串写成了 `TensorDumpCollector::initialize`,但本类在 #1270(TensorDump→ArgsDump 改名)后已叫 `ArgsDumpCollector`——`TensorDumpCollector` 现在全仓已不是任何真实符号(仅剩这一处字符串)。建议改为 `ArgsDumpCollector::initialize`,否则排障时会指向一个不存在的类,也局部回退了 #1270 的改名。

src/common/platform/include/host/profiler_base.h
```diff
@@ -630,6 +641,34 @@ class ProfilerBase {
...
+     *                  threads can produce. Always `aicpu_thread_num`.
+     *   shard_count_ — how many drain/collector threads (== host shards) to
+     *                  run. Capped by Module::kMaxCollectorThreads, which is 1
+     *                  for the orchestrator-only subsystems (DepGen,
+     *                  ScopeStats): they have a single device-side producer, so
+     *                  one drain thread scanning all `queue_count_` queues is
+     *                  the whole job.
+     *
+     * The two are equal for the subsystems whose producers are the scheduler
+     * threads (L2Swimlane, TensorDump, PMU).
```
ChaoZheng109: 新增注释里的 `TensorDump` 是 #1270 之前的旧名;改名后该子系统统一叫 `ArgsDump`(本文件其余注释,如第 14/167/218 行,已是 ArgsDump)。建议这里也改成 `ArgsDump` 保持一致。

ChaoZheng109 (commented):
整体 LGTM:三量拆分(kMaxCollectorThreads 容量 / queue_count_ / shard_count_)清晰,SPSC 不变式(drain 数 = collector 数 = 扫描 stride = shard_count_,drain i 只喂 shard i)对任意 (shard, queue) 组合都成立;回收水位按 shard_count 缩放、L2 sched-phase 缓冲按 aicpu_thread_num 分配、空 shard 不再误触 idle-timeout 三处修复到位;测试含真机 dropped==0 门禁 + UT 回归,CI 全绿。

两处小问题(见 inline):局部回退了 #1270 的 TensorDump→ArgsDump 改名。

另外:PR 描述提到新增 `docs/dfx/dynamic-dfx-thread-scaling.md`,但本 PR 中未找到该文件(只新增了 test),麻烦补充该文档或更正描述。

src/common/platform/shared/host/args_dump_collector.cpp
yanghaoran29: Fixed in `c977871f` — log string 改为 `ArgsDumpCollector::initialize`,与 #1270 改名后的类名一致。

src/common/platform/include/host/profiler_base.h
yanghaoran29: Fixed in `c977871f` — 注释改为 `ArgsDump`,与本文件其余注释保持一致。

ChaoZheng109 (approved):

PR #1359, opened 2026-07-14
Add selective task-timing slots for lightweight dispatch/finish timing

### LINKED ISSUE #1325: [Feature] Add lightweight selective task dispatch/finish timing slots
### Summary

Add 16 fixed, per-run task-timing slots to the existing lightweight AICPU device-phase transport. Orchestration can attach an optional slot id (`0..15`) to selected L2 tasks, and the Scheduler records the task/window's AICPU `dispatch` and `finish` timestamps into that slot.

The host should reset the slots before each run, read them back once after stream synchronization, and expose each valid slot on the existing device-clock `[STRACE]` timeline. This path must work with L2 swimlane disabled and must not start L2 collector threads or write per-task AICore profiling records.

### Motivation / Use Case

L2 swimlane can capture every task's AICore start/end and AICPU dispatch/finish timestamps, but that is much heavier than many day-to-day measurements require. Enabling it allocates rotating record pools, runs host collector threads, stamps every task on AICPU/AICore, and performs AICore record write-back. The measured observer effect for back-to-back AICore tasks is about 0.8 us per switch, dominated by timestamp reads and record write-back.

A common question is much smaller: measure the interval from task A's first Scheduler dispatch to task B's final Scheduler-observed completion. The orchestrator knows which tasks bound that interval, but it cannot observe their asynchronous execution. The Scheduler already owns the exact dispatch and FIN-observation points, so selected fixed slots provide the needed answer without collecting a full timeline.

The repository already has the right transport: the device-phase buffer is host-allocated, reset with one H2D copy before a run, written by AICPU, and read with one D2H copy after the streams synchronize. Extending that fixed buffer by 16 small records is sufficient.

### Proposed API / Behavior

A suggested RT2 API is an optional field on `L0TaskArgs`, which avoids changing the `PTO2RuntimeOps::submit_task` function-pointer ABI:

```cpp
L0TaskArgs first_args;
first_args.set_task_timing_slot(0);
rt_submit_aiv_task(FIRST_KERNEL_ID, first_args);

L0TaskArgs last_args;
last_args.set_task_timing_slot(1);
rt_submit_aiv_task(LAST_KERNEL_ID, last_args);
```

`L0TaskArgsWithDeps` should forward the same setter. The legacy a5 `host_build_graph` orchestration API needs an equivalent way to set the slot on a submitted task.

Required behavior:

- Provide 16 slots. The default sentinel means "not recorded"; valid ids are `0..15`. Out-of-range ids fail through the existing invalid-argument path and never write out of bounds.
- Carry the id from Orch to Scheduler with the task metadata. In RT2, the 4-byte alignment gap after `PTO2TaskDescriptor::kernel_id[3]` can hold the id without growing the descriptor; protect that property with size/offset assertions.
- Extend the existing device-phase allocation with a dedicated per-Scheduler-thread `TaskTimingRecord[16]` tail rather than turning these records into named `AicpuPhase` enum values. This preserves the current once-per-run phase contract while reusing the same base pointer and H2D/D2H copies.
- For each slot, aggregate `min(dispatch_cycle)` and `max(finish_cycle)` across participating Scheduler threads. Reusing a slot for multiple tagged tasks intentionally produces the window from the earliest tagged dispatch to the latest tagged finish; using distinct slots preserves each task's own window and lets tooling compute `finish(B) - dispatch(A)`.
- `dispatch` means the earliest Scheduler publication of `DATA_MAIN_BASE`, after payload publication and immediately before the register write. Pending/speculative early dispatch therefore records the initial gated publication, not the later doorbell release or exact AICore kernel start.
- `finish` means the latest Scheduler observation of FIN across every required block/subtask, sampled after the COND load + `rmb()` and before fanin/fanout/deferred-completion processing. It matches the existing L2 swimlane `finish_time` definition; it is not the AICore kernel-end timestamp or completion of later deferred conditions.
- Emit every complete slot as a stable device-clock span such as `simpler_run.runner_run.device_wall.task_slot_0`, retaining start and end so cross-slot intervals are recoverable. Keep `Worker.run()` returning `None`; `[STRACE]` remains the timing source of truth.
- Reset all slots every run. Unset or incomplete slots are skipped, so a failed/short run cannot leak stale data from the previous run.
- Untagged tasks must not call `get_sys_cnt_aicpu()`, write timing records, or enable L2 collectors. Their only incremental hot-path work should be the cache-hot sentinel check. Do not add a new environment variable or compile-time behavior gate; explicit task tagging is the opt-in. The feature should remain available in `SIMPLER_DFX=0` builds.
- Support a2a3/a5, onboard/sim, and both `tensormap_and_ringbuffer` and `host_build_graph`. The a5 `host_build_graph` path uses its legacy `Runtime::Task` representation, so it needs a small compatibility implementation rather than assuming the RT2 descriptor path.

### Alternatives Considered

- **Use L2 swimlane level 2.** It already exposes dispatch/finish, but still records every task and retains the AICore timing/rotation path. That is the overhead this feature is intended to avoid.
- **Measure in Orch.** Orch only creates/submits the DAG; execution is asynchronous, so an Orch timestamp cannot represent task completion.
- **Store the records in the PTO2 shared-memory header.** This duplicates host/device transport logic and complicates a5 host-shadow handling. The existing fixed device-phase buffer already supplies reset, publication, conversion, and readback on both architectures.
- **Add 16 ordinary `AicpuPhase` entries.** Task slots can receive multiple block/subtask events per Scheduler thread, while `AicpuPhase` is explicitly a small set of phases that fires once per run/thread. A dedicated fixed tail keeps those contracts separate.

### Additional Context

Relevant implementation seams:

- Fixed device-phase contract and records: `src/common/platform/include/common/device_phase.h`
- AICPU phase-buffer accessors: `src/common/platform/include/aicpu/device_phase_aicpu.h`
- Onboard H2D reset / D2H reduce: `src/common/platform/onboard/host/device_runner_base.cpp`
- Host marker emission: `src/common/platform/{onboard,sim}/host/c_api_shared.cpp`
- Orch API and task metadata: `src/{arch}/runtime/*/{orchestration,runtime}/`
- RT2 dispatch/finish boundaries: `runtime/scheduler/scheduler_dispatch.cpp` and `scheduler_completion.cpp`
- Legacy a5 host-build-graph task path: `src/a5/runtime/host_build_graph/`
- Existing timing docs: `docs/dfx/device-phases.md`, `docs/dfx/l2-timing.md`, and `docs/dfx/l2-swimlane-profiling.md`

Acceptance coverage should include:

1. Unit coverage for Arg-to-descriptor propagation, slot bounds/sentinel, fixed-buffer layout, cross-thread min/max reduction, incomplete slots, and per-run reset.
2. End-to-end sim coverage with L2 swimlane off for a normal chain plus MIX/SPMD tasks, proving earliest-dispatch/latest-FIN aggregation and stable `task_slot_N` markers.
3. Parity coverage for a2a3/a5 and both runtime variants, including the legacy a5 host-build-graph path.
4. At least one onboard check of the real H2D/D2H path, following the repository's architecture precheck and `task-submit` rules.
5. Updated device-phase/L2-timing/runtime documentation, including early-dispatch, deferred-completion, duplicate-slot, dummy-task, and `SIMPLER_DFX=0` semantics.

Related: #995, #1299

doraemonmj:
## Summary

Adds 16 fixed, per-run task-timing slots on the existing lightweight
device-phase transport. Orchestration tags selected tasks with a slot id
(`0..15`); the Scheduler folds each tagged task's AICPU `dispatch`/`finish`
cycles into that slot. The host resets slots before each run, reads them
back once after stream sync, and emits each complete slot as a device-clock
`simpler_run.runner_run.device_wall.task_slot_N` `[STRACE]` span.

Works with L2 swimlane **disabled** and in `SIMPLER_DFX=0` builds — it starts
no collector threads and writes no per-task AICore records. Untagged tasks
pay only a cache-hot sentinel check (`!= TASK_TIMING_SLOT_NONE`); they never
read the sys counter.

Closes #1325.

## Design

- **Transport**: a fixed `TaskTimingRecord[16]` tail appended after the
  `AicpuPhaseRecord` region in the same device buffer — same base pointer,
  same per-run H2D reset and post-sync D2H copy.
- **ABI**: RT2 carries the id in the 4-byte pad after
  `PTO2TaskDescriptor::kernel_id[3]`, so the descriptor does not grow;
  `static_assert`s pin its size and `packed_buffer_base` offset.
- **Aggregation**: per slot, `min(dispatch)` / `max(finish)` across Scheduler
  threads and across a MIX task's subtasks / an SPMD task's blocks. Reusing a
  slot merges into one window; distinct slots recover `finish(B) - dispatch(A)`.
- **Boundaries** match the L2 swimlane: `dispatch` = earliest gated
  `DATA_MAIN_BASE` publication; `finish` = latest FIN observation (after
  `rmb()`, before fanin/deferred-completion). Incomplete/unset slots are
  skipped, so a short/failed run leaks no stale data.
- **a5 host_build_graph** uses its legacy `Runtime::Task` path, so it gets a
  small `set_task_timing_slot(runtime, task_id, slot)` compatibility setter
  rather than the RT2 `L0TaskArgs` field.

## Coverage

Supports a2a3/a5, onboard/sim, and both `tensormap_and_ringbuffer` and
`host_build_graph`, plus unit / sim-e2e / onboard tests and device-phase /
L2-timing docs.

## Verification

**Unit — a2a3** (`tests/ut/cpp/a2a3/test_task_timing_slots.cpp`, 9/9 pass):
arg->descriptor propagation, slot bounds/sentinel, fixed-buffer layout +
offsets, cross-thread min/max reduction, incomplete-slot skip, per-run reset.

**Build:** a2a3sim + a5sim + a2a3 (onboard) all build; a2a3 cpput subset plus
the 9 new cases pass 100%.

**e2e — a2a3, sim + onboard** (`tests/st/task_timing/task_timing_slots/`, L2
swimlane **OFF**): all pass on both sim and onboard. The first four cases run
under `tensormap_and_ringbuffer`; the a2a3 `host_build_graph` fold path (the
Scheduler `PublishHandle` in `scheduler_context.h` / `scheduler_dispatch.cpp`,
distinct from the inline tensormap dispatch) is covered by two
`host_build_graph`-tagged cases that reuse the same RT2 orch:

| test | runtime | checks |
|---|---|---|
| `test_distinct_slots_emit_markers` | tensormap | chain, `dispatch(slot1) >= finish(slot0)` |
| `test_duplicate_slot_merges_window` | tensormap | 3 tasks, same slot -> one merged window > single-task baseline |
| `test_spmd_task_aggregates_across_threads` | tensormap | `block_num=8` -> cross-thread min/max -> one complete slot |
| `test_mix_task_aggregates_across_subtasks` | tensormap | MIX AIC+AIV0+AIV1 subtasks fold into one slot |
| `test_hbg_distinct_slots_emit_markers` | host_build_graph | a2a3 hbg fold path, distinct-slot dependency ordering |
| `test_hbg_duplicate_slot_merges_window` | host_build_graph | a2a3 hbg fold path, same-slot min/max merge |

Measured `[STRACE]` (a2a3, ns, device clock):
- **sim:** `task_slot_0 ts=14600 dur=428500`, `task_slot_1 ts=444240 dur=49560`
  -> `dispatch(0) < finish(0) < dispatch(1)` (dependency respected),
  whole-chain `finish(1) - dispatch(0) = 479200`.
- **onboard** (Ascend910 via `task-submit`, real H2D/D2H):
  `task_slot_0 ts=270620 dur=8280`, `task_slot_1 ts=284260 dur=7979`,
  whole-chain `21619` — confirms H2D reset -> AICPU fold -> D2H tail readback
  -> emit on real silicon.

**e2e — a5 host_build_graph, onboard (passing):**
`tests/st/task_timing/task_timing_a5hbg/test_task_timing_a5hbg.py::test_a5hbg_task_timing_slots_emit_markers`,
~17.5s — the legacy `add_task` path end to end: both `task_slot_0`/`task_slot_1`
markers with positive duration, `dispatch(slot1) >= finish(slot0)`, golden
correct, swimlane off.

**Static / lint — local pre-commit (all pass):** clang-format, clang-tidy,
cpplint, ruff (check + format), pyright, plus header / English-only /
trailing-whitespace / large-file gates.

**Integration:** rebases cleanly onto current `main`. The RT2 tag rides in
`PTO2TaskDescriptor`'s existing 4-byte pad after `kernel_id[3]` (no descriptor
growth), and `Arg` gains a plain `task_timing_slot_` member (defaulted to
`TASK_TIMING_SLOT_NONE`, reset in `clear()`); the `PTO2TaskDescriptor` size /
`packed_buffer_base`-offset `static_assert`s hold.

**Not run locally — relies on CI:** a5 `tensormap_and_ringbuffer` (its a5sim
e2e needs g++-15; this box has g++-12).

### DIFF
--- a/src/a2a3/runtime/host_build_graph/runtime/pto_types.h
+++ b/src/a2a3/runtime/host_build_graph/runtime/pto_types.h
@@ -37,6 +37,7 @@
 #endif
 
 #include "aicpu/dump_arg_selection.h"
+#include "common/device_phase.h"
 #include "data_type.h"
 #include "profiling_config.h"
 #include "pto_submit_types.h"
@@ -231,6 +232,20 @@ struct Arg : TaskArgsTpl<TensorRef, uint64_t, MaxT, MaxS, TensorArgType> {
     void set_allow_early_resolve(bool v = true) { allow_early_resolve_ = v; }
     bool allow_early_resolve() const { return allow_early_resolve_; }
 
+    // Selective task-timing slot: tag this task to have the scheduler record its
+    // AICPU dispatch/finish cycles into fixed slot `slot` (0..15). Untagged by
+    // default. An out-of-range id fails through the standard invalid-arg path so
+    // the scheduler never stamps out of bounds.
+    int32_t task_timing_slot_{TASK_TIMING_SLOT_NONE};
+    void set_task_timing_slot(int32_t slot) {
+        if (slot < 0 || slot >= NUM_TASK_TIMING_SLOTS) {
+            set_error("task_timing_slot out of range (valid: 0..15)");
+            return;
+        }
+        task_timing_slot_ = slot;
+    }
+    int32_t task_timing_slot() const { return task_timing_slot_; }
+
     void clear() {
         Base::clear();
 #if SIMPLER_DFX
@@ -239,6 +254,7 @@ struct Arg : TaskArgsTpl<TensorRef, uint64_t, MaxT, MaxS, TensorArgType> {
         explicit_deps_ = nullptr;
         explicit_dep_count_ = 0;
         allow_early_resolve_ = false;
+        task_timing_slot_ = TASK_TIMING_SLOT_NONE;
     }
 
     void reset() {
--- a/src/a2a3/runtime/tensormap_and_ringbuffer/runtime/pto_types.h
+++ b/src/a2a3/runtime/tensormap_and_ringbuffer/runtime/pto_types.h
@@ -37,6 +37,7 @@
 #endif
 
 #include "aicpu/dump_arg_selection.h"
+#include "common/device_phase.h"
 #include "data_type.h"
 #include "profiling_config.h"
 #include "pto_submit_types.h"
@@ -262,6 +263,20 @@ struct Arg : TaskArgsTpl<TensorRef, uint64_t, MaxT, MaxS, TensorArgType> {
     void set_predicate(const L0TaskPredicate &pred) { predicate_ = pred; }
     const L0TaskPredicate &predicate() const { return predicate_; }
 
+    // Selective task-timing slot: tag this task to have the scheduler record its
+    // AICPU dispatch/finish cycles into fixed slot `slot` (0..15). Untagged by
+    // default. An out-of-range id fails through the standard invalid-arg path so
+    // the scheduler never stamps out of bounds.
+    int32_t task_timing_slot_{TASK_TIMING_SLOT_NONE};
+    void set_task_timing_slot(int32_t slot) {
+        if (slot < 0 || slot >= NUM_TASK_TIMING_SLOTS) {
+            set_error("task_timing_slot out of range (valid: 0..15)");
+            return;
+        }
+        task_timing_slot_ = slot;
+    }
+    int32_t task_timing_slot() const { return task_timing_slot_; }
+
     void clear() {
         Base::clear();
 #if SIMPLER_DFX
@@ -271,6 +286,7 @@ struct Arg : TaskArgsTpl<TensorRef, uint64_t, MaxT, MaxS, TensorArgType> {
         explicit_dep_count_ = 0;
         allow_early_resolve_ = false;
         predicate_ = L0TaskPredicate{};
+        task_timing_slot_ = TASK_TIMING_SLOT_NONE;
     }
 
     void reset() {
--- a/src/a2a3/runtime/tensormap_and_ringbuffer/runtime/scheduler/scheduler_completion.cpp
+++ b/src/a2a3/runtime/tensormap_and_ringbuffer/runtime/scheduler/scheduler_completion.cpp
@@ -400,21 +400,18 @@ void SchedulerContext::check_running_cores_for_completion(
         }
 #endif
 
-        // Task-timing finish: latest FIN observation for a tagged task, folded as
-        // max at the same point L2 captures finish_time (after rmb, before fanin/
-        // deferred-completion work). Independent of L2 swimlane level so it works
-        // in SIMPLER_DFX=0 builds; untagged tasks pay only the cache-hot compare.
-        if (t.pending_done && core.pending_slot_state->task->task_timing_slot != TASK_TIMING_SLOT_NONE) {
-            aicpu_task_timing_finish(core.pending_slot_state->task->task_timing_slot, thread_idx);
-        }
-        if (t.running_done && core.running_slot_state->task->task_timing_slot != TASK_TIMING_SLOT_NONE) {
-            aicpu_task_timing_finish(core.running_slot_state->task->task_timing_slot, thread_idx);
-        }
-
         // --- Apply phase: execute actions based on transition ---
 
         // 1. Complete finished tasks (capture pointers before modifying core state)
         if (t.pending_done) {
+            // Task-timing finish: latest FIN observation for a tagged task, folded
+            // as max. Sampled after the rmb above and before complete_slot_task runs
+            // fanin / deferred-completion (which may also clear pending_slot_state),
+            // matching L2's finish_time point. Independent of L2 swimlane level, so
+            // it works in SIMPLER_DFX=0 builds; untagged tasks pay only the compare.
+            if (core.pending_slot_state->task->task_timing_slot != TASK_TIMING_SLOT_NONE) {
+                aicpu_task_timing_finish(core.pending_slot_state->task->task_timing_slot, thread_idx);
+            }
             complete_slot_task(
                 *core.pending_slot_state, core.pending_reg_task_id, core.pending_subslot, thread_idx, core_id, hank,
                 completed_this_turn, deferred_release_slot_states, deferred_release_count
@@ -426,6 +423,9 @@ void SchedulerContext::check_running_cores_for_completion(
             cur_thread_completed++;
         }
         if (t.running_done) {
+            if (core.running_slot_state->task->task_timing_slot != TASK_TIMING_SLOT_NONE) {
+                aicpu_task_timing_finish(core.running_slot_state->task->task_timing_slot, thread_idx);
+            }
             complete_slot_task(
                 *core.running_slot_state, core.running_reg_task_id, core.running_subslot, thread_idx, core_id, hank,
                 completed_this_turn, deferred_release_slot_states, deferred_release_count
@@ -616,7 +616,7 @@ SchedulerContext::drain_stage_cores(PTO2TaskSlotState *slot_state, int32_t block
             // contention).
             uint64_t my_mask[PTO2_EARLY_DISPATCH_CORE_MASK_WORDS] = {0};
             for (int i = 0; i < handle_count; i++) {
-                publish_subtask_to_core(handles[i], dispatch_ts);
+                publish_subtask_to_core(handles[i], dispatch_ts, thread_idx);
                 if (gated) {
                     int32_t cid = tracker.get_core_id_by_offset(handles[i].core_offset);
                     sched_->early_dispatch_doorbell_table[cid].addr = handles[i].reg_addr;
--- a/src/a2a3/runtime/tensormap_and_ringbuffer/runtime/scheduler/scheduler_completion.cpp
+++ b/src/a2a3/runtime/tensormap_and_ringbuffer/runtime/scheduler/scheduler_completion.cpp
@@ -400,6 +400,17 @@ void SchedulerContext::check_running_cores_for_completion(
         }
 #endif
 
+        // Task-timing finish: latest FIN observation for a tagged task, folded as
+        // max at the same point L2 captures finish_time (after rmb, before fanin/
+        // deferred-completion work). Independent of L2 swimlane level so it works
+        // in SIMPLER_DFX=0 builds; untagged tasks pay only the cache-hot compare.
+        if (t.pending_done && core.pending_slot_state->task->task_timing_slot != TASK_TIMING_SLOT_NONE) {
+            aicpu_task_timing_finish(core.pending_slot_state->task->task_timing_slot, thread_idx);
+        }
+        if (t.running_done && core.running_slot_state->task->task_timing_slot != TASK_TIMING_SLOT_NONE) {
+            aicpu_task_timing_finish(core.running_slot_state->task->task_timing_slot, thread_idx);
+        }
+
         // --- Apply phase: execute actions based on transition ---
 
         // 1. Complete finished tasks (capture pointers before modifying core state)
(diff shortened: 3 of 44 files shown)


src/a2a3/runtime/tensormap_and_ringbuffer/runtime/pto_types.h
```diff
@@ -262,6 +263,21 @@ struct Arg : TaskArgsTpl<TensorRef, uint64_t, MaxT, MaxS, TensorArgType> {
...
+    void set_task_timing_slot(int32_t slot) {
+        if (slot < 0 || slot >= NUM_TASK_TIMING_SLOTS) {
+            has_error = true;
+            error_msg = "task_timing_slot out of range (valid: 0..15)";
+            return;
+        }
+        task_timing_slot_ = slot;
+    }
+    int32_t task_timing_slot() const { return task_timing_slot_; }
+
```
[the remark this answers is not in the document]
doraemonmj: Fixed — the out-of-range case now routes through set_error(), preserving the sticky first-error invariant. Applied to all three PTO2 forks that carry this setter.

src/a2a3/runtime/host_build_graph/runtime/pto_types.h
```diff
@@ -231,6 +232,21 @@ struct Arg : TaskArgsTpl<TensorRef, uint64_t, MaxT, MaxS, TensorArgType> {
...
+    void set_task_timing_slot(int32_t slot) {
+        if (slot < 0 || slot >= NUM_TASK_TIMING_SLOTS) {
+            has_error = true;
+            error_msg = "task_timing_slot out of range (valid: 0..15)";
+            return;
+        }
+        task_timing_slot_ = slot;
+    }
+    int32_t task_timing_slot() const { return task_timing_slot_; }
+
```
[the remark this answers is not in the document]
doraemonmj: Fixed — same set_error() change applied here.

src/a2a3/runtime/tensormap_and_ringbuffer/runtime/scheduler/scheduler_completion.cpp
```diff
@@ -400,6 +400,17 @@ void SchedulerContext::check_running_cores_for_completion(
         }
 #endif
 
+        // Task-timing finish: latest FIN observation for a tagged task, folded as
+        // max at the same point L2 captures finish_time (after rmb, before fanin/
+        // deferred-completion work). Independent of L2 swimlane level so it works
+        // in SIMPLER_DFX=0 builds; untagged tasks pay only the cache-hot compare.
+        if (t.pending_done && core.pending_slot_state->task->task_timing_slot != TASK_TIMING_SLOT_NONE) {
```
ChaoZheng109: 这个 finish 折叠块重复评估了 `t.pending_done`/`t.running_done` —— 下面 apply 阶段的 `if (t.pending_done)` / `if (t.running_done)`(以及上面 L2 `finish_ts` 采集条件里的 `||`)已经判过。建议把折叠并入下面已有的 apply 块、置于 `complete_slot_task` **之前**,这样 done 谓词只判一次,本块可整体删除:

```cpp
if (t.pending_done) {
    if (core.pending_slot_state->task->task_timing_slot != TASK_TIMING_SLOT_NONE)
        aicpu_task_timing_finish(core.pending_slot_state->task->task_timing_slot, thread_idx);
    complete_slot_task(*core.pending_slot_state, ...);
    // ...
}
```

**位置约束(load-bearing):** `aicpu_task_timing_finish` 内部自采 `get_sys_cnt_aicpu()`,必须放在 `complete_slot_task` 之前 —— 后者执行 fanin/deferred-completion,若置其后,finish 采样点会漂到 fanin 之后,违反 issue #1325 "sampled after rmb, before fanin" 的定义;且 `complete_slot_task` 可能改动 `core.pending_slot_state`,读 `->task` 也应在其之前。

此改动不触碰 L2 的 `#if SIMPLER_DFX` 采集、保持隔离性,只消除重复判断。同样适用于逐字节相同的 a2a3/hbg 副本。

src/a2a3/runtime/tensormap_and_ringbuffer/runtime/scheduler/scheduler_context.h
```diff
@@ -248,6 +249,8 @@ class SchedulerContext {
         uint32_t reg_task_id;
         int32_t core_offset;
         uint64_t *dispatch_timestamp_slot;
+        int32_t task_timing_slot;  // TASK_TIMING_SLOT_NONE unless the task is tagged
+        int32_t thread_idx;        // publishing Scheduler thread, for the task-timing record
```
ChaoZheng109: `PublishHandle.thread_idx` 是冗余的:两个 `publish_subtask_to_core` 调用点的外层函数本就有 `thread_idx` 形参 —— `flush_publish` 在 `dispatch_shape(int32_t thread_idx, ...)` 内,early-dispatch 循环在 `stage_consumer_blocks(int32_t thread_idx, ...)` 内。可去掉此字段,改为:(a) 折叠挪到那两个循环里做(`thread_idx` 直接可用),或 (b) 给 `publish_subtask_to_core` 加 `thread_idx` 参数。

注:同结构里的 `task_timing_slot` 字段是**必需**的(每 handle 各一,flush 循环只持有 `handles[]`、无 task 指针),只有 `thread_idx` 冗余。SchedulerContext 方法参数里流转的 `thread_idx` 也不冗余 —— context 单实例多线程共享,靠它索引 per-thread 数组。

src/a2a3/runtime/tensormap_and_ringbuffer/runtime/scheduler/scheduler_completion.cpp
ChaoWao: 已在 `7a8b1dc9` 修复:把 finish 折叠并入下面的 apply 块(`if (t.pending_done)` / `if (t.running_done)`),置于 `complete_slot_task` **之前** —— done 谓词只判一次,采样点仍在 rmb 之后、fanin/deferred-completion 之前(符合 #1325 "sampled after rmb, before fanin"),且在 `complete_slot_task` 可能清空 `pending_slot_state` 之前读 `->task`。行为与原来完全一致。同样应用到逐字节相同的 a2a3/hbg 与 a5/tmr 副本。a2a3sim 上 tmr(4)+hbg(2)e2e 通过。

src/a2a3/runtime/tensormap_and_ringbuffer/runtime/scheduler/scheduler_context.h
ChaoWao: 已在 `7a8b1dc9` 修复:采用方案 (b) —— 删除 `PublishHandle.thread_idx` 字段,给 `publish_subtask_to_core` 加 `thread_idx` 形参。三个调用点(`dispatch_shape` 的 `flush_publish`、`stage_consumer_blocks` 的 early-dispatch、`drain_stage_cores` 的 publish 循环)本就有 `thread_idx` 在作用域内。`task_timing_slot` 字段保留(flush 循环只持 `handles[]`、无 task 指针)。a2a3 tmr + hbg 都已改;a5 用内联 dispatch、无 PublishHandle,不受影响。

ChaoWao (approved):

PR #1371, opened 2026-07-16
ci: pin ut-a2a3 to the x86 (X64) self-hosted runner

ChaoWao:
## What
Pin `ut-a2a3` to the x86 (`X64`) self-hosted runner.

## Why
The `a2a3` runner pool is mixed-arch:
- **x86 8P**: `infra-gpu-npu-021[-2]` (label `X64`)
- **arm64 4P**: `runner-dev4-7` / `runner-dev8-11` (label `ARM64`)

`runs-on: [self-hosted, a2a3]` can land `ut-a2a3` on a 4-NPU arm64 box, but the job assumes the 8-NPU x86 host — `DEVICE_RANGE` is `0-7`/`8-15`, and the local pto-isa mirror (which makes the build's `pto-isa` clone reliable behind the pod proxy) only exists there. Adding the `X64` label targets it deterministically.

This PR is also the validation run for that pto-isa local-mirror setup — the `Set up environment` step (the `pto-isa` clone) should now succeed.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -460,7 +460,11 @@ jobs:
   # ---------- Unit tests (a2a3 hardware, Python + C++) ----------
   ut-a2a3:
     needs: pre-commit
-    runs-on: [self-hosted, a2a3]
+    # The a2a3 pool mixes x86 8P hosts (infra-gpu-npu-021[-2]) and arm64 4P
+    # hosts (runner-dev*) under the same `a2a3` label. ut-a2a3 assumes the
+    # 8-NPU x86 host (DEVICE_RANGE 0-7/8-15; local pto-isa mirror lives here),
+    # so pin to the X64 label instead of matching the 4-NPU arm64 runners.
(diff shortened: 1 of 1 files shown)


ChaoWao:
Closing as test-only.

The actual fix lives on the runner, not in this repo: `a2-github-ci` (`infra-gpu-npu-021`) now hosts a local pto-isa mirror with a `url.insteadOf` redirect, so the build's `git clone` of pto-isa resolves to a same-FS hardlink clone (~20 ms) instead of going through the pod proxy — which was dropping the HTTP/2 pack with `curl 56 GnuTLS recv error / early EOF`. (pto-isa is only ~17 MB; the failure was the proxy, not size.)

This PR's `ut-a2a3` run validated it: green on `infra-gpu-npu-021-2` (the same runner that failed in #1267), `pto-isa` clone step passed.

The `X64` label pin here is optional hardening — it keeps `ut-a2a3` off the arm64 4P runners (which have no mirror). Happy to re-send as a real change if we want that pin permanent.

Issue #1351, opened 2026-07-13
[Code Health] Extend PTO-ISA build/run version guard to a5 onboard SDMA overlay (missed after #1179)


ChaoZheng109:
### Category

Missing Validation

### Component

Build System

### Description

The PTO-ISA build/runtime version guard added in #1096 (record the pto-isa
commit runtimes were built with, and reject stale binaries at load) and #1194
(fold the pto-isa commit into the cmake cache stamp + ccache key so a pin bump
forces a real recompile) was **deliberately scoped to a2a3 onboard**, because
a2a3 onboard was the only variant that baked pto-isa headers into
`host_runtime.so` at the time.

**#1179 changed that assumption.** It made a5 onboard `host_runtime.so` embed
pto-isa headers (and link `libnnopbase`) whenever
`SIMPLER_ENABLE_PTO_SDMA_WORKSPACE=ON` — i.e. a5 now has exactly the
compile-time pto-isa dependency the guard exists to protect. But #1179 only
wired the *enable* path (CMake option, `PTO_ISA_ROOT` requirement, header
include, nnopbase link); it did **not** extend any of the version-guard
machinery to a5. Every guard still keys on `arch == "a2a3" and variant ==
"onboard"`, so an a5 build with the overlay enabled embeds pto-isa headers with
**none** of the pin/staleness protections. When someone flips the overlay ON
(the re-enable path tracked in #1315), a pto-isa pin bump will silently serve a
stale a5 `host_runtime.so` built against the old headers — the exact
507018 / 507899-class failure #1194 was written to prevent, now un-guarded on
a5.

A second, subtler point: the a2a3 guard's coarse `arch/variant` scoping is only
correct because a2a3 **forces** `SIMPLER_ENABLE_PTO_SDMA_WORKSPACE` ON
unconditionally (`_requires_pto_isa_metadata_validation` docstring says as
much). On a5 the overlay is **opt-in** (default OFF), so an a5 guard must key on
the env var / CMake option, not on `arch == "a5"` — otherwise a default (OFF)
a5 build that embeds no pto-isa headers would be forced to clone pto-isa and
record metadata for nothing. So this is not a one-line `or arch == "a5"` fix.

Concretely, the gaps are:

1. **No managed-pin resolution for a5.** `build_runtimes.py` resolves the
   pinned managed checkout (`ensure_pto_isa_root()`) and exports `PTO_ISA_ROOT`
[...]

Issue #1315, opened 2026-07-10
[Code Health] Require upgraded a5 environment supporting sdma


jvjhfhg:
### Category

Technical Debt (cleanup, refactor)

### Component

Host Runtime

### Description

PR #1179 landed the a5 SDMA workspace overlay (`ensure_sdma_workspace` → `aclnnShmemSdmaStarsQuery`) but gated it behind an opt-in macro because the available a5 CANN drops do not expose working SDMA primitives:

- CANN 9.1.T500: `aclnnShmemSdmaStarsQuery` creates STARS streams but `aclrtSynchronizeStream` fails with AICPU exception `0x715002a`, which poisons the AICPU context and turns every subsequent kernel launch into `507018`. This breaks **all** a5 communication cases, not just the SDMA demo.
- CANN 9.1.0 (`timestamp=20260625`): `HcclCommInitRootInfo` itself returns `HCCL_E_INTERNAL (4)` — base HCCL never comes up.

The overlay was verified working on a separate a5 box whose CANN does expose the primitive (`sdma_async_completion_demo` passes). So this is purely an environment gating issue, not a code defect.

Currently the overlay defaults OFF via `option(SIMPLER_ENABLE_PTO_SDMA_WORKSPACE ... OFF)` and the `SIMPLER_ENABLE_PTO_SDMA_WORKSPACE` env var. With it OFF: `ensure_sdma_workspace` is a no-op, `CommContext.workSpace` stays 0, the SDMA demo self-skips, and all other comm cases run normally.

*(revised 2026-07-22)* PR #1392 (`ca000263`) added the URMA deferred completion backend behind the same gating pattern (`SIMPLER_ENABLE_PTO_URMA_WORKSPACE`, default OFF), and PR #1421 (`a7a29ee3`) added its two-device demo. The two overlays are **mutually exclusive** — see the 2026-07-22 update below.

### Location

*(revised 2026-07-22 — extended to both overlays and the post-#1403 mechanism)*

```markdown
[...]

PR #1392, opened 2026-07-17
feat(a5): add URMA deferred completion backend for tensormap_and_ringbuffer

wxwnnzdyd:
## Summary
- Add AICore-side URMA async request submission for TGET/TPUT.
- Register URMA async events as deferred completion tokens.
- Add AICPU-side URMA CQ polling and task retirement.
- Wire URMA completion type through the existing async wait framework.
- Initialize PTO-ISA URMA workspace in A5 onboard host communication setup.

## Testing
- Real A5 deferred demo validated manually on devices 0,1.
- `git diff --check`

### DIFF
--- a/simpler_setup/runtime_builder.py
+++ b/simpler_setup/runtime_builder.py
@@ -346,16 +346,12 @@ class RuntimeBuilder:
                 defines: dict[str, str] = {}
                 if build_pto_isa_commit:
                     defines["SIMPLER_PTO_ISA_BUILD_COMMIT"] = build_pto_isa_commit
-                # Mirror the SIMPLER_ENABLE_PTO_SDMA_WORKSPACE env var into the
-                # CMake option (src/{arch}/platform/onboard/host/CMakeLists.txt)
-                # so the a5 SDMA workspace overlay is built only when opted in.
-                if os.environ.get("SIMPLER_ENABLE_PTO_SDMA_WORKSPACE", "").upper() in {
-                    "1",
-                    "ON",
-                    "TRUE",
-                    "YES",
-                }:
-                    defines["SIMPLER_ENABLE_PTO_SDMA_WORKSPACE"] = "ON"
+                for opt_in_define in (
+                    "SIMPLER_ENABLE_PTO_SDMA_WORKSPACE",
+                    "SIMPLER_ENABLE_PTO_URMA_WORKSPACE",
+                ):
+                    if os.environ.get(opt_in_define, "").upper() in {"1", "ON", "TRUE", "YES"}:
+                        defines[opt_in_define] = "ON"
                 cmake_defines = defines or None
             # compile() adds a {target}/ subdirectory inside build_dir
             cache_dir = self._CACHE_DIR / arch / variant / name
--- a/simpler_setup/runtime_compiler.py
+++ b/simpler_setup/runtime_compiler.py
@@ -23,7 +23,7 @@ from .toolchain import Aarch64GxxToolchain, CCECToolchain, GxxToolchain, Toolcha
 logger = logging.getLogger(__name__)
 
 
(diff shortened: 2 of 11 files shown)


jvjhfhg:
**Requesting changes: the host-side URMA workspace needs the same opt-in macro gating as the SDMA overlay — the a5 CI environment is not ready for it (see #1315).**

The reason for gating is environmental, not code quality: as documented in #1315, the current st-onboard-a5 CI CANN drops do not expose working async primitives (CANN 9.1.T500: `aclrtSynchronizeStream` AICPU exception `0x715002a` poisons the context and turns every subsequent kernel launch into `507018`; CANN 9.1.0 `timestamp=20260625`: `HcclCommInitRootInfo` returns `HCCL_E_INTERNAL (4)`). #1179's SDMA overlay was gated behind `SIMPLER_ENABLE_PTO_SDMA_WORKSPACE` (default OFF) for exactly this reason. This PR wires the PTO-ISA URMA workspace in **unconditionally**, going further than the SDMA overlay did:

1. **CMake**: `PTO_ISA_ROOT` becomes a hard `FATAL_ERROR` requirement for *every* a5 host_runtime build (`src/a5/platform/onboard/host/CMakeLists.txt`). The SDMA overlay guards this behind `option(SIMPLER_ENABLE_PTO_SDMA_WORKSPACE ... OFF)`.
2. **`comm_hccl.cpp`**: the `#include "pto/comm/async/urma/urma_workspace_manager.hpp"`, the `urma_workspace` members on `CommHandle_` / `DomainAllocation`, and the init calls are all unconditional. Compare with the SDMA paths, which are `#ifdef SIMPLER_ENABLE_PTO_SDMA_WORKSPACE` / no-op when OFF.
3. **Failure blast radius**: `comm_alloc_windows` now does `if (!ensure_base_urma_workspace(h)) return -1;` (+ `file_barrier`), and `domain_alloc_via_ipc` returns -1 when URMA init fails on dense mappings. On the current CI CANN, **every** a5 communication case would break — the exact failure mode #1315 tracks for SDMA.
4. **No escape hatch**: no env-var→CMake-define forwarding in `runtime_compiler.py` / `runtime_builder.py`, no `_urma_workspace_enabled()`-style check, no test skip gate. The `PTO_URMA_SUPPORTED` ifdefs in `urma_completion_kernel.h` are a pto-isa capability macro, not a simpler-side opt-in, and the host side doesn't consult it.
5. **Version guard gap**: per #1315 / #1351, compiling pto-isa headers into a5 `host_runtime.so` today happens with no pin verification, no build-metadata recording, no load-time staleness rejection, and no cmake-cache/ccache invalidation on a pin bump — the #1096/#1194 guard is still `arch == "a2a3"` only. This PR makes that exposure unconditional.

Asks:

- Gate the host-side URMA workspace behind an opt-in macro (e.g. `SIMPLER_ENABLE_PTO_URMA_WORKSPACE`, or fold into the existing SDMA option), mirroring the SDMA pattern: CMake option default OFF, `#ifdef`-guarded include/members/init, no-op `ensure_*` when OFF, env-var forwarding in `runtime_builder.py`, conditional `PTO_ISA_ROOT` ensure in `runtime_compiler.py _init_a5`, and a `pytest.skip` gate in the demo test.
- When OFF, URMA-init failure must not fail `comm_alloc_windows` / `comm_alloc_domain_windows`.
- Please leave a comment on #1315 supplementing the URMA-side follow-ups that must happen once the a5 CI CANN is upgraded (re-enable steps / verification for URMA alongside the SDMA Option A/B checklist), so #1315 remains the single tracking point for "re-enable a5 async overlays after CI env is ready" — including "#1351 resolved" as an explicit prerequisite before any default-ON flip.

jvjhfhg (approved):

PR #1401, opened 2026-07-20
Fix: bind runner TLS during arena prewarm

hashiqiqixian:
## Summary

- bind the active `DeviceRunner` in thread-local storage for the full `simpler_init()` call
- clear the binding on every return path with `RAIIScopeGuard`
- prevent eager arena prewarming from dereferencing a null runner on sim and onboard
- add an in-repo a2a3sim TRB regression test for L2 `Worker.init(prewarm_config=...)`

## Root cause

`HostApi` callbacks recover the current runner through thread-local storage. The eager prewarm path runs during `simpler_init()`, before the existing register/run entry points establish that binding.

## Validation

- `git diff --check`
- `clang-format --dry-run --Werror` on both modified C++ files
- Python syntax validation for `test_prewarm_config.py`
- Simpler PR CI runs the new test under the existing a2a3sim scene-test job
- downstream PyPTO #2085 previously passed the five system jobs that crashed before the fix

A full local runtime rebuild was not run because the development host had only 3.07 GB available RAM; PR CI provides the runtime build and simulator execution.

## Related

- https://github.com/hw-native-sys/pypto/pull/2085
- https://github.com/hw-native-sys/pypto/pull/2068

### DIFF
--- a/src/common/platform/onboard/host/c_api_shared.cpp
+++ b/src/common/platform/onboard/host/c_api_shared.cpp
@@ -330,6 +330,13 @@ int simpler_init(
     if (ctx == NULL) return -1;
 
     DeviceRunnerBase *runner = static_cast<DeviceRunnerBase *>(ctx);
+    // HostApi callbacks, including the prewarm path below, recover their
+    // DeviceRunner from this thread-local binding.
+    pthread_once(&g_runner_key_once, create_runner_key);
+    pthread_setspecific(g_runner_key, ctx);
+    auto tsd_guard = RAIIScopeGuard([]() {
+        pthread_setspecific(g_runner_key, nullptr);
+    });
 
     // CANN dlog must be levelled BEFORE the device context is opened
     // (rtSetDevice inside attach_current_thread): CANN snapshots the
--- a/src/common/platform/sim/host/c_api_shared.cpp
+++ b/src/common/platform/sim/host/c_api_shared.cpp
@@ -331,6 +331,14 @@ int simpler_init(
     if (ctx == NULL) return -1;
 
     SimDeviceRunnerBase *runner = static_cast<SimDeviceRunnerBase *>(ctx);
+    // HostApi callbacks, including the prewarm path below, recover their
+    // DeviceRunner from this thread-local binding.
+    pthread_once(&g_runner_key_once, create_runner_key);
+    pthread_setspecific(g_runner_key, ctx);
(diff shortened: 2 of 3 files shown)


hashiqiqixian:
@coderabbitai review

ChaoZheng109:
**Suggest adding an in-repo regression test.** This is a bugfix, but validation currently rests entirely on downstream pypto#2085 — simpler's own CI stays green because nothing in-repo passes `prewarm_config` into `simpler_init` (the scene-test framework always calls `worker.init()` without it). A future refactor of the TLS/prewarm path could silently reintroduce this null-runner segfault with simpler CI still green.

The repro is small (~20–40 lines, no golden/args boilerplate needed) since L2 trb prewarm runs synchronously inside `cw.init`:

- Build an L2 `Worker(runtime="tensormap_and_ringbuffer", ...)`, call `w.init(prewarm_config=CallConfig(<ring sizing>))`, assert it succeeds. Pre-fix this segfaults on the null-runner deref; post-fix it returns cleanly.
- Must be on the **trb** runtime — `host_build_graph` links the weak no-op `prewarm_config_impl` and won't crash, so it can't catch the regression there. A pure `a2a3sim` test already covers the sim half of this fix and needs no device lock.
- `prepared_callable/conftest.py` already shows the fresh-L2-`Worker` construction pattern to copy.

If it can't land in this PR, worth filing an issue to track it.

hashiqiqixian:
Added in `7f7082cf`. The new `tests/st/a2a3/tensormap_and_ringbuffer/test_prewarm_config.py` constructs a fresh L2 TRB worker on `a2a3sim`, sets `ring_task_window=64`, and calls `init(prewarm_config=config)`. This exercises the synchronous `simpler_init` prewarm path without device hardware or scene-test callable/golden boilerplate.

hashiqiqixian:
CI is green except [`st-onboard-a5`](https://github.com/hw-native-sys/simpler/actions/runs/29729334561/job/88311000787). That job failed during checkout before setup/tests because the self-hosted runner could not remove `.pytest_cache/v/cache` (`EACCES: permission denied`). All code-related jobs, including both a2a3sim scene-test jobs with the new regression test, passed. Could a maintainer please rerun the failed job? The PR author account does not have the repository admin permission required by GitHub to rerun it.

ChaoZheng109 (approved):

PR #1404, opened 2026-07-20
Fix: extend PTO-ISA version guard to a5 async workspace overlays

### LINKED ISSUE #1351: [Code Health] Extend PTO-ISA build/run version guard to a5 onboard SDMA overlay (missed after #1179)

yanghaoran29:
## Summary
- Generalize the PTO-ISA pin/metadata/staleness/ccache guard from a2a3-onboard-only to "host embeds pto-isa headers", via a new `platform_embeds_pto_isa` predicate: a2a3 onboard always; a5 onboard only when an async workspace overlay is ON — SDMA (`SIMPLER_ENABLE_PTO_SDMA_WORKSPACE`) **or** URMA (`SIMPLER_ENABLE_PTO_URMA_WORKSPACE`, added in #1392).
- Resolve the managed pin in `build_runtimes` / `_init_a5` when embedding, write a5 provenance metadata, and bake `SIMPLER_PTO_ISA_BUILD_COMMIT` into the a5 host ccache key whenever an overlay populates it.
- Extend `test_runtime_builder` coverage for a5 SDMA/URMA overlay ON (validate / metadata / stamp / cmake define / pin) and OFF (no-op); OFF cases now also clear the URMA env var.
- Docs: refresh the `docs/a5-sdma-overlay.md` "Gating points" table for the guard sites that activate when the overlay is ON, and note the a5 case in `getting-started.md` / `developer-guide.md`.

Rebased onto latest `main`; composes with the URMA overlay (#1392) so both a5 async overlays get the same stale-binary guard.

Fixes #1351

### DIFF
--- a/docs/a5-sdma-overlay.md
+++ b/docs/a5-sdma-overlay.md
@@ -52,11 +52,18 @@ Controlled by the build-time macro / env var `SIMPLER_ENABLE_PTO_SDMA_WORKSPACE`
 
 ### Gating points
 
-| File | What |
-| ---- | ---- |
-| `src/a5/platform/onboard/host/CMakeLists.txt` | `option(SIMPLER_ENABLE_PTO_SDMA_WORKSPACE ... OFF)`; `PTO_ISA_ROOT` check + pto-isa include guarded |
+Everything below keys on the a5 async workspace overlays. When SDMA
+(`SIMPLER_ENABLE_PTO_SDMA_WORKSPACE`) or URMA (`SIMPLER_ENABLE_PTO_URMA_WORKSPACE`)
+is ON the a5 onboard host `.so` embeds pto-isa headers, so the same pin /
+metadata / staleness / ccache guard that protects a2a3 onboard applies to a5
+too (#1351); when both are OFF none of it runs.
+
+| File / mechanism | What |
+| ---------------- | ---- |
+| `src/a5/platform/onboard/host/CMakeLists.txt` | `option(SIMPLER_ENABLE_PTO_SDMA_WORKSPACE ... OFF)`; under the option: `PTO_ISA_ROOT` check + pto-isa include. After the option blocks, the shared `SIMPLER_PTO_ISA_BUILD_COMMIT` cache-string → compile-define block bakes the pinned commit into the `host_runtime` ccache key (mirror of the a2a3 block); it is a no-op until an overlay populates the cache var |
 | `src/a5/platform/onboard/host/comm_hccl.cpp` | `ensure_sdma_workspace` body under `#ifdef SIMPLER_ENABLE_PTO_SDMA_WORKSPACE` |
-| `simpler_setup/runtime_compiler.py` | `_init_a5` `PTO_ISA_ROOT` ensure + `_sdma_workspace_enabled()` |
+| `simpler_setup/runtime_compiler.py` | `_init_a5` resolves the pinned managed pto-isa checkout via `ensure_pto_isa_root` (same contract as `_init_a2a3`) when the overlay is ON, then `env_manager.ensure("PTO_ISA_ROOT")` |
+| `simpler_setup/runtime_builder.py` (`platform_embeds_pto_isa`) | Single predicate every guard site keys on: a2a3 onboard always, a5 onboard only when the SDMA or URMA overlay is on (`_sdma_workspace_enabled() or _urma_workspace_enabled()`). Drives pin resolution (`_resolve_build_pto_isa_commit`), the `SIMPLER_PTO_ISA_BUILD_COMMIT` stamp, `pto_isa_build.json` metadata write, and load-time staleness validation (`_requires_pto_isa_metadata_validation`) |
 | `simpler_setup/runtime_builder.py` | `_compile_target` forwards the env var to the host CMake define |
 | `examples/a5/.../sdma_async_completion_demo/test_sdma_async_completion_demo.py` | `pytest.skip` when env var unset |
 
@@ -91,7 +98,7 @@ pip install -e .
 The env-var→CMake-define forwarding and the test skip gate pick it up
 automatically.
 
-**Option B — make SDMA the a5 default** (revert the 5 gating points above):
+**Option B — make SDMA the a5 default** (revert the gating points above):
 flip `option(... OFF)` → `set(... ON)`, make `_init_a5`'s `PTO_ISA_ROOT`
 ensure unconditional (mirror `_init_a2a3`), and remove the test skip gate.
 
--- a/docs/developer-guide.md
+++ b/docs/developer-guide.md
@@ -185,7 +185,7 @@ This builds the nanobind `_task_interface` extension **and** pre-builds all runt
 | Nanobind bindings (`python/bindings/`) | Re-run `pip install --no-build-isolation -e .` (no rebuild-on-import; `editable.rebuild = false`) |
 | Python-only code (`python/*.py`, `simpler_setup/*.py`) | No rebuild needed (editable install) |
 | Examples / kernels (`examples/{arch}/`, `tests/st/`) | No rebuild needed, just re-run |
-| `pto_isa.pin` changed | Re-run `pip install`. The cmake cache stamp and the `host_runtime` ccache key include the pinned PTO-ISA commit for a2a3 onboard, so a pin bump invalidates stale runtime objects automatically. |
+| `pto_isa.pin` changed | Re-run `pip install`. The cmake cache stamp and the `host_runtime` ccache key include the pinned PTO-ISA commit for a2a3 onboard (and a5 onboard when the SDMA overlay is enabled), so a pin bump invalidates stale runtime objects automatically. |
 
 A `pto_isa.pin` bump changes the SDMA headers embedded by
 `host_runtime.so`. Install-time runtime builds and run-time kernel compilation
--- a/docs/getting-started.md
+++ b/docs/getting-started.md
@@ -55,8 +55,9 @@ To use a different PTO-ISA revision, update `pto_isa.pin` to the desired
 diff and applies the same revision to install-time runtime builds and run-time
 kernel compilation.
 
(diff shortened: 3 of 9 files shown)


ChaoZheng109:
**Doc consistency: update `docs/a5-sdma-overlay.md` "Gating points" table**

The code changes look correct and the test coverage is solid (a5 overlay ON/OFF both exercised). One thing to fix before merge — the doc most directly tied to this change is now stale.

This PR extends the PTO-ISA pin / metadata / staleness / ccache guard to the a5 SDMA overlay, but the **"Gating points"** table in `docs/a5-sdma-overlay.md` (the canonical map of what the overlay gates) no longer matches the code:

- The `simpler_setup/runtime_compiler.py` row still reads `_init_a5` "`PTO_ISA_ROOT` ensure + `_sdma_workspace_enabled()`". Under this PR `_init_a5` now resolves the **pinned managed** checkout (`ensure_pto_isa_root`) — same contract as a2a3 — not just an `ensure` of a user-provided var.
- There is **no row** for the guard machinery this PR newly activates under the overlay:
  - `platform_embeds_pto_isa` predicate in `simpler_setup/runtime_builder.py` (now the single driver of every guard site)
  - pin/metadata write + load-time staleness validation for a5 onboard binaries
  - the new `SIMPLER_PTO_ISA_BUILD_COMMIT` cache-string / compile-define block in `src/a5/platform/onboard/host/CMakeLists.txt`

Please update the table in the same commit (per `.claude/rules/doc-consistency.md` §1/§4) so the overlay doc reflects the protections that now apply when the overlay is ON.

---

_Optional, non-blocking:_ `docs/getting-started.md:58` and `docs/developer-guide.md:188` both describe the metadata/ccache guard as "a2a3 onboard" only. A parenthetical "(and a5 onboard when the SDMA overlay is enabled)" would make them accurate, but they self-correct once a5 unifies to a2a3, so feel free to defer.

_Also non-blocking (transitional only):_ `_requires_pto_isa_metadata_validation()` recomputes `_sdma_workspace_enabled()` at load time, so a5 load-time staleness validation is skipped if the loading process doesn't set the env var. This is strictly better than the pre-PR state (no a5 validation at all) and disappears entirely once a5 forces the workspace ON like a2a3 — no action needed here.

yanghaoran29:
Thanks — all addressed in the pushed update:

- **`docs/a5-sdma-overlay.md` "Gating points"** — rewrote the table: fixed the `runtime_compiler.py` row (`_init_a5` now resolves the pinned managed checkout via `ensure_pto_isa_root`, same contract as `_init_a2a3`), and added rows for `platform_embeds_pto_isa`, the pin/metadata/staleness validation it drives, and the `SIMPLER_PTO_ISA_BUILD_COMMIT` cache-key block.
- **Optional parentheticals** — added "(and a5 onboard when the SDMA overlay is enabled)" to `docs/getting-started.md` and `docs/developer-guide.md`.

While rebasing onto latest `main` I also composed with the URMA overlay (#1392): `platform_embeds_pto_isa` now keys on **SDMA or URMA**, so both a5 async overlays get the same guard (previously URMA would have hit the identical stale-binary gap). Table and UT updated to match.

ChaoZheng109 (approved):

PR #1406, opened 2026-07-20
Add: opt-in async-DMA SDMA workspace via Worker enable_sdma

lyfne123:
## Summary

Provision the PTO-ISA async-SDMA workspace **only** when a Worker is created with `enable_sdma=True`, replacing the earlier demand-driven design.

- `Worker(level=2, …, enable_sdma=True)` → `ChipWorker.init(dma_workspace_mask)` → `simpler_provision_dma_workspace` → `dma_workspace_provision`. Provisions the 48 STARS streams **once at init**, latches the addresses into `InitArgs`, and re-latches `simpler_aicpu_init` so `g_dma_workspace_addr` feeds every core's `GlobalContext` (`get_dma_workspace`).
- A Worker **without** the flag creates no SDMA streams; its kernels read a zero workspace address. `a5`/sim/host_build_graph have no SDMA provider (`dma_workspace_supported_mask()==0`), so `enable_sdma` there fails fast — the shared plumbing still compiles for every arch.

## Why (issue #1425)

The 48 device-only STARS streams sit in the device fault/sync domain, so a fault on an SDMA-provisioned device pays a multi-minute CANN teardown. Making SDMA an explicit **per-Worker opt-in** confines that risk to Workers that ask for it; ordinary Workers keep main's fast (~0.3s) recovery and ordinary `507046` error code.

## What changed vs the demand-driven design

- Replace the comm run-gate (`dma_workspace_run_begin/run_end/reset_begin/reset_end/abandon` + `DmaWorkspaceState`) with `dma_workspace_provision` / `dma_workspace_release`.
- Add `simpler_provision_dma_workspace` to the runtime C ABI.
- Drop the callable `required_dma_workspaces` declaration and the `requires_device_generation_recycle` pool signal; revert arch `finalize()` and `chip_worker` to their upstream teardown.

## CI isolation

Both SDMA demos run in a dedicated `task-submit` task, excluded from the general a2a3 sweep — so their teardown never shares a device with the `aicore_op_timeout` fault-injection test.

## Testing

- [x] Clean rebuild of all 4 platforms (a2a3sim/a5sim/a2a3/a5)
- [x] `tools/verify_packaging.sh` (clean wheel build + packaging) passes
- [x] 60 cpp UT, 726 non-hardware Python UT
- [x] Hardware (a2a3): prefetch demo provisions 48 STARS streams at init, injected workspace reaches the kernel, output bit-exact

Mitigates #1425. Faults after genuine SDMA use still require a CANN runtime/driver teardown fix.


lyfne123:
Added a single-card ST so the new accessor is covered inside this repo, not only via the PyPTO consumer.

**`examples/a2a3/tensormap_and_ringbuffer/prefetch_async_demo/`** — kernel + orchestration + test:

- The host provisions the workspace and threads its device address into the kernel **as a scalar**, the same way `sdma_async_completion_demo` threads its `CommContext` pointer. That avoids needing a `child_memory` tensor path: the workspace is a runtime-owned device buffer, not user data to stage H2D.
- The kernel prefetches `in` into L2, waits on the returned event, then copies `in` to `out` through a tile. Since the prefetch changes no value, bit-exact `out == in` is the property under test — plus the wait actually completing rather than hanging.
- Needs no comm domain (unlike the SDMA completion demo), because the accessor is comm-independent. One device is enough.
- Mirrors the pto-isa single-card ST at `tests/npu/a2a3/src/st/testcase/tprefetch_async`.
- When the address is 0 the kernel skips the prefetch and still runs the copy, so the demo stays green — with reduced coverage — where SDMA is unavailable.

Verified on a2a3 hardware that it exercises the real SDMA path, not the skip branch:

```
[SDMA] Created 48 STARS streams OK
[SDMA] STARS query AICPU kernel completed OK
[prefetch_async_demo] sdma workspace addr = 0x12c0c001a000
[INFO] prefetch_async_demo: out matches src after prefetch + copy
1 passed
```

a5 is intentionally not covered: its SDMA overlay is gated off by default (`SIMPLER_ENABLE_PTO_SDMA_WORKSPACE`, see `docs/a5-sdma-overlay.md`), so the demo would only ever take the skip branch there.

ChaoWao (approved):

PR #1409, opened 2026-07-20
perf(bench): quiet CANN driver log on multi-round timing runs

noabauma:
**NOTE: This is not an optimization PR! This only fixes the problem of reducing noise when benchmarking!**

## Summary

Multi-round timing runs (`--rounds N`, N>1) inherit the CANN driver's default
INFO log level (`ASCEND_GLOBAL_LOG_LEVEL` unset), which writes hundreds of
device-side `[INFO]` lines per run into `~/ascend/log/debug`. Those device_log
writes land inside the measured device wall — **~4–6%** on `paged_attention`
(n=8, a2a3), recurring every round (~178 lines/round + ~870-line one-time init).
simpler's own AICPU logging contributes **0** lines (codestyle rule 7).

Default `ASCEND_GLOBAL_LOG_LEVEL=3` (ERROR) when `--rounds > 1`, so timing runs
aren't taxed by driver log spam while real failures still surface; an
explicitly-set value is respected. Also documents the level knob in
`running-onboard.md` (which previously covered only the log *path*).

Verified onboard (a2a3, device 2): `--rounds 8` drops the driver log 1092 → 1
line; `--rounds 1` unchanged; explicit `ASCEND_GLOBAL_LOG_LEVEL` respected.

## Benchmark

`tools/benchmark_rounds.sh -d 3 -n 100`, a2a3 onboard, **same device, sequential**
(baseline first, then this branch). Baseline = `main` (`321b88ee`); current = this
branch. Metric: `[STRACE]` device-domain spans via `strace_timing --rounds-table`.
Negative Δ = faster on this branch.

Driver log confirmed quieted on this branch (`--rounds 100`, same example):

| build | driver-log lines / run |
| ----- | ---------------------- |
| baseline (`main`) | 12,257 |
| this branch | **1** |

| Example | Device Δ | Effective Δ | Orch Δ |
| ------- | -------- | ----------- | ------ |
| `batch_paged_attention` (Case1) | **−4.56%** | **−4.94%** | **−6.82%** |
| `alternating_matmul_add` (Case1) | **−2.95%** | **−3.04%** | −3.04% |
| `benchmark_bgemm` (Case0) | −0.55% | −0.42% | −0.46% |
| `paged_attention_unroll` (Case1) | −0.03% | +0.04% | +1.02% |
| `paged_attention_unroll` (Case2) | −0.47% | −0.37% | −1.26% |
| `paged_attention_unroll_manual_scope` (Case1) | −0.44% | −0.38% | +0.26% |
| `paged_attention_unroll_manual_scope` (Case2) | −0.22% | −0.07% | +0.30% |

Improvement is example-dependent — largest on dispatch-heavy workloads
(`batch_paged_attention`: −4.9% Effective) and within noise on the light
`paged_attention_unroll` cases — because the CANN driver-log tax scales with
per-round device-log volume. **0 regressions > 2%.** (Host-wall reductions were
larger, up to −16%, but Host is noisy so Device/Effective are the headline.)

### DIFF
--- a/.claude/rules/running-onboard.md
+++ b/.claude/rules/running-onboard.md
@@ -156,6 +156,28 @@ task-submit --device auto --device-num 1 --run "python ... -d \$TASK_DEVICE"
 Then read it directly (`$LOGDIR/device-*/device-*.log`) — no more `grep`-ing
 `~/ascend/log/debug/` and guessing which file is yours.
 
+## Device log *level*: quiet the CANN driver log for timing runs
+
+`ASCEND_PROCESS_LOG_PATH` (above) sets *where* the log goes;
+**`ASCEND_GLOBAL_LOG_LEVEL`** sets *how much*. Unset, it defaults to **INFO**,
+so the CANN driver/runtime writes hundreds of `[INFO]` lines per run into
+`device-<id>/*.log` (device-side `aicpu-sd`) and `plog/*.log` (host `RUNTIME`).
+The device-side writes land inside the measured device wall: on
+`paged_attention` they cost **~4–6%** (measured n=8, a2a3) and recur every
+`--rounds` iteration (~178 lines/round on top of a ~870-line one-time init).
+
+For any timing/benchmark run, set it to **ERROR** so only real failures log:
+
+```bash
+export ASCEND_GLOBAL_LOG_LEVEL=3   # 0=DEBUG 1=INFO(default) 2=WARNING 3=ERROR 4=NULL
+```
+
+It is a **severity floor** — higher = quieter — and is a *different* system from
+simpler's `--log-level` V0..V9 verbosity (independent knobs). Use `3` (ERROR),
+not `4` (NULL), so `507018`-class errors still surface. Scene tests set this
+automatically when `--rounds > 1` (a timing run, standalone or pytest) unless you
+pin it.
+
 ## `507018` triage: classify the mechanism before concluding
 
 `507018` (and `507014` / `507899`) is a **generic host-side code** —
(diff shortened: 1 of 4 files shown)


ChaoZheng109:
Thanks for digging into this — you've surfaced a real issue, and the root of it is worth recording.

The verbose device log on timing runs isn't the CANN driver's default: `simpler_init` (`c_api_shared.cpp`) calls `dlog_setlevel(-1, HostLogger::level(), 0)` whenever `ASCEND_GLOBAL_LOG_LEVEL` is unset, and since `HostLogger::level()` defaults to INFO and `-1` covers every CANN module, we end up opening the driver's own INFO stream ourselves — that's the cost landing inside the measured device wall. (The `[STRACE]` benchmark numbers themselves are unaffected either way: they ride host stderr + the `rtMemcpy` device-phase readback, not the CANN log.)

We're going to address this at the root, as part of a broader logging cleanup rather than a per-run env var, in phases:

1. Slim down the current logging and reduce its footprint.
2. Remove the `V0..V9` sub-tiers and move to a single `debug / info / timing / warn / error` ladder with `timing` as the default — STRACE emitted at `timing`, and the device level mapped so `timing` stays at WARN. Default runs then carry no driver-log tax on host or device, WARN/ERROR stay visible, and device INFO remains reachable via `info` / `debug`.

This resolves the device-side timing overhead you found as a side effect, and keeps a single `--log-level` driving both host and device. We'll open a follow-up issue to track the log-system rework.

Thanks again for pinpointing where the cost was coming from.

ChaoZheng109:
Following the discussion above, we're going to fix this at the root as part of a broader logging refactor rather than the per-run env var. That work is tracked in:

- #1428 — reduce redundant / hot-path logging footprint (phase 1)
- #1429 — re-tier log levels (`debug/info/timing/warn/error`, default `timing`) and stop `simpler_init` from seeding CANN dlog to INFO (phase 2)

Since it will land as a new PR series, I'm closing this one for now. Thanks again for surfacing the issue and the measurements — they directly informed the plan, and the device-side timing overhead will be resolved as part of #1429.

PR #1430, opened 2026-07-22
Add: reject kind4 device pointers dispatched or freed on the wrong worker

ChaoWao:
## Scope

Device-address guard ② (`worker-memory-model.md` §9). A child (kind4) device
pointer is a bare `int` today — no owner, no generation (`tensor.h`
`child_memory` is one bit). Once it enters a `TaskArgs` it can be freed,
copied, or run on a worker that never allocated it; a raw device VA is not
globally unique, so worker B could silently touch worker A's (or illegal)
device memory. Guard ① (#1397) only stopped host-addr rewrite — nothing checked
the owning worker.

This tracks live child pointers by their exact `(worker_id, ptr)` provenance
and validates every kind4 **consumer** — the device memory ops and dispatch
(L2 `Worker.run` is not a consumer). **Python-only; no wire / C++ / ABI change.**

## Delivered

- **Typed provenance table** on the `Worker`: `(worker_id, ptr) -> {malloc_owned,
  domain_allocation_ids}`. Each op is atomic under one private lock. Ordering is
  **safety-first**: malloc records **after** the native alloc; `free` and domain
  release revoke **before** the native free.
- **malloc / free / copy_to / copy_from**: `free` validates an exact live malloc
  base, **revokes provenance, then** does the native free (a commit barrier, L2
  and L3 identical). A native-free failure is a **terminal leak** — provenance is
  **not** restored and an explicit retry is rejected; an async unwind after a
  successful free can never leave a freed address live. `copy_*` require the
  device-side pointer live on that worker. Covers both `Worker.malloc` and the
  direct `orch.malloc/copy/free` path (single `Orchestrator` choke).
- **submit_next_level / submit_next_level_group**: a `child_memory` argument must
  resolve to exactly one eligible target worker (resolved eligibility, not raw
  `-1`) and be live there; `0` or `>= 2` candidates is rejected. The resolved
  owner is passed to C++ as the **effective affinity** so the child `TensorKey`
  is keyed by its owner, not `-1`. The group path materialises a full per-member
  affinity only when `workers` is empty/None; a non-empty list is length-checked
  (never padded), and a group that resolves two members to the **same** worker is
  rejected (a group must dispatch to distinct workers).
- **CommDomain** window / buffer pointers enter provenance on `allocate_domain`
  and are revoked in `_release_domain_now` **before** the backend free (commit
  barrier): the deferred window stays dispatchable, but once release begins the
  pointers are undispatchable and a partial backend-release failure leaves them
  dropped rather than "live forever".

## Boundary (explicit)

- Catches **stale-before-reuse**; does **not** prevent strict **ABA** — needs P1
  generation handles.
- Raw-C++ deep-bypass (`worker._orch._o.malloc`) out of scope (needs C++).
- Remote path (`RemoteBufferHandle`) already owner+generation guarded — exempt.

## Testing

- [x] **40 device-free UT** (`test_child_addr_guard.py`): provenance table,
  target resolution, orchestrator memory ops incl. **free commit-barrier +
  native-error terminal-leak**, single/group dispatch + resolved affinity +
  **group length/duplicate-worker validation**, CommDomain dispatch and
  **release-before-free** commit barrier, L2 path. Fix-specific tests verified
  failing against pre-change source.
- [x] Broad UT sweep (`test_worker/` + task/callable): **498 passed / 2 skipped**
  — no regression. ruff + pyright clean.
- [x] Static non-regression on every kind4/domain example (all pin `worker=`).
- [ ] Simulation / onboard via CI (`st-sim-*`, `st-onboard-*`).

### DIFF
--- a/examples/workers/l3/child_memory/README.md
+++ b/examples/workers/l3/child_memory/README.md
@@ -16,7 +16,10 @@ pattern in miniature.
 
 There is no `orch.free` — the buffer is reclaimed by `DeviceRunner::finalize`
 when `worker.close()` runs. That asymmetry is the whole point of
-`child_memory`: you opt out of the runtime's bookkeeping.
+`child_memory`: you opt out of the runtime's **automatic lifecycle** (no
+auto-malloc, no auto-free). The pointer is still tracked for **ownership**: its
+`(worker_id, ptr)` provenance is validated on every `copy_*` and dispatch, so it
+cannot be freed, copied, or run on a worker that did not allocate it.
 
 ## Run
 
--- a/python/simpler/orchestrator.py
+++ b/python/simpler/orchestrator.py
@@ -194,11 +194,33 @@ class Orchestrator:
             self._worker._stage_host_buffers_for_chip_submit(c_args)
         final_worker_ids = _remote_data_eligible_worker_ids(remote_sidecar, eligible_worker_ids)
         cpp_worker_id = int(worker)
-        captured_refs = self._worker._capture_remote_sidecar_refs(remote_sidecar) if self._worker is not None else []
+        worker = self._worker
+        # Do the (fallible) kind4 provenance analysis BEFORE capturing remote slot
+        # refs, so an exception here can never leave captured refs neither
+        # released nor adopted (which would defer a remote free forever). Capture
+        # is the last step before the rollback try.
+        child_ptrs = worker._child_ptrs_in_args(c_args) if worker is not None else []
+        prov_guard: Any = contextlib.nullcontext()
+        candidates: set[int] = set()
+        if child_ptrs and worker is not None:
+            candidates = self._child_dispatch_candidates(cpp_worker_id, final_worker_ids)
+            prov_guard = worker._child_prov_lock
+        captured_refs = worker._capture_remote_sidecar_refs(remote_sidecar) if worker is not None else []
         try:
-            self._o.submit_next_level(
-                digest, kind, target_namespace, c_args, cfg, cpp_worker_id, final_worker_ids, remote_sidecar
-            )
+            with prov_guard:
+                if child_ptrs and worker is not None:
+                    worker._child_prov_check_dispatch(child_ptrs, candidates, api="submit_next_level")
+                    # The child_memory arg resolved to a unique owner; pass that
+                    # worker as the effective affinity so C++ keys the child
+                    # TensorKey by its owner (worker_id), not the raw -1.
+                    # Otherwise the same buffer submitted once as -1 and once as W
+                    # yields two keys (ptr,-1) / (ptr,W) and its dependency is
+                    # missed. The unique target is exactly what the scheduler
+                    # would have picked, so pinning it changes no scheduling.
+                    cpp_worker_id = next(iter(candidates))
+                self._o.submit_next_level(
+                    digest, kind, target_namespace, c_args, cfg, cpp_worker_id, final_worker_ids, remote_sidecar
+                )
         except BaseException:
             if self._worker is not None:
                 self._worker._release_remote_slot_refs(captured_refs)
@@ -206,7 +228,23 @@ class Orchestrator:
         if self._worker is not None:
             self._worker._adopt_remote_slot_refs(captured_refs)
 
-    def submit_next_level_group(
+    def _child_dispatch_candidates(self, cpp_worker_id: int, eligible_ids: Any) -> set[int]:
+        """Resolve the set of eligible target workers for a kind4 dispatch.
+
+        A pinned ``worker`` is the sole candidate; an unpinned ``-1`` falls back
+        to the callable's eligible set, or the full next-level pool when the
+        callable is unconstrained. ``_child_prov_check_dispatch`` rejects any
+        result that is not a single unique target.
+        """
+        if cpp_worker_id >= 0:
+            return {cpp_worker_id}
+        if eligible_ids:
+            return {int(w) for w in eligible_ids}
+        if self._worker is None:
+            return set()
+        return set(self._worker._next_level_target_ids())
+
+    def submit_next_level_group(  # noqa: PLR0912 -- linear per-member sidecar + eligibility + kind4-provenance passes, one branch each
         self,
         callable_handle: Any,
         args_list: list,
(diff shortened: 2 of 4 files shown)


ChaoWao:
Thanks — addressed all four findings (force-pushed).

**1 [high] `worker=-1` resolved owner not passed to C++ → wrong `TensorKey`.** Fixed. After the child_memory arg narrows to a unique target, that worker is now passed as the effective affinity to `_o.submit_next_level` (`cpp_worker_id = next(iter(candidates))`), and the group path materialises a full per-member affinity vector, pinning each child member to its resolved owner (`[-1, W, …]`). So the child `TensorKey` is keyed `(ptr, W)`, not `(ptr, -1)` — the missed-dep and same-VA-collision cases are closed. New tests assert the affinity actually handed to C++ (`call_args.args[5]`), not just that submit was called.

**2 [high] CommDomain release had no provenance commit barrier.** Fixed. The revoke moved to the **start** of `_release_domain_now`, **before** the backend free, under the provenance lock. Through the deferred-release window (logical release → post-drain physical teardown) the pointers stay live for in-flight tasks; once physical release begins they are undispatchable. A partial/failed backend release now leaves them **dropped** (a recoverable leak) instead of stranded-live (use-after-free). New tests cover both the drop-before-dispatch ordering and the backend-failure-still-dropped path against the real `_release_domain_now`.

**3 [medium] transaction-failure tests.** Added: native `malloc` raises → nothing recorded (orch + L2); native `free` raises → provenance **kept** for retry (contract now explicit + tested); CommDomain partial-release failure; and the deterministic revoke-before-free interleaving.

**4 [low] doc scope.** PR body now says "every kind4 **consumer** (L2 `Worker.run` is not a consumer)". `examples/workers/l3/child_memory/README.md` now distinguishes "no automatic lifecycle (no auto-malloc/free)" from "still ownership-tracked: `(worker_id, ptr)` provenance validated on every copy/dispatch".

Verification: **36 device-free UT** (all 4 fix-specific tests verified failing against the pre-fix source), 494/2-skip broad sweep, ruff + pyright clean, no wire/C++/ABI change.

ChaoWao:
Round 2 addressed (force-pushed).

**1 [high] free() not transactional → UAF on async unwind.** Fixed with a safety-first commit barrier: `free` now validates the exact malloc base, **revokes provenance, then** calls the native free — under the lock, both L2 (`Worker.free`) and L3 (`Orchestrator.free`). An async unwind (KeyboardInterrupt after the binding returns) can no longer leave a freed address live. A native-free failure is now a **terminal leak**, never a re-authorized maybe-freed address; the earlier "keep for retry" contract is gone. New tests: `free` revokes before the native call (spy asserts already-revoked at native time, L2 + L3), and native-error → revoked + retry rejected.

**2 [medium] group swallowed illegal `workers` length.** Fixed: the full per-member affinity is materialised **only** when `workers` is empty/None. A non-empty list is length-checked (`len(workers) == len(args_list)`) and copied — never padded — so `2 args + workers=[0]` now raises instead of becoming `[0, -1]` and bypassing the C++ length check. New test covers the rejection.

**Should-fix (both done):**
- The positive group affinity test is now a **single-member** group (schedulable) — no more `{1}`-eligibility-for-two-distinct-workers case.
- Stale "revoke after backend free" comment on `allocate_domain`, the `_child_prov_drop_domain` docstring, and `.docs/tasks/g2-design.md` §2 all now say "revoke **before** the backend free (commit barrier)".

Verification: **39 device-free UT** — I reverted both round-2 hunks and confirmed the 4 fix-specific tests fail with the right assertions, then restored. 497/2-skip broad sweep, ruff + pyright clean, no wire/C++/ABI change.

ChaoWao:
Round 3 addressed (force-pushed `49367442`).

**1 [blocker] group could pin two members to the same worker.** Fixed. After the per-member affinity write-back, a group with two members resolving to the same owner (a duplicate non-negative affinity — e.g. two child args on the same chip) is now **rejected** ("a group must dispatch to distinct workers") rather than emitting `[W, W]` and serializing on one WorkerThread. New regression: two child members owned by the same worker → reject.

**2 [blocker] free contract vs docs mismatch.** Fixed everywhere to the implemented safety-first semantics (**validate → revoke → native free**; native-free failure = terminal leak, no retry):
- PR body (also corrected counts to **40 UT / 498 broad**).
- `worker.py` `_child_alloc_prov` init comment, the provenance-helpers block comment, and `_child_prov_clear_malloc` docstring.
- `tests/…/test_child_addr_guard.py` section header.
- `.docs/tasks/g2-design.md` §2 (free ordering + linearization).
- (`task-device-addr-guard2.md` is the original task prompt / spec — left as historical; happy to annotate if you'd prefer.)

**Non-blocking:** `_next_level_target_ids()` L4 fallback now returns the stable `_next_level_worker_ids` instead of `range(len(_next_level_workers))`.

Verification: **40 device-free UT** — the group-duplicate fix verified failing (DID NOT RAISE) against reverted source; 498/2-skip broad sweep; ruff + pyright clean.

**CI:** agreed on your read — last round's `st-onboard-a5` (bidirectional-ring allreduce) passed on `c5d61e20` and my diff doesn't touch comm, so it's a hardware flake; `ut-a5` died at runner "Set up job" (SSL/action-download infra). I'm watching this run and will rerun those if they recur, and only conclude "flake" once a rerun is green.

ChaoWao:
Round 4 addressed (force-pushed `8e959593`).

**1 [high] empty entry could re-authorize a freed pointer.** Fixed, fail-closed:
- New entries set their role **before** insertion; dropping the **last** role deletes the key directly — a role-less entry now never exists.
- `_ChildProvEntry.is_live()` (`malloc_owned or domain_allocation_ids`); `_child_prov_require_live` and the dispatch guard both test `is_live()`, not key presence. So even if an entry were momentarily left empty by an interrupted revoke, no `copy_*`/dispatch passes.
- Regression: an injected role-less entry is rejected by require_live / require_malloc_base / dispatch; and dropping the last role deletes outright.

**2 [medium] provenance analysis outside the remote-ref rollback window.** Fixed: the (fallible) kind4 analysis now runs **before** `_capture_remote_sidecar_refs` — capture is the last step before the `try` — in both the single and group paths. An analysis failure can no longer strand captured refs (deferring a remote free forever). Regression asserts capture is not called when the analysis raises.

**Should-fix / track:**
- **Deterministic contention tests** added: `free` holds `_child_prov_lock` across the native free (mutual exclusion), and a domain-release-vs-dispatch check confirms a dispatch landing during the native release is rejected (already revoked).
- **distinct-workers scope narrowed** in comment + design doc: the check rejects duplicate **pinned** owners (the hazard the affinity write-back creates); full injective feasibility (a wildcard member colliding with a pinned one) is the scheduler's pre-existing capacity concern, explicitly not claimed here.
- Doc numbers synced to **45 UT / 503 broad**; `_next_level_target_ids()` L4 fallback → stable `_next_level_worker_ids`; and `task-device-addr-guard2.md` §1 free-order updated to the implemented revoke-before-native / terminal-leak / fail-closed semantics.

Verification: **45 device-free UT** — both blocker fixes verified failing against reverted source (empty-entry DID-NOT-RAISE; capture called-1-time); 503/2-skip broad sweep; ruff + pyright clean; no wire/C++/ABI change.

PR #1446, opened 2026-07-22
Enable -cce-vf-aa-between-iters by default for A5 CCEC builds

yanghaoran29:
## Summary
- Add `-mllvm -cce-vf-aa-between-iters=true` to the default CCEC flags for A5 hardware.
- This option enables alias analysis across loop iterations. With it on, AA is stricter and directly affects BiSheng VF fusion scale and correctness.
- Context: without this option, BiSheng fusion can misbehave in some cases; fusion still needs to run under the stricter AA setting. BiSheng plans to make this always-on later; simpler enables it by default on A5 CCEC now to align with that direction.
- a5sim is unchanged: `KernelCompiler` leaves `ccec` unset and uses `Gxx15Toolchain`, so this LLVM option does not apply there.
- Board check: DeepSeek V4 `qkv_proj_rope.py` (decode, `-p a5`) fails with VF fusion alone (AICore / scheduler stall); adding this option makes it PASS. Turning fusion off also PASSes.

## Test plan
- [ ] Confirm `CCECToolchain(platform="a5").get_compile_flags()` includes `-mllvm` / `-cce-vf-aa-between-iters=true`
- [ ] Confirm `CCECToolchain(platform="a2a3")` and a5sim kernel path do **not** add this option
- [ ] A5 board: `qkv_proj_rope.py -p a5 --mode decode` PASSes

### DIFF
--- a/simpler_setup/toolchain.py
+++ b/simpler_setup/toolchain.py
@@ -167,7 +167,7 @@ class CCECToolchain(Toolchain):
         else:
             raise ValueError(f"Unknown platform: {self.platform}. Supported: a2a3, a2a3sim, a5, a5sim")
 
-        return [
+        flags = [
             "-c",
             "-O3",
             "-g",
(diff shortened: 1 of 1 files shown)


yanghaoran29:
Addressed CodeRabbit feedback: narrowed the flag to A5 hardware only (`platform == "a5"`). a5sim uses Gxx15Toolchain and does not go through CCECToolchain.

ChaoZheng109 (approved):

PR #1452, opened 2026-07-23
host_build_graph: all AICPU threads schedule (drop orch/sched split)

noabauma:
## What

`host_build_graph` builds the task graph on the **host**, so no AICPU thread orchestrates on-device at runtime. Despite that, the runtime reserved thread `N-1` as a no-op "orchestrator": it owned **zero** AICore cores and never dispatched, leaving only `aicpu_thread_num - 1` scheduling threads (**3 of 4** in every scene test).

This PR removes the vestigial orch/sched split (a `tensormap_and_ringbuffer` / device-orch concept) so **all `aicpu_thread_num` threads schedule**, matching what `a5` host_build_graph already does:

- Delete `orch_to_sched_`, `runtime.orch_to_sched`, and the `PTO2_ORCH_TO_SCHED` env read — meaningless in HBG (no device orchestration).
- Delete the `sched_thread_num_` split; distribute cores over all threads (`active_sched_threads_ = aicpu_thread_num_`). The drain barrier, completion counting, and cluster-id math already key off `active_sched_threads_`, so they scale automatically.
- The last thread still performs the one-time host-orch boot preamble, then joins scheduling like every other thread.

Net effect at `aicpu_thread_num=4`: **4 scheduling threads instead of 3.** NUMA locality is unchanged — `compute_allowed_cpus` still pins all threads to a single AICPU cluster (cpu_ids `{4,5,6,7}`) because `active_count` stays 4. Do **not** raise `aicpu_thread_num` past a single cluster's core count, or threads spill across the NUMA boundary.

## Hardware validation (a2a3, 910B2)

`bgemm` onboard passes with **4 scheduler summaries — one per thread, including thread 3**, which now schedules real work instead of sitting idle:

| Thread | tasks_scheduled | total_time |
| ------ | --------------- | ---------- |
| 0 | 36 | 88.40 µs |
| 1 | 34 | 88.38 µs |
| 2 | 31 | 88.26 µs |
| 3 | **27** (was 0) | 87.82 µs |

## Benchmark 1: 3 vs 4 scheduling threads

Same build, `aicpu_thread_num` toggled at runtime (**3 → cpus `{4,5,6}`, 4 → cpus `{4,5,6,7}`** — same NUMA cluster, isolating only the thread count). Metric = per-thread device-side scheduler `total_time`, 15 rounds; lower is faster.

| Example | 3 threads (mean / min) | 4 threads (mean / min) | Δ mean |
| ------- | ---------------------- | ---------------------- | ------ |
| **bgemm** (block_dim 24) | 81.8 / 74.1 µs | 65.1 / 57.9 µs | **−20%** |
| matmul | 22.5 / 16.4 µs | 22.1 / 15.2 µs | −2% |
| paged_attention | 22.4 / 17.1 µs | 23.1 / 17.8 µs | +3% |
| vector_example | 21.1 / 13.9 µs | 21.0 / 13.8 µs | ~0% |

Dispatch-bound compute (**bgemm, ~20% faster**) is the clear win: 4 threads shorten each thread's serial MMIO poll round (~95 ns/core), feeding cores sooner. `matmul` gains modestly; `vector_example` is neutral (only ~3 clusters, so the 4th thread has no cores); `paged_attention` at scene-test scale (~a dozen tasks) is fixed-overhead-dominated, so ±3% is within noise — its production-shape win is documented in the PyPTO2 report. Consistent with that report directionally.

## Benchmark 2: 3 vs 4 scheduling threads

Same build, `aicpu_thread_num` toggled at runtime (3 → cpus `{4,5,6}`, 4 → `{4,5,6,7}` — same NUMA cluster). Metric = device-wall (`[STRACE] device_wall` via `strace_timing`), lower is faster. Measured across `paged_attention` (Case1 params) scaled by `batch` to span the ms range (ring/heap bumps applied equally to both configs):

| Workload (device @ 3 thr) | batch | 3 threads | 4 threads | Δ |
| ------------------------- | ----- | --------- | --------- | - |
| paged_attention (~2.6 ms) | 16  | 2.65 ms  | 1.99 ms  | **−25%** |
| paged_attention (~5 ms)   | 32  | 5.23 ms  | 3.77 ms  | **−28%** |
| paged_attention (~10 ms)  | 64  | 10.39 ms | 7.43 ms  | **−28%** |
| paged_attention (~21 ms)  | 128 | 21.21 ms | 15.03 ms | **−29%** |
| paged_attention (~43 ms)  | 256 | 43.36 ms | 30.23 ms | **−30%** |

n=8/point, σ<0.5% at scale. The 4th scheduler thread delivers a **consistent ~29% device-time reduction across the whole ms range**, growing with dispatch pressure (task count). Sub-millisecond scene tests (`bgemm` ~0.3 ms, `matmul`/`vector` ~0.1–0.2 ms) are overhead-bound — too few tasks for a 4th scheduler to help — so they're flat-to-slightly-slower, which is immaterial for production shapes. Corroborates the PyPTO2 report.

### DIFF
--- a/src/a2a3/runtime/host_build_graph/aicpu/aicpu_executor.cpp
+++ b/src/a2a3/runtime/host_build_graph/aicpu/aicpu_executor.cpp
@@ -79,9 +79,6 @@ static int32_t read_pto2_runtime_status(Runtime *runtime) {
 static PTO2Runtime *rt{nullptr};
 
 struct AicpuExecutor {
-    int32_t sched_thread_num_;
-    bool orch_to_sched_{false};
-
     // ===== Thread management state =====
     std::atomic<int32_t> thread_idx_{0};
     std::atomic<bool> init_done_{false};
@@ -166,13 +163,11 @@ int32_t AicpuExecutor::init(Runtime *runtime) {
 
     if (is_leader) {
         LOG_INFO_V0("AicpuExecutor: Initializing");
-        // The 0 → 1 fixup already applied above; derive scheduler count from it.
+        // The 0 → 1 fixup already applied above.
         aicpu_thread_num_ = nthreads;
-        sched_thread_num_ = nthreads - 1;
-        orch_to_sched_ = runtime->orch_to_sched;
 
         hs_arrived_.store(0, std::memory_order_relaxed);
-        if (sched_ctx_.pre_handshake_init(runtime, aicpu_thread_num_, sched_thread_num_, get_platform_regs()) != 0) {
+        if (sched_ctx_.pre_handshake_init(runtime, aicpu_thread_num_, get_platform_regs()) != 0) {
             init_failed_.store(true, std::memory_order_release);
             hs_setup_done_.store(true, std::memory_order_release);
             return -1;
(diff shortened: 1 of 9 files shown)


ChaoZheng109:
The core logic looks correct: `active_sched_threads_` now drives core assignment, the drain/ack barrier (`all_acked = (1 << active_sched_threads_) - 1`), and the sched-phase pool count, so they all scale to N threads automatically. Every drain spin still bails on `is_completed()`, so a thread that short-circuits dispatch (all tasks inline-completed) can't hang the ack barrier. The bgemm onboard table (thread 3 scheduling 27 tasks, 4 scheduler summaries) confirms the boot thread now schedules end-to-end. Removing the `PTO2_ORCH_TO_SCHED` gate is also in line with the repo's env-gating convention. 👍

One thing to fix in the same PR: a few **docs/comments** still describe the old "3 sched + 1 orch" split that this change falsifies (per the `doc-consistency` rule):

1. **`src/a2a3/runtime/host_build_graph/runtime/runtime.h:158-163`** (right above the `orch_to_sched` field this PR deletes) still says AICPU splits into "one orchestrator thread ... and the remaining `aicpu_thread_num-1` scheduler threads ... The orch thread also dispatches when env `PTO2_ORCH_TO_SCHED` is set" — it references the env var this PR removes.
2. **`docs/RUNTIME_LOGIC.md` §8 core-assignment table** — the `Thread 3 | Orchestrator | none` row and "divided equally among the **3** scheduler threads" now contradict the new behavior (thread 3 owns cores and schedules).
3. **`docs/device_log_profiling.md:89`** — "Each of the **3** scheduler threads (Thread 0, 1, 2) prints its own summary" contradicts the **4**-summary output shown in this PR's own validation table.

Updating these three spots is all that's left.

ChaoZheng109 (approved):

PR #1454, opened 2026-07-24
Fix: initialize dep-gen before orchestrator submission

Little-oil:
## Summary
- initialize the dep-gen buffer during leader `pre_handshake_init`, before the decoupled orchestrator can submit task 0
- keep PMU and dump setup post-handshake so the barrier-free scheduler/orchestrator overlap remains intact
- run the a2a3 and a5 dep-gen smoke with full-device worker counts to cover the initialization window

## Root cause
The barrier-free scheduler initialization lets the orchestrator build the graph while scheduler threads are still handshaking. Dep-gen initialization was left in the post-handshake profiling path, so early `dep_gen_aicpu_record_submit` calls saw a null collector state and returned without recording the leading tasks or their dependencies.

## Testing
- [x] pre-commit hooks on all changed files
- [x] a2a3sim dep-gen smoke (`1 passed`)
- [x] a5sim dep-gen smoke (`1 passed`)
- [x] a2a3sim L2 swimlane suite with dep-gen (`2 passed`)
- [x] a5sim L2 swimlane suite with dep-gen (`2 passed`)
- [x] original a2a3 DeepSeek V4 Flash MoE reproduction succeeds with this runtime worktree (requester confirmed)

### DIFF
--- a/tests/ut/py/test_worker/test_orphan_child_reaping.py
+++ b/tests/ut/py/test_worker/test_orphan_child_reaping.py
@@ -40,6 +40,10 @@ NUM_SUB_WORKERS = 3
 # enumerating children externally also catches the transient shell that runs
 # pgrep, and inside a container's shallow pid namespace it picks up unrelated
 # low pids — which the test would then wait on and, worse, SIGKILL.
+#
+# The pids go to a dedicated file named on the command line, not to stdout:
+# the parent's stdout and stderr carry log lines and warnings, and scraping
+# integers out of that stream picks up any bare number they happen to contain.
 _PARENT_SRC = textwrap.dedent(
     """
     import os, sys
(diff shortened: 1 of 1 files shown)


ChaoWao:
Drive-by data point: this race also fires **onboard** on a2a3, not just in the sim smokes.

`st-onboard-a2a3` on an unrelated PR (#1482, a Python-only `RunHandle` change) failed its `dep_gen smoke` step with exactly the signature this PR's root cause predicts — the *leading* task missing from the sidecar:

```
tests/st/a2a3/tensormap_and_ringbuffer/dfx/dep_gen/test_dep_gen.py::TestDepGen::test_run
AssertionError: tasks[] missing expected ids: {0}
  assert {0, 4294967296, ...} <= {4294967296, ...}
```

Run: https://github.com/hw-native-sys/simpler/actions/runs/30184309146/job/89746344476

Confirmed intermittent: a plain re-run of the same job on the same commit passed, and the same test passed 3/3 locally on a2a3 silicon under CI-equivalent flags (`--device-num 4 -p no:xdist --require-pto-isa --enable-dep-gen`). Consistent with the initialization window being timing-dependent rather than deterministic.

No action needed here — just noting that the onboard job is an additional (flaky) surface for the same bug, so this fix should quiet it too.

PR #1459, opened 2026-07-24
Refactor: isolate hierarchical worker run state

Crane-Liu:
## Summary

- add per-run identity, lifecycle state, completion counts, and first-error isolation to the hierarchical worker
- namespace TensorMap entries by run and serialize construction/consumption access
- remove completion dependence on the global drain/error slot while keeping `Worker.run()` blocking and returning `None`
- compact Ring storage only when the worker is globally quiescent

## Series scope

This is PR 1 of the worker async-preparation series. It intentionally does not expose `RunHandle`, admit overlapping runs, or change device ordering. Those behaviors remain in later independently reviewable PRs.

## Review follow-ups applied

- **Keep run bookkeeping non-throwing on the scheduler and worker threads.** `on_task_failed_cb` is invoked from `Scheduler::worker_done`, which runs on a `WorkerThread` whose `on_complete_` call sits outside its try/catch, and `on_consumed_cb` runs inside `Scheduler::run()`, which has no try/catch. Both reach run accounting, so an unknown run id there would terminate the process. `find_run()` now resolves without throwing, and a count underflow is recorded as the run's error instead of thrown.
- **Carry the orchestration failure message onto the run fence.** `_fail_run_submission` takes an optional message, so a submission failure is retrievable from the run and not only from the raising caller — required once `RunHandle` lands in PR 2.
- **Close the allocator-compaction race.** `begin_run()` takes only `runs_mu_`, so testing quiescence under the scheduler loop mutex alone still allowed: test quiescent → release `runs_mu_` → another thread `begin_run()` + allocate its first slot → `reset_to_empty()` invalidates it. `compact_if_quiescent()` now holds `runs_mu_` across **both** the quiescence test and `reset_to_empty()`. This is sufficient because both allocator call sites (`alloc()`, `reserve_outputs_and_slot()`) are gated behind `current_building_run()`, so no slot can be allocated while no run is building. Lock order is scheduler loop mutex then `runs_mu_`, matching the scheduler's `on_consumed` path.
- **Close the run inside the block that releases it**, so a failure in `close`/`fail_run_submission` cannot strand the run id or skip run-owned resource cleanup.
- Document that no transition assigns `RunPhase::PREPARED` while one run builds at a time.

Added regression tests: error reporting and consumption for an already-released run are ignored rather than fatal, and a failed submission carries its message to the fence. The compaction race is fixed structurally (the test and the reset are now under one mutex); a deterministic unit test would need an injection hook to stop a thread mid-compaction, and a timing-based one would be flaky, so none was added.

## Testing

- pre-commit: all hooks passed
- C++ unit tests: 60/60 passed (`ctest --test-dir tests/ut/cpp/build -LE requires_hardware`)
- Python unit tests: 804 passed, 2 skipped (`pytest tests/ut`)

Tested commit: `7b7b206d`.

The architecture-probe and hardware runs quoted in earlier revisions were against `4d7eb574`, which is no longer this branch's head, and have not been re-run. GitHub simulation CI remains the automated regression stage.

### DIFF
--- a/docs/comm-domain.md
+++ b/docs/comm-domain.md
@@ -58,26 +58,26 @@ The handle is a context manager. Its lifecycle has **two distinct states**:
   exits). Further indexing (`handle[i]`) raises. This is the *user-visible*
   state: "do not hand this domain to any new `submit_*`."
 - **`freed`** — the backend `comm_release_domain_windows` has actually run and
-  the device memory is gone. This happens **after** `Worker.run` drains the
-  DAG, never inside the `with` block.
+  the device memory is gone. This happens **after** the owning run's completion
+  fence, never inside the `with` block.
 
 This split exists because `submit_next_level()` only *enqueues* DAG work;
-`Worker.run()` does not drain until the orch function returns. If `release()`
-freed memory immediately on `with`-exit, a still-queued task that captured the
-domain's `device_ctx` / `buffer_ptrs` would read freed memory. So **release is
-deferred**: `release()` flips `released` and queues the backend free; the real
-free runs after drain, when every task that could reference the window has
-completed.
+`Worker.run()` does not wait for completion until the orch function returns.
+If `release()` freed memory immediately on `with`-exit, a still-queued task that
+captured the domain's `device_ctx` / `buffer_ptrs` would read freed memory. So
+**release is deferred**: `release()` flips `released` and queues the backend
+free; the real free runs after the run fence, when every task that could
+reference the window has completed.
 
 Mental model: like `with open(f) as fh: ...` — the user-visible close is
 lexical (end of block), the physical teardown is managed for you. Use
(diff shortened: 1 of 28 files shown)


Crane-Liu:
CI triage update: `ut-a5` and `st-onboard-a5` both failed before checkout or test execution. In each job, `actions/checkout@v5` could not clean the runner workspace:

```text
EACCES: permission denied, rmdir .../simpler/.pytest_cache/v/cache
```

Affected jobs:
- https://github.com/hw-native-sys/simpler/actions/runs/30082144278/job/89446583483
- https://github.com/hw-native-sys/simpler/actions/runs/30082144278/job/89446583568

The fork token cannot re-run upstream jobs (`403 Must have admin rights to Repository`). Please repair/clean the a5 runner workspace ownership and re-run these two jobs. No PR code ran in either failure.

ChaoWao (approved):

PR #1460, opened 2026-07-24
Add: Worker.submit with per-run completion handles

Crane-Liu:
## Summary

- add `Worker.submit(...) -> RunHandle`; graph construction stays synchronous, but L3 device work may still be active when `submit()` returns
- keep `Worker.run(...)` source-compatible as `submit(...).wait()`, still blocking and still returning `None`
- `RunHandle` exposes `done`, `wait(timeout)` and `result(timeout)`; the terminal result is cached so repeated and concurrent waits observe the same outcome
- exactly one waiter performs fence-owned cleanup and `_release_run`; the others replay the cached result
- bind `wait_run_for`, a timed native fence wait that releases the GIL
- retain the callback, args and config on the handle until the fence fires, so run-owned resources outlive `submit()`
- `close()` fences admission and drains accepted handles before tearing down the worker tree
- L2 submission stays blocking and returns an already-completed handle

Graph-construction errors remain synchronous in `submit()`. Device and endpoint
errors are owned by the originating handle, surface from `wait()`/`result()`,
and do not poison a later submission.

## Scope

Only one live device run is admitted: a later submission drains the previous
handle before building another DAG. This deliberately does **not** add
prepared-run FIFO admission, a Python dispatcher thread, cancellation, or any
device overlap — preparation of run N+1 still does not overlap execution of
run N.

## Series context

Part of the worker async-preparation series; follows #1459, which introduced
per-run identity, the completion fence and per-run error isolation. Prepared-run
admission, K=2 resources and device overlap come later.

## Rebase

#1459 has landed, so this branch is rebased onto merged `main` and the
superseded run-fence commit is dropped. The diff is now this PR's own increment
only: 9 files, +582/-76.

The `worker.py` conflict against the final #1459 implementation was resolved by
keeping this PR's `_submit_l3_locked` restructure (submission no longer waits;
cleanup moved into `RunHandle`) and re-applying #1459's failure-message
plumbing, so a graph-construction failure still reaches the run fence:
`_fail_run_submission(run_id, _format_exc("orchestration", e))`.

## Validation

Re-run after the rebase onto merged `main`:

- changed-file pre-commit hooks: passed
- C++ unit tests: 60/60 passed (`ctest --test-dir tests/ut/cpp/build -LE requires_hardware`)
- Python unit tests: 821 passed, 2 skipped (`pytest tests/ut`)

The architecture-probe and hardware runs quoted before the rebase were against
`8e56fa6a` and have not been re-run against the current head. GitHub CI is the
automated regression stage.

### DIFF
--- a/docs/orchestrator.md
+++ b/docs/orchestrator.md
@@ -51,11 +51,12 @@ public:
     // --- Intermediate-buffer allocation (runtime-owned lifetime) ---
     Tensor alloc(const std::vector<uint32_t> &shape, DataType dtype);
 
-    // --- Internal lifecycle (invoked by Worker::run only) ---
+    // --- Internal lifecycle (invoked by Python Worker.submit/RunHandle) ---
     RunId begin_run();
     void close_run_submission(RunId run_id);
     void fail_run_submission(RunId run_id, std::exception_ptr error);
     void wait_run(RunId run_id);
+    bool wait_run_for(RunId run_id, double timeout_seconds);
     bool run_done(RunId run_id) const;
     void release_run(RunId run_id);
     void scope_begin();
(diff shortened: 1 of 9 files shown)


Crane-Liu:
`st-onboard-a2a3` needs a maintainer rerun. The failure is isolated to the
standalone `dep_gen smoke`: `tasks[]` omitted task id 0 while all expected
dependency edges were present. The same `test_dep_gen.py` passed earlier in
this exact CI job as part of the full a2a3 scene suite.

I reran the smoke command on queued a2a3 hardware against this PR's exact
commit. It passed:

- task: `task_20260724_032702_351647729229`
- result: 1 passed, exit 0
- command scope: `test_dep_gen.py --platform a2a3 -p no:xdist --require-pto-isa --enable-dep-gen`

The job rerun REST API returns 403 for the fork author's token, so I cannot
rerun it directly. No product-code change was made for this non-reproducible
capture failure.

ChaoWao (dismissed):

ChaoWao (approved):

Issue #1197, opened 2026-06-29
[Code Health] Teardown ordering: release RTS resources before aclFinalize (load_aicpu_op_ stopgap is ad-hoc)


ChaoZheng109:
### Category

Robustness (potential edge-case failure)

### Component

Platform (a2a3 / a2a3sim)

### Description

The onboard `DeviceRunner` teardown has a structural **ordering hazard**: an RTS
interface gets called *after* the RTS/ACL context is torn down.

- `aclFinalize` (which tears down the RTS/ACL context) runs inside the **derived**
  `DeviceRunner::finalize()` (a2a3/a5 `device_runner.cpp`); the error path's
  `force_reset_device()` also does `aclrtResetDeviceForce` + `aclFinalize`.
- But RTS-holding objects like `load_aicpu_op_` (`host::LoadAicpuOp`) are
  **base-class members** of `DeviceRunnerBase`. By C++ destruction rules, a base
  member's destructor runs **after** the derived `finalize()` completes — i.e.
  **after** `aclFinalize`.
- `~LoadAicpuOp()` calls `rtsBinaryUnload(binary_handle_)` on an already-torn-down
  RTS handle. **a5's driver segfaults** at process exit; a2a3's driver tolerates
  the call (the UB happened to no-op), which is why it stayed hidden.
- Only manifests on the path that calls `aclFinalize` — the error / force-reset
  path that the device-error negative STs exercise. The common rt-path only does
  `rtDeviceReset` (no `aclFinalize`), so the late unload is benign there.

**Current state (stopgap shipped with #1180):** `finalize_common()` now calls
`load_aicpu_op_.Finalize()` explicitly (before `aclFinalize`, while RTS is live),
which unloads and nulls `binary_handle_`, so the later `~LoadAicpuOp` `Finalize()`
no-ops. This is a **point-fix for one member** and does not fix the underlying
invariant: explicit-Finalize + idempotent-destructor is awkward, and any other
base-class member whose destructor touches an RTS/ACL/`rt*` API after `finalize()`
would hit the same crash.

### Location

- `src/common/platform/onboard/host/device_runner_base.cpp` — `finalize_common()`, the `load_aicpu_op_.Finalize()` stopgap
- `src/common/platform/onboard/host/device_runner_base.h` — `load_aicpu_op_` is a base member (`~line 766`)
[...]

PR #1464, opened 2026-07-24
Refactor: isolate A2A3 run stream sets per pipeline slot

Crane-Liu:
Depends on #1463 (merged).

## Summary

- give the A2/A3 device runner an AICPU/AICore stream set per pipeline slot and submit every run on the selected set
- keep the persistent pair for bootstrap and control work — the `Init` and `RegisterCallable` payloads and the block-dim query
- size the slot array from `PTO_PIPELINE_MAX_DEPTH`; K=1 selects slot 0 only
- create a slot's set on first use and reuse it; `destroy_run_stream_sets()` in `finalize()` is the sole release point
- split `sync_run_streams()` so the wait can name a pair; A5 keeps the no-argument form and the persistent pair

## Why a set per slot, not per run

The pipeline contract classifies both streams as `EXEC_HANDLE`, which asks for
one instance *per in-flight run* — that is one set per slot, i.e.
`pipeline_depth` sets, not one set per `run()` invocation. A stream set carries
no per-run content, so rebuilding it every run buys nothing.

It also costs a great deal. `rtStreamCreate`/`rtStreamDestroy` are ~300 us each
on this silicon, and both sit on the synchronous host path around
`KernelLaunch`. Measured with `vector_example` at 40 rounds, three alternating
repetitions of each binary under a single device lock (only
`libhost_runtime.so` swapped between repetitions), host time per round as the
p50 of the per-repetition p50 with warmup rounds dropped:

| | host p50 | vs `main` |
| --- | --- | --- |
| `main` | 10316.8 us | — |
| per-run create/destroy | 12607.6 us | +2290.8 us (+22.2%) |
| this change | 10481.1 us | +164.3 us (+1.6%) |

`main`'s own three repetitions span 2.7%, so the remaining +1.6% is inside the
run-to-run spread. Device time is unchanged (~250 us) in all three.

Reusing the set also removes the teardown questions the per-run shape raised: a
stream destroy can no longer fail an otherwise successful run, and there is no
window in which a run's streams are destroyed while a launch error left an
AICore kernel spinning on them.

## Teardown

`run_stream_sets_` is this subclass's own RTS-owning member, so
`destroy_run_stream_sets()` runs in `finalize()` while RTS is live and before
the device reset — the window `finalize_common()` uses for the bootstrap pair,
per the teardown invariant that file documents (#1197). It is idempotent, so
the destructor's `finalize()` call is a safe backstop. As there, no pre-destroy
sync: `rtStreamDestroy` is the supported teardown for a stream an op-timeout
left in the error state.

## Observability and the reuse test

Reuse is the invariant this PR turns on, and nothing outside `DeviceRunner`
could see it: streams are opaque handles, the `run stream set created` line sits
below the default host log threshold, and a leak needs ~990 worker lifetimes to
exhaust a device's 1981 streams. So the runner now reports how many sets it has
built, through the same C API → `ChipWorker` → nanobind path the
`aicpu_dlopen_count` / `host_dlopen_count` counters already use for exactly this
kind of assertion. It is diagnostic only and gates nothing; simulation and A5
report 0.

`tests/st/a2a3/host_build_graph/run_stream_reuse` runs one worker four times and
requires exactly one set. Reinstating the per-run create/destroy makes it report
four, so the regression this PR fixes can no longer land green.

## Scope

K=1 throughout. `TASK_ACCEPTED`, asynchronous flight, Gate A/B, K=2 and arena
banking are later changes. `run()` still calls `launch_run` and `reap_run`
back-to-back.

## Validation

Rebased onto merged `main`; the four superseded ancestor commits are dropped,
so this is a single commit of 13 files, +307/-15.

- changed-file pre-commit hooks: passed
- C++ unit tests: 61/61 passed
- Python unit tests: 824 passed, 2 skipped
- `a2a3sim`, full `examples tests/st`: 108 passed, 0 failed
- `a2a3` onboard, the exact CI command (`examples tests/st` minus the two SDMA
  demos, 4 devices under `task-submit`): 115 passed, 1 skipped, 0 failed
- the new reuse test was mutation-checked: with the per-run create/destroy put
  back it fails with `expected 1 run stream set for 4 runs, got 4`
- `a2a3` onboard SDMA step (`prefetch_async_demo`,
  `sdma_async_completion_demo`, 2 devices): 2 passed
- reuse across runs is exercised by the multi-run `prepared_callable` cases in
  the onboard suite; the code is onboard-only, so `st-onboard-a2a3` is its only
  CI coverage

### DIFF
--- a/python/bindings/task_interface.cpp
+++ b/python/bindings/task_interface.cpp
@@ -1465,6 +1465,14 @@ NB_MODULE(_task_interface, m) {
             "host_build_graph variants. Mirrors aicpu_dlopen_count for the "
             "host-orchestration path; 0 on device-orch variants."
         )
+        .def_prop_ro(
+            "run_stream_set_create_count", &ChipWorker::run_stream_set_create_count,
+            "Number of run stream sets the bound runner has created. A set "
+            "belongs to a pipeline slot and is reused for every run on that "
+            "slot, so a worker that has served any number of runs reports 1; "
+            "platforms whose runs use the persistent bootstrap pair report 0. "
+            "Tests assert this to verify repeated runs do not rebuild the set."
+        )
         .def("malloc", &ChipWorker::malloc, nb::arg("size"))
         .def("free", &ChipWorker::free, nb::arg("ptr"))
         .def("copy_to", &ChipWorker::copy_to, nb::arg("dst"), nb::arg("src"), nb::arg("size"))
--- a/python/simpler/task_interface.py
+++ b/python/simpler/task_interface.py
@@ -1196,6 +1196,11 @@ class ChipWorker:
         """Number of host-side orch SO dlopens (host_build_graph variants)."""
         return self._impl.host_dlopen_count
 
+    @property
+    def run_stream_set_create_count(self):
+        """Number of run stream sets the bound runner has created."""
+        return self._impl.run_stream_set_create_count
+
     def malloc(self, size):
         """Allocate memory. Returns a pointer (uint64)."""
         return int(self._impl.malloc(int(size)))
--- a/python/simpler/worker.py
+++ b/python/simpler/worker.py
@@ -5955,6 +5955,19 @@ class Worker:
             return 0
         return self._chip_worker.host_dlopen_count
 
+    @property
+    def run_stream_set_create_count(self) -> int:
+        """L2 only: number of run stream sets the bound runner has created.
+
+        A set belongs to a pipeline slot and is reused for every run on that
+        slot, so a worker that has served any number of runs reports 1.
+        Returns 0 on non-L2 workers and on platforms whose runs use the
+        persistent bootstrap stream pair (simulation, a5).
+        """
+        if self.level != 2 or self._chip_worker is None:
+            return 0
+        return self._chip_worker.run_stream_set_create_count
+
     # ------------------------------------------------------------------
     # close
     # ------------------------------------------------------------------
--- a/src/a2a3/platform/onboard/host/device_runner.cpp
+++ b/src/a2a3/platform/onboard/host/device_runner.cpp
@@ -192,6 +192,7 @@ int DeviceRunner::destroy_comm_stream(void *stream) {
 // `src/common/platform/onboard/host/device_runner_base.cpp`.
 
 int DeviceRunner::run(Runtime &runtime, const CallConfig &config) {
+    constexpr unsigned kPipelineSlot = 0;
     // Latch this run's diagnostic enables onto the runner before the collector
     // paths below read them; block_dim/aicpu_thread_num are consumed locally.
     apply_call_config(config);
@@ -221,6 +222,12 @@ int DeviceRunner::run(Runtime &runtime, const CallConfig &config) {
     int rc = ensure_device_initialized();
     if (rc != 0) {
         LOG_ERROR("ensure_device_initialized failed: %d", rc);
+        return rc;
+    }
+
+    rc = ensure_run_stream_set(kPipelineSlot);
+    if (rc != 0) {
+        LOG_ERROR("ensure_run_stream_set(%u) failed: %d", kPipelineSlot, rc);
         return rc;
     }
 
@@ -432,10 +439,10 @@ int DeviceRunner::run(Runtime &runtime, const CallConfig &config) {
         l2_swimlane_collector_.set_core_types(core_types.data(), num_aicore);
     }
 
-    rc = launch_run(runtime, num_aicore, launch_aicpu_num);
+    rc = launch_run(runtime, num_aicore, launch_aicpu_num, kPipelineSlot);
     if (rc != 0) return rc;
 
-    rc = reap_run();
+    rc = reap_run(kPipelineSlot);
     if (rc != 0) return rc;
 
     // Print handshake results (reads from device memory, must be before free)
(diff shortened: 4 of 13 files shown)


ChaoWao:
## Revision: stream sets are per slot, not per run

Pushed `e56d50bb`, rebased onto merged `main` (the four superseded ancestor commits drop out, leaving 3 files / +109/-14). Review found one blocking issue and three that fall out of it.

### Blocking: per-run create/destroy cost +22% host time

`create_run_stream_set` in `run()` and `destroy_run_stream_set` in `reap_run` are both synchronous host calls on the critical path, and `rtStreamCreate`/`rtStreamDestroy` measure ~300 us each on this silicon.

Measured on a2a3, `vector_example` at 40 rounds, three alternating repetitions of each binary inside one device lock with only `libhost_runtime.so` swapped, host p50 per round (p50 of per-repetition p50, warmup dropped):

| | host p50 | vs `main` |
| --- | --- | --- |
| `main` | 10316.8 us | — |
| previous revision | 12607.6 us | +2290.8 us (+22.2%) |
| this revision | 10481.1 us | +164.3 us (+1.6%) |

`main`'s own repetitions span 2.7%, so +1.6% is inside the noise. Device time is ~250 us in all three. An intermediate control build — the previous revision with only the create made idempotent and the destroys removed — also landed at +0.7%, which pins the whole regression on create/destroy rather than on anything else in the diff.

The fix is to keep the slot shape and drop the per-run lifetime: a set is created on first use and reused. That is what `EXEC_HANDLE` in #1463 already asks for — one instance *per in-flight run*, i.e. per slot, not per `run()` invocation. A set carries no per-run content, so there was never a reason to rebuild it.

### Falls out of the same change

- **`run_stream_sets_` had no release point.** It is an RTS-owning member released neither by `finalize_common()` nor by a destructor, against the invariant documented at `device_runner_base.cpp:979` (#1197). Now `destroy_run_stream_sets()` runs in `finalize()` while RTS is live, before the device reset, and is idempotent so `~DeviceRunner() { finalize(); }` is a safe backstop.
- **A teardown failure could fail a successful run.** `reap_run` returned the destroy rc and `run()` returned on it before `print_handshake_results()`, so a benign `rtStreamDestroy` failure aborted a run that had already computed and reconciled. `reap_run` returns 0 again.
- **Duplicate cleanup in `reap_run`.** The RAII guard plus the explicit destroy plus the dismiss are all gone with the per-run lifetime.

There was also a narrower hazard: on the `launch_aicpu_kernel` failure path the AICore kernel is already launched and spinning, and the guard destroyed its stream right after a bounded drain that may itself have timed out. Reuse removes that window.

### Other

- `kRunStreamSetCount` is now `PTO_PIPELINE_MAX_DEPTH` instead of a literal `2`. `src/common/worker` is already on this target's include path.
- The `DEVICE_SCRATCH`-adjacent comment on the base's persistent pair was reworded — they are per-slot sets, not per-run.

### Validation

61/61 C++, 824 passed / 2 skipped Python, `a2a3sim` full suite 54 passed, and the exact onboard CI command on 4 devices: **113 passed, 1 skipped, 0 failed**, plus the SDMA step 2 passed. Reuse across runs is exercised by the multi-run `prepared_callable` cases.

Coverage note: all three files are onboard-only, so `st-onboard-a2a3` is the only CI job that can reach them — sim cannot.

ChaoWao:
## Revision: a regression barrier for the reuse invariant, and a question about slot 1

Pushed `47c70da7`.

### The reuse invariant now has a test

I said in the last revision that the create-once invariant had no observable and therefore no test. That was true but not a good place to stop — the +22% regression this PR fixes would have landed green again. Streams are opaque handles, the `run stream set %u created` line sits below the default host log threshold, and a leak needs ~990 worker lifetimes to exhaust a device's 1981 streams (measured), so none of the passive routes work.

So the runner reports how many sets it has built, through the same C API → `ChipWorker` → nanobind path that `aicpu_dlopen_count` / `host_dlopen_count` already use — whose binding docstring says verbatim "Tests assert this to verify register_callable + repeated run do not redundantly dlopen." Same purpose, same shape. Diagnostic only, gates nothing; sim and A5 report 0.

`tests/st/a2a3/host_build_graph/run_stream_reuse` runs one worker four times and requires exactly one set. Mutation-checked by reinstating the per-run create/destroy:

```
E  AssertionError: expected 1 run stream set for 4 runs, got 4 — the set is being rebuilt per run
```

### Slot 1 is unreachable through the whole series

Worth raising before this merges. I checked the `codex/worker-async-pr6-run-resources` and `codex/worker-async-pr7-ack-flight` branches: both still have `constexpr unsigned kPipelineSlot = 0`. So through #1467 — the last PR in the announced split — `ensure_run_stream_set(1)` is never called, slot 1's pair is never created, launched on, or synced, and the four-run-stream configuration never exists.

That is consistent with what #1467 actually enables: `TASK_ACCEPTED` lets a later `submit()` **build** while the prior run executes on the device, and K=1 device ownership is held until `TASK_DONE`. The overlap is host-prepare(N+1) ∥ device-execute(N), not stream ∥ stream. Two runs' kernels are never in flight together, so one pair would still suffice.

The second slot only starts paying when a run's `KernelLaunch` may be issued before the prior run is reaped — a stream is FIFO and a sync waits on everything enqueued, so at that point run N's reap must not be able to block on run N+1's kernels. No PR in the series does that.

So `kRunStreamSetCount = PTO_PIPELINE_MAX_DEPTH` ships an array whose second entry no test can reach and no announced change will select. I have left it as-is because you argued for fixing the ownership shape once, and it is 32 bytes — but the alternative, a single pair here and the array in the PR that first selects slot 1, would carry no untestable structure. Your call.

One thing the slot split does **not** buy, in case it is load-bearing anywhere downstream: separate streams give ordering and independent completion, not AICore exclusivity. Two runs on two slots would still contend for the same cores; serializing that is Gate B's job, not the stream layer's.

### Validation

61/61 C++, 824 passed / 2 skipped Python, `a2a3sim` full suite 108 passed / 0 failed, and the exact onboard CI command on 4 devices: **115 passed, 1 skipped, 0 failed**.

PR #1475, opened 2026-07-25
Fix: replace verbose log tiers with severity levels

### LINKED ISSUE #1429: [Code Health] Re-tier log levels: replace V0..V9 with debug/info/timing/warn/error; fix CANN dlog INFO seeding (a2a3 + a5)
### Category

### Component

Other — logging subsystem (host `HostLogger` + AICPU `device_log` + `--log-level` surface), spanning **both a2a3 and a5**.

### Description

Two coupled problems with the current logging design:

**1. simpler self-seeds CANN's driver INFO stream (device performance tax).**
When `ASCEND_GLOBAL_LOG_LEVEL` is unset, `simpler_init` calls `dlog_setlevel(-1, HostLogger::level(), 0)`. `HostLogger::level()` defaults to INFO and `-1` targets **every** CANN module, so we open the CANN driver's own INFO logs as a side effect of wanting our own host verbosity. Those device-side writes land inside the measured device wall (~4–6% on some workloads; see #1409). It is not a CANN default — it is self-inflicted.

**2. The `V0..V9` INFO sub-tier scheme is a two-axis (severity, info_v) design that is more complex than what the code uses.** In practice device logging is binary (V0/V9 only — see #1428), and the v-tiers force a `--log-level` surface (`v0..v9`) wired through `conftest.py`, `_log.py::_split_threshold`, `log_config.py`, and the docs.

Target: a single-axis ladder **`debug / info / timing / warn / error`**, default **`timing`**, replacing `V0..V9`. STRACE moves from `LOG_INFO_V9` to the `timing` level; `simpler_init` maps the device (CANN) level so `timing → WARN`. Result:

- default `timing` → host emits only STRACE (+ warn/error); device level = WARN → **no driver-log tax on any run** (not just `rounds > 1`);
- `info` / `debug` → device INFO back on for debugging, without the DEBUG flood;
- WARN/ERROR always visible; one `--log-level` still drives both host and device.

Benchmark measurement is unaffected either way — `[STRACE]` rides host stderr + the `rtMemcpy` device-phase readback, not the CANN log.

This supersedes the per-run `ASCEND_GLOBAL_LOG_LEVEL` env-var mitigation in #1409, and depends on the footprint reduction in #1428 landing first.

### Location

- Device seeding: `src/common/platform/onboard/host/c_api_shared.cpp` (`simpler_init` → `dlog_setlevel(-1, HostLogger::level(), 0)`)
- STRACE emission: `src/common/log/include/common/strace.h` (`LOG_INFO_V9` → `timing`)
- Two-axis machinery: `python/simpler/_log.py` (`_split_threshold`), `simpler_setup/log_config.py` (`LOG_LEVEL_CHOICES`), `conftest.py` (V0..V9 registration), `src/common/log/host_log.cpp` (`info_v`), `src/common/platform/onboard/aicpu/device_log.cpp` (`g_log_info_v`), `src/common/log/include/common/unified_log.h` (`LOG_INFO_V0..V9`)
- Call-site migration, both arches: `src/a2a3/**` and `src/a5/**` `LOG_INFO_V*` sites
- Docs: `docs/logging.md`, `docs/testing.md`, `docs/troubleshooting/device-error-codes.md`, `.claude/rules/running-onboard.md`

### Proposed Fix

Phased (after #1428):

1. **Introduce the ladder** `debug / info / timing / warn / error` (single severity axis, `timing` between `info` and `warn`); make `timing` the default. Emit STRACE at `timing`.
2. **Wire the device mapping** in `simpler_init`: `debug→DEBUG, info→INFO, timing→WARN, warn→WARN, error→ERROR` — so default `timing` no longer seeds CANN INFO.
3. **Remove `V0..V9`** (or keep as deprecated aliases: `v9→timing`, `v5–v8→info`, `v0–v4→debug`) and migrate remaining call sites + the `--log-level` surface + docs.

Open decisions to settle in the issue: hard-remove vs deprecated aliases for `v0..v9`; whether the 109 must-see device diagnostics map to `info` or `debug`.

### Priority

Medium (real device-side overhead, but #1409 provides a stopgap; depends on #1428)

---

Context / root-cause discussion: hw-native-sys/simpler#1409. Depends on: #1428. Related: #409 (hot-path logging overhead).

zmnobug:
## Summary

- replace the V0-V9 verbosity tiers with `DEBUG`, `INFO`, `TIMING`, `WARN`, and `ERROR`, defaulting to `TIMING`
- carry one threshold consistently through Python, host logging, simulation AICPU libraries, and onboard CANN configuration
- move STRACE markers to `TIMING`, migrate affected call sites and documentation, and retain the log-footprint reduction from #1447
- add regressions for CANN threshold mapping and simulation AICPU threshold propagation

## Testing

- editable package build
- 78 focused Python logging/worker tests
- 60 C++ unit tests
- a2a3sim STRACE and AICPU debug propagation tests
- a5sim AICPU debug propagation test
- a2a3 onboard STRACE/device-wall test
- full Python, a2a3sim, and a5sim suites before the final rebase

## Performance

On a2a3 `batch_paged_attention` Case1 over 100 rounds:

- device time: 4561.0 us -> 4349.7 us (-4.63%)
- effective time: 4536.7 us -> 4324.9 us (-4.67%)
- device INFO lines: 6299 -> 0

Fixes #1429


ChaoZheng109:
**Stale reference to a removed symbol — `tests/st/a5/tensormap_and_ringbuffer/bench_prewarm_timing.py:69`**

```python
# line 25
from simpler import _log
# line 69
logging.getLogger("simpler").setLevel(_log.V9)
```

This PR deletes `V0..V9` from `python/simpler/_log.py` (only `TIMING` / `NUL` remain), but `bench_prewarm_timing.py` still references `_log.V9`. Running the script now raises `AttributeError: module 'simpler._log' has no attribute 'V9'`.

The file isn't part of this PR's diff, so a line-anchored inline comment can't be attached here — flagging it at the conversation level instead. It's also why CI stays green: the filename isn't `test_*`, pytest doesn't collect it, and nothing else invokes it — so it's dead-on-arrival only if run manually. A repo-wide grep for `V0..V9` / `_log.V` references (per `.claude/rules/doc-consistency.md` §1) would have caught it.

Suggested fix: `_log.V9` -> `_log.TIMING` (V9 was the must-see tier STRACE used; `TIMING` is its successor and keeps the benchmark's "maximize timing visibility" intent).

src/common/platform/onboard/aicpu/device_log.cpp
```diff
@@ -30,14 +28,14 @@
...
+bool g_is_log_enable_timing = false;
 bool g_is_log_enable_warn = false;
 bool g_is_log_enable_error = false;
 
-int g_log_info_v = 5;
-
 void init_log_switch() {
     g_is_log_enable_debug = CheckLogLevel(AICPU, DLOG_DEBUG);
     g_is_log_enable_info = CheckLogLevel(AICPU, DLOG_INFO);
+    g_is_log_enable_timing = CheckLogLevel(AICPU, DLOG_WARN);
```
ChaoZheng109: **host/device log-level should stay aligned: onboard can't suppress TIMING at `--log-level warn`.**

Gating TIMING on `CheckLogLevel(AICPU, DLOG_WARN)` makes `g_is_log_enable_timing` identical to `g_is_log_enable_warn`, so onboard can never separate the two. At `--log-level warn` (threshold 30):

- host: `is_enabled(TIMING)` -> 25 >= 30 is false -> TIMING suppressed ✓
- sim device: this PR forwards the threshold into `set_log_level`, which gates `g_is_log_enable_timing = level <= 25` independently -> TIMING suppressed ✓
- **onboard device: TIMING still emits** (shares the `DLOG_WARN` flag) ✗

So after this PR sim and onboard disagree on the timing/warn boundary, and neither matches the host only for onboard. There are no device `LOG_TIMING` call sites today, but the level semantics should still line up across host and both device backends so a future device-side TIMING log behaves consistently.

Cheap fix, no new plumbing — the threshold already reaches the device: `kernel.cpp` calls `set_log_level(init_args->log_level)`. Instead of making onboard's `set_log_level` a no-op, store the threshold in a simpler-owned global and gate `is_log_enable_timing()` on `threshold <= TIMING(25)` (mirroring sim, and the old `g_log_info_v` simpler-managed axis). Keep DEBUG/INFO/WARN/ERROR on CANN as-is; only TIMING needs the simpler-side cut, since CANN has no level between INFO and WARN.

zmnobug: Fixed — onboard now keeps DEBUG/INFO/WARN/ERROR under CANN while deriving TIMING from the host-provided threshold. Added a no-hardware regression covering WARN suppressing TIMING and TIMING keeping it enabled.

zmnobug:
@ChaoZheng109 Addressed the stale benchmark reference: `bench_prewarm_timing.py` now uses `_log.TIMING`. The repo-wide audit also covers `_log.V*`; it is clean after migrating the three `LOG_INFO_V0` calls introduced on the latest main.

ChaoZheng109:
## Verification: functional (per-level) + performance (#1409 benefit)

Tested on real **a2a3** silicon (`Ascend910_93`, via `task-submit`) against a **clean rebuild** of this PR's head. Both parts pass.

> Note on setup: my first build silently reused a stale cache (whole tree kept the old 2-arg `simpler_log_init`), which surfaced as `simpler_log_init failed with code -1` at runtime. A clean `build/` wipe + reinstall fixed it — the PR source is correct. Flagging in case others hit the same stale-cache symptom.

### 1. Functional — every `--log-level` behaves

- **ut-py** `test_log_config.py`: 15 passed; name→threshold mapping correct (`debug=10 / info=20 / timing=25 / warn=30 / error=40 / null=60`).
- **ut-cpp** `test_host_log_off` + `test_onboard_device_log_level`: pass (host gating, CANN mapping + external-override, and the onboard `warn`-suppresses-`timing` / `timing`-keeps-`timing` gate).
- **a2a3sim sweep** (dummy_task, all 6 levels, 6/6 `passed`) — clean monotonic ladder:

  | level | STRACE | DEBUG | INFO | TIMING |
  | ----- | ------ | ----- | ---- | ------ |
  | debug | ✓ | ✓ | ✓ | ✓ |
  | info | ✓ | ✗ | ✓ | ✓ |
  | **timing** (default) | ✓ | ✗ | ✗ | ✓ |
  | warn | ✗ | ✗ | ✗ | ✗ |
  | error / null | (silent) | | | |

  Default `timing` keeps host STRACE while suppressing INFO — as intended.

- **a2a3 onboard**, all 6 levels, 6/6 `passed`. The `debug/` (applog) device stream — the one `dlog_setlevel` controls — is gated exactly right: `info` → 266 INFO/round, `timing`/`warn` → **0**, `null` → empty file.

> One nuance worth recording: a fixed ~20 INFO lines remain in the **`run/` device partition** (`device ccpu system process run log`, from the resident `aicpu_scheduler` daemon: `ae_so_manager GetApi`, TSD events). These are **invariant** across `--log-level` (timing→null) **and** `ASCEND_DEVICE_LOG_LEVEL=3` — verified empirically. They are not reachable by `dlog_setlevel` (which governs applog/debuglog only), so this is **not** a gap in this PR; simpler mutes everything the applog level mechanism can mute.

### 2. Performance — the #1409 device-INFO tax is still eliminated

`batch_paged_attention` Case1, 100 rounds, same die, **A-B-A** (timing → info → timing) to bound drift:

| run | level | Avg Device (us) | Avg Effective (us) |
| --- | ----- | --------------- | ------------------ |
| A | timing | 4247.7 | 4223.0 |
| B | **info** | **4810.0** | **4424.4** |
| C | timing | 4314.5 | 4290.0 |

timing baseline (A,C mean): Device **4281.1**, Effective **4256.5**.

| metric | timing → info penalty |
| ------ | --------------------- |
| Device (device_wall) | **−528.9 us / −11.0%** |
| Effective (orch∪sched) | **−167.9 us / −3.8%** |

Drift between the two timing runs is only 66.8 us (1.6%) — far below the info penalty, so the signal is robust. **Effective −3.8% closely matches this PR's reported −4.67%**; Device shows an even larger gap here (my absolute info-Device is higher, likely shared-box contention during the info run). Mechanism confirmed: `info` opens ~266 device INFO/round (the 6299-over-100-rounds tax) that land inside the device wall; default `timing` drives it to 0.

**Conclusion:** all `--log-level` values are functional, host/device stay aligned, and the #1409 optimization is intact (default `timing` avoids the device-INFO tax; ~−3.8% Effective, in line with the PR).

ChaoZheng109 (approved):

PR #1491, opened 2026-07-26
Fix: pace the idle mailbox poll in forked sub/child workers

ChaoWao:
## Summary

`_sub_worker_loop` and `_child_worker_loop` polled their mailbox state word in a bare `while True:` with **no sleep and no yield**, so every forked child held a core for its entire lifetime whether or not it had work. Once the live sub/child workers outnumber the available cores they starve each other and parallel dispatch degrades toward serial.

**Measured, 4 sub workers idle for 3 s (aarch64 Linux):**

| | CPU burned per idle child |
| --- | --- |
| before | **80% of a core** |
| after | **1% of a core** |

## Why this is the child side's bug

The parent already paces itself. `LocalMailboxEndpoint::run` sleeps 50 us between `TASK_DONE` polls (`src/common/hierarchical/worker_manager.cpp`), and `WorkerThread` blocks on a condition variable. The child side was the asymmetry, not a deliberate latency trade.

## Fix

Pace the idle branch to match. The first `_MAILBOX_SPIN_POLLS` (200) consecutive idle passes still spin, so a child dispatched again right away answers at spin latency; past that the poll sleeps for the **same 50 us quantum the parent already accepts**. A burst of back-to-back dispatches is therefore unaffected, while a genuinely idle child releases its core.

The counter resets to 0 on any pass that finds work, so the hot-spin window is fresh after every dispatch.

### Deliberately not changed

`_run_chip_main_loop` (the device-owning chip process) keeps spinning: it is latency-critical on the NPU dispatch path, and its count is bounded by the physical devices, so it is not a source of the oversubscription this fixes. Extending the pacing there would trade real dispatch latency for a saving that does not apply.

## Regression test

`tests/ut/py/test_worker/test_idle_worker_cpu.py` asserts on **CPU time consumed** (per-child `utime+stime` over an idle window), not on the wall time of a parallel workload. That observes the spin directly instead of inferring it from scheduling outcomes, so it does not depend on the scheduler's choices and cannot flake under load. The 25% threshold sits an order of magnitude away from both measured states.

Linux-only (reads `/proc`); the loops under test are platform-independent.

**Red-to-green verified:**

```
# without the fix
AssertionError: 5 idle children burned 4.00 CPU-seconds over 1.0s
(80% of a core each, limit 25%) - the mailbox poll is spinning instead of pacing itself

# with the fix
1 passed in 1.21s
```

## How this surfaced

Triaging the recurring red `ut (macos-latest, 3.10)` check. `test_hostsub_fork_shm.py::test_parallel_wall_time` failed 4 times in 218 sampled macOS jobs (0 in 101 Ubuntu jobs) with wall times of 0.49-0.57 s against a fully-serial baseline of 0.60 s — three children that only `time.sleep(0.2)`, needing no CPU to finish, getting near-zero effective concurrency on a narrow-core runner.

The cost is not confined to CI. While running these tests I found six orphaned pytest processes on this shared dev box, each pinned at ~99.6% CPU for **5-7 days** — leaked children of long-dead test runs, still spinning. An unbounded busy-wait turns every leaked child into a permanently occupied core.

## Verification

- `pytest tests/ut/py -q` -> **825 passed, 2 skipped**.
- `pre-commit` clean.

`_sub_worker_loop` crosses ruff's 15-branch limit with the added idle branch; it carries the same `# noqa: PLR0912` with a reason that its sibling `_run_chip_main_loop` already uses for the identical state-machine shape.

### DIFF
--- a/python/simpler/worker.py
+++ b/python/simpler/worker.py
@@ -952,6 +952,28 @@ def _buffer_field_addr(buf, offset: int) -> int:
     return ctypes.addressof(ctypes.c_char.from_buffer(buf)) + offset
 
 
+# Pacing for the idle branch of the forked sub/child worker mailbox polls.
+# The first _MAILBOX_SPIN_POLLS consecutive idle passes spin, so a child that
+# is dispatched again right away still answers at spin latency; past that the
+# poll sleeps, matching the 50 us granularity the parent's dispatch poll
+# already uses (worker_manager.cpp::LocalMailboxEndpoint::run).
+_MAILBOX_SPIN_POLLS = 200
+_MAILBOX_IDLE_POLL_S = 50e-6
+
+
+def _mailbox_idle_wait(idle_polls: int) -> int:
+    """Pace one idle pass of a child mailbox poll; returns the new idle count.
+
+    A forked child whose poll never pauses holds a core for its entire
+    lifetime, so once the live sub/child workers outnumber the available cores
+    they starve each other and parallel dispatch degrades toward serial.
+    Callers reset the count to 0 whenever a pass finds work.
+    """
+    if idle_polls >= _MAILBOX_SPIN_POLLS:
+        time.sleep(_MAILBOX_IDLE_POLL_S)
+    return idle_polls + 1
+
+
 def _write_error(buf, code: int, msg: str = "") -> None:
     """Write an (error code, message) tuple into the mailbox error region.
 
@@ -1003,7 +1025,7 @@ def _read_args_from_mailbox(buf) -> TaskArgs:
     return read_args_from_blob(mailbox_addr + _OFF_TASK_ARGS_BLOB)
 
 
-def _sub_worker_loop(
+def _sub_worker_loop(  # noqa: PLR0912 -- unified TASK_READY / CONTROL_REQUEST / SHUTDOWN / idle state machine
     buf,
     registry: dict[int, Any],
     identity_table: dict[bytes, int],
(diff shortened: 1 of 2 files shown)


ChaoWao:
Closing in favour of replacing the poll with a real blocking trigger.

Two things came out of review that undercut this PR as framed:

**1. The macOS evidence in the description is wrong.** I cited `test_hostsub_fork_shm.py::test_parallel_wall_time` as measuring the starvation this fixes. That test drives its own test-local `_SubWorkerPool` (its own `while True:` at `test_hostsub_fork_shm.py:79`) and never imports `simpler.worker` — so this change cannot affect it. The claim was inherited from an investigation summary and I did not verify it before writing it down. The idle-CPU measurement (80% of a core per idle child -> 1%) stands on its own; the link to any macOS flake does not.

**2. Pacing tunes the tradeoff instead of removing it.** Review raised that a mailbox poll wants to be as fast as possible, and that a spin window delays a cold dispatch. Measuring both:

| | cold `run()` median | cold p99 | burst per task | idle CPU per child |
| main | 158.1 us | 264.5 us | 110.0 us | 80% of a core |
| window 200 (as pushed) | - | - | **128.9 us (+17%)** | 1.0% |
| window 10000 | 161.0 us | 274.4 us | 105.7 us | 1.0% |

The pushed 200-poll window covers only ~19 us of wall time against a parent turnaround of ~110 us, so a burst slept between every task — a real 17% regression this PR would have shipped. Widening the window fixes that, but `time.sleep(50e-6)` actually costs ~102 us, and the whole scheme is still a window that has to be tuned per host.

A blocking trigger removes the tradeoff: zero CPU while waiting, ~5-20 us wake, no window to tune. It also covers `_run_chip_main_loop`, which this PR deliberately left spinning because pacing it would have cost device-dispatch latency.

Follow-up work, in order:
1. Unify `_sub_worker_loop` / `_child_worker_loop` / `_run_chip_main_loop` — all three drive the identical TASK_READY/CONTROL_REQUEST/SHUTDOWN protocol and differ only in their per-task action and control vocabulary. Doing this first leaves one wait site rather than three.
2. Replace the wait on both sides with a pipe-backed trigger, keeping the existing child-liveness sampling via `poll()` with a timeout so a dead peer still cannot hang the parent.

Most of the latency is on the parent side anyway: `LocalMailboxEndpoint::run` sleeps 50 us per `TASK_DONE` poll, which is on the critical path of every dispatch and dominates the ~150 us round trip. Fixing only the child side would not have moved it.

PR #1492, opened 2026-07-26
Fix: capture dep_gen on the host orchestrator for host_build_graph

### LINKED ISSUE #1487: [Bug] host_build_graph inits dep_gen but never sets its orch thread index, leaving queues[-1]
### Platform

All / Unknown

### Runtime Variant

host_build_graph

### Description

`host_build_graph` initialises the dep_gen collector but never completes its
wiring:

- `src/a2a3/runtime/host_build_graph/runtime/scheduler/scheduler_cold_path.cpp:899`
  calls `dep_gen_aicpu_init()`.
- Nothing in the hbg tree calls `dep_gen_aicpu_set_orch_thread_idx()` or
  `dep_gen_aicpu_flush()`. `tensormap_and_ringbuffer` calls both, from the
  orchestrator block of `aicpu_executor.cpp`.

`s_orch_thread_idx` therefore stays at its initialiser, `-1`
(`src/common/platform/shared/aicpu/dep_gen_collector_aicpu.cpp`), and any
`enqueue_ready` record would index `queues[-1]` — an out-of-bounds write on the
AICPU. Without the flush, whatever did get captured is never handed to the host
either.

Not reachable today: hbg has no dep_gen consumer path, so no record is produced
and the negative index is never used. It is latent, not live — but the init call
implies a working subsystem that is not there, and the next person to enable
dep_gen on hbg gets a silent memory corruption rather than a missing feature.

Two ways out, both reasonable, and picking one is the point of this issue:

- **Finish the wiring** — call `set_orch_thread_idx()` from hbg's orchestrator
  path and `flush()` at teardown, matching `tensormap_and_ringbuffer`.
- **Remove the init** — drop `dep_gen_aicpu_init()` from hbg's cold path so the
  subsystem is honestly absent rather than half-present.

Either way `s_orch_thread_idx` should be validated before use so a `-1` can
never reach an array subscript.

### Steps to Reproduce

Static; no runtime repro exists because the path is unreachable.

```bash
# init is called
rg -n 'dep_gen_aicpu_init' src/a2a3/runtime/host_build_graph/

# the other two halves are not
rg -n 'dep_gen_aicpu_set_orch_thread_idx|dep_gen_aicpu_flush' src/a2a3/runtime/host_build_graph/

# tensormap_and_ringbuffer, for contrast, calls all three
rg -n 'dep_gen_aicpu_init|dep_gen_aicpu_set_orch_thread_idx|dep_gen_aicpu_flush' \
   src/a2a3/runtime/tensormap_and_ringbuffer/
```

### Expected Behavior

Either dep_gen is fully wired on host_build_graph, or it is not initialised
there at all. `s_orch_thread_idx` is never used as an index while negative.

### Actual Behavior

dep_gen is initialised but left with `s_orch_thread_idx == -1` and no flush, so
enabling it on host_build_graph would write `queues[-1]` and produce no output.

### Git Commit ID

bddd321a

### CANN Version

N/A — static defect, not version-specific

### Host Platform

Linux (aarch64)

### Additional Context

Found while root-causing #1486 (dep_gen initialised behind the AICore handshake
on `tensormap_and_ringbuffer`, so the orchestrator's first submits raced it).
That is a different defect on a different runtime; this one is being filed
rather than folded in, since fixing it means choosing whether hbg should support
dep_gen at all.

ChaoWao:
Fixes #1487

## What

`host_build_graph` called `dep_gen_aicpu_init()` from its scheduler cold path but never called `dep_gen_aicpu_set_orch_thread_idx()` or `dep_gen_aicpu_flush()` — the subsystem was initialised and then left half-wired.

It cannot be finished the `tensormap_and_ringbuffer` way: hbg orchestrates on the **host** (`run_host_orchestration` dlsym's and runs the whole orchestration before any scheduler thread starts; the AICPU boot thread only attaches the prebuilt arena), so there is no device orchestrator thread to own a ready queue and nothing device-side to flush.

So this PR gives hbg dep_gen for real, in the shape its architecture allows: **capture the graph as the host orchestrator builds it.**

- `compute_task_fanin` takes an `Annotate` hook that fires alongside its existing `emit`, so creator-retention and tensormap-lookup producers are recorded *as the runtime resolves them*; `submit_task_common` opens the task entry and records declared dependencies. The graph is the runtime's own answer, not a replay's reconstruction, so it cannot drift from `compute_task_fanin` semantics.
- No ring, no collector, no reconcile — orchestration completes before scheduling starts, so nothing can be dropped under back-pressure. The runner picks the shape via `dep_gen_host_graph_active()` and skips device collector init/start/reconcile (and the device DFX flag) for the host-orch runtime.
- `deps.json` keeps the exact schema the device-orch replay emits, so `deps_viewer` and the swimlane join read both runtimes identically.
- hbg's copy of `dep_gen_replay.{h,cpp}` is removed — which is what the device runners' weak stubs already claimed ("host_build_graph has no replay implementation today").

Separately, the shared AICPU writer now guards the negative orchestrator index: a record arriving before `dep_gen_aicpu_set_orch_thread_idx()` has no ready queue to reach, so it is charged to `dropped_record_count` instead of filling a buffer nothing can publish and surfacing later as a misleading "ready_queue full" error. Note the `queues[-1]` write the issue predicted was already impossible — `DeviceProfilerEngine::enqueue_ready` → `wait_for_ready_queue_space` rejects a negative index before `write_ready_entry`.

## Equivalence check

The new st case runs the **same** `vector_example` orchestration the `tensormap_and_ringbuffer` dep_gen test uses, and asserts the same 6 edges. Cross-checking the two artifacts directly (topology by submit order, since hbg keeps the inner manual scope on ring 0 where trb moves it to ring 1):

```text
tmr (device-orch replay): [(0,1,creator), (0,2,creator), (0,4,creator), (1,3,creator), (2,3,creator), (3,4,creator)]
hbg (host-orch direct)  : [(0,1,creator), (0,2,creator), (0,4,creator), (1,3,creator), (2,3,creator), (3,4,creator)]
identical: True
```

## Testing

| Suite | Result |
| ----- | ------ |
| cpput (`-LE requires_hardware`) | 61/61 pass, incl. new `test_dep_gen_collector_aicpu` (written failing first) |
| pyut (`tests/ut/py`) | 824 passed, 2 skipped |
| sim — `tests/st/a2a3/host_build_graph` | 13 passed |
| sim — hbg dep_gen case with `--enable-dep-gen` | passed |
| sim — tmr dep_gen cases with `--enable-dep-gen` | 2 passed |
| onboard (a2a3, via `task-submit`) — hbg dep_gen + vector_example with `--enable-dep-gen` | 2 passed |
| onboard (a2a3) — tmr dep_gen with `--enable-dep-gen` | 2 passed |
| onboard (a2a3) — full `tests/st/a2a3/host_build_graph` | 13 passed, 1 skipped |

The full `--runtime tensormap_and_ringbuffer` sim sweep shows 7 failures / 25 errors on this box (a5 cases collected under `a2a3sim`, multi-device L3 cases against a 1-device sim pool). A baseline worktree at `upstream/main` **without** this change reproduces the identical set (7 failed, 66 passed, 2 skipped, 25 errors), so they are pre-existing environment limits, not regressions.


ChaoWao:
CI note — the first attempt of both no-hardware `ut` jobs failed on
`tests/ut/py/test_worker/test_l4_recursive.py::TestL4WithOwnSubs::test_l4_sub_and_l3_dispatch`
with `assert 1 == 2` (the shared counter saw only one of the two sub-callable
increments). They pass on retry, and everything is green now.

Not this PR: the same assertion fails on an unrelated branch —
[run 30186923896](https://github.com/hw-native-sys/simpler/actions/runs/30186923896)
(`fix-depgen-init-race`, macos `ut`) — and this PR touches no Python worker code
(`git diff upstream/main...HEAD -- python/ simpler_setup/` is empty); the test is
a pure multi-process fork/shared-memory case with no dep_gen involvement. It also
passed 8/8 locally.

Flagging rather than waving it through: it looks like a real race in the L4→L3
dispatch teardown (one path not having completed when `w4.close()` returns) that
happens to reproduce more readily on the hosted runners. Happy to open a separate
issue for it if maintainers want it tracked.

ChaoWao:
Self-review pass (with codex + gemini cross-check) turned up findings; all fixed and re-pushed.

**Must fix — the new capture had zero CI coverage.** All four dep_gen smoke steps hardcoded the `tensormap_and_ringbuffer` path, and the default `pytest examples tests/st` passes no `--enable-dep-gen`, so the new st case was collected, passed, and asserted nothing. Exactly the failure mode the smoke steps were added for in #742. Both a2a3 smoke steps (sim + onboard) now run the host_build_graph dep_gen test alongside the tmr one.

**Should fix — process-global capture state.** `HostGraphState` was a function-local `static`, but both `c_api_shared.cpp` files key the runner off a `pthread_key_t`; two runners on two threads would have wiped and raced each other's graph. Not reachable from Python today (`ChipWorker.run` holds the GIL, sub-workers are forked), but the device-orch counterpart is per-runner (`dep_gen_collector_` is a member) and there was no reason for this half to be weaker. Now `thread_local`, with the isolation contract documented in the header. *Independently flagged by gemini.*

**Should fix — silent empty graph.** `emit()` wrote whatever was in the tables without checking capture was ever armed, so a mis-wired arm produced an empty-but-valid `deps.json` that looks like success (I hit exactly this during development). It now returns non-zero and says so.

**Should fix — st covered only one of three edge sources.** The shared `vector_example` orchestration produces only `creator` edges, leaving the tensormap (Step B) and explicit (STEP 1) hooks unasserted. Added `TestDepGenHostBuildGraphEdgeSources`, which runs `predicated_dispatch` and asserts `{(0,2,explicit), (1,2,tensormap), (2,3,tensormap)}` plus the producer-side slice annotation — the latter being what would catch the `entry` read drifting after the `INOUT`+`COVERED` `remove_entry()`.

**Should fix — doc drift I introduced.** `docs/dfx/dep_gen.md` §1 still defined dep_gen as "capture inputs + replay offline", true of only one of the two shapes after §2 split them.

Also folded in: `block_num` now guards before the widening cast (a negative `int16_t` would have become ~4e9 instead of clamping to 1); `reset()` releases the graph's memory instead of `clear()`ing it (a captured graph is one heap-allocated arg vector per task — bgemm already hits 128 tasks); an explicit `end_task()` closes the task entry so a stray edge can't attach to the previous task; and the dedup-key asymmetry (runtime keys fanin on `(ring, slot)`, capture on producer task id — equal only because hbg never reuses a slot at build time) is now stated in the header.

Re-verified after the fixes: hbg sim suite 14 passed; both dep_gen st cases pass with and without `--enable-dep-gen`; tmr dfx suite 7 passed with `--enable-dep-gen`; cpput 62/62; onboard a2a3 run of the exact new CI command (both runtimes' dep_gen tests) passes. A sweep of every hbg example with capture on exercises all three edge sources (bgemm: 128 tasks / 112 edges, creator 64 + tensormap 48; predicated_dispatch: explicit 1 + tensormap 2) with no edge referencing a `tensor_id` absent from `tensors[]`.

One note on the earlier `ut` flake: still believed pre-existing (same assertion fails on `fix-depgen-init-race`), unrelated to this diff.

PR #1444, opened 2026-07-22
Add Graph Execution to host_build_graph

TaoZQY:
## Summary

- add Graph Execution to the a2a3 `host_build_graph` runtime as a composite incore task alongside AIC, AIV, MIX, and SPMD;
- use the existing `L0TaskArgs` boundary directly; there are no public `GraphArgs`, `GraphBindings`, `Patch`, or `ScalarRef` types;
- record and execute the first structurally compatible invocation, then submit one outer `GRAPH` task for each cache hit;
- build and upload the complete host task image before launching the resident Scheduler; this PR does not add intra-run Host-Orchestrator/Scheduler concurrency;
- allocate Graph execution storage from host-owned, Worker-accounted GM and reuse it across runs by pipeline slot, Graph key, and occurrence index;
- keep host bookkeeping in C++ STL containers and compact it into exact-size, contiguous, pointer-free POD Definitions at the host-device boundary;
- add AIC/AIV/MIX/SPMD coverage, DFX conversion tests, execution-storage/empty-boundary regressions, and a three-layer Qwen3-14B hardware integration case.

## Public API

```cpp
void qwen_decoder_layer(const L0TaskArgs &args) {
    const Tensor &hidden = args.tensor(0).ref();
    const Tensor &weight = args.tensor(1).ref();
    const Tensor &output = args.tensor(2).ref();

    L0TaskArgs task_args;
    task_args.add_input(hidden, weight);
    task_args.add_output(output);
    task_args.add_scalar(uint32_t{16});  // fixed Definition data
    rt_submit_aic_task(FUNC_MATMUL, task_args);
}

void submit_qwen_decoder_layer(const L0TaskArgs &args) {
    rt_submit_graph(&qwen_decoder_layer, args);
}
```

The function pointer is the default Graph identity. Optional trailing integral, floating-point, or boolean construction parameters are forwarded to the function and hashed by value into the Definition key.

Callers may use an explicit `GRAPH_KEY`, but that key must be unique for every distinct Graph function in an orchestration callable. The explicit-key overload deliberately excludes the Graph function pointer so the key remains stable.

Boundary Tensor addresses may change between invocations, while Tensor metadata and aliasing must match the recorded Definition. Scalars used by internal tasks are fixed Definition data. Dynamic scalars and empty boundaries are not cached and execute through the ordinary submit path.

[...]

PR #1495, opened 2026-07-26
Fix: exit a forked worker child once its parent is gone

### LINKED ISSUE #1493: [Bug] Forked worker children survive parent death and spin at 100% CPU forever
### Platform

All / Unknown

### Runtime Variant

All / Unknown

### Description

A forked sub/child worker never checks whether its parent is still alive. If the parent dies without running `Worker.close()` — SIGKILL from a `timeout`, an OOM kill, a cancelled CI job, an IDE/agent killing a wedged pytest — every child is reparented to `init` and keeps running its mailbox poll **forever**. Because that poll is an unbounded busy-wait, each survivor holds a full core indefinitely.

This is a resource leak with no upper bound in time. It is not theoretical: on a shared dev box I found **6 such processes belonging to 2 different users, alive for 5-7 days each at ~99.6% CPU** — 6 cores permanently consumed. Four of the six had `tests/ut/py/test_worker/test_startup_readiness.py` on their command line, i.e. runs whose parent was killed rather than allowed to finish.

### Steps to Reproduce

```python
# pdeath.py
import os, sys
from simpler.worker import Worker

w = Worker(level=3, num_sub_workers=3)
w.register(lambda args: None)
w.init()
kids = sorted(int(p) for p in os.popen(f"pgrep -P {os.getpid()}").read().split())
print(f"parent={os.getpid()} children={kids}", flush=True)
os.kill(os.getpid(), 9)   # parent dies without close()
```

```bash
python pdeath.py            # note the printed child PIDs
sleep 4
ps -o pid=,ppid=,pcpu=,stat= -p <those PIDs>
```

Deterministic — reproduces on the first attempt.

### Expected Behavior

A child whose parent is gone exits. No process outlives the Worker that forked it.

### Actual Behavior

**4 of 4 children survive, reparented to init, three of them spinning:**

```text
2205532       1  0.0 SN
2205533       1  100 RN
2205535       1  101 RN
2205537       1  100 RN
    PID    PPID %CPU STAT
```

`ppid=1` and ~100% CPU is exactly the signature of the week-old orphans found on the shared box.

### Git Commit ID

e17b29aeb1e0e0b8e59e34e5c3ab0c69b2c1ad3f

### CANN Version

9.0.0

### Driver Version

npu-smi 26.0.rc1

### Host Platform

Linux (aarch64)

### Additional Context

**Two independent defects compound here, and the second is not the primary one:**

1. **No parent-death detection** (this issue). The child has no way to learn its parent is gone.
2. **The mailbox poll is an unbounded busy-wait** (`_sub_worker_loop` / `_child_worker_loop`, and `_run_chip_main_loop`). This is what turns a leaked process into a *permanently held core*.

Fixing only (2) would make orphans free on CPU but still leave leaked processes holding fds and shm segments, so it is not a substitute for (1). Fixing (1) is what actually bounds the leak.

**Suggested fix.** A `getppid() == 1` check on the idle path of the mailbox loop is portable and sufficient: the child is already looping there, and reparenting to init is the one observable, race-free signal that the parent is gone. On Linux `prctl(PR_SET_PDEATHSIG, SIGTERM)` right after `os.fork()` makes it immediate rather than poll-latency-bounded, but it is Linux-only (no macOS equivalent) so it is an optimisation on top, not the mechanism.

`python/simpler/worker.py` now routes all three child loops through a single `_run_mailbox_loop`, so the check has exactly one home.

**What I could not substantiate.** I first suspected `TestNextLevelStartupFailure::test_second_child_failure_reaps_first` of leaking children from its reap path, because 4 of the 6 orphans carried that file's command line. It does not: **0 leaks in 20 consecutive runs** once candidate survivors are confirmed by reading their own `/proc/<pid>/cmdline`. An earlier "12/20 leaked" measurement was an artifact of a concurrently running st suite being caught by a PID-set difference. The test reaps correctly when it is allowed to finish; the orphans came from runs that were killed.

ChaoWao:
Fixes #1493. **Stacked on #1494** (`unify-child-mailbox-loops`) — the check needs the single `_run_mailbox_loop` that PR introduces, so review/merge that one first. The diff here is 27 lines plus a test.

## Summary

A forked sub/child/chip worker never checked whether its parent was still alive. When the parent died without `Worker.close()` — SIGKILL from a `timeout`, an OOM kill, a cancelled CI job, an editor killing a wedged pytest — every child was reparented and kept polling a mailbox nobody would ever write to. Because that poll is an unbounded busy-wait, each survivor held a full core **indefinitely**.

Not theoretical: on a shared dev box, **six such processes belonging to two different users had been alive 5-7 days each at ~99.6% CPU** — six cores permanently consumed. Four of the six carried `tests/ut/py/test_worker/test_startup_readiness.py` on their command line: runs whose parent was killed rather than allowed to finish.

## Fix

`_run_mailbox_loop` samples `getppid()` on its idle path and leaves by the same SHUTDOWN route once it changes, so a child tears down exactly as it would on a clean shutdown.

**Why not test for pid 1.** A subreaper — container init, a systemd user session — adopts orphans instead of init, so the pid changes but never becomes `1`. Comparing against the pid captured at loop entry covers both cases, and a live parent's pid cannot change, so the check cannot fire spuriously.

**Cost.** Sampling every `_PARENT_LIVENESS_POLL_INTERVAL` (1000) idle polls puts a `getppid()` roughly every 100 us against a ~0.1 us poll — fast enough that an orphan goes away before it is noticeable, cheap enough to be lost in the noise of the poll itself.

**Residual gap, stated rather than papered over.** A parent that dies between `os.fork()` and the child reaching the loop is already gone when the pid is captured, so that child is not detected. Closing it needs the pid handed in from before the fork, or `PR_SET_PDEATHSIG` — which is Linux-only and therefore an optimisation on top of this, not the mechanism.

## Regression test

`test_orphan_child_reaping.py` SIGKILLs a subprocess parent and asserts every child is gone within 20 s.

Two details are load-bearing, both learned by getting them wrong first:

- **It writes to a file and never reads a pipe.** Orphans inherit the parent's stdout, so on regression they hold a pipe's write end open — my first version used `subprocess.run(capture_output=True)` and *hung for 120 s* instead of failing.
- **Its `finally` kills whatever survived.** That first version raised before it had parsed the pids, so its cleanup never ran and it leaked three spinning processes per run — a leak test that leaks.

**Red-to-green:**

```
# without the fix
AssertionError: 4 of 4 children outlived their SIGKILLed parent by 20s:
[2453757, 2453758, 2453759, 2453760] - a child is not noticing that its parent is gone
1 failed in 20.41s          (and no orphans left behind)

# with the fix
1 passed in 0.35s
```

## Verification

- `pytest tests/ut/py -q` -> **825 passed, 2 skipped** (824 baseline + the new test).
- `pytest examples tests/st --platform a2a3sim --device 0-7 -q` -> **108 passed, 2 skipped, 0 failed, 60/60 groups** — identical to the #1494 baseline. The chip loop shares this code path, so the device run is the one that matters.
- `pre-commit` clean.

## What this does not fix

The busy-wait itself. With the poll replaced by a blocking trigger an orphan would cost 0% CPU — but it would still be a leaked process holding fds and shm, so that work is complementary, not a substitute. This PR is what bounds the leak in time.


ChaoWao:
CI caught a defect in the regression test itself — the fix under test is unaffected:

```
AssertionError: 1 of 6 children outlived their SIGKILLed parent by 20s: [3]
```

Two problems, both from enumerating children with `pgrep -P` inside the parent:

- The transient shell that `os.popen` spawns to run `pgrep` **is itself a child of the parent**, so it landed in the list.
- Inside the CI container's shallow pid namespace the list also picked up unrelated low pids — hence `[3]`. The test then waited 20 s on a container process and, in its `finally`, sent it `SIGKILL`.

Fixed by taking the pids from the Worker's own `_sub_pids` bookkeeping, and tightening `>= NUM_SUB_WORKERS` to `== NUM_SUB_WORKERS` so a wrong count fails loudly instead of being absorbed.

With `pgrep` gone the test is no longer Linux-specific, so the `skipif` is dropped — reparenting on parent death is POSIX behaviour, and the macOS `ut` job (which failed for the same reason) now covers it too.

Re-verified red-to-green against the parent branch's `worker.py`, since the fix is committed and `git stash` had nothing to remove — my first attempt at this "passed" only because it had silently run *with* the fix:

```
# without the fix
AssertionError: 3 of 3 children outlived their SIGKILLed parent by 20s: [2881160, 2881161, 2881162]
1 failed in 20.26s

# with the fix
1 passed in 0.24s   (5/5 consecutive runs)
```

`pytest tests/ut/py -q` → 825 passed, 2 skipped.

ChaoWao:
Rebased onto `main` now that #1494 has merged, and this PR now also **removes `src/a2a3/runtime/host_build_graph/docs/GRAPH_EXECUTION.md`**.

That file should never have been in #1494. It is documentation from the **open #1444** (`feat/graph-execution`, @TaoZQY) that was sitting untracked in my working tree, and a `git add -A` swept it into the refactor commit — so #1494 merged 418 lines of somebody else's unmerged PR into `main`. Removing it here restores `main` to the state #1444 expects to merge into. It is unrelated to the fix and can be reviewed independently.

Also hardened the regression test after it failed once in ~22 local runs with an unhelpful `expected exactly 3 forked sub-worker pids, parent reported []`. I could not reproduce that (0 failures in 27 subsequent runs), so rather than guess at a cause I made the test name it: the parent is supposed to die from its own `SIGKILL`, so the test now asserts `returncode == -SIGKILL` and dumps the parent's own output otherwise. A parent that fails before forking now reports *why*, instead of surfacing as a pid-count mismatch that reads like a defect in the code under test.

Verification after the rebase:

```
# without the fix (main's worker.py)
AssertionError: 3 of 3 children outlived their SIGKILLed parent by 20s
1 failed in 20.35s

# with the fix
1 passed   (12/12 consecutive runs)
```

`pre-commit` clean.

Issue #1496, opened 2026-07-26
[Bug] test_parallel_wall_time: forked workers run near-serially on macOS CI (~2% of runs)


ChaoWao:
### Platform

All / Unknown

### Runtime Variant

All / Unknown

### Description

`tests/ut/py/test_hostsub_fork_shm.py::TestParallelExecution::test_parallel_wall_time` fails on macOS CI in roughly 2% of runs (**4 of 218 sampled `ut (macos-latest, 3.10)` jobs** over Jul 20-26; **0 of 101** Ubuntu jobs).

The test forks 3 workers, dispatches a `time.sleep(0.2)` callable to each, and asserts the wall time is under 80% of fully-serial.

**This is not a marginal threshold.** Measured on aarch64 Linux over 7 consecutive runs:

| | elapsed |
| --- | --- |
| ideal parallel | 0.200 s |
| **Linux actual** | **median 0.203 s, max 0.204 s** |
| assertion threshold | 0.480 s |
| **macOS when failing** | **0.49 - 0.57 s** |
| fully serial | 0.600 s |

Linux lands 1.5% above ideal with a 2 ms spread. The macOS failures are 2.5-2.9x that, and 85-95% of *fully serial* — parallelism essentially did not happen. Raising the threshold would silence exactly the signal this test exists to produce.

### Steps to Reproduce

```bash
python -m pytest tests/ut/py/test_hostsub_fork_shm.py::TestParallelExecution::test_parallel_wall_time
```

Only observed on macOS runners, ~1 in 50. **I have no macOS box and could not reproduce it**; everything below is from CI logs plus controlled Linux experiments.

### Expected Behavior

Three children that each only `time.sleep(0.2)` finish in ~0.2 s wall time, as they do on Linux.

### Actual Behavior

```text
expected parallel wall time < 0.48s (serial would be 0.60s), got 0.57s
```

### Git Commit ID

b28bd58f2825d571edeff2353e9925a178a4c23d

### CANN Version

N/A (pure-Python unit test, no NPU)

### Driver Version

N/A

### Host Platform

macOS (aarch64)

### Additional Context

**Ruled out, with the evidence:**

- **Not the product's worker code.** `_SubWorkerPool` and its `_worker_loop` are defined locally in the test file; it never imports `simpler.worker`. Neither #1494 (unifying the child mailbox loops) nor #1495 (parent-death detection) can affect it.
[...]

Issue #1489, opened 2026-07-26
[Bug] Allreduce L3 cases intermittently hit an AICore UB out-of-bounds in full-suite onboard runs


ChaoWao:
### Platform

a2a3 (Ascend 910B/C hardware)

### Runtime Variant

tensormap_and_ringbuffer

### Description

Two L3 collective cases — `TestAllreduceOnephaseP2` (2 ranks) and `TestAllreduceRingP4` (4 ranks) — intermittently fail with an **AICore VEC addressing fault**, not a timeout:

```text
VEC instruction error: the ub address out of bounds
  core id 15, blk:1, errcode 0x4000000000000000
→ aicpu exception → 507018 → sched_error_code=100 SCHEDULER_TIMEOUT
```

The `507018` / `SCHEDULER_TIMEOUT` tail is the *consequence*, not the mechanism: per `.claude/rules/running-onboard.md`, `507018` is a generic host-side code, and here the device log names a UB out-of-bounds as the thing that actually fired. A vector unit reading outside its UB allocation is a real addressing bug, so this should not be filed away as scheduling noise even though it presents as flaky.

Only reproduced in a **full** `pytest examples tests/st` run with many devices in flight; never in isolation.

**This is likely to get more frequent, not less.** The failures were seen while validating a wide parallel scheduler, and #1309 has since removed the user-facing `block_dim` knob — every run now takes the whole device (24 clusters on a2a3) instead of whatever a case pinned. Whatever contention widens this window is now the default configuration for every case in the suite.

### Steps to Reproduce

```bash
# Full suite, 8 devices — the failing cases are multi-device L3 running
# concurrently with unrelated cases under the resource scheduler.
task-submit --timeout 3600 --max-time 3600 --device auto --device-num 8 \
    --run "python -m pytest examples tests/st --platform a2a3 --device \$TASK_DEVICE"
```

Observed once in a full-suite run; a second full-suite run immediately after passed 54/54, and `pytest tests/st/worker/collectives/allreduce --platform a2a3` alone passed 10/10 on the same build. Expect to repeat the full suite several times to hit it.

To capture the device-side evidence, redirect the device log out of the shared default first:

```bash
LOGDIR="$PWD/outputs/allreduce_ub/ascend"; mkdir -p "$LOGDIR"
export ASCEND_PROCESS_LOG_PATH="$LOGDIR"
```

### Expected Behavior

[...]

PR #1503, opened 2026-07-26
Add: per-worker timeline to the parallel wall-time assertion

ChaoWao:
Makes the next occurrence of #1496 self-diagnosing. **No behaviour change, no threshold change.**

## Why

`test_parallel_wall_time` fails on macOS CI in ~2% of runs (4 of 218 sampled `ut (macos-latest, 3.10)` jobs; **0 of 101** on Ubuntu) with a total of 0.49-0.57 s against a **0.203 s Linux median** (max 0.204 s over 7 runs). That is near-serial, not a marginal threshold.

But the assertion reports only the total, which cannot separate the two shapes:

- the parent was descheduled between its three sequential `dispatch()` calls, so w1/w2 started a full sleep period behind, **or**
- every worker was dispatched on time and finished late.

Those are different bugs. The total is identical either way.

## What

Record each worker's dispatch and completion offset and include them — plus the visible CPU count — in the failure message. Verified by temporarily tightening the threshold:

```
AssertionError: expected parallel wall time < 0.48s (serial would be 0.60s), got 0.20s
  cpu_count=320 affinity=320
  w0: dispatched +0.1ms, done +200.6ms, w1: dispatched +0.1ms, done +200.6ms, w2: dispatched +0.1ms, done +204.5ms
```

On Linux all three dispatch at +0.1 ms and finish together — the healthy shape. On the next macOS failure the same line says which worker started late, and the question in #1496 is answered from the CI log alone.

`sched_getaffinity` is Linux-only, hence the guard. `cpu_count()` alone can overstate what a container actually lets the process run on, and the runner's real width is one of the unknowns #1496 could not close.

## What this deliberately does not do

**It does not touch the assertion.** #1496 asks explicitly that the threshold not be relaxed and no retry be added: at 0.49-0.57 s against 0.203 s, something is serialising three processes that are each only in `time.sleep(0.2)`. A wider bound would make the symptom disappear and leave the cause running.

This is the same approach as #1501 for #1489 — when a failure is rare and cannot be reproduced on demand, the highest-value move is to guarantee the next natural occurrence yields the evidence, rather than spending exclusive hardware or CI time trying to force one.

## Verification

- `pytest tests/ut/py/test_hostsub_fork_shm.py -q` → 6 passed.
- Failure path exercised by temporarily tightening the bound (output above), then restored.
- `pre-commit` clean.

### DIFF
--- a/tests/ut/py/test_hostsub_fork_shm.py
+++ b/tests/ut/py/test_hostsub_fork_shm.py
@@ -300,19 +300,39 @@ class TestParallelExecution:
         pool = _SubWorkerPool(num_workers=n_workers, registry=registry)
 
         try:
+            # Per-worker timestamps, so a failure says *which* worker ran late
+            # rather than only that the total was too high. Workers are
+            # dispatched in a loop, so if the parent is descheduled between
+            # iterations the later workers start a full sleep period behind —
+            # a shape the total alone cannot distinguish from every worker
+            # being slow. See #1496.
+            dispatched: list[float] = []
+            completed: list[float] = []
+
             start = time.monotonic()
             for i in range(n_workers):
                 pool.dispatch(i, i)
+                dispatched.append(time.monotonic() - start)
             for i in range(n_workers):
                 result, err = pool.wait(i, timeout=5.0)
+                completed.append(time.monotonic() - start)
                 assert err == 0
                 assert result == int(sleep_sec * 1000)
             elapsed = time.monotonic() - start
 
             serial_time = n_workers * sleep_sec
+            timeline = ", ".join(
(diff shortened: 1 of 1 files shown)


ChaoWao:
Superseded by #1509.

This PR existed to instrument the wall-clock failure message so the next macOS CI failure could be diagnosed blind. With a macOS box now available, the failure is traced to `time.sleep` **over-sleeping under load** (measured 0.2 s → 0.24–0.31 s under `taskpolicy -b` + contention) — not a code defect, and not reproducible on bare metal. So the wall-clock assertion is **dropped outright** (and the no-wall-clock-proxy principle codified in `docs/testing.md`) rather than instrumented. Closing in favor of #1509.

PR #1506, opened 2026-07-26
Update: re-harvest qwen3_14b_decode and scale it to all 40 layers

### LINKED ISSUE #1484: [Code Health] qwen3_14b_decode: re-harvest stale 2-layer codegen and scale the case to 40 layers
### Category

### Component

Tests

### Description

`examples/a2a3/tensormap_and_ringbuffer/qwen3_14b_decode/` is simpler's only in-repo Qwen3-14B workload — the "no cross-repo descent" reference case a developer runs to exercise a realistic fused decode. Its C++ is **harvested pypto-lib codegen** (36 sources: orchestration + 8 AIC + 27 AIV, ~15.2k LOC) plus a hand-ported golden (`simpler_setup/goldens/qwen3_14b_decode.py`, 263 lines). Vendored codegen does not age well, and this copy has now drifted from its upstream source in three concrete ways.

**1. The stated provenance file no longer exists upstream.** The example (README + module docstring) says it is a port of pypto-lib `models/qwen3/14b/decode_layer.py`, entry `decode_fwd_layers`. `decode_layer.py` was **deleted** in pypto-lib (#670, restored as #747); the entry now lives in `models/qwen3/14b/decode_fwd.py`. The README's "To regenerate:" recipe therefore cannot be followed as written — the regeneration path is documented against a file that is gone.

**2. The chunk size is pinned two generations back.** The harvest is `_CHUNK_NLAYERS == 2`. Upstream's default is now **8** (`decode_fwd.py:1125`), and the driver exposes `--validate-fwd --fwd-layers N`. So the vendored kernels no longer represent the shape pypto-lib actually generates.

**3. The codegen has not been refreshed in a month.** Last real harvest was #1088 (2026-06-25, 1-layer → 2-layer). Every subsequent commit touching the directory (#1104, #1123, #1181, #1265, #1270) was a mechanical repo-wide refactor (ArgsDump rename, positional dump / `func_id` array, TPUSH/TPOP normalization) — not a re-harvest. Anything pypto-lib changed in the decode path since (e.g. the recent split-K on-core seed and indexer/SWA-store work) is absent here.

Consequences: the example under-covers the decode stack (**2 of 40 layers = 5%**), and it silently misrepresents current pypto codegen to anyone using it as the reference for how generated kernels look or as a perf/scheduling workload.

**Scale target.** Qwen3-14B is `num_layers=40` (`build/pypto-lib/models/qwen3/14b/constants.py:105`). Running the full 40 layers is the ask here. Footprint at the existing fixture regime (BATCH=16, MAX_SEQ=5500, decode seq_len=3500), weights + paged KV, bf16:

| layers | weights | KV pool | total |
| -----: | ------: | ------: | ----: |
| 2 (today) | 1.23 GiB | 0.67 GiB | **1.90 GiB** |
| 8 (upstream chunk) | 4.92 GiB | 2.69 GiB | **7.61 GiB** |
| 40 (full model) | 24.61 GiB | 13.44 GiB | **38.05 GiB** |

Die HBM is 64 GiB, so 40 layers **fits** — but it leaves ~26 GiB for ring heap, workspace and dumps, so ring/heap sizing (`PTO2_RING_HEAP`, task window, dep pool) and CI wall-time both need to be re-checked rather than assumed. Note the fixture reuses layer-0 weights stacked along dim 0 (the lib's const-layer-0 stacked-fwd reference), so no real 14B checkpoint is needed — the 24.6 GiB is stacked duplicates, and aliasing the slices is one lever if the footprint proves too tight.

### Location

- `examples/a2a3/tensormap_and_ringbuffer/qwen3_14b_decode/` (whole directory — `README.md`, `test_qwen3_14b_decode.py`, `kernels/orchestration/decode_fwd_layers.cpp`, `kernels/aic/` ×8, `kernels/aiv/` ×27)
- `simpler_setup/goldens/qwen3_14b_decode.py` (`N_LAYERS = 2` at line 52)
- Upstream source: `pypto-lib models/qwen3/14b/decode_fwd.py` (`_CHUNK_NLAYERS = 8` at line 1125; `--fwd-layers` at line 1564)

### Proposed Fix

1. **Re-harvest from the current entry.** Regenerate orchestration + kernels + `kernel_config.py` from `decode_fwd.py`'s `decode_fwd_layers` at today's pypto-lib main, and update the golden to match. Fix the README/docstring provenance to name `decode_fwd.py` (not the deleted `decode_layer.py`) and record the exact pypto-lib commit the harvest came from, so the next drift is detectable instead of archaeological.

2. **Decide the 40-layer shape** — this is the one open design question and is worth settling in this issue before implementation:
   - **(a) One fused chunk of 40** (`_CHUNK_NLAYERS=40`): a single dispatch, but produces a very large kernel set and diverges from how the lib actually runs the model.
   - **(b) 5 dispatches of the upstream chunk-8 codegen** *(recommended)*: matches production decode exactly, keeps codegen size at the upstream default, and turns the example into a real multi-dispatch scheduling workload — the case loops the callable 5× over per-chunk weight/KV slices with the FP32 residual carried across chunks.

3. **Re-check the one hand-edit.** The current harvest patches `fa_fused_aiv` to replace a `[[block_local]] static` sub-block id (rejected by the strict `.text`-only loader, #900/#830/#831) with `setEntryOffset(get_sub_block_id(args) * …)`. Verify whether current codegen still emits that construct; if it does, prefer the direct `TPUSH`/`TPOP` form the README already flags as the long-term fix (`docs/tpush-tpop.md`, `tests/st/a2a3/tensormap_and_ringbuffer/spmd_paged_attention/`).

4. **Size and gate the run.** Re-tune ring heap / task window / dep pool for the 40-layer live set, measure wall time, and pick the CI tier deliberately — a 38 GiB, full-depth case may belong behind a heavier level than the current `level=2` if it slows the default suite.

5. **Update the docs that describe it.** `.claude/skills/multi-repo-qwen-setup/SKILL.md` documents this example as "Path 0" with a 2-layer description; it needs the same refresh (it also still refers to pypto-lib `decode_layer.py`).

Verification stays as-is: the case validates against the torch golden on device (current status: passing, `max_abs_diff ≈ 0.03`), so a correct re-harvest at N=40 should keep matching output and all per-layer KV caches.

### Priority

_Observed at simpler commit `f8e206700ab349a90f9db4e57b50fe30b4a4b53d`; upstream pypto-lib `45be52c`._

ChaoWao:
Fixes #1484.

## What

Re-harvests `examples/a2a3/tensormap_and_ringbuffer/qwen3_14b_decode/` from
pypto-lib `45be52c` and scales it from a 2-layer chunk to **all 40 Qwen3-14B
decode layers in one dispatch**.

Two commits: the `CALLABLE` include-dirs hook the harvest needs, then the
harvest itself.

## Settling the issue's open question

The issue asked whether to run 40 layers as one fused chunk or as 5 dispatches
of the upstream chunk-8, and recommended the latter to keep codegen size down.
**That reason does not hold.** The generated orchestration *loops* over layers —
`for (int64_t i = 0; i < 40; i += 1)` — so going 2 → 40 moves a trip count and
leaves the kernel set the same size. One dispatch it is.

The lib chunks because a 40-layer dispatch used to exceed a 2000 ms stream-sync
timeout (`decode_fwd.py:1153`). simpler raised that to 50 s in #1175 and a
40-layer step is ~40 ms, so the constraint is gone.

Two other framings in the issue turned out not to apply: `@scene_test(level=)`
is the runtime hierarchy level (L2 chip vs L3 worker), not a CI cost tier, and
layer count is baked into the orchestration `.cpp`, so it cannot be a per-`CASES`
param — one layer count per class.

## What actually changed: attention is no longer codegen

Since pypto-lib #765/#796 the attention stage is a **CANN
`FusedInferAttentionScore` extern**, not generated kernels. It folds per-head
QK-norm, RoPE, the paged KV write, the flash inner loop and the online softmax
into one mixed AIC+2×AIV task behind an FFTS barrier.

- **Seven kernels disappear**: `rope_qkv`, `qk_gamma`, `qk_recip`,
  `fa_work_build`, `fa_fused_aic`/`_aiv`, `online_softmax`, `attn_fence`.
- The extern arrives under `kernels/paged_attention_cce/`, copied **verbatim**.
  `kernel/fai_body.hpp` reaches its dependencies by relative include, so
  splitting the tree would mean patching the source and re-patching on every
  refresh. Its `vendor/` subtree is CANN source under Huawei copyright
  (~16k LOC). Verbatim also means our header/format/language lint must leave it
  alone — hence the carve-out in `.pre-commit-config.yaml` and
  `tests/lint/check_headers.py`.
- Paged KV moves from NSND to vLLM's **BSND** (`[page, token, kv_head, dim]`, so
  `slot_mapping` is directly the row index). The golden models the new layout on
  both the attention read and the KV write-back.
- Building it needs CANN devkit headers, which is what the first commit's
  `CALLABLE["incores"][i]["extra_include_dirs"]` is for. `$ASCEND_HOME_PATH`
  keeps them off machine-specific paths.

This depends on #1497 — AscendC's block-local globals carry relocations no
amount of inlining removes, so these kernels could not load before it.

## Issue item 3: the hand-edit is gone

The previous harvest patched `fa_fused_aiv` to replace a `[[block_local]]
static` the loader rejected. That kernel no longer exists and the current
codegen emits no such construct, so **this harvest is unmodified codegen
throughout** — a refresh is now a plain copy.

## Verification

- **On device: PASSES.** Output *and all 40 layers' KV caches* match the torch
  reference at `RTOL=5e-2 / ATOL=1e-1`. ~5 min wall on an idle die.
- **No ring-sizing overrides needed.** Each layer's intermediates live inside
  that iteration's scope, so the live set does not grow with layer count — the
  default heap / task window / dep pool carry the 40-layer graph.
- **Footprint 38.05 GiB** (24.61 weights + 13.44 paged KV, bf16) against 64 GiB
  HBM, matching the issue's estimate.
- a2a3 onboard sweep: **52 PASS / 0 FAIL**. a2a3sim: **60 PASS / 0 FAIL**.
  Python UTs: 829 passed, 2 skipped. Full pre-commit clean on the diff.

## Docs

README records the exact pypto-lib commit plus the ptoas/pto-isa pins and a
regeneration recipe that works (`codegen_only=True`, no device). Also updated:
`multi-repo-qwen-setup` Path 0, `docs/testing.md` for the new key.

The l0-swimlane docs used this example as their "real SPMD workload", shrinking
the generated `fa_fused` via `--set-arg 0=96`. The extern takes its work
distribution from runtime tiling metadata, so no `--set-arg` shrinks it — those
references are removed rather than repointed at something unverified. The
dfx buffer-capacity row measured at 2 layers is dropped for the same reason.

## Notes for review

- One deliberate deviation from `kernel_config.py`: it marks `k_cache`/`v_cache`
  as plain inputs, but the extern writes the current token's KV into them. The
  `CALLABLE` marks them `INOUT` so simpler copies the pools back and the golden
  can verify all 40 layers' KV writes — the stronger check the issue asks for.
- The diff is large (~30k lines) because of the vendored CANN tree; the
  hand-written surface is the test file, the golden, and the lint carve-outs.


ChaoWao:
CI triage: everything is green except `ut` on ubuntu/macos, which is **not from this PR**.

It is `test_orphan_child_reaping` failing with `expected exactly 3 forked sub-worker pids, parent reported [.., .., .., 3]` — the test scrapes pids from the parent's combined stdout+stderr, and a `resource_tracker: There appear to be 3 leaked shared_memory objects` warning contributes a stray digit. #1504 is already fixing exactly this and states that main is currently red on `ut` because of it. Nothing here touches worker forking or shared memory, and the test passes 5/5 locally.

Will rebase once #1504 lands. Everything else on [this run](https://github.com/hw-native-sys/simpler/actions/runs/30213051116) passes, including `st-onboard-a2a3` — so the 40-layer case builds its CANN attention extern and runs on real hardware in CI.

PR #1539, opened 2026-07-27
Fix: bind AICore run streams to code images

Crane-Liu:
## Summary

- derive a stable AICore image identity from child kernel bytes and function mapping during callable registration
- keep AICPU streams slot-owned while recreating the AICore stream whenever the loaded image changes
- preserve same-image stream reuse and expose code-generation transitions through the existing creation counter
- add A2/A3 hardware coverage that alternates observably different add/sub images and checks every output

## Testing

- `python -m pytest tests/ut -q`: 832 passed, 6 skipped
- non-hardware CTest: 62/62 passed
- pre-commit on all changed files: passed
- A2/A3 architecture precheck: passed
- A2/A3 hardware regression: `task_20260727_061452_383131022458`, 3/3 passed
- failing regression on `main`: `task_20260727_060331_37075211981` reached the A-to-B transition with correct outputs but the stream-generation counter remained unchanged

## Stack

This is the A0b prerequisite from the worker async pipeline plan. It should land before the rewritten acceptance fence in #1467 and before depth-two slot leases are enabled.

### DIFF
--- a/python/simpler/task_interface.py
+++ b/python/simpler/task_interface.py
@@ -1264,7 +1264,7 @@ class ChipWorker:
 
     @property
     def run_stream_set_create_count(self):
-        """Number of run stream sets the bound runner has created."""
+        """Number of run stream generations the bound runner has created."""
         return self._impl.run_stream_set_create_count
 
     def malloc(self, size):
--- a/python/simpler/worker.py
+++ b/python/simpler/worker.py
@@ -6155,12 +6155,12 @@ class Worker:
 
     @property
     def run_stream_set_create_count(self) -> int:
-        """L2 only: number of run stream sets the bound runner has created.
+        """L2 only: number of run stream generations the runner has created.
 
-        A set belongs to a pipeline slot and is reused for every run on that
-        slot, so a worker that has served any number of runs reports 1.
-        Returns 0 on non-L2 workers and on platforms whose runs use the
-        persistent bootstrap stream pair (simulation, a5).
+        AICPU streams belong to pipeline slots. AICore streams are reused only
+        while the loaded AICore image is unchanged, so each code transition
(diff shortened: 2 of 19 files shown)


ChaoWao:
Closing — this landed with #1467.

#1467 was merged as a squash of both commits (`b072b5b9`, 35 files, +933/-137), so the A0b work went in underneath the acceptance change rather than ahead of it. The intended A0b → A1 ordering did not survive the merge, but nothing was dropped.

Verified file by file against `upstream/main` rather than by patch-id — `git cherry` reports `+` here only because the commit was rebased, so its patch-id no longer matches.

**On main already:**

- `aicore_image_hash` / `has_aicore_image` on the run stream set
- `ensure_run_stream_set(slot, aicore_image_hash)`
- `tests/st/a2a3/host_build_graph/run_stream_reuse/kernels/aiv/kernel_sub.cpp`
- `test_aicore_stream_tracks_code_image`
- `tests/ut/cpp/types/test_chip_callable_upload_immutable.cpp`

**Only on this branch — all of it stale, none of it a loss:**

| Line | Why main is right |
| --- | --- |
| `config` carries `block_dim (0 = auto)` | #1309 removed the user-facing `block_dim` knob |
| `_MAILBOX_ARGS_CAPACITY` without the `- 8`, four-argument `run_from_blob` | pre-dates the sticky acceptance word carved out in #1467 |
| "Only one live device run is admitted", `_wait_for_serialization` in the submit drain | pre-dates the acceptance fence; `submit()` now drains on acceptance |

So the branch is behind main, not ahead of it.

Sorry for the collapsed stack — that was mine. I pushed the rebased #1467 with this commit still below it, which is what let the squash swallow both.

PR #1544, opened 2026-07-28
host_build_graph: dedicated resolution thread (3S+1P)

ChaoZheng109:
## What

Split the `host_build_graph` AICPU threads into **3 core-owning schedulers (S) + 1 core-less resolution thread (P)** (generally: `aicpu_thread_num - 1` S threads + 1 P, enabled at ≥2 threads).

- **S** polls the COND registers of its **own** AICore cores, detects FIN, frees the core immediately (core reuse never waits on P), and hands the finished task to P over a **per-S SPSC queue** (single-producer/single-consumer, plain release/acquire, no CAS).
- **P** owns no cores. It drains the S→P queues and runs `on_task_complete` for every finished task — publish `completion_flags`, drain the wake list, advance `completed_watermark`. P owns `completed_tasks_` and the terminal `completed_` flip, so the S threads keep dispatching until P has resolved the whole graph (the host's `wait_for_consumers` never sees a stranded prefix).

Because **P is the sole producer of the ready queues**, the enqueue CAS never contends and the wake-list drain is single-threaded. Async deferred completions and dependency-only retirement (dummy + predicate-failed tasks) also move to P, so the S loop stays purely core-local — poll own COND, dispatch own cores — with no per-iteration polling of the shared mailbox / dummy queue.

Scope: **a2a3 `host_build_graph` only.**

## Why

Under the polling-completion model (#1435), the 4 symmetric scheduler threads each do completion **and** resolution **and** dispatch, so the wake-list CAS and the MPMC `ready_queues[]` enqueue contend across all 4. Concentrating resolution on one thread removes that contention and keeps the dispatch threads' hot loop core-local.

## Performance (a2a3, single die, pinned device, L2-swimlane device-execution span)

Same binary A/B (4S = this change reverted, 3S+1P = this change):

| workload | shape | util | 4S | 3S+1P | Δ |
| --- | --- | --- | --- | --- | --- |
| `paged_attention` Case1 (65536 tasks) | wide, scheduler-bound | 67% | 29.5 ms | **25.4 ms** | **−14%** (4 runs, 26.0–26.4→25.4 stable) |
| dep-bound A/B **wide** (matmul/add, 128 chains) | wide, parallel | 88% | 0.44 ms | **0.43 ms** | **−3.8%** |
| qwen3-14B decode, 40 layers, batch16/seq3500 | wide, **compute-bound** | 98% | 738 ms | 740 ms | ~0% (no headroom) |
| dep-bound A/B **deep** (single serial chain) | serial, latency-bound | 60% | 2.39 ms | 2.44 ms | +2.3% |
| `deep_chain` (512-deep add-scalar) | serial, latency-bound | 47% | 3.2 µs/edge | 3.35 µs/edge | +4.7% |

Reading:
- **Scheduler-bound + parallel → speedup** (`pa` −14%): the scheduler-thread completion phase roughly halves as resolution contention leaves the hot path (3.55M→1.74M cyc).
- **Compute-bound → neutral** (qwen batch16 util 98%): the scheduler is not the bottleneck; 3S+1P still cuts scheduler work ~13% but there is no idle headroom to reclaim.
- **Serial/latency-bound → small cost** (deep +2–5%): each dependency edge pays one extra S→P→S cross-core round-trip that a serial chain cannot hide behind parallel work.

The dep-bound A/B holds kernels and tile size constant and varies only the dependency structure, isolating the direction split to the resolution path.

Correctness: `matmul`, `bgemm`, `paged_attention` (Case1 + Case2 SPMD), `vector_example`, `predicated_dispatch`, and the 2-layer qwen3 decode graph all pass on 3S+1P.

## Notes / open questions

- Async completions are polled by P but a workload with no async (all cases here) never exercises that path; correctness holds either way (`on_task_complete` is internally CAS-safe).
- The serial cost is inherent to splitting detect→resolve→dispatch across two cores (one 64-byte slot-state cache line + the ready_queue line bounce S↔P per edge). A conditional single-consumer inline fast path was considered but not included — it risks re-introducing ready-queue contention on the wide case (real workloads are majority fanout-1), so it needs its own measurement before landing.

### DIFF
--- a/src/a2a3/runtime/host_build_graph/aicpu/aicpu_executor.cpp
+++ b/src/a2a3/runtime/host_build_graph/aicpu/aicpu_executor.cpp
@@ -288,7 +288,11 @@ int32_t AicpuExecutor::run(Runtime *runtime) {
             LOG_ERROR("Thread %d: rt is null after orchestrator error, skipping dispatch", thread_idx);
         } else {
             sched_ctx_.bind_runtime(rt);
-            int32_t completed = sched_ctx_.resolve_and_dispatch(runtime, thread_idx);
+            // 3S+1P: the last thread is the core-less resolution (P) thread; the
+            // rest are core-owning schedulers (S).
+            int32_t completed = (sched_ctx_.p_thread_mode() && thread_idx == sched_ctx_.p_thread_idx()) ?
+                                    sched_ctx_.run_resolution_thread(runtime, thread_idx) :
+                                    sched_ctx_.resolve_and_dispatch(runtime, thread_idx);
             if (completed < 0) {
                 LOG_ERROR("Thread %d: Scheduler failed with rc=%d", thread_idx, completed);
                 run_rc = completed;
--- a/src/a2a3/runtime/host_build_graph/aicpu/aicpu_executor.cpp
+++ b/src/a2a3/runtime/host_build_graph/aicpu/aicpu_executor.cpp
@@ -290,7 +290,7 @@ int32_t AicpuExecutor::run(Runtime *runtime) {
             sched_ctx_.bind_runtime(rt);
             // 3S+1P: the last thread is the core-less resolution (P) thread; the
             // rest are core-owning schedulers (S).
-            int32_t completed = (sched_ctx_.p_thread_mode() && thread_idx == sched_ctx_.p_thread_idx()) ?
+            int32_t completed = (thread_idx == sched_ctx_.p_thread_idx()) ?
                                     sched_ctx_.run_resolution_thread(runtime, thread_idx) :
                                     sched_ctx_.resolve_and_dispatch(runtime, thread_idx);
             if (completed < 0) {
--- a/src/a2a3/runtime/host_build_graph/runtime/scheduler/scheduler_cold_path.cpp
+++ b/src/a2a3/runtime/host_build_graph/runtime/scheduler/scheduler_cold_path.cpp
@@ -700,7 +700,14 @@ void SchedulerContext::handshake_partition(Runtime *runtime, int32_t tidx, int32
 bool SchedulerContext::assign_cores_to_threads() {
     // Cluster-aligned round-robin assignment: cluster ci -> sched thread ci % active_sched_threads_.
     // Each cluster = 1 AIC + 2 adjacent AIV; the triple is always kept together.
-    active_sched_threads_ = aicpu_thread_num_;
+    //
+    // 3S+1P: with >= 2 AICPU threads the last one is the core-less resolution
+    // thread (P). Cores partition across the remaining (aicpu_thread_num_ - 1)
+    // scheduler threads only, so P never owns a cluster and never polls a COND
+    // register.
+    p_thread_mode_ = (aicpu_thread_num_ >= 2);
+    p_thread_idx_ = p_thread_mode_ ? (aicpu_thread_num_ - 1) : -1;
+    active_sched_threads_ = p_thread_mode_ ? (aicpu_thread_num_ - 1) : aicpu_thread_num_;
     int32_t cluster_count = aic_count_;
 
     // Max clusters any single sched thread can hold: ceil(cluster_count / active_sched_threads_).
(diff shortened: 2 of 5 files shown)


ChaoZheng109:
Thanks for the review. Addressed in `d0f145aa`:

**Actionable**
- **P false-timeout guard (scheduler_dispatch.cpp)** — fixed. P's idle timeout now latches only when work is outstanding *and* `no_thread_owns_running_task()`, mirroring `resolve_and_dispatch`; a task that legitimately runs longer than the budget refreshes `last_progress_ts` instead of aborting the run. This was a real robustness bug (a long kernel / startup skew could false-latch `PTO2_ERROR_SCHEDULER_TIMEOUT`), so thanks for catching it.
- **P DFX/PMU teardown (scheduler_dispatch.cpp)** — fixed. P resets its per-thread swimlane state on entry and flushes swimlane / dump-args / PMU on exit (the AICore-keyed flushes iterate an empty core list since P owns none; the sched-phase-buffer flush is the one that matters).
- **SCHED_PROFILING `tasks_completed` (scheduler_dispatch.cpp)** — fixed. P now mirrors `completed_tasks_` into the profiling counter, since the scheduler threads' `completed_this_turn` no longer feeds it in P mode.

**Nitpicks**
- **Queue capacity** — now sized to `min(total_tasks, ring task window)`, no `1<<20` ceiling. The window already bounds in-flight completions, so capacity is never clamped below the working set and the backpressure spin stays a pure correctness backstop. Header comment softened accordingly.
- **`init` hardening** — added a power-of-two `debug_assert`. The `std::atomic` cursors already make `CompletedTaskQueue` (and therefore `SchedulerContext`) non-copyable/non-movable, so the accidental-copy double-free can't occur; called that out in the comment rather than adding redundant deleted members.

**Performance (this fixed commit `d0f145aa` vs 4S baseline).** Re-ran all five reference workloads on the same die (dev6), paired same-session, L2-swimlane device-execution span. The fixes are on init / idle-timeout / teardown / profiling paths only — never the hot SP-drain + resolve + dispatch path — so the 3S+1P benefit is intact (a same-binary fixed-vs-pre-fix check was within run-to-run noise across all five):

| workload | shape | 4S | 3S+1P (fixed) | Δ |
| pa Case1 (wide, 65536 tasks) | scheduler-bound | 29.68 ms | **24.44 ms** | **−17.6%** |
| dep-bound wide | parallel | 0.43 ms | 0.42 ms | −1.1% |
| qwen3-14B decode, 40L batch16 | compute-bound (util 98%) | 35.85 ms | 35.78 ms | −0.2% |
| dep-bound deep | serial / latency | 2.39 ms | 2.46 ms | +2.8% |
| deep_chain (512 serial) | serial / latency | 1.61 ms | 1.70 ms | +6.1% |

Direction is unchanged from the original write-up: scheduler-bound wide graphs speed up (−17.6% on pa here), compute-bound is neutral (no headroom), and a pure serial chain pays a few percent for the extra S→P hop that it cannot hide behind parallel work. All five pass (`--skip-golden` for the perf runs; correctness verified separately on `matmul`/`bgemm`/`paged_attention`/`vector_example`/`predicated_dispatch`).

ChaoZheng109:
## Follow-up: rigorous 100-round re-measurement (corrects the single-run numbers above)

The per-workload percentages I posted earlier were **single-run** and, for the
sub-3 ms dependency cases, dominated by noise. Re-measured properly:

**Method** — same physical card (a2a3 silicon, device 6, `task-submit`-locked),
100 rounds per workload, drop round 0 (warm-up), IQR-trim outliers (1.5×IQR
fence), then mean. Device time from the `runner_run.device_wall` STRACE marker.
`cv` = coefficient of variation of the trimmed set. 3S+1P is main @ `453ff71b`;
4S baseline is the pre-#1544 runtime (`96d02914`) rebuilt against the same HEAD.

| workload | scale | 4S | 3S+1P | Δ | cv |
| --- | --- | --- | --- | --- | --- |
| **paged_attention** (scheduler-bound) | 65792 tiny tasks | 28.016 ms | **25.013 ms** | **−10.7%** | 0.2% |
| qwen3 2-layer (compute-bound) | — | 36.222 ms | 35.952 ms | −0.7% | 0.1% |
| deep_big (serial chain, latency-bound) | 4096 serial | 19.048 ms | 19.091 ms | +0.2% | 0.3% |
| wide_big (parallel, many tasks) | 49152 tasks | 21.266 ms | 21.288 ms | +0.1% | 0.2% |

**The earlier small-case results were measurement artifacts.** At 1–2 ms the same
workloads showed deep_chain +9.4%, dep-deep +5.7%, dep-wide +3.8% — which read
like a regression. Scaled to the same ~20 ms magnitude as pa/qwen, those deltas
**collapse to ±0.2% (noise)**. The gap was a one-off fixed cost (one fewer
dispatch thread + a single S→P handoff per completion) that is a large fraction
of a 1 ms run and negligible in a 20 ms one — **not** a per-task cost that
accumulates with scale. There is no real regression on dependency-bound work.

**Why pa is the only real win, and it is not diluted:** the win requires the
*scheduler* to be the bottleneck — many tiny tasks, low per-task compute
(pa: util 67%). There, eliminating the multi-producer contention on the ready
queue / wake lists (single P producer) is worth −10.7%. My synthetic dep cases
use 128×128 matmul, so even at 49152 tasks the *cores* stay the bottleneck
(compute-bound, like qwen) → neutral. Heavier or longer tasks push **toward**
compute-bound, away from the regime 3S+1P helps.

**Net:** 3S+1P is a scenario-specific optimization — a solid, reproducible
−10.7% on scheduler-throughput-bound graphs (many small tasks), neutral on
serial-latency-bound and compute-bound graphs, and no measured regression at
realistic scale.

PR #1557, opened 2026-07-29
Fix: accept interior copy_to/copy_from ranges within a live allocation

### LINKED ISSUE #1537: [Bug] provenance guard rejects valid interior ranges in copy_to/copy_from
### Diagnosis

simpler runtime provenance guard - `copy_to()` / `copy_from()` only accepts an exact live allocation base, but the public runtime API has no supported base-plus-offset copy operation. This prevents valid partial updates to a persistent device allocation.

### Description

A caller can allocate a long-lived device buffer and safely update a page inside it, but the current provenance check rejects the page address as an "interior pointer" before the native copy is issued.

This surfaced while synchronizing newly-cleared pages of a resident DeepSeek V4 decode cache. The same issue is reproducible without serving or NPU hardware using the Python orchestration guard.

### Minimal reproduction

```python
base = orch.malloc(worker_id=0, size=64)

# A 16-byte write wholly inside the live 64-byte allocation.
orch.copy_to(worker_id=0, dst=base + 32, src=host_ptr, size=16)
# Current result: ValueError: device pointer ... is not a live allocation
```

The guard currently looks up provenance by the exact `(worker_id, ptr)` key, so it does not recognize `base + 32` as part of the allocation.

### Expected behavior

- `copy_to()` and `copy_from()` should accept a range wholly contained in a live allocation on the target worker.
- Wrong-worker, stale, and out-of-range pointers must still be rejected.
- `free()` must continue to require the exact allocation base.

### Suggested fix

Record allocation sizes in provenance. Add a range validator for copy operations:

```text
base <= ptr && ptr + nbytes <= base + allocation_size
```

Use overflow-safe arithmetic. Keep the existing exact-base validator for `free()` and APIs that require a base pointer.

CommDomain windows should similarly carry their live extent if copy APIs support window/buffer offsets.

### Regression coverage

A device-free mocked-orchestrator test should cover:

1. `malloc(64)` then `copy_to(base + 32, ..., 16)` succeeds.
2. `copy_to(base + 63, ..., 2)` is rejected as out of range.
3. The same address on another worker is rejected.
4. A copy after `free(base)` is rejected.
5. `free(base + 32)` remains rejected.

### Environment

| Component | Version |
|---|---|
| pypto-serving | `5ab2a0c` |
| pypto-lib | `b183933` |
| pypto | `937ebbb7` |
| simpler | `9922afdb` |
| PTOAS | `0.48` |
| Device / platform | Ascend A2A3, 8 devices |

### Additional context

The immediate serving workaround is a local `copy_to_offset(base, offset, ...)` helper that validates the base before applying the offset. That helper is not present in `origin/main`; a first-class runtime solution would avoid every caller needing to reimplement this pattern.

ChaoZheng109:
## Problem

Issue #1537: the child-pointer provenance guard keyed every allocation by its exact `(worker_id, base)` and admitted a copy only when the address hit that base. A partial update of a persistent device buffer — `copy_to(base + offset)` into one page of a resident cache (e.g. the DeepSeek V4 decode KV cache) — was rejected as though the interior address were unknown, because the guard stored no allocation extent to range-check against.

```python
base = orch.malloc(worker_id=0, size=64)
orch.copy_to(worker_id=0, dst=base + 32, src=host_ptr, size=16)
# before: ValueError: device pointer ... is not a live allocation
```

## Fix

Record each allocation's byte extent in provenance and validate copies by **containment** instead of exact base:

```
base <= ptr && ptr + nbytes <= base + extent
```

- `_ChildProvEntry` now carries `malloc_size` and maps each CommDomain allocation id to the extent of the window / buffer recorded at that base; `live_extent()` returns the widest live role.
- `copy_to` / `copy_from` (Worker L2 path and Orchestrator L3 path) call the new `_child_prov_require_live_range`, which scans the worker's live allocations and admits any range wholly contained in one. Python ints are unbounded, so the `ptr + nbytes` bound is exact (no wraparound).
- CommDomain windows / carved buffers carry their live extent so interior copies into them are validated too.

`free()` and kind4 dispatch are **unchanged** — they still require the exact base, since they act on the whole allocation, not an interior slice.

## Behavior

| Operation | Before | After |
| --------- | ------ | ----- |
| `copy_to(base + 32, ..., 16)` within a `malloc(64)` | rejected | **accepted** |
| `copy_to(base + 63, ..., 2)` (overruns) | accepted (guard blind to size) | **rejected** — out of range |
| same address on another worker | rejected | rejected |
| copy after `free(base)` | rejected | rejected |
| `free(base + 32)` | rejected | rejected |

## Tests

Device-free, mocked-orchestrator regression coverage in `tests/ut/py/test_worker/test_child_addr_guard.py` for all five cases in the issue, at both the L3 orchestrator and L2 worker entry points. `tests/ut/py/test_worker/` — 349 passed, 2 skipped (pre-existing).

close #1537

### DIFF
--- a/docs/user/reference/python-api.md
+++ b/docs/user/reference/python-api.md
@@ -62,8 +62,8 @@ else raises. Remote-worker and remote-memory calls require `level >= 4`.
 | ------ | ----- |
 | `malloc(size, worker_id=0) -> int` | Returns a device pointer as an integer |
 | `free(ptr, worker_id=0)` | |
-| `copy_to(dst, src, size, worker_id=0)` | H2D; `dst` is a device pointer, `src` a host address |
-| `copy_from(dst, src, size, worker_id=0)` | D2H; `dst` is the host address |
+| `copy_to(dst, src, size, worker_id=0)` | H2D; `dst` is a device pointer, `src` a host address. `dst` may be a `base + offset` interior address as long as `[dst, dst + size)` lies within one live allocation (partial update of a persistent buffer) |
+| `copy_from(dst, src, size, worker_id=0)` | D2H; `dst` is the host address. `src` may be an interior address whose `[src, src + size)` lies within one live device allocation |
 | `create_host_buffer(nbytes) -> HostBuffer` / `free_host_buffer(handle)` | Host-side buffer the device can reach |
 | `remote_malloc` / `remote_free` / `remote_copy_to` / `remote_copy_from` / `remote_export` / `remote_import` / `remote_release_import` | L4 only |
 
--- a/python/simpler/orchestrator.py
+++ b/python/simpler/orchestrator.py
@@ -491,7 +491,7 @@ class Orchestrator:
             return int(self._o.malloc(wid, sz))
         with self._worker._child_prov_lock:
             ptr = int(self._o.malloc(wid, sz))
-            self._worker._child_prov_record_malloc(wid, ptr)
+            self._worker._child_prov_record_malloc(wid, ptr, sz)
             return ptr
 
     def free(self, worker_id: int, ptr: int) -> None:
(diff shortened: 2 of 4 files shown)


ChaoZheng109:
@coderabbitai review

ChaoWao (approved):

PR #1379, opened 2026-07-16
feat: support pipelined orchestration and scheduling across tokens and requests

TaoZQY:
## Summary

This PR enables host-side orchestration and device-side scheduling to run as a pipeline.

- Allows orchestration for token N+1 to overlap with scheduling and execution for token N.
- Supports concurrent requests through overlapping `Worker.run()` calls. While one request is running, another request can be accepted and prepared.
- Adds per-token L3–L2 interaction instead of batching multiple tokens into one response.
- Ensures each generated token is delivered to the user immediately. The producer waits until the consumer receives the current token before continuing.
- Adds trace events for token receipt and user delivery.
- Simplifies the Perfetto swimlane to focus on token-level and request-level overlap.

## Profiling Results

The dual-request trace demonstrates:

- O/S overlap between adjacent tokens.
- Overlap between two independent requests.
- Two concurrent `Worker.run()` calls.
- Per-token L3 receipt and user delivery events.
- No batching of several generated tokens before returning them to the user.

img_v3_0213l_888affc9-dd0e-4b22-81c7-1e5913d8be5g

PR #1565, opened 2026-07-29
Fix: reserve NEXT_LEVEL workers for ready groups

### LINKED ISSUE #1551: [Bug] READY NEXT_LEVEL group can be overtaken indefinitely by later singles (no Tier-0 equivalent at L3)
### Platform

All / Unknown

### Runtime Variant

All / Unknown

### Description

A `NEXT_LEVEL` group and the per-worker `NEXT_LEVEL` singles live in separate,
unordered queues, and the scheduler gives a blocked group **no hold on its
already-idle targets**. A single that becomes READY *after* the group can
therefore take a worker the group is waiting for, and it can do so on every
scheduler round. A group of N targets only launches if all N happen to be idle
within the *same* round, so under sustained single traffic on its targets a
READY group can be delayed without bound. The larger the group, the less likely
the simultaneous-idle window ever opens — a TP-8 group needs 8 workers idle in
one round.

The expected policy is the one L2 already implements for `sync_start`: an
all-or-nothing cohort is a **strict Tier-0** that takes resources before any
regular task, and when it cannot fit it **waits and retries** rather than
yielding its share back.

`src/common/hierarchical/scheduler.cpp`:

```cpp
void Scheduler::dispatch_ready() {
    dispatch_next_level_group();     // :353
    dispatch_next_level_singles();   // :354
    dispatch_sub_ready();
}

void Scheduler::dispatch_next_level_group() {
    ...
    for (int32_t i = 0; i < group_size; ++i) {
        ...
        if (!worker->idle()) return;   // :420 — gives up, leaves no trace
        workers.push_back(worker);
    }
```

`dispatch_next_level_singles()` (`:437-458`) then walks every worker and
dispatches to any that is `idle()`. It has no knowledge that a group is waiting
on that worker, so it refills the exact worker the group needs.

Group-before-singles ordering therefore only holds *within one round*, and each
round is triggered by a single completion — so whichever worker just went idle
is refilled in the same round it became available.

### Steps to Reproduce

Two `NEXT_LEVEL` workers; a group targeting both; singles arriving on both after
the group.

```text
T1 -> worker0 (single)             # running
T2 -> group {worker0, worker1}     # group FIFO head, blocked on worker0
T3 -> worker1 (single)             # submitted AFTER T2

round A (T1 running):  group: worker0 busy -> return
                       singles: worker1 idle -> dispatch T3      <-- overtakes T2
round B (T1 done):     group: worker1 busy  -> return
                       singles: worker0 idle -> dispatch T4
round C (T3 done):     group: worker0 busy  -> return
                       singles: worker1 idle -> dispatch T5
...
```

`T2` launches only if worker0 and worker1 are idle in the same round. With
singles continuing to become READY on both targets, that never happens.

Note the queues are filled at **READY** time, not submit time (`PENDING tasks
live in task-slot state, not in a ready queue`), so a dependency chain feeding
either worker is enough to sustain the traffic — no adversarial submission
pattern is needed. Pipelined multi-token / multi-request orchestration (#1379)
produces exactly this shape.

The overtaking half is already an asserted behavior in the suite —
`tests/ut/cpp/hierarchical/test_scheduler.cpp:777`
`GroupSchedulerFixture.BlockedGroupDoesNotDispatchPartiallyOrReserveIdleWorker`:

```cpp
auto group = orch.submit_next_level_group(C(71), {group_a, group_b}, cfg, {0, 1});  // first
...
auto independent = orch.submit_next_level(C(72), ..., cfg, 1);                      // later
worker_b.wait_running();
EXPECT_EQ(worker_b.dispatched[0].callable_hash0, 72u);            // later task runs first
EXPECT_EQ(S(group.task_slot).state.load(), TaskState::READY);     // group still waiting
```

`LaunchableGroupPrecedesConflictingSingles` (`:817`) does cover group-first
dispatch, but only by holding `sched.loop_mutex()` while completing both
workers, forcing the scheduler to observe both idle in one round. Nothing covers
targets going idle one at a time with singles pending.

### Expected Behavior

**When a group is READY, singles pause and wait for it.**

Concretely: while the group-FIFO head is READY, its target workers are reserved;
`dispatch_next_level_singles` skips a reserved worker; the reservation is
released when the group launches. Reserved workers drain naturally as their
running tasks complete, the group launches, and singles resume.

This is L2's `sync_start` contract applied at L3. From
`src/a2a3/runtime/tensormap_and_ringbuffer/docs/RUNTIME_LOGIC.md:641-660`:

> the sync lane is drained as a strict **Tier-0** before the regular lane
> (`sync_start > MIX > C/V`)
>
> 1. **Single ownership** — a CAS on `sync_start_pending` (0 -> -1) makes drains
>    mutually exclusive; only one cohort drains at a time.
> 2. **All-or-nothing** — thread 0 checks `count_global_available >= block_num`
>    *before* staging. **If short, it advances the attempt and retries after
>    completions free cores.** A cohort is fully staged or not at all.
>
> Single ownership + all-or-nothing make the drain **deadlock-free** across
> multiple cohorts: at most one drains, and it fully stages or waits, so two
> cohorts can never each half-occupy the cluster set.

L3 groups are the same problem — all-or-nothing co-occupancy of N resources —
and should get the same answer. Restricting the reservation to the **group-FIFO
head** is the direct analogue of L2's single-ownership CAS: at most one reserver
exists, so the wait-for graph cannot contain a cycle.

Liveness argument for the reservation: a reservation only blocks *new* dispatch,
it never keeps a busy worker busy; the task currently running on a reserved
worker depends on nothing that is still queued; and a READY group has fanin 0,
so it depends on no queued single. Every reserved worker therefore reaches idle
in bounded time and the head group launches.

Precondition, which should be stated as an explicit invariant: a running task
must not block waiting on a not-yet-dispatched task at the same level. Splitting
a collective into independent singles already violates the intent of the group
API (nothing co-schedules independent singles), so this is making an existing
implicit requirement explicit rather than adding a new constraint.

### Actual Behavior

The group yields. `dispatch_next_level_group` returns without recording that a
group wants those workers, and `dispatch_next_level_singles` immediately gives
one away.

This is currently documented as intended. `docs/directed-next-level-scheduling.md`:

```text
Only the group FIFO head is examined, and a launchable group is tried before
conflicting singles. The runtime adds no fairness, aging, priority, or
reservation policy; callers choose worker sets that make acceptable progress.

## Non-goals
- No priorities, work stealing, rebinding, queue scanning, aging, quotas,
  preemption, starvation prevention, or partial group reservation.
```

Pushing "make acceptable progress" onto callers is not workable for the target
use case: a TP/EP group's worker set is fixed by the parallelism strategy, and
the caller has no way to stop unrelated singles on those same workers from
overtaking it.

Note the two properties bundled under "no partial reservation" in #1436 are
separable:

| property | meaning | keep? |
| --- | --- | --- |
| no partial **dispatch** | never launch only part of a group | yes — required |
| no partial **reservation** | never hold an idle worker for a waiting group | no — this is the defect |

L2 keeps the first and rejects the second; L3 currently rejects both.

### Git Commit ID

f0bd241cb6a2bf5bf54cf1d7bda3a7232d53c68c

### Host Platform

Linux (aarch64)

### Additional Context

Scope of the fix:

- `src/common/hierarchical/scheduler.cpp` — a reserved-worker set owned solely
  by `sched_thread_` (no lock needed); set by `dispatch_next_level_group` for
  the head group, honored by `dispatch_next_level_singles`, cleared on launch.
- `tests/ut/cpp/hierarchical/test_scheduler.cpp:777` — this test asserts the
  behavior being removed and must be rewritten (its name encodes the old
  contract). Add a real regression: targets going idle **one at a time** with
  singles pending on each, asserting the group launches within a bounded number
  of rounds.
- `docs/directed-next-level-scheduling.md` — drop `starvation prevention` and
  `partial group reservation` from Non-goals; document the head-only reservation
  rule and the same-level no-blocking-on-undispatched invariant.

Only the head group reserves. Head-of-line blocking between groups (a group
needing {0,1} blocks one needing {2,3}) is retained as-is; relaxing that is a
separate FIFO-semantics question and should not be bundled here.

Related: #1436 (introduced the current model), #1379 (pipelined orchestration —
first workload that sustains the traffic pattern), #1548 (Tier-0 `sync_start`
drain fast path at L2).

puddingfjz:
A blocked NEXT_LEVEL group now reserves every target from later singles until
the complete set becomes idle. The reservation is round-local and passed
directly from the group pass to the singles pass, preserving all-or-nothing
group dispatch while unrelated worker queues progress.

Scheduler regressions cover targets draining one at a time, unrelated-worker
progress, and consecutive groups with overlapping targets.

In-tree collectives whose ranks wait for same-level peers now submit the
complete cohort through `submit_next_level_group()`, matching the reservation
contract and preventing a running rank from waiting on a peer withheld behind
a ready group.

Fixes #1551

### DIFF
--- a/docs/directed-next-level-scheduling.md
+++ b/docs/directed-next-level-scheduling.md
@@ -44,14 +44,15 @@ routing operation.
 
 Each Scheduler iteration:
 
-- launches the group FIFO head only when every one of its target workers is
-  idle — all members or none, with no partial reservation;
-- dispatches the head of each idle worker's single FIFO;
+- checks the group FIFO head and either launches every member together or
+  reserves all of its target workers until the complete set is idle;
+- dispatches the head of each idle, unreserved worker's single FIFO;
 - dispatches READY SUB work through the existing free-selection path.
 
-Only the group FIFO head is examined, and a launchable group is tried before
-conflicting singles. The runtime adds no fairness, aging, priority, or
-reservation policy; callers choose worker sets that make acceptable progress.
+Only the group FIFO head reserves workers. A reservation blocks new singles on
+the group's targets while their running work drains; singles on other workers
+continue normally. The reservation is released when the complete group
+launches. Later groups remain behind the FIFO head and do not reserve workers.
 
 ## Invariants
 
@@ -62,13 +63,15 @@ reservation policy; callers choose worker sets that make acceptable progress.
   terminal state.
 - The NEXT_LEVEL worker set is fixed after `init()`; a worker lookup that fails
   after submit-time validation is an invariant violation, not a fallback.
+- A NEXT_LEVEL single does not wait for a not-yet-dispatched peer at the same
+  scheduler level. Work that requires concurrent placement uses the group API.
 - No queue or scheduler mutex is held while an endpoint executes.
 
 ## Non-goals
 
 - No SUB worker selection.
 - No priorities, work stealing, rebinding, queue scanning, aging, quotas,
-  preemption, starvation prevention, or partial group reservation.
+  preemption, or fairness beyond the group-head reservation.
 - No compatibility flags, environment variables, or macros.
 
 ## Related documents
--- a/docs/orchestrator.md
+++ b/docs/orchestrator.md
@@ -239,6 +239,10 @@ A group task is a single DAG node that executes in parallel on N workers.
 Each worker gets its own `TaskArgs`; the node only reaches COMPLETED when all
 N finish.
 
+Callers submit tasks that wait for same-level peers as one complete group.
+Submitting those members as independent singles can start one member before
+its peers are READY, leaving the running member unable to finish.
+
 ```cpp
 SubmitResult Orchestrator::submit_next_level_group(
     const CallableIdentity &callable, const std::vector<TaskArgs> &args_list,
@@ -259,10 +263,11 @@ routing.
 
 At dispatch time the Scheduler checks the group FIFO head and resolves every
 entry in `workers` to that exact stable worker ID. It dispatches only if the
-entire target set is idle; a blocked group reserves no partial worker set and
-does not cause a scan past the FIFO head. Each WorkerThread runs `worker->run`
-with its own `task_args_list[i]`. Completion remains aggregated at the group
-slot, so downstream consumers are released once after every member is terminal.
+entire target set is idle. A blocked group reserves all of its targets against
+new singles but does not cause a scan past the FIFO head. Each WorkerThread
+runs `worker->run` with its own `task_args_list[i]`. Completion remains
+aggregated at the group slot, so downstream consumers are released once after
+every member is terminal.
 
 ---
 
--- a/docs/scheduler.md
+++ b/docs/scheduler.md
@@ -54,8 +54,8 @@ while (true) {
         on_task_complete(item);
     }
 
-    dispatch_next_level_group();
-    dispatch_next_level_singles();
+    reserved_workers = dispatch_next_level_group();
+    dispatch_next_level_singles(reserved_workers);
     dispatch_sub_ready();
 
     if (stop_requested && all workers are idle) {
@@ -86,8 +86,8 @@ if target worker is idle and its FIFO is non-empty:
 ```
 
 There is no idle-worker search, rebinding, work stealing, or scan into another
-worker's queue. FIFO is independent per worker, so a busy worker A does not
-block READY work for worker B.
+worker's queue. Outside a group-head reservation, each worker FIFO progresses
+independently, so a busy worker A does not block READY work for worker B.
 
 ### Group tasks
 
@@ -103,13 +103,15 @@ if every target is idle:
     dispatch member i to target_worker_ids[i]
 else:
     leave the group at the FIFO head
+    reserve every target against single-task dispatch
 ```
 
-The check is all-or-nothing: a blocked group reserves no partial worker set.
-The Scheduler does not scan later groups. It continues to single-task queues,
-so the runtime adds no fairness, aging, priority, or reservation policy beyond
-trying a launchable group before singles in each iteration. Users are
-responsible for choosing worker sets that make acceptable progress.
+Dispatch remains all-or-nothing: no group member launches until the complete
+target set is idle. A blocked FIFO head reserves every target against new
+single-task dispatch while existing work drains. Singles on other workers
+continue normally, and later groups do not reserve workers because the
+Scheduler does not scan past the FIFO head. The reservation is released when
+the group launches.
 
 ## 4. SUB dispatch
 
@@ -184,6 +186,9 @@ The scheduling invariants are:
 5. Only the Scheduler calls `WorkerThread::dispatch`.
 6. Only one successful PENDING-to-READY transition enqueues a consumer.
 7. A group produces one aggregate DAG completion regardless of member count.
+8. A NEXT_LEVEL single does not wait for a not-yet-dispatched peer at the same
+   scheduler level; work that requires concurrent placement is submitted as a
+   group.
 
 ## 8. Related documents
 
--- a/examples/workers/l3/allreduce/README.md
+++ b/examples/workers/l3/allreduce/README.md
@@ -9,8 +9,8 @@ Walks through the full flow:
 3. **`worker.init()`** — fork chip children; lazy base-communication init
 4. **`orch.allocate_domain(...)`** — allocate a communication domain with a
    `CommBufferSpec` scratch window
-5. **`orch.submit_next_level(chip_handle, chip_args, cfg, worker=i)`** —
-   submit the allreduce task for each rank
+5. **`orch.submit_next_level_group(chip_handle, args_list, cfg, workers=...)`** —
+   submit every mutually waiting allreduce rank as one group
 6. **`worker.run(orch_fn, ...)`** — execute the DAG and golden-check against
    the known expected sum
 
--- a/examples/workers/l3/allreduce/main.py
+++ b/examples/workers/l3/allreduce/main.py
@@ -151,6 +151,7 @@ def run(device_ids: list[int], platform: str = "a2a3") -> int:
                 window_size=window_size,
                 buffers=[CommBufferSpec(name="scratch", dtype="float32", count=float_elems, nbytes=scratch_nbytes)],
             ) as handle:
+                args_list = []
                 for i in range(nranks):
                     domain = handle[i]
                     chip_args = TaskArgs()
@@ -167,7 +168,8 @@ def run(device_ids: list[int], platform: str = "a2a3") -> int:
                     )
                     chip_args.add_scalar(domain.domain_size)
                     chip_args.add_scalar(domain.device_ctx)
-                    orch.submit_next_level(chip_handle, chip_args, cfg, worker=i)
+                    args_list.append(chip_args)
+                orch.submit_next_level_group(chip_handle, args_list, cfg, workers=list(range(nranks)))
 
         print(f"[allreduce] running {nranks}-chip allreduce DAG...")
         worker.run(orch_fn, args=None, config=CallConfig())
--- a/examples/workers/l3/domain_rank_map/README.md
+++ b/examples/workers/l3/domain_rank_map/README.md
@@ -23,7 +23,9 @@ tail = workers [1, 2]        # chip 2 is in both
 After the inspection pass, each domain runs its own small allreduce — in its
 **own `worker.run()`**, so chip 2 never juggles two collectives at once. That
 separation is deliberate; see `dual_domain_overlap` for the case where both
-domains are live across the same DAG.
+domains are live across the same DAG. The ranks within each allreduce are one
+`submit_next_level_group`, so every peer that participates in the device
+barrier is dispatched as a complete set.
 
 ## Run
 
--- a/examples/workers/l3/domain_rank_map/main.py
+++ b/examples/workers/l3/domain_rank_map/main.py
@@ -210,13 +210,15 @@ def run(platform: str, device_ids: list[int]) -> int:
                 window_size=WINDOW_SIZE,
                 buffers=_scratch_buffers(),
             ) as handle:
+                args_list = []
                 for worker_idx in DOMAINS[domain_name]:
                     domain = handle[worker_idx]
                     args = TaskArgs()
                     args.add_tensor(make_tensor_arg(host_inputs[worker_idx]), TensorArgType.INPUT)
                     args.add_tensor(make_tensor_arg(outputs[domain_name][worker_idx]), TensorArgType.OUTPUT_EXISTING)
                     _add_domain_scratch(args, domain)
-                    orch.submit_next_level(allreduce_handle, args, cfg, worker=worker_idx)
+                    args_list.append(args)
+                orch.submit_next_level_group(allreduce_handle, args_list, cfg, workers=DOMAINS[domain_name])
 
         return _orch_fn
 
--- a/examples/workers/l3/dual_domain_overlap/README.md
+++ b/examples/workers/l3/dual_domain_overlap/README.md
@@ -20,7 +20,7 @@ through both and then computes on the results.
 | ------- | --- |
 | **Per-domain identity** | Chip 1 is rank 1 in `left` and rank 0 in `right`. The `workers` list order defines the dense rank, so the same chip legitimately holds two different ranks at once. |
 | **Domains allocated inside the orch function** | `with orch.allocate_domain(name=..., workers=..., window_size=..., buffers=[CommBufferSpec(...)])` — created and released within one orchestration, not configured on the `Worker`. |
-| **`submit_next_level_group`** | The affine stage submits one `TaskArgs` per member in a single call, with `workers=worker_indices`, instead of a loop of `submit_next_level`. |
+| **`submit_next_level_group`** | Both the peer-waiting allreduce and the affine stage submit one `TaskArgs` per member in a single call with `workers=worker_indices`. |
 | **Compute that depends only on its own domain's result** | Each affine task reads `reduce_out[domain][chip]`. The dependency is implicit — same `buffer.addr` as the reduce output — so `left`'s affine work can never consume `right`'s reduction. |
 
 ## Run
--- a/examples/workers/l3/dual_domain_overlap/main.py
+++ b/examples/workers/l3/dual_domain_overlap/main.py
@@ -208,6 +208,7 @@ def run(platform: str, device_ids: list[int]) -> int:
                     window_size=WINDOW_SIZE,
                     buffers=_scratch_buffers(),
                 ) as handle:
+                    args_list = []
                     for worker_idx in worker_indices:
                         domain = handle[worker_idx]
                         print(
@@ -221,7 +222,8 @@ def run(platform: str, device_ids: list[int]) -> int:
                             make_tensor_arg(reduce_out[domain_name][worker_idx]), TensorArgType.OUTPUT_EXISTING
                         )
                         _add_domain_scratch(args, domain)
-                        orch.submit_next_level(allreduce_handle, args, cfg, worker=worker_idx)
+                        args_list.append(args)
+                    orch.submit_next_level_group(allreduce_handle, args_list, cfg, workers=worker_indices)
 
             return _orch_fn
 
--- a/examples/workers/l3/ep_dispatch_combine/README.md
+++ b/examples/workers/l3/ep_dispatch_combine/README.md
@@ -56,6 +56,7 @@ scatters into `routed_y_buf` without clearing it first.
 | Concept | How |
 | ------- | --- |
 | **Three children under one `ChipCallable`** | `children=[(0, dispatch), (1, local_expert), (2, combine)]` — the integers are `func_id`s, matching the `rt_submit_aiv_task(0/1/2, …)` calls in the orchestration. Each child declares only the args it consumes; the orchestration signature is the union. |
+| **Collective group dispatch** | The complete per-rank callable is submitted as one NEXT_LEVEL group because its dispatch and combine phases wait on peer ranks. |
 | **Ordering without dependencies** | The three tasks run back-to-back because `rt_submit_aiv_task` dispatches in submission order, not because any tensor edge forces it. |
 | **Chaining through host-backed tensors** | `recv_x_out` / `recv_w_out` / `recv_count_out` are `OUTPUT_EXISTING` for dispatch and inputs to `local_expert`; `recv_y` likewise feeds `combine`. |
 | **A hand-laid-out window** | `SCRATCH_NBYTES` sums every region — counts table, two signal areas, three receive windows, the combine push destination, a third signal — and must match the `kOff*` offsets in the kernels. |
--- a/examples/workers/l3/ep_dispatch_combine/main.py
+++ b/examples/workers/l3/ep_dispatch_combine/main.py
@@ -548,6 +548,7 @@ def run(
                     )
                 ],
             ) as handle:
+                args_list = []
                 for i in range(nranks):
                     domain = handle[i]
                     print(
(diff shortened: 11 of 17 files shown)


ChaoWao:
## Review — approve with nits

Reviewed `321954ba...be82742b` (6 files, +93/−45; Core is 13 lines). The fix does exactly what #1551 specified and I could not break it. Notes below are ordered by what I think is worth acting on.

### What I verified rather than assumed

- **The regression test is a real barrier.** I rebuilt the PR's `test_scheduler.cpp` against the *pre-fix* `scheduler.{cpp,h}` in a scratch build dir. It fails 5/5 on exactly the two intended assertions:

  ```text
  test_scheduler.cpp:797: Failure  Value of: single_a_overtook       Actual: true   Expected: false
  test_scheduler.cpp:810: Failure  Value of: group_preceded_singles  Actual: false  Expected: true
  ```

  `LaunchableGroupPrecedesConflictingSingles` still **passes** pre-fix, so the new test is not a duplicate of it. Post-fix all 3 scheduler tests pass, and `ctest --repeat until-fail:200` is 200/200 clean in 37.8 s — not flaky on this box.
- **Reservation lifetime is sound.** `dispatch_ready()` is the only caller of both helpers and runs solely on `sched_thread_` (`scheduler.cpp:201,219`), so the set needs no synchronization; the unconditional `clear()` at function entry means a reservation can never leak into a later round, including across `stop()`/`start()`.
- **`#include <algorithm>` removal is safe** — `std::find` was its only use in the file.
- `pto_isa.pin` (`83d0131…`) untouched, no pto-isa include changes. No action.

---

### Should fix

**1. "Unrelated worker queues progress" has zero test coverage.**

That claim is in the PR body and now in three docs, but `GroupSchedulerFixture` has exactly two `NEXT_LEVEL` workers (`test_scheduler.cpp:640-641,657-658`) and the new test makes *both* of them group targets. A regression that over-reserves — say reserving every worker id instead of the head group's — would pass the entire suite. Adding a third worker and asserting that a single on the non-target worker still dispatches while the group is blocked is the assertion that separates "reserves the right workers" from "stalls the scheduler".

**2. `reserved_next_level_worker_ids_` carries two load-bearing invariants and no comment.**

`scheduler.h:101` declares it bare. Two non-obvious facts hold it up:

- The set is meaningful **only** between `dispatch_next_level_group()` and `dispatch_next_level_singles()` within one `dispatch_ready()`. Reorder those calls, or add a second caller of the singles pass, and reserved workers are silently starved forever.
- The `clear()` at `:430` is **not** redundant with the one at `:395`. It is what stops a second group in the same round, sharing a target with the first, from hitting `insert().second == false` and throwing `"duplicate target worker"` — which this file's own comment (`:344-350`) notes is fatal to the whole worker tree via `std::terminate`, not a per-task failure. A future reader is quite likely to delete `:430` as dead.

No test covers that path either — zero tests in the file submit more than one group.

Optionally, keeping the duplicate check on the local `workers` vector would decouple *reservation* from *malformed-group detection* and make `:430` obviously non-load-bearing:

```cpp
if (std::find(workers.begin(), workers.end(), worker) != workers.end()) {
    throw std::runtime_error("Scheduler::dispatch_next_level_group: duplicate target worker");
}
reserved_next_level_worker_ids_.insert(worker_id);
```

(The check is unreachable either way — `Orchestrator::validate_worker_eligibility` already rejects duplicate targets at submit time, `orchestrator.cpp:622`.)

### Consider

**3.** The `if (!group_preceded_singles)` branch (`test_scheduler.cpp:812-824`) is dead under correct behavior. It doesn't mask the failure (`EXPECT_TRUE` already records it), but it duplicates the assertions and reads as though both orderings were acceptable.

**4.** `single_a_overtook` is `worker_a.dispatched_count() > 1`, which is equally true if the group *partially* dispatched to the reserved-idle worker. The old test asserted no-partial-dispatch explicitly (`EXPECT_EQ(worker_b.dispatched_count(), 0)`); that now survives only implicitly under a misleading name. Partial group dispatch is a distinct and more serious failure.

**5. Pre-existing, out of scope, flagging because this PR is why people will look at that loop:** `Scheduler::run()`'s cv predicate includes `!ready_next_level_queues->empty()`, and `NextLevelReadyQueues::empty()` returns false whenever `group_queue_` is non-empty (`types.cpp:127-133`). A blocked group is deliberately left at the head, so the predicate is permanently satisfied — `wait()` returns immediately every iteration and the scheduler thread spins at 100% CPU until some worker completes. Not introduced here, and this PR shortens the window if anything.

---

## The one thing that needs a decision

This PR adds a new invariant to `docs/scheduler.md:189` and `docs/directed-next-level-scheduling.md:66`:

> A running task does not wait for a not-yet-dispatched task at the same scheduler level. Work that requires concurrent placement uses the group API.

Documenting it is the right call — it *is* the precondition the fix now depends on. The problem is that **the repo currently violates it in 10 places.**

### Why the invariant is load-bearing

`TWAIT` is an unbounded device-side spin. From `examples/workers/l3/allreduce/kernels/aiv/allreduce_onephase_kernel.cpp:112-126`:

```cpp
// Phase 2: device barrier — each rank notifies every peer that its
// stage-in is visible, then waits until every peer has notified us.
for (int peer = 0; peer < nranks; ++peer) { ... TNOTIFY(sig, 1, AtomicAdd); }
for (int peer = 0; peer < nranks; ++peer) { ... TWAIT(sig, 1, GE); }
```

A collective rank does not return until **every** peer rank has been dispatched. So "all N ranks co-resident" is a hard precondition — which is the definition of the group API's all-or-nothing dispatch. Yet every collective in the tree is submitted as N independent `submit_next_level()` calls.

### How that deadlocks after this PR

Two workers, a 2-rank collective `C` submitted as singles, a group `G` targeting `{w0, w1}`:

```text
t1  C_rank0 READY; G still PENDING, so no reservation
    → singles: w0 idle → C_rank0 dispatched
    ★ w0 enters TWAIT waiting for rank1. w0->idle() is false from here on, forever

t2  C_rank1 still PENDING (its own producer hasn't finished)

t3  G becomes READY, targets {w0, w1}
    → group: w0 busy → all_workers_idle = false → reserve {w0, w1}   ← new in this PR

t4  C_rank1 becomes READY, lands in w1's single FIFO
    → group: w0 still in TWAIT → reservation {w0, w1} re-derived
    → singles: w1 is reserved → continue                              ← C_rank1 never dispatched

t5  C_rank1 not dispatched → w0's TWAIT never returns → w0 never idle
    → G never launches → reservation never lifts → C_rank1 not dispatched
```

`t1` is the hinge: rank0 must get on-device *before* `G` becomes READY. Once it is in `TWAIT`, w0 is permanently non-idle, so the group can never assemble and the reservation is permanent. If rank0 becomes READY *after* `G`, it is simply withheld, w0 stays idle, `G` launches normally and everything drains — so this is timing-dependent, not deterministic.

**Before this PR the same sequence made progress**: at `t4` w1 was idle and unreserved, `C_rank1` dispatched, the barrier converged, w0 freed, `G` launched.

### It is not reachable today

The shape needs a group and a `TWAIT` single **coexisting in one DAG**, and nothing in the tree does that:

| site | group | collective single | coexist in one DAG? |
| :--- | :--- | :--- | :--- |
| `dual_domain_overlap/main.py` | ✅ `:238` | ✅ `:224` | ❌ — separate `worker.run` calls, and `Worker.run()` is `submit(...).wait()` (`worker.py:6115`), so the reduce DAG is fully terminal before the affine DAG is submitted |
| `dynamic_register/test_dynamic_register.py` | ✅ | ❌ no `TWAIT` | — |
| `test_l3_group.py` | ✅ | ❌ | — |
| `allreduce`, `ffn_tp_parallel`, `domain_rank_map`, `ep_dispatch_combine`, `st/worker/collectives/_helpers.py` | ❌ no group at all | ✅ | — |

**So this PR cannot deadlock anything in-tree.** I'm not asking to hold the merge on it. But three fairly natural next steps introduce it:

1. **Fusing `dual_domain_overlap`'s three `worker.run` calls into one DAG.** The example is literally named *overlap* but executes serially — making it actually overlap is the obvious follow-up, and its affine group targets are exactly the domain members.
2. **Adding an epilogue group to `ffn_tp_parallel`,** which already has the per-rank dependency skew that produces `t1`/`t2`.
3. **Pipelined multi-round orchestration (#1379)** — round N's group coexisting with round N+1's collective singles. This is the motivation #1551 itself cites.

### Why it would be nasty to debug

The scene at hang time: **reserved workers sitting idle** (resources free, nothing running), one worker at 100% AIV spinning in `TWAIT` (looks like it's computing), the group slot stuck at `READY` forever, and eventually `HandleTaskTimeout` killing `aicpu-sd` → surfaces as a plain `507018`. No allocator/dep deadlock detector fires, because it isn't a capacity problem — and per the triage table in `.claude/rules/running-onboard.md`, "only `HandleTaskTimeout` fired" routes you toward "long or stalled op", i.e. away from the real cause. The actual cause is one `unordered_set` on the host scheduler and the device log never mentions it.

---

## Proposed follow-ups

### A. Move collectives onto the group API

Criterion: **kernel contains `TWAIT` ⇒ must be a group; no `TWAIT` ⇒ stays a single.**

| priority | site | why |
| :--- | :--- | :--- |
| P1 | `tests/st/worker/collectives/_helpers.py:139,208` | highest leverage — one file backs allgather / allreduce / all_to_all / broadcast / reduce_scatter, **9 `TWAIT` kernels** |
| P2 | `domain_rank_map/main.py:219`, `dual_domain_overlap/main.py:224` | lets `domain_rank_map` drop its serialization workaround at `:230-232` (*"separate runs keep the overlapping chip 2 from juggling two collectives concurrently"*) |
| P3 | `allreduce/main.py:170`, `ffn_tp_parallel/main.py:236`, `ep_dispatch_combine/main.py:580` | contract consistency |
| **don't** | `ffn_tp_parallel/main.py:217` (AIC matmul, no `TWAIT`) | ranks are independent; a group would force them co-resident and cost the stage1/stage2 overlap |

The edit is mechanical — accumulate in the loop, submit once after:

```python
        args_list = []
        for i in range(nranks):
            chip_args = TaskArgs()
            ...
            args_list.append(chip_args)
        orch.submit_next_level_group(chip, args_list, config, workers=list(range(nranks)))
```

Two things worth noting, both in favour:

- **It turns a device spin into a scheduler wait.** Today in `ffn_tp_parallel`, rank *i*'s allreduce is dispatched the moment rank *i*'s matmul finishes, then burns AIV in `TWAIT` waiting for the others. As a group it stays READY-pending until every rank's `host_partial` is produced (`infer_deps` consumes the whole `args_list`), leaving the worker free meanwhile.
- **It removes a hidden dependence on FIFO ordering.** With two overlapping domains, correctness currently rests on each chip's FIFO order being mutually consistent; an inversion is a wait cycle. All-or-nothing dispatch removes the precondition structurally.

This should be its own PR — 10 sites, touches the ST suite, needs onboard validation. Not for #1565.

### B. A stall watchdog on the reservation

Time alone can't separate deadlock from a legitimately long op. The discriminator is **structural**: a reserved worker that is `IDLE` *and* has a non-empty single FIFO. All-targets-busy is an ordinary long-op wait and must not warn; an idle reserved target with queued work means the scheduler is actively withholding runnable work from an idle resource — exactly state `t4` above. Even when it isn't a deadlock, that state is worth reporting.

Sketch (host-side, so `codestyle.md` §7 doesn't bind, but the report is edge-triggered anyway — the dispatch loop runs far too often to log per round):

```cpp
// Well below the 45s op-execute timeout, so the report always precedes the
// 507018 that would otherwise be the only symptom.
constexpr auto kReservationStallWarnAfter = std::chrono::seconds(5);

void Scheduler::check_reservation_stall(TaskSlot group_slot) {
    const auto now = std::chrono::steady_clock::now();
    if (reserved_next_level_worker_ids_ != reported_reservation_) {   // set changed → progress
        reported_reservation_ = reserved_next_level_worker_ids_;
        reservation_since_ = now;
        reservation_stall_reported_ = false;
        return;
    }
    if (reservation_stall_reported_) return;
    if (now - reservation_since_ < kReservationStallWarnAfter) return;

    bool withholding = false;   // an idle reserved target with queued singles
    for (int32_t id : cfg_.ready_next_level_queues->worker_ids()) { ... }
    if (!withholding) return;

    reservation_stall_reported_ = true;
    std::fprintf(stderr, "[simpler][scheduler] NEXT_LEVEL group slot=%d has held its target "
        "reservation for >5s without launching. targets:%s\n"
        "  If a task on a BUSY target is waiting for a peer task queued on a reserved target, "
        "this will never resolve. The usual cause is a collective submitted as N separate "
        "submit_next_level() calls instead of one submit_next_level_group().\n", ...);
}
```

Hook it on the blocked return in `dispatch_next_level_group()`:

```cpp
        if (!all_workers_idle) {
            check_reservation_stall(slot);
            return;
        }
```

Needs a small `NextLevelReadyQueues::single_empty(worker_id)` accessor; `ReadyQueue::empty()` already exists. No reset code needed on launch — `clear()` makes the next round's set differ, which re-arms the watchdog. It must be a log, not a `throw`: an exception on `sched_thread_` is `std::terminate` for the whole tree.

Testable in the existing fixture by simply never calling `complete()` on the mock worker holding the busy target.

**I'd argue against making it self-healing** (releasing the reservation on timeout). That contradicts #1551's *"waits and retries rather than yielding its share back"*, brings the starvation bug back in intermittent form, and would permanently mask the submission bug that caused it. Better to hang, but hang legibly.

---

## Suggested split

- **This PR:** Should-fix 1 and 2 (third worker in the fixture, invariant comment). Both are small and in scope.
- **This PR or immediately after:** follow-up B, the stall watchdog — it's the diagnostic for the failure mode this PR introduces, and the two together are the complete change. If split, please leave an issue link here.
- **Separate PR:** follow-up A, the collective → group migration.

Nice work on this one — the core change is 13 lines, the docs landed in the same commit across all three affected files, and documenting the new precondition instead of leaving it latent is what let the rest of this analysis happen at all.

<sub>Cross-checked with `codex review` and `gemini`; both returned no findings. Codex independently reproduced the pre-fix failure and ran the 200× stress loop. Gemini's "no missing tests" conclusion was dropped — it read the new test's singles as landing on unrelated workers, but they're on the group's own targets (the fixture only has two workers), which is Should-fix 1.</sub>

puddingfjz:
@ChaoWao Thanks for the detailed review. I folded the changes that belong to
this fix into the same commit:

- made the blocked-head reservation round-local and pass it directly from the
  group dispatch pass to the single dispatch pass;
- added explicit coverage for no partial dispatch, unrelated-worker progress,
  and consecutive groups with overlapping targets;
- migrated the in-tree collectives whose ranks wait for same-level peers to
  submit each complete cohort through `submit_next_level_group()`;
- documented the corresponding caller and scheduler contracts.

I kept the stall watchdog and the pre-existing scheduler condition-variable
spin behavior out of this PR because they are independent diagnostics/runtime
policy changes and are not required for the reservation fix.

ChaoWao:
Approving — the scheduler change is sound, the collectives migration landed with
full onboard coverage (18 collectives ST + 6 L3 examples, all PASS on
`st-onboard-a2a3`), and cpput is clean at 26 tests × 100 iterations.

Two notes on the update itself before the follow-ups:

- Returning the reservation set from `dispatch_next_level_group()` and passing it
  as a `const&` is a better answer than the comment I asked for. It makes both
  invariants I was worried about unstateable rather than merely documented: the
  set is a local, so it cannot leak into a later round, and the load-bearing
  `clear()` no longer exists for a future reader to delete.
- Sharpening the doc invariant to *"a NEXT_LEVEL **single** does not wait for a
  not-yet-dispatched **peer**"* is the right scoping — group members legitimately
  wait on each other.

Recording the follow-ups here so they don't get lost. None of them block this PR.

---

### P1 — `ConsecutiveGroupsReserveOnlyBlockedHeadTargets` hangs instead of failing

The new test detects a regression correctly but then never exits. Reproduced by
building it against the pre-fix scheduler; confirmed with gdb:

```text
Thread 1: GroupSchedulerFixture::TearDown() → Scheduler::stop() → std::thread::join()  ← blocked forever
Thread 8: Scheduler::run()                                                              ← never returns
```

`ASSERT_EQ(worker_b.dispatched_count(), 2)` (`test_scheduler.cpp:889`) aborts the
test body with `worker_c` still running its task — nobody completes it. `run()`
exits only when `stop_requested && !manager.any_busy()`
(`scheduler.cpp:203-204`), so `any_busy()` stays true and the join never returns.

This is a pre-existing fixture hazard, but this test is the first to trigger it,
and it matters on the merged code too: the 500 ms deadlines are wall-clock, so on
a loaded runner **any** assertion failure becomes a 60-minute CI job timeout with
the failure messages buried in a truncated log, instead of a clean red test.

Draining the mocks in `TearDown` fixes it — `complete()` only sets a flag and
notifies (`test_scheduler.cpp:102-108`), so it is safe to call unconditionally:

```cpp
    void TearDown() override {
        // A failed ASSERT can abort the test body with mock workers still
        // running; Scheduler::run() exits only once no worker is busy, so
        // stop() would join forever.
        for (MockMailboxWorker *w : {&worker_a, &worker_b, &worker_c}) {
            for (int i = 0; i < 8 && w->is_running.load(); ++i) {
                w->complete();
                std::this_thread::sleep_for(std::chrono::milliseconds(1));
            }
        }
        sched.stop();
        manager.stop();
        allocator.shutdown();
    }
```

Cheapest item on this list, and it protects every future test in the fixture.

### P2 — Stall watchdog on the reservation

Converting the collectives removed today's trigger, but the failure mode this PR
introduces is still reachable for future code: a group whose target set covers a
worker holding a not-yet-dispatched peer of a task already spinning in `TWAIT`.
Two realistic ways it comes back — adding an epilogue group to `ffn_tp_parallel`
(which already has the per-rank dependency skew), or the pipelined multi-round
orchestration in #1379, which is the motivation #1551 itself cites.

The scene at hang time is actively misleading: reserved workers sitting **idle**,
one worker at 100% AIV in `TWAIT`, the group slot stuck at `READY`, and eventually
`HandleTaskTimeout` → a plain `507018`. No allocator/dep detector fires, so the
triage table in `.claude/rules/running-onboard.md` routes you toward "long or
stalled op" — away from the real cause, which is one `unordered_set` on the host.

Time alone can't separate a deadlock from a long op. The discriminator is
**structural**: a reserved worker that is `IDLE` *and* has a non-empty single FIFO.
All-targets-busy is an ordinary long-op wait and must not warn; an idle reserved
target with queued work means the scheduler is withholding runnable work from an
idle resource — worth reporting even when it isn't a deadlock.

Sketch, hooked on the blocked return in `dispatch_next_level_group()`:

```cpp
// Well below the 45s op-execute timeout, so the report precedes the 507018
// that would otherwise be the only symptom.
constexpr auto kReservationStallWarnAfter = std::chrono::seconds(5);
```

Edge-triggered — the dispatch loop runs far too often to log per round — and it
must be a log rather than a `throw`, since an exception on `sched_thread_` is
`std::terminate` for the whole worker tree. Needs a small
`NextLevelReadyQueues::single_empty(worker_id)` accessor; `ReadyQueue::empty()`
already exists. Testable in the existing fixture by never completing the mock
holding the busy target.

I'd argue against making it self-healing (releasing the reservation on timeout):
that contradicts #1551's *"waits and retries rather than yielding its share
back"*, brings the starvation bug back in intermittent form, and would
permanently mask the submission bug that caused it. Better to hang, but legibly.

**Related coverage gap.** Two collective groups with overlapping targets in one
DAG is currently exercised only by the cpput mock test
`ConsecutiveGroupsReserveOnlyBlockedHeadTargets` — there is no onboard equivalent.
If that is worth covering on hardware, a dedicated case under
`tests/st/worker/collectives/` is the right vehicle. Please don't retrofit it
onto `dual_domain_overlap`: that example documents *overlapping domain
membership* (chip 1 in two domains), and bending it into a scheduler test would
cost its clarity for a signal a purpose-built ST gives more cleanly.

### P3 — `Scheduler::run()` spins at 100% CPU while a group is blocked (pre-existing)

The cv predicate includes `!ready_next_level_queues->empty()`, and
`NextLevelReadyQueues::empty()` returns false whenever `group_queue_` is
non-empty (`types.cpp:127-133`). A blocked group is deliberately left at the
head, so the predicate is permanently satisfied — `wait()` returns immediately
every iteration and the scheduler thread burns a core until some worker
completes.

Not introduced by this PR (a blocked group was always left queued), and this PR
shortens the window if anything. Worth its own issue.

Subordinate note: now that `dispatch_next_level_group()` returns by value, the
blocked-group path allocates and frees an `unordered_set` **per spin iteration** —
exactly the path that spins. The empty-return case is free (a default-constructed
`unordered_set` does not allocate). Fixing the spin makes this moot, and
returning by value is the right call regardless; just noting the trade.

---

Suggested split: **P1** folded into any nearby test PR; **P2** on its own, since
it is the diagnostic for the failure mode this PR introduces; **P3** as an issue,
being independent and pre-existing.

ChaoWao (approved):

PR #1566, opened 2026-07-29
host_build_graph: fail-fast boot barriers on an out-of-range AICPU thread

ChaoZheng109:
## What

Make the `host_build_graph` device-boot barriers **fail fast** instead of hanging when an AICPU thread gets an out-of-range `thread_idx`.

Follow-up to #1553 — addresses the boot-barrier deadlock flagged in review there ([discussion](https://github.com/hw-native-sys/simpler/pull/1553#discussion_r3670678518)).

## Why

`run()` starts by validating `thread_idx` (from the affinity mapping, or the `thread_idx_++` counter). On an out-of-range index it used to `LOG_ERROR` and `return -1` **before** reaching either boot barrier:

- the boot leader then spins forever on `classify_arrived_ < aicpu_thread_num_` (the missing thread never arrives), and
- `finished_count_` never reaches `aicpu_thread_num_`, so `finished_` never publishes.

Either way the whole op hangs into the op-execute timeout (**507018**) — the generic host code that *masks* the real cause, so it reads as a runtime hang rather than the misconfiguration it is.

The path is only reachable via a fatal platform misconfiguration (a broken affinity mapping, or `run()` invoked more than `aicpu_thread_num_` times). It cannot happen under correct configuration — every thread gets a distinct valid index. This change is purely about turning that hang into a legible, fast failure.

## How

One shared abort, closing **both** boot barriers (`classify_arrived_` and `finished_count_`):

- The invalid thread no longer returns early. It publishes a `boot_abort_` flag, records the error, and falls through.
- Every boot wait (`classify_ready_`, `classify_arrived_`, `runtime_init_ready_`) also breaks on `boot_abort_`, so a missing participant can never leave the rest spinning.
- **Dispatch is skipped under `boot_abort_` on every thread.** A run missing a core-owning thread can never reach `is_completed()`, so merely un-hanging the boot barriers would just move the hang into the dispatch loop — the survivors must fail fast, not run a graph that can never complete.
- The invalid thread skips the index-keyed work (boot attach, its classify slice, `shutdown`) but still reaches the shared **finish barrier**, so `finished_count_` always reaches `aicpu_thread_num_` and `finished_` publishes.
- Every thread returns `-1` on abort, so the host reads no partial success.

`boot_abort_` is one-shot per run, reset in `deinit()` alongside the other boot flags. Boot/init path only — no dispatch-hot-path cost.

### Note on the review suggestion

The reviewer proposed "a shared abort state checked by every boot wait, and route participants through normal shutdown." This does exactly that, with one refinement worth calling out: an out-of-range thread is routed to the **finish barrier** (so `finished_` still publishes) but **not** through `shutdown(thread_idx)`, because it owns no core slice keyed by that index — calling `shutdown` with the bad index would be the same class of bug. Valid threads still `shutdown` normally.

## Scope

`a2a3` `host_build_graph` only.

## Test

`matmul`, `vector_example`, `predicated_dispatch` pass on a2a3 (device-locked) — the normal (all-valid-index) boot path is unchanged. The abort path is a misconfiguration guard; it is not exercised by the functional suite, which never produces an out-of-range index.

### DIFF
--- a/src/a2a3/runtime/host_build_graph/aicpu/aicpu_executor.cpp
+++ b/src/a2a3/runtime/host_build_graph/aicpu/aicpu_executor.cpp
@@ -102,6 +102,13 @@ struct AicpuExecutor {
     std::atomic<bool> classify_ready_{false};
     std::atomic<int32_t> classify_arrived_{0};
 
+    // Set by any run() invocation that cannot participate in the boot barriers
+    // (an out-of-range thread_idx). Every boot wait polls it so a missing
+    // participant cannot leave the rest spinning, and dispatch is skipped when it
+    // is set — a run missing a core-owning thread can never complete. One-shot per
+    // run, reset in deinit().
+    std::atomic<bool> boot_abort_{false};
+
     int32_t aicpu_thread_num_{0};
 
     // ===== Task queue state (managed by scheduler ready queues) =====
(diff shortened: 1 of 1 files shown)


ChaoZheng109:
Closing — the guard this hardens is unreachable by design, so the change adds boot-path complexity for a case the architecture already precludes one layer up.

Tracing the AICPU thread model settled it:

- Threads reach `run()` only after passing `platform_aicpu_affinity_gate_filter` (`src/common/platform/onboard/aicpu/platform_aicpu_affinity.cpp`). That filter assigns each surviving thread an `exec_idx` strictly in `[0, allowed_count)`, drops surplus threads (they `return 0` in `simpler_aicpu_exec` and never enter `run()`), and carries a forward-progress guard that **fills any missing exec slot by report order** precisely "so sched/orch roles are all present instead of timing out the AICore side." So the set of threads entering `run()` is exactly `allowed_count`, each with a distinct valid index.

- `thread_idx >= aicpu_thread_num_` can therefore only be reached if `allowed_count > aicpu_thread_num_` — i.e. the host handed down an affinity table wider than the runtime's thread count. That is a host-config self-inconsistency, not a runtime-reachable state, and even then the op-execute timeout reaps it.

- The scheduler-timeout precedent that suggested a hang here is not analogous: there all N threads are present, one trips the no-progress watchdog and sets the shared `completed_` latch, the rest exit their dispatch loops cooperatively, and every thread still reaches the finish barrier — so `finished_` publishes. The finish barrier only stalls when a participant is *missing*, and the affinity filter is what guarantees none is.

Net: the out-of-range `thread_idx` guard is defensive against a config invariant the affinity gate already upholds. Hardening it with a new atomic + abort-aware boot waits is negative-value on the boot path. Leaving the existing `return -1` as-is.

PR #1592, opened 2026-07-30
feat: add committed_device_memory() API (L1 byte counter + L2/L3 query)

### LINKED ISSUE #1457: [Feature] Expose simpler's committed NPU memory (rtMalloc pool is invisible to aclrtGetMemInfo)
## Summary

Downstream runtimes (we run **pypto-serving**, a vLLM-style Ascend serving stack on top of simpler) size their paged KV cache the way vLLM does: after loading weights and warming up, they compute the non-KV peak as `total_HBM − free_HBM` via `torch.npu.mem_get_info` (which wraps `aclrtGetMemInfo`) and hand the rest of HBM to the KV cache.

That scheme is only correct if `free_HBM` reflects **all** device memory already committed — including the ring-heap arena that simpler allocates itself via `rtMalloc` (`aclrtMalloc`). It does not appear to. As a result `total − free` under-reports real committed memory, the KV cache is over-provisioned, and the device runs out of HBM at runtime. We need either a simpler API to query its committed device memory, or confirmation of how the arena is reserved so we can account for it.

## How we size the KV cache (consumer side)

In `pypto_serving/model/qwen/npu_runner.py`, after the L3 worker is created, weights are uploaded, and a profile warm-up has run (so simpler's arena is already allocated):

```python
free_bytes, total_bytes = torch.npu.mem_get_info(f"npu:{device_id}")
peak_non_kv = total_bytes - free_bytes                 # ← the assumption lives here
kv_budget   = int(total_bytes * utilization - peak_non_kv)   # utilization default 0.90
num_pages   = max(kv_budget // bytes_per_page, 1)
```

We deliberately order everything (worker creation, static tensors, decode scratch, warm-up) **before** this measurement so that `peak_non_kv = total − free` is meant to capture weights + simpler arena + compiled buffers + scratch. The KV cache is then allocated into `kv_budget`; on alloc OOM we halve the page count and retry.

The entire estimate hinges on one assumption: **`total − free` is a true measure of committed HBM.**

## The problem

simpler's worker arena is its own `rtMalloc` (`aclrtMalloc`) allocation, sized on the order of `PTO2_RING_HEAP × 4 + 128 MB` (≈ 1.1 GB at the default `PTO2_RING_HEAP = 256 MB`). torch and simpler are **two independent NPU allocators**. `torch.npu.mem_get_info` (`aclrtGetMemInfo`) does not appear to subtract simpler's `rtMalloc` pool from `free`, so:

- `free` is over-reported → `peak_non_kv` is under-reported → `kv_budget` is too large → **the KV cache is allocated into HBM that simpler has already reserved but torch still believes is free.**

This produces two distinct failure shapes:

1. **Alloc-time (partly mitigated).** The torch KV allocation itself OOMs and our halve-and-retry shrinks the cache. This only catches the instantaneous torch-allocator failure.
2. **Runtime (NOT mitigated).** The KV allocation *succeeds* (torch thought the memory was free), then simpler uses / grows its arena during real serving and collides with the already-resident KV cache → a runtime OOM or device fault that the halve-retry cannot catch.

A second-order uncertainty makes this worse: **is the arena fully reserved at worker init, or grown lazily on demand?** If it is lazy, the `free` we measure after warm-up reflects only a fraction of the arena's eventual footprint, so the over-provisioning is larger and the runtime collision in (2) is more likely.

## How to reproduce / confirm

Run this probe on the device and report the `free₁ − free₂` delta:

```python
import torch, simpler   # simpler worker / allocator handle

free1, total = torch.npu.mem_get_info("npu:0")

# Trigger a known-size simpler rtMalloc allocation (e.g. init a worker,
# or allocate one arena slab of S bytes).
S = ...                      # bytes simpler is asked to commit
...                          # simpler allocation here

free2, _ = torch.npu.mem_get_info("npu:0")
print("delta:", free1 - free2, "expected:", S)
```

- If `free₁ − free₂ ≈ S` → `mem_get_info` sees simpler's `rtMalloc` pool (no problem from our side; we'd look elsewhere).
- If `free₁ − free₂ ≈ 0` → simpler's pool is invisible to `aclrtGetMemInfo` and this issue is the root cause.

We can run this and attach numbers, but the allocation internals are simpler's, so we'd like the simpler side to (a) confirm the expected behavior and (b) tell us whether `rtMalloc` here is plain `aclrtMalloc` or the virtual-memory path (`aclrtReserveMemAddress` + `aclrtMallocPhysical`), which `aclrtGetMemInfo` is known not to count.

## What we need from simpler

Any one of the following unblocks us; the first is the clean fix:

1. **An API to query simpler's committed device memory** — the sum of `rtMalloc`-backed pools / the resident arena size — so consumers can subtract it explicitly: `peak_non_kv = (total − free) + simpler_committed_bytes`.
2. **Confirmation** of whether `aclrtGetMemInfo` is *expected* to reflect simpler's allocations, and which CANN allocation API the ring heap uses (`aclrtMalloc` vs. the virtual-memory reserve/map path).
3. **Arena lifecycle guidance**: is `PTO2_RING_HEAP × 4 + 128 MB` reserved up-front at worker init, or grown lazily? This determines whether a warm-up-time `mem_get_info` measurement is ever safe to rely on.

## Environment

- Device: Ascend 910B/C
- CANN: `<fill in>`
- torch_npu: `<fill in>`
- simpler / pypto-lib: `<commit>`
- `PTO2_RING_HEAP`: default 256 MB (arena ≈ 1.1 GB)

## Workaround we can adopt in the meantime

Stop relying on `mem_get_info` for the simpler portion and subtract the arena explicitly:

```python
arena = int(os.environ.get("PTO2_RING_HEAP", 256 * 1024 * 1024)) * 4 + 128 * 1024 * 1024
kv_budget = int(total_bytes * utilization - peak_non_kv) - arena
```

This works only if the arena is reserved up-front; it also duplicates simpler's internal sizing formula and will silently drift whenever simpler changes how it sizes the arena — which is exactly why a real query API (item 1 above) would be far more robust.

### Motivation / Use Case

"`aclrtGetMemInfo` / `torch.npu.mem_get_info` does not reflect simpler's `rtMalloc` ring-heap arena; please expose simpler's committed device memory".

### Proposed API / Behavior

_No response_

### Alternatives Considered

_No response_

### Additional Context

_No response_

sunghajung6688:
L1: MemoryAllocator tracks committed bytes (ptr->size map + counter). L2: ChipWorker.committed_device_memory property + C ABI committed_device_memory_ctx. L3: control_committed_device_memory control op (mirrors control_malloc across the endpoint hierarchy) + Orchestrator::committed_device_memory + nanobind binding; Python _CTRL_COMMITTED_DEVICE_MEMORY, child-handler, Worker facade L3 branch, orchestrator wrapper. Lets downstream runtimes read simpler's authoritative HBM total (which aclrtGetMemInfo may undercount).

Tests: tests/ut/cpp/common/test_memory_allocator.cpp (C++), tests/ut/py/test_worker/test_committed_memory_sim.py (Python sim).

Closes https://github.com/hw-native-sys/simpler/issues/1457

### DIFF
--- a/python/bindings/task_interface.cpp
+++ b/python/bindings/task_interface.cpp
@@ -1532,6 +1532,14 @@ NB_MODULE(_task_interface, m) {
             "run creates and retires its own AICore stream, so this advances once per "
             "run; platforms whose runs use the persistent bootstrap pair report 0."
         )
+        .def_prop_ro(
+            "committed_device_memory", &ChipWorker::committed_device_memory,
+            "Total device HBM (bytes) currently committed by this worker's "
+            "MemoryAllocator (user tensors + pooled arenas + runtime buffers). "
+            "Excludes HCCL/VMM comm windows. 0 when not "
+            "initialized. Lets downstream runtimes subtract simpler's own HBM "
+            "from their cache budget (it may be invisible to aclrtGetMemInfo)."
+        )
         .def("malloc", &ChipWorker::malloc, nb::arg("size"))
         .def("free", &ChipWorker::free, nb::arg("ptr"))
         .def("copy_to", &ChipWorker::copy_to, nb::arg("dst"), nb::arg("src"), nb::arg("size"))
--- a/python/bindings/worker_bind.h
+++ b/python/bindings/worker_bind.h
@@ -315,6 +315,15 @@ inline void bind_worker(nb::module_ &m) {
             nb::arg("worker_id"), nb::arg("size"), "Allocate memory on next-level worker."
         )
         .def(
+            "committed_device_memory",
+            [](Orchestrator &self, int worker_id) {
+                return self.committed_device_memory(worker_id);
+            },
+            nb::arg("worker_id"),
+            "Total device HBM (bytes) committed by next-level worker's MemoryAllocator "
+            "(tensors + pooled arenas + runtime buffers)."
+        )
+        .def(
             "free",
             [](Orchestrator &self, int worker_id, uint64_t ptr) {
                 self.free(worker_id, ptr);
--- a/python/simpler/orchestrator.py
+++ b/python/simpler/orchestrator.py
@@ -494,6 +494,10 @@ class Orchestrator:
             self._worker._child_prov_record_malloc(wid, ptr, sz)
             return ptr
 
+    def committed_device_memory(self, worker_id: int) -> int:
+        """Total device HBM (bytes) committed by next-level worker *worker_id*'s ``MemoryAllocator``."""
+        return int(self._o.committed_device_memory(int(worker_id)))
+
     def free(self, worker_id: int, ptr: int) -> None:
         """Free memory on next-level worker *worker_id*."""
         wid, p = int(worker_id), int(ptr)
--- a/python/simpler/task_interface.py
+++ b/python/simpler/task_interface.py
@@ -1447,3 +1447,8 @@ class ChipWorker:
     def initialized(self):
         """Whether the underlying native worker has completed init."""
         return self._impl.initialized
+
+    @property
+    def committed_device_memory(self) -> int:
+        """Total device HBM (bytes) committed by this chip worker's MemoryAllocator."""
+        return int(self._impl.committed_device_memory)
--- a/python/simpler/worker.py
+++ b/python/simpler/worker.py
@@ -312,6 +312,7 @@ _BLOB_TENSOR_STRIDE = 128
 _BLOB_HEADER_BYTES = 8
 _CTRL_L3_L2_REGION_CREATE = 16
 _CTRL_L3_L2_REGION_RELEASE = 17
+_CTRL_COMMITTED_DEVICE_MEMORY = 18
 
 # Layout of the CTRL_COMM_INIT request shm.
 _COMM_INIT_HEADER = struct.Struct("<II")  # rank (u32), nranks (u32)
@@ -1679,6 +1680,8 @@ def _run_chip_main_loop(  # noqa: PLR0913, PLR0915 -- fork-child entry: every de
                 _handle_ctrl_l3_l2_region_create(cw, buf, chip_platform, l3_l2_region_store)
             elif sub_cmd == _CTRL_L3_L2_REGION_RELEASE:
                 _handle_ctrl_l3_l2_region_release(buf, l3_l2_region_store)
+            elif sub_cmd == _CTRL_COMMITTED_DEVICE_MEMORY:
+                struct.pack_into("Q", buf, _CTRL_OFF_RESULT, cw.committed_device_memory)
             else:
                 raise RuntimeError(f"unknown control sub-command {int(sub_cmd)}")
         except Exception as e:  # noqa: BLE001
@@ -5802,6 +5805,26 @@ class Worker:
             self._check_chip_worker_id(worker_id)
             assert self._orch is not None
             self._orch.free(worker_id, ptr)
+
+    def committed_device_memory(self, worker_id: int = 0) -> int:
+        """Total device HBM (bytes) committed by chip worker *worker_id*'s
+        ``MemoryAllocator`` (tensors + pooled arenas + runtime buffers; excludes
+        HCCL/VMM comm windows). Useful for downstream
+        runtimes to subtract simpler's own HBM from their cache budget.
+
+        Level 2 returns the in-process chip worker's committed bytes directly;
+        level 3 forwards a ``CTRL_COMMITTED_DEVICE_MEMORY`` query to the forked
+        chip child *worker_id* (sum across worker_ids for a multi-chip total).
+        """
+        with self._operation_lease("committed_device_memory"):
+            if self.level == 2:
+                assert self._chip_worker is not None
+                return int(self._chip_worker.committed_device_memory)
+            if not self._chip_shms:
+                raise NotImplementedError("committed_device_memory requires at least one forked chip worker")
+            self._check_chip_worker_id(worker_id)
+            assert self._orch is not None
+            return int(self._orch.committed_device_memory(worker_id))
 
     def copy_to(self, dst: int, src: int, size: int, worker_id: int = 0) -> None:
         """Copy *size* bytes from host *src* to chip worker *dst*."""
--- a/src/a2a3/platform/onboard/host/memory_allocator.cpp
+++ b/src/a2a3/platform/onboard/host/memory_allocator.cpp
@@ -34,7 +34,8 @@ void *MemoryAllocator::alloc(size_t size) {
     }
 
     std::scoped_lock<std::mutex> lk(mu_);
-    ptr_set_.insert(ptr);
+    ptr_size_map_[ptr] = size;
+    committed_bytes_ += size;
     return ptr;
 }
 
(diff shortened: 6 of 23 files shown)


sunghajung6688:
@coderabbitai review

ChaoZheng109:
### Docstring: "excludes profiling buffers" is platform-dependent and wrong for a2a3

*(Correcting my earlier comment — the runtime call chain I cited was inaccurate; the conclusion holds, but the mechanism and, more importantly, the a2a3-vs-a5 split are below.)*

`committed_device_memory()` docstrings (in `pto_runtime_c_api.h`, `python/bindings/task_interface.cpp`, `python/simpler/worker.py`) claim it *"Excludes HCCL/VMM comm windows and profiling buffers."* The "profiling buffers" half is **platform-dependent**, and wrong on a2a3:

**a2a3 — DFX buffers ARE counted.** `init_pmu` / `init_dep_gen` / `init_scope_stats` / `init_args_dump` in `src/a2a3/platform/onboard/host/device_runner.cpp` pass `alloc_cb = [this](size_t s){ return mem_alloc_.alloc(s); }` to each collector, so their device buffers go through the tracked `MemoryAllocator`.

**a5 — DFX buffers are NOT counted.** `src/a5/platform/onboard/host/device_runner.cpp` wires every collector (`l2_swimlane`, `dump`, `pmu`, `scope_stats`, `dep_gen`) to `prof_alloc_cb`, which is a **raw `rtMalloc`** ("a5 onboard goes directly through CANN runtime — no per-allocation tracking"). These bypass `mem_alloc_` entirely.

So the same docstring is correct for a5 and wrong for a2a3. Please either make the two platforms consistent, or state the exclusion is platform-specific.

**These allocations also happen at runtime, not just init (a2a3 path).** A background thread started in `ProfilerBase::start()` runs `mgmt_replenish_loop()` (`profiler_base.h`), which each iteration calls `replenish_recycled_pools()` → for any `(kind, shard)` whose `recycled_count < clamped_recycled_warm_target` → `BufferPoolManager::allocate_recycled_batch()` → `alloc_and_register_block()` → `ops_.alloc` (= `mem_alloc_.alloc` on a2a3). So on a2a3 the committed counter genuinely grows during serving as the recycled pools warm up to their (bounded) target — it is not a one-shot init cost.

**Consequence for the consumer (#1457).** `committed_device_memory()` is a live, point-in-time value, not a high-water-mark. On a2a3 it will keep rising after warm-up as DFX pools fill; on a5 it never reflects DFX at all. A KV-cache size computed from a single warm-up reading can therefore still under-account — the runtime-collision failure mode (2) in #1457. Worth deciding whether a `peak_committed_bytes()` high-water-mark (and platform-consistent DFX tracking) is in scope.

ChaoZheng109:
### Coverage scope of `committed_device_memory()` — points to be aware of in this PR (completeness can follow up)

`committed_bytes()` counts **only** device memory routed through the host-side `MemoryAllocator` (`mem_alloc_.alloc`). I audited every device-allocation site in `src/` and classified them. This PR is fine to land as the L1/L2/L3 plumbing, but the number it reports is a **subset** of simpler's real committed HBM. Flagging the gaps so they're a conscious decision, not a silent one — the missing pieces can be closed in a later PR.

**Counted today (through `mem_alloc_`):** the three arenas (`gm_heap` / `gm_sm` / `runtime_arena_pool` — the ~1.1GB in #1457), user tensors, chip-callable buffers, AICore reg buffers, host-side `api->device_malloc` mid-run pooled-arena growth, and **a2a3** DFX collectors.

**Not counted (bypass `mem_alloc_`):**

| # | Site | Memory | Recommendation |
|---|------|--------|----------------|
| 1 | `comm_hccl.cpp` (a2a3+a5): `aclrtMalloc` (CommContext) + `aclrtMallocPhysical`/`aclrtReserveMemAddress` (VMM comm windows / peer buffers) | Can be large under distributed serving | **Follow-up.** Do *not* force through `mem_alloc_` — the VMM reserve/map path is a different lifecycle than `rtMalloc` and HCCL windows have RDMA alignment/registration constraints. Add a `comm_committed_bytes_` counter in the comm layer and **sum it at the query layer**. |
| 2 | a5 `device_runner.cpp` `prof_alloc_cb` = raw `rtMalloc` for all DFX collectors | DFX pools | **Should fix (small, ideally in this PR).** a2a3 already routes DFX through `mem_alloc_`; a5 just needs the same capturing-lambda `alloc_cb`/`free_cb`. Removes the platform inconsistency and matches the docstring. |
| 3 | `load_aicpu_op.cpp` `DeviceBuf::aclrtMalloc` (inner-`.so` upload staging) | Transient, RAII-freed at init | **Leave out.** Counting a short-lived init buffer would make the number spike-then-drop and misrepresent steady state. |
| 4 | AICPU-side `aicpu_device_malloc` (`halMemAlloc`, device-tier) | Device-side dynamic alloc | **Out of scope (architectural).** The AICPU cannot reach the host `MemoryAllocator`; tracking would need a device-side counter mirrored back via shared memory. No runtime caller found in trb today. |
| 5 | `force_reset_device` 64-byte probe (`device_runner.cpp`) | 64 B, immediately freed | **Leave out.** Health-check poke, not real footprint. |

**Framing recommendation:** treat `committed_device_memory()` as *"sum of simpler's committed-HBM counters"* rather than *"one allocator's byte count"*. Today that sum has one term (`mem_alloc_`); #1 adds a comm term and #4 could add a device-side term later. For **this** PR the minimal asks are: (a) do #2 or at least document the a2a3/a5 asymmetry, and (b) have the docstring state that comm/VMM windows are excluded (already partly there) and that the value is a live snapshot, not a peak — so downstream (#1457) knows it must add comm accounting and headroom itself until the follow-up lands.

ChaoZheng109:
### Two smaller items still open on the latest revision

**1. (should fix) The L2 sim test doesn't verify the arena is counted — the one path #1457 actually cares about.**

`tests/ut/py/test_worker/test_committed_memory_sim.py`:

```python
baseline = worker.committed_device_memory()
assert isinstance(baseline, int)
assert baseline >= 0  # arenas are committed during init
```

The comment claims "arenas are committed during init", but `>= 0` is trivially true and passes even if `baseline == 0`. The later assertions (`after_alloc >= baseline + size`, `after_free <= after_alloc - size`) only verify delta-tracking of the *explicit* `malloc`/`free` — they'd pass just as well if the arena were invisible. So the core issue-driving path (simpler's own arena is included in the reported number) is unverified. Suggest asserting `baseline > 0` (the three arenas are committed by `init()`), or explicitly triggering an arena commit and asserting the counter rose.

**2. (consider) The L3 `NotImplementedError` message conflates "no children" with "aggregation not implemented".**

`python/simpler/worker.py`:

```python
if not self._chip_shms:
    raise NotImplementedError("committed_device_memory is not yet aggregated across level-3 forked workers")
self._check_chip_worker_id(worker_id)
return int(self._orch.committed_device_memory(worker_id))
```

The exception fires only when there are **no** forked chip workers (nothing to aggregate). When there *are* children — the case where aggregation would actually matter — it silently returns a **single** worker's bytes (the docstring already tells the caller to "sum across worker_ids"). So the wording is backwards relative to the behavior. Suggest something like `"level-3 committed_device_memory requires at least one forked chip worker"` (or, if a true aggregate is intended, actually sum across `worker_id`s here instead of leaving it to the caller).

ChaoZheng109 (approved):

PR #1594, opened 2026-07-30
Fix: preserve dep-gen tail buffers on the host

Little-oil:
## Problem

Dep-gen's management-to-collector host ready ring inherited the
device-side `kReadyQueueSize` (8), even though the host buffer-pool
capacity is larger. During the final stop-time drain, a burst of flushed
buffers can fill that host ring. An excess buffer is then retired before
collection, leaving reconciliation incomplete and preventing `deps.json`
from being emitted.

## Fix

- Add an optional `kHostReadyQueueSize` module trait for the host hand-off
  ring. It defaults to `kReadyQueueSize`, preserving existing collector
  behavior.
- Size `DepGenModule`'s host ready ring to `kHostPoolQueueSize`, covering
  the full retained-buffer bound during the final drain.
- Document the trait and extend the buffer-pool unit test to cover
  independent ready, done, and recycled capacities.

This fixes dependency-graph completeness only. Clean swimlane timing is
provided separately by PyPTO's graph/timing two-pass flow, whose timing
pass disables dep-gen. This change does not remove dep-gen timing
perturbation or address cross-rank launch skew.

## Testing

- `python -m pip install --no-build-isolation -e .` — passed.
- `python -m simpler_setup.build_runtimes --platforms a2a3` — passed for
  `host_build_graph` and `tensormap_and_ringbuffer`.
- `ctest --test-dir tests/ut/cpp/build -LE requires_hardware --output-on-failure`
  — 66/66 passed.
- `pre-commit run --from-ref origin/main --to-ref HEAD` — passed.
- A2A3 four-card integration of the same source diff before rebasing onto
  current `main` (`device=0,2,4,6`, EP=4, TP=4) — the L3 decode run exited
  successfully, and all `rank0..3/{d0,d1}` directories contained
  `deps.json`, L2 swimlane records, and merged swimlanes (8/8 each). This
  validates execution and DFX artifact completeness, not numerical
  correctness.

### DIFF
--- a/docs/dfx/dep-gen.md
+++ b/docs/dfx/dep-gen.md
@@ -48,9 +48,11 @@ inputs to each submit are captured and the graph is reconstructed afterwards.
 - **In-memory drain.** The host `DepGenCollector` (background mgmt
   thread, ProfilerBase machinery shared with PMU / L2 Perf / Tensor
   Dump) drains the ring into a `std::vector<DepGenRecord>` resident on
-  the runner. No `submit_trace.bin` lands on disk — the host already
-  has the records once the run ends, and going through the filesystem
-  would just be extra I/O.
+  the runner. Its host hand-off queue is sized independently from the
+  smaller device ready queue so a stop-time burst of flushed buffers can
+  drain without losing the tail. No `submit_trace.bin` lands on disk —
+  the host already has the records once the run ends, and going through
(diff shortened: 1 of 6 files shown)


Little-oil:
改成等待host ready queue有空闲，而不是直接调大容量

PR #1603, opened 2026-07-30
Fix: split A5 L2 record reserve and commit

yanghaoran29:
## Summary

- reserve the A5 L2 task-record slot before ACK and commit that exact slot after execution, so buffer rotation cannot move a record into a later generation
- avoid reading the AICore system timer when L2 swimlane DFX is disabled
- preserve A5 PMU per-task staging order and lazy L2 head resolution
- add a focused unit test for reserve/commit across buffer generation changes

Related to #1582 (missing item #5).

## Correctness

The reserved L2 record remains bound to the buffer generation that was active before ACK. PMU data is still published before FIN for AICPU consumption, while the L2 record is committed after FIN using the reserved pointer.

## Validation

- pre-commit on all changed files: passed
- focused C++ generation test: passed
- A5Sim L2-only, level 4 + dep-gen: 2 passed
- A5Sim PMU-only: 1 passed
- A5Sim L2 + PMU, level 4: 2 passed
- A5Sim DFX-off vector smoke: 1 passed
- A5 onboard L2 + PMU, level 4 + dep-gen: 2 passed
- full A5Sim tensor-map/ring-buffer suite: 72 passed, 2 skipped; 3 timing-marker tests failed in the sweep and passed when rerun in isolation (existing capture flake, outside this L2 path)

## Performance (`--enable-l2-swimlane 4`)

Same main commit (`ce81c973`), device 4, serial execution. Each variant used 10 independent processes with `--rounds 1 --enable-l2-swimlane 4`, because the scene-test runner disables L2 swimlane when `--rounds` is greater than 1. All 40 valid case runs per variant passed.

`Device` is the complete on-NPU AICPU `run_prepared.runner_run.device_wall` span emitted by `[STRACE]`; it excludes Python startup and host-side swimlane export/validation. Values below are arithmetic means over the 10 independent runs:

```text
mean Device = sum(Device for 10 runs) / 10
delta = (PR mean - main mean) / main mean * 100%
```

A negative delta means the PR is faster. For example, paged-attention Case1 is `(2457.97 - 2467.15) / 2467.15 = -0.37%`.

| Case | Metric | main mean | this PR mean | delta |
| --- | --- | ---: | ---: | ---: |
| paged_attention_unroll / Case1 | Device | 2467.15 us | 2457.97 us | -0.37% |
| paged_attention_unroll / Case1 | Effective | 2050.99 us | 2054.23 us | +0.16% |
| paged_attention_unroll / Case2 | Device | 1526.57 us | 1523.77 us | -0.18% |
| paged_attention_unroll / Case2 | Effective | 1109.29 us | 1106.36 us | -0.26% |
| manual_scope / Case1 | Device | 2461.55 us | 2406.86 us | -2.22% |
| manual_scope / Case1 | Effective | 2017.20 us | 1968.11 us | -2.43% |
| manual_scope / Case2 | Device | 1506.94 us | 1507.77 us | +0.06% |
| manual_scope / Case2 | Effective | 1075.78 us | 1079.31 us | +0.33% |

No level-4 performance regression observed; three cases are effectively flat and manual-scope Case1 improves by about 2.2%-2.4%.

### DIFF
--- a/src/a5/runtime/tensormap_and_ringbuffer/aicore/aicore_executor.cpp
+++ b/src/a5/runtime/tensormap_and_ringbuffer/aicore/aicore_executor.cpp
@@ -146,7 +146,7 @@ __aicore__ __attribute__((weak)) void aicore_execute(__gm__ Runtime *runtime, in
             //
             // Early-dispatch (src_payload != 0): receive_time stays at pickup —
             // before the doorbell wait — so it may precede the producer's end.
-            uint64_t receive_time = get_sys_cnt_aicore();
+            uint64_t receive_time = l2_swimlane_enabled ? get_sys_cnt_aicore() : 0;
 
             uint32_t task_id = reg_val;  // Decode: register holds task_id directly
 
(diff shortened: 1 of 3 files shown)


yanghaoran29:
@coderabbitai Reviewed the FIN/commit ordering against the shared AICore buffer protocol. Keeping the existing ordering intentionally: FIN does not publish the L2 buffer to the host. AICPU defers the old generation behind `s_aicore_pending_enqueue` and releases it only after ACK of the next generation’s first task; because AICore processes tasks serially, that ACK occurs after the prior record’s `commit + dcci + dsb`. This preserves record visibility while excluding DFX commit overhead from task completion timing, matching the current a2a3 lifecycle. The branch is also rebased onto current main.

ChaoZheng109 (approved):

PR #1610, opened 2026-07-31
fix(ring-buffer): make structural deadlock detection scope-aware

Leaf-Salix:
## Summary

This change makes the `tensormap_and_ringbuffer` structural-deadlock check use
scope identity: it reports an immediate fatal only for a cycle that the
blocked orchestrator itself makes impossible to break. Other no-progress
states remain covered by the existing timeout.

### Issue

The affected allocation paths are:

- `PTO2TaskAllocator::alloc()`, which reserves both a task-window slot and the
  task's output bytes from the ring heap and diagnoses task-window
  (`PTO2_ERROR_FLOW_CONTROL_DEADLOCK`) or heap
  (`PTO2_ERROR_HEAP_RING_DEADLOCK`) exhaustion;
- `PTO2DepListPool::ensure_space()`, which reserves dependency-list entries for
  a task with live fanin and reports `PTO2_ERROR_DEP_POOL_OVERFLOW` when it
  gives up.

Insufficient space does not immediately mean allocation has failed. Both paths
wait for the scheduler to make the ring's reclaim head `CONSUMED`, advance
`last_task_alive`, reclaim space, and then retry. This is normal backpressure.
They return failure only after another fatal is observed or their own
structural/timeout detector concludes that waiting cannot continue.

The task window, ring heap, and dependency-list pool hold different data, but
all three reclaim in task order using the same ring-local `last_task_alive`
watermark. Therefore the same head task controls whether any of these blocked
allocations can recover.

The merge-base structural detector turns that wait into an immediate fatal
when the reclaim-head task appears `COMPLETED`, has released all consumer
references, but has not released its scope reference. The same lifecycle-based
predicate is used for task-window exhaustion, heap exhaustion, and
dependency-pool exhaustion.

Those lifecycle fields do not identify which scope owns the reclaim head.
The head may belong to a closed child, sibling, or earlier serial scope whose
`scope_end` has already executed. The scheduler can still transition such a
head to `CONSUMED`, advance `last_task_alive`, and unblock the allocation.
Because the merge-base check reads the lifecycle fields while the scheduler is
progressing them, it can report a structural fatal immediately before that
valid reclaim.

The missing proof is whether this allocation wait prevents a still-pending
`scope_end` that is required for this particular head to become `CONSUMED`.
Lifecycle state alone cannot establish that relationship.

## Versions

- Fix: `b7ad6e97e383868ba34326caa3ff2ada733349f2`.
- Upstream merge base: `21e82d5e54eb8dbb922557a102afca8f58a89ea0`.
- PTO-ISA: `83d01313d9bfc247c4b7c8bcf969d1019f0d106f`.

## Fix

- Replace the lifecycle/refcount structural predicate in both
  `PTO2TaskAllocator::alloc()` and `PTO2DepListPool::ensure_space()` with an
  oldest-open-task identity check.
- Derive the oldest task retained by any open scope on the blocked ring in
  O(1) from the orchestrator's existing `scope_begins` and `scope_tasks`
  bookkeeping.
- Pass its slot pointer into both allocation paths.
- When task-window, heap, or dependency-pool space is unavailable, report an
  immediate structural fatal only if that pointer equals the reclaim-head
  slot. Otherwise, keep waiting for reclaim and use the existing approximately
  500 ms no-progress timeout as the fallback.
- Apply the same behavior to A2/A3 and A5
  `tensormap_and_ringbuffer`.
- Distinguish a proven open-scope cycle from a generic reclaim timeout in
  diagnostics and troubleshooting documentation.

No shared-memory fields, task-slot fields, locks, atomics, runtime hooks, or
linear scope scans are added.

### Definitions

- **Blocked ring** is the ring on which the current task-window, heap, or
  dependency-pool allocation cannot obtain space.
- An **open scope** has executed `scope_begin` but not its matching
  `scope_end`. `scope_tasks` records emitted task-slot pointers in order, and
  `scope_begins[depth]` records where each open scope starts in that array.
- **`H` (head)** is the slot addressed by the blocked ring's currently observed
  `last_task_alive`: `slot_states[last_task_alive & window_mask]`. It is the
  oldest slot that this ring has not reclaimed. The reclaim boundary cannot
  move past `H` until the scheduler makes `H` `CONSUMED`.
- **`O` (oldest open task)** is obtained from
  `begin = scope_begins[current_ring_id()]`. If `begin < scope_tasks_size`,
  then `O = scope_tasks[begin]`; otherwise `O = nullptr`. Because task pointers
  are recorded in emission order, `O` is the oldest task on the blocked ring
  retained by the shallowest open scope mapped to that ring, including its
  still-open descendants. Its scope reference cannot be released until the
  blocked orchestrator reaches the corresponding `scope_end`.

`H == O` compares task-slot pointers, not task IDs or independently sampled
lifecycle fields.

### Cases

The same `H/O` decision applies whether the unavailable resource is a
task-window slot, ring-heap space, or dependency-list capacity.

1. **Current-scope head: `H == O`**

   The current scope has emitted `H` and then blocks in one of the affected
   allocation paths on the same ring. The orchestrator cannot reach this
   scope's `scope_end`, so `H` cannot release its scope reference or become
   `CONSUMED`. The fix reports an immediate structural fatal.

2. **Open-parent head: `H == O`**

   A nested scope blocks while allocating on a ring also used by an open
   parent, and `H` is the parent's oldest retained task on that ring. The
   orchestrator cannot leave the child and reach the parent's `scope_end`, so
   `H` cannot become `CONSUMED`. The fix also reports an immediate structural
   fatal. This includes nesting deeper than the number of physical rings,
   where multiple scope depths share the deepest ring.

3. **Earlier sibling head on the same ring: `H != O`**

   Two sibling scopes execute serially on the same ring. The earlier scope has
   executed `scope_end`, but its head `H` has not yet transitioned to
   `CONSUMED`. The later scope is now open and blocks while allocating on that
   ring; `O` belongs to the later scope, so `H != O`. The scheduler may still
   consume `H` and unblock allocation, which makes an immediate structural
   fatal unsafe. The fix keeps waiting. If `H` never advances, the existing
   timeout reports the sustained no-progress deadlock.

## Why this proves a deadlock

When `H == O`, the blocked allocation prevents the orchestrator from reaching
the `scope_end` that would release `O`'s scope reference. `O` therefore cannot
become `CONSUMED`, while ordered reclamation cannot advance past the same slot
as `H`. The allocation is waiting for a reclaim that the blocked allocation
itself prevents, so this is a structural cycle and can be reported
immediately.

When `H != O`, the blocked orchestrator is not proven to pin `H`. The scheduler
may still consume it and advance the reclaim boundary. The allocator therefore
continues waiting; if the boundary remains stuck, the timeout reports sustained
no progress without misclassifying a transient scheduler state.

Scope depth maps one-to-one to rings until the deepest ring; deeper scopes
share that ring. For that ring, `scope_begins` selects the shallowest open
scope, so `O` still includes tasks retained by any open ancestor.

## Concurrency and ABA safety

`scope_tasks` and `scope_begins` have one orchestrator writer, and the scheduler
does not modify them. While blocked in allocation, that orchestrator cannot
close a scope or change `O`. Because `O` remains scope-retained, its slot cannot
be consumed and reused. A concurrent reclaim-boundary read may therefore delay
detection when `H != O`, but it cannot turn `H == O` into a false positive.

The dependency-list path uses the new task's `slot_state.ring_id`, while the
scope helper uses `current_ring_id()`; `prepare_task()` establishes that these
refer to the same ring before live-fanin wiring can block.

## Assumptions and unchanged concerns

The proof relies on existing runtime contracts: scope references are retained
until `scope_end`, each scope stack has one orchestrator writer, and a live task
slot is not reused. Those contracts would need to be revisited if early scope
release or multiple orchestrator writers are introduced.

This change does not alter the existing `int32_t` task-ID lifetime limit,
`current_task_index` publication ordering, or the allocator's possible
diagnostic-code overwrite by a concurrent fatal. None is used to establish
`H == O`.

## Regression coverage

Focused A2/A3 and A5 tests cover:

- an allocation blocked by the current open-scope head;
- an allocation blocked by an open ancestor when nested scopes share the
  deepest ring;
- a different-scope head that must use the timeout instead of an immediate
  structural fatal;
- both heap/task allocation and dependency-pool exhaustion.

These cases distinguish oldest-open-task slot identity from the old
`COMPLETED + consumers released + scope bit unset` approximation.
They use the normal scope-end and scheduler-consume lifecycle and add no
production test hooks or AICPU hot-path logging.

## Validation

- A2/A3 and A5 scope-deadlock and orchestrator-fanin tests passed.
- The focused scope detector and nested-scope regressions passed 20 repeated
  runs on both architectures.
- `git diff --check origin/main...HEAD` and modified-file pre-commit checks
  passed.

## Performance

Computing `O` is O(1): one ring lookup, one `scope_begins` load, one bounds
check, and at most one `scope_tasks` load. It adds no scan, lock, atomic
operation, or shared-memory layout change. The allocator compares slots once
per 1024 blocked spins; live-fanin computes `O` once before the dependency
pool's existing availability check.

An on-device ABBA benchmark compared the two validated commits:

- Platform: `a2a3`.
- Workload:
  `tests/st/a2a3/tensormap_and_ringbuffer/multi_round_paged_attention/`
  `test_multi_round_paged_attention.py`, `Case1`, `--skip-golden`.
- Sequence: 10-round base/fix warm-ups, then base/fix/fix/base with 500 rounds
  per measurement.
- All four measurements reported 500/500 valid device rounds.
- The measured fix commit was `19fca94a0ae0ccb86203173a1f03b8bfcf0bb846`.
  The final commit only amends its message; both commits have source-tree hash
  `b04e4b339cc260f55414c3ac9c3e43cd657cfa4a`.

| Run | Host (us) | Device (us) | Effective (us) | Orch (us) | Sched (us) |
| --- | --------: | ----------: | -------------: | --------: | ---------: |
| Base 1 | 1394.5 | 58.0 | 35.7 | 15.3 | 31.2 |
| Fix 1 | 1296.3 | 57.9 | 35.7 | 14.8 | 30.8 |
| Fix 1 vs. base 1 | -7.04% | -0.17% | 0.00% | -3.27% | -1.28% |
| Fix 2 | 1363.6 | 57.7 | 35.7 | 14.8 | 30.9 |
| Base 2 | 1307.6 | 57.1 | 35.3 | 14.6 | 30.7 |
| Fix 2 vs. base 2 | +4.28% | +1.05% | +1.13% | +1.37% | +0.65% |
| Base mean | 1351.05 | 57.55 | 35.50 | 14.95 | 30.95 |
| Fix mean | 1329.95 | 57.80 | 35.70 | 14.80 | 30.85 |
| Combined delta | -1.56% | +0.43% | +0.56% | -1.00% | -0.32% |

The paired deltas change direction between the two measurements, and the
combined device, effective, orchestrator, and scheduler deltas remain within
1%. The benchmark therefore shows no stable performance regression from this
change.

### DIFF
--- a/.claude/rules/running-onboard.md
+++ b/.claude/rules/running-onboard.md
@@ -172,16 +172,19 @@ for the signature that actually fired:**
 
 | device-log signature | mechanism | note |
 | -------------------- | --------- | ---- |
-| `FATAL: Task Allocator Deadlock` / `Provable head-of-line` | ring/heap or dep-pool **deadlock** (alloc can't reclaim) | AICPU detector: 500ms backstop (`PTO2_ALLOC_DEADLOCK_TIMEOUT_CYCLES`) or immediate structural `head_blocked_on_scope_end`. Real capacity/scope deadlock. |
+| `FATAL: Task Allocator Deadlock` / `FATAL: Dependency Pool Deadlock` / `FATAL: Fanin Spill Pool Deadlock` | an allocator or pool could not reclaim enough space | This header identifies the blocked resource, not the root cause. Classify it using the structural/timeout line that follows. |
+| `Provable head-of-line deadlock` | **proven open-scope structural deadlock** | TRB only: the reclaim head is the oldest task owned by an open scope on that ring. The blocked orchestrator cannot end that scope, so the head cannot become consumed. |
+| `No reclaim progress for ~500 ms` / `cannot reclaim space after ~500 ms` | allocator/pool **reclaim timeout** | The 500ms backstop (`PTO2_ALLOC_DEADLOCK_TIMEOUT_CYCLES`) exists in HBG and TRB. It proves prolonged lack of reclaim progress, not why progress stopped; check capacity, the dumped head, consumers and scheduler state. |
 | `Timeout (N cycles): producer/consumers ...` | **SPIN** wait on a specific producer/consumer | `pto_runtime2.cpp`. |
 | `HandleTaskTimeout` / `kill aicpu-sd` | **OS op-execute timeout** | STARS/tsdaemon, default 45s (`PLATFORM_OP_EXECUTE_TIMEOUT_US`). **A 45s kill ≠ deadlock** — the op was merely long or stalled. Raise this constant to measure true on-device duration. |
 | `log_stall_diagnostics` (cores idle + tasks `state=WAIT fanin 0/N` + `completed` frozen) | **forward-progress stall** | No dedicated detector — intermittent races (often contention-triggered) land here and are reaped only by the op-timeout above. |
 
-Decisive rule: **if a capacity/deadlock detector did NOT fire (counts of
-`Task Allocator Deadlock` / `Timeout (cycles)` are 0) and only
-`HandleTaskTimeout` did, it is NOT a proven deadlock or capacity bug** — it is a
-long/stalled op. A capacity exhaustion would trip its own detector; silence ⇒
-look for a race or just-too-slow, not "the ring is too small". For the
+Decisive rule: **if no allocator/pool fatal, producer/consumer
+`Timeout (N cycles)`, or scheduler timeout diagnostic fired and only
+`HandleTaskTimeout` did, it is NOT a proven deadlock or capacity bug** — it is
+a long/stalled op.
+A blocked allocator or pool would print its own fatal; silence means look for
+a race or just-too-slow, not "the ring is too small". For the
 host/device timing breakdown of a (completed) run, parse its `[STRACE]` markers
 with `python -m simpler_setup.tools.strace_timing <log> --rounds-table` (see
 `simpler_setup/tools/README.md`); for the per-thread `loops`/`tasks_scheduled`
--- a/docs/troubleshooting/device-error-codes/capacity.md
+++ b/docs/troubleshooting/device-error-codes/capacity.md
@@ -2,11 +2,20 @@
 
 ← [Device Error Codes](../device-error-codes.md)
 
-SCOPE_DEADLOCK, HEAP_RING_DEADLOCK and DEP_POOL_OVERFLOW all say the same thing —
-"some resource ran out" — without saying *which* resource or *which* scope. Do not
-guess at the ring sizes. Turn on `scope_stats`, which records the high-water mark of
-all four resources (task-window slots, heap bytes, dep-pool entries, tensormap
-entries) per `PTO2_SCOPE`:
+SCOPE_DEADLOCK, HEAP_RING_DEADLOCK and DEP_POOL_OVERFLOW all mean that an
+allocation could not obtain enough space. The adjacent device-log line determines
+how strong that diagnosis is:
+
+- `Provable head-of-line deadlock` is a structural proof: in TRB the reclaim head
+  is the oldest task owned by an open scope on that ring, and the blocked
+  orchestrator cannot end the scope that pins it.
+- `No reclaim progress for ~500 ms` or `cannot reclaim space after ~500 ms` is
+  the backstop. It proves that reclaim remained stalled, but not whether the root
+  cause is undersizing, a stuck consumer, or a stalled scheduler.
+
+Do not guess at the ring sizes from the error code alone. Turn on `scope_stats`,
+which records the high-water mark of all four resources (task-window slots, heap
+bytes, dep-pool entries, tensormap entries) per `PTO2_SCOPE`:
 
 ```python
 cfg = CallConfig()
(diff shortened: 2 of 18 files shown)


Leaf-Salix:
@coderabbitai review

Leaf-Salix:
Regarding the CodeRabbit maintainability nit: I am keeping the two checks local to their allocation paths. The task allocator derives the head slot from its private `slot_states_`/`window_mask_`, while the dependency-list pool derives it from the ring header. The only common operation left after those path-specific lookups is a null check plus pointer equality; extracting that into a production helper would widen the helper surface without centralizing an additional invariant. Both paths intentionally compare the same `oldest_open_task` slot identity, and the focused tests cover the structural and timeout decisions for each path.

ChaoZheng109 (approved):

PR #1643, opened 2026-08-03
Add: select A5 AICPU cores from runtime topology

yanghaoran29:
## Scope

This change adds topology-aware AICPU core selection for A5.

The runtime combines the authoritative device-side AICPU scheduler pool with Host CPU_TOPO metadata, classifies known FG, PG1, and PG2 topologies, and applies scenario placement for every supported active thread count from 2 through 5. Scheduler SMT availability is recorded as an independent topology property; it does not define a separate FG scenario.

A5 now supports at most five active AICPU threads. The independent launch population remains capped at 14 so the affinity gate can cover the device-visible scheduler pool.

## Changes

- Query device-side OCCUPY, PF_OCCUPY, and OS_SCHED before normal AICPU work.
- Prefer driver CPU_TOPO metadata and lock Host/AICPU wire layouts with compile-time ABI assertions.
- Match packaged fallback entries by SoC and every constraint declared by that entry. Ascend950PR_9599 requires exact OCCUPY `0x1f8` on either Host architecture; Ascend950PR_9579 also constrains Host architecture.
- Apply parameterized FG, PG1, and PG2 placement to manual and automatic active counts from 2 through 5.
- For UNKNOWN topology, use topology order when metadata is valid and CPU-ID order for OCCUPY-only metadata. Manual counts must be satisfied exactly; automatic mode may shrink from five but keeps at least one Scheduler and one Orchestrator.
- Keep explicitly generic verified fallbacks on the generic topology-aware packer.
- Cache successful merged topology probes for one DeviceRunner lifetime and invalidate them on recovery, reset, and finalize.
- Keep Runtime, DFX, affinity, and actual launch-count state consistent.
- Extend `aicpu-device-query` with A5 classification and diagnostic JSON output.
- Update design, hardware, runtime, and DFX documentation to the five-thread cap.

## Verification

- Final code review: passed with no remaining findings.
- `pre-commit run --from-ref upstream/main --to-ref HEAD`: passed.
- C++ unit tests: 80/80 passed; A5 topology/selection target contains 23 tests.
- A5 onboard Host/AICPU/AICore builds passed for both `host_build_graph` and `tensormap_and_ringbuffer`.
- `aicpu-device-query` Host build passed.

### DIFF
--- a/src/a5/platform/onboard/host/aicpu_topology_probe.cpp
+++ b/src/a5/platform/onboard/host/aicpu_topology_probe.cpp
@@ -14,10 +14,18 @@
 #include <dlfcn.h>
 
 #include <algorithm>
+#include <cctype>
+#include <cstddef>
 #include <cstring>
-#include <mutex>
+#include <filesystem>
+#include <fstream>
+#include <set>
+#include <sstream>
+#include <string>
+#include <tuple>
 #include <unordered_map>
 
+#include "common/platform_config.h"
 #include "common/unified_log.h"
 #include "common/acl_hal_device.h"
 
@@ -28,9 +36,7 @@ namespace {
 // HAL constants — replicated locally so we don't need to pull
 // driver/ascend_hal_base.h into the public header.  Values match
 // `tools/cann-examples/query/query.cpp`'s usage.
-constexpr int32_t kModuleAicpu = 1;
 constexpr int32_t kModuleSystem = 0;
-constexpr int32_t kInfoOccupy = 8;
 constexpr int32_t kInfoCpuTopo = 59;
 
 // DSMI constants for SOC_INFO + CPU_TOPO. Verified against
@@ -61,21 +67,21 @@ struct DsmiCpuTopo {
     DsmiSingleCpu cpus[kCpuTopoMaxLogical];
 };
 
+static_assert(sizeof(DsmiSingleCpu) == 16, "DsmiSingleCpu ABI size drift");
+static_assert(offsetof(DsmiSingleCpu, cpu_id) == 8, "DsmiSingleCpu::cpu_id offset drift");
+static_assert(offsetof(DsmiSingleCpu, hyperthread_id) == 11, "DsmiSingleCpu::hyperthread_id offset drift");
+static_assert(offsetof(DsmiCpuTopo, cpus) == 8, "DsmiCpuTopo::cpus offset drift");
+
+bool fill_dsmi_topo_from_json(
+    const char *soc_name, uint64_t occupy, DsmiCpuTopo &out, bool &out_generic_selection_only
+);
+
 // dlsym helpers — keep error reporting at WARN, callers fall back.
-using HalGetDeviceInfoFn = int (*)(uint64_t deviceId, int32_t moduleType, int32_t infoType, int64_t *value);
 using HalGetDeviceInfoByBuffFn =
     int (*)(uint64_t deviceId, int32_t moduleType, int32_t infoType, void *buf, int32_t *size);
 using DsmiGetDeviceInfoFn =
     int (*)(uint32_t device_id, unsigned int main_cmd, unsigned int sub_cmd, void *buf, unsigned int *size);
 using AclrtGetSocNameFn = const char *(*)();
-
-HalGetDeviceInfoFn load_hal_get_device_info() {
-    auto fn = reinterpret_cast<HalGetDeviceInfoFn>(dlsym(nullptr, "halGetDeviceInfo"));
-    if (fn == nullptr) {
-        LOG_WARN("aicpu_topology_probe: halGetDeviceInfo not found via dlsym");
-    }
-    return fn;
-}
 
 HalGetDeviceInfoByBuffFn load_hal_get_device_info_by_buff() {
     return reinterpret_cast<HalGetDeviceInfoByBuffFn>(dlsym(nullptr, "halGetDeviceInfoByBuff"));
@@ -109,19 +115,6 @@ const char *query_soc_name() {
     return fn();
 }
 
-bool query_occupy(uint32_t device_id, uint64_t &out_mask) {
-    auto fn = load_hal_get_device_info();
-    if (fn == nullptr) return false;
-    int64_t v = 0;
-    int rc = fn(static_cast<uint64_t>(pto::acl_to_hal_device_id(device_id)), kModuleAicpu, kInfoOccupy, &v);
-    if (rc != 0) {
-        LOG_WARN("aicpu_topology_probe: halGetDeviceInfo(AICPU,OCCUPY) rc=%d", rc);
-        return false;
-    }
-    out_mask = static_cast<uint64_t>(v);
-    return true;
-}
-
 bool query_cpu_topo(uint32_t device_id, DsmiCpuTopo &out) {
     std::memset(&out, 0, sizeof(out));
     if (auto fn = load_hal_get_device_info_by_buff(); fn != nullptr) {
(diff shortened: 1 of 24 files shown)


src/a5/platform/onboard/host/device_runner.cpp
```diff
@@ -137,14 +137,72 @@ int DeviceRunner::destroy_comm_stream(void *stream) {
...
+        launch_aicpu_payload(stream_aicpu_, &args, sizeof(args), host::KernelNames::TopologyQueryName, /*aicpu_num=*/1);
+    if (rc != 0) {
+        LOG_ERROR("AICPU device occupancy query launch failed: %d", rc);
+        return rc;
+    }
+    rc = aclrtSynchronizeStreamWithTimeout(stream_aicpu_, PLATFORM_STREAM_SYNC_TIMEOUT_MS);
+    if (rc != 0) {
+        LOG_ERROR("AICPU device occupancy query sync failed: %d", rc);
+        return rc;
+    }
```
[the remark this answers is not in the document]
yanghaoran29: Fixed — topology-query launch and synchronization failures now call recover_device_or_mark_unusable before returning, so the persistent AICPU stream cannot remain poisoned.

tools/cann-examples/aicpu-device-query/host/query_device_hal.cpp
```diff
@@ -494,13 +495,48 @@ int main(int argc, char **argv) {
...
-            kInfoName(requests[i].info_type), results[i].rc, (long)results[i].value, (long)results[i].value
-        );
+    if (json_output) {
+        pto::a5::AicpuDeviceOccupancy occupancy;
+        occupancy.os_sched = static_cast<uint64_t>(results[0].value);
+        occupancy.os_sched_valid = results[0].rc == 0;
+        occupancy.occupy = static_cast<uint64_t>(results[1].value);
+        occupancy.occupy_valid = results[1].rc == 0;
+        occupancy.pf_occupy = static_cast<uint64_t>(results[4].value);
+        occupancy.pf_occupy_valid = results[4].rc == 0;
```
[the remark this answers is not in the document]
yanghaoran29: Fixed — named constants now bind the OS_SCHED, OCCUPY, and PF_OCCUPY result positions to the request list and are used for both values and validity flags.

yanghaoran29:
@coderabbitai Addressed the remaining review suggestions:

- `topology_key` now returns a value tuple.
- topology JSON escapes the SoC name and maps selection policies with an exhaustive switch.
- the fixed 4S+1O thread count has a named constant.
- the driver library directory is an overridable CMake cache path.
- JSON fallback helpers are grouped at the end of `aicpu_topology_probe.cpp`.

Validation passed: the 16 A5 topology unit tests, pre-commit, and a separate clang-tidy run.

yanghaoran29:
Clarification on the A5 AICPU launch count and the constant 14:

- Onboard first launches a separate one-thread preflight kernel to read the device-side AICPU OCCUPY mask.
- The main-kernel launch count is the number of stable reachable CPU IDs after merging CPU_TOPO with that OCCUPY mask. If live CPU_TOPO is unavailable, the packaged JSON fallback is tried before the generic OCCUPY-only fallback. In the normal case, the resulting launch count is popcount(OCCUPY).
- PLATFORM_MAX_AICPU_THREADS_JUST_FOR_LAUNCH=14 is a compile-time capacity bound, not the default or an unconditional launch count. It also sizes the per-launched-thread phase and task-timing storage.
- For example, six stable reachable CPUs cause a six-thread main launch; the default policy keeps five active roles (4 Scheduler + 1 Orchestrator) and drops one. Fourteen stable reachable CPUs cause a fourteen-thread main launch and retain five active roles. More than fourteen is rejected instead of clamped.
- A5 simulation is different: it deliberately creates fourteen simulated AICPU threads and applies the gate afterward because there is no real device OCCUPY query.
- The value 14 corresponds to the expected widest 0x7ffe occupancy (7 physical CPUs x 2 SMT threads). Whether that SKU can complete a real fourteen-thread production launch still requires hardware validation; the code does not silently reduce a failed fourteen-thread launch.
- Automatic active count defaults to five and may shrink when fewer CPUs are stably reachable; explicit active counts are exact in the supported 2-5 range. The runtime warns both when stable reachable capacity is below five and when live CPU_TOPO is unavailable.

src/a5/platform/onboard/host/aicpu_topology_probe.cpp
```diff
@@ -237,6 +255,21 @@ bool derive_topology_from_occupy(const char *soc_name, uint64_t occupy, std::vec
...
+        cpu.phy_cpu_id = -1;
+        cpu.hyperthread_id = -1;
+        cpu.cluster_id = -1;
+        cpu.die_id = -1;
+        out_user_cpus.push_back(cpu);
+    }
+    return !out_user_cpus.empty();
+}
+
 bool probe_aicpu_topology(uint32_t device_id, std::vector<AicpuLogicalCpu> &out_user_cpus) {
```
ChaoZheng109: **死代码 — 旧 2 参 `probe_aicpu_topology` 已无生产调用者**

本 PR 后,这个旧的 2 参重载 `probe_aicpu_topology(uint32_t, std::vector<AicpuLogicalCpu>&)`(连同 `probe_aicpu_topology_uncached`,以及第 175–176 行的静态 `s_topo_cache` / `s_topo_cache_mu`)在生产路径已无调用者:`device_runner.cpp` 与 `aicpu-device-query` 工具都改用了新的 3 参重载 `probe_aicpu_topology(device_id, occupancy, topology)`。目前它仅被单测通过 `derive_topology_from_occupy` 间接触及,等于保留了一条带独立静态缓存的不可达路径,读者需要同时理解两个同名重载(其中一个已死)。

建议:删除旧重载 + `probe_aicpu_topology_uncached` + `s_topo_cache`/`s_topo_cache_mu`(单测若仍需覆盖 OCCUPY-only 回退,直接测 `derive_topology_from_occupy` / `enumerate_cpus_from_occupy` 即可);或至少在此处注释说明「仅为兼容/测试保留」。非阻塞。

src/a5/platform/onboard/host/aicpu_topology_probe.cpp
```diff
@@ -430,4 +583,756 @@ bool compute_allowed_cpus(
...
+        std::string key;
+        if (!parse_json_string(p, key)) return false;
+        skip_json_ws(p);
+        if (*p != ':') return false;
+        ++p;
+        skip_json_ws(p);
+        if (*p != '{') return false;
+
+        if (key != soc_name) {
+            // Skip this soc object by brace depth.
```
ChaoZheng109: **手写 JSON 解析器:裸数花括号 + 不解码转义,受控文件之外不鲁棒**

这里(以及下方约第 1165 行 `obj_end` 的计算)按 `{`/`}` 逐字符数深度,但**不跳过字符串内容**;`parse_json_string`(第 999 行)也**不解码转义**(`"\n"` 会被解析成字母 `n`)。当前 `aicpu_cpu_topo_fallback.json` 恰好不含「字符串内落单的花括号」与转义,所以能正确解析。

风险:这是数据驱动的,编译器/linter 拦不住。将来有人给 fallback 文件加一个带落单花括号的说明字段(如 `"_comment": "mask 0x{"`)或带转义的值,`depth` 计数就会错位 → 跳过 soc 冲过头 / `obj_end` 算错 → 约束键(host_arch/occupy_mask)误判、或 `cpus` 数组被截断 → 读出错误拓扑 → 选错核,且难排查。

建议(从省到彻底):① 顶部加一句契约注释,声明「仅解析仓库内受控文件,不支持字符串内花括号与 `\` 转义」;② 更稳:把这两处裸数括号改成字符串感知——遇到 `"` 就用 `parse_json_string` 跳过整段再继续数。非阻塞。

src/a5/platform/onboard/host/aicpu_topology_probe.cpp
yanghaoran29: Fixed — removed the unused two-argument probe, its uncached implementation and per-device cache, plus the HAL OCCUPY helpers that only served that dead path.

src/a5/platform/onboard/host/aicpu_topology_probe.cpp
yanghaoran29: Fixed — object-boundary scanning now skips complete JSON strings through parse_json_string, so escaped quotes and braces inside documentation fields do not affect depth. The packaged fallback includes a regression field that failed with the old scanner and now passes.

ChaoZheng109 (approved):

Issue #1425, opened 2026-07-21
[Performance] Provisioning the PTO-ISA async-SDMA workspace degrades AICore fault recovery ~1000x (0.3s → ~306s) on a2a3


ChaoWao:
### Platform

a2a3 (Ascend 910B/C hardware)

### Runtime Variant

tensormap_and_ringbuffer

### Summary

Provisioning the PTO-ISA async-SDMA scratch workspace on an a2a3 device makes **AICore fault recovery ~1000× slower** (~0.3 s → ~306 s), and also changes the surfaced fault error code (507046 → 507015).

`SdmaWorkspaceManager::Init()` creates **48 STARS streams** via `aclrtCreateStreamWithConfig(..., ACL_STREAM_DEVICE_USE_ONLY)`. Those streams are device-visible and sit in the device **fault/sync domain**. Once any AICore task faults and poisons the card, tearing the 48 streams down — whether by `aclrtResetDeviceForce`/`rtDeviceReset` at finalize or by `aclrtDestroyStream` — **blocks ~306 s**. Merely *provisioning* the workspace (no SDMA op ever submitted) is enough to trigger this on the *next* AICore fault, because the cost is in creating/holding the streams, not using them.

This is latent on `main` (the SDMA workspace is provisioned only on the HCCL comm path today), and was surfaced while making single-card provisioning eager in PR #1406 (branch `feat-sdma-prefetch-workspace-addr`), which pointed the `tests/st/aicore_op_timeout` / `tests/st/runtime_fatal_codes` fault-injection STs at an SDMA-provisioned device.

### Git Commit ID

simpler `upstream/main` `4dbcc9c8`; pto-isa pin `83d01313` (root cause is in `SdmaWorkspaceManager::CreateStarsStreams`). Reproduced on PR #1406 head.

### CANN Version

cann-9.0.0

### Host Platform

Linux (aarch64)

### Reproduction

1. On a2a3 hardware, get the PTO-ISA async-SDMA workspace provisioned on a device (the eager per-device path in PR #1406, or the HCCL comm path on `main`).
2. Run a kernel that deliberately hangs an AICore op and let the timeout chain reap it, e.g.:
   ```
   task-submit --device auto --device-num 1 --run \
     "python -m pytest tests/st/aicore_op_timeout/test_aicore_op_timeout.py --platform a2a3 --device \$TASK_DEVICE"
   ```
[...]

PR #1664, opened 2026-08-03
Fix: avoid remote teardown after fatal AICore errors

sunkaixuan2018:
## Summary

- Preserve the existing stream error-code and DFX contract for ordinary a2a3/a5 runs; apply `ACL_STOP_ON_FAILURE` only to the a2a3 per-run stream when a real SDMA workspace is provisioned.
- Publish fatal state before completion so every teardown thread observes the poisoned-device state, with one thread owning the emergency shutdown.
- On fatal teardown, stop host collectors, force-reset the card, and abandon device-backed handles without destroy/free/unregister calls that can block in `DEV_RUNNING_DOWN`.
- Keep the full resource-release path unchanged for healthy teardown, and add focused regression coverage for ordering, reset budget, and no-device-touch cleanup.

The a2a3 SDMA fatal path waits `max(10 seconds, configured op-execute timeout + 5 seconds)` before reset. This is an empirical workaround for CANN 9.0.0 and driver 26.0.rc1, not a portable runtime completion fence; it is deliberately limited to workers that actually provision the SDMA workspace.

### Reset budget follows the stream population

`force_reset_device()` drains the card before resetting it and returns 0 only when its post-reset probe confirms a usable generation, so a second pass runs against a settled card and can recover a poison the first could not. Ordinary poison therefore keeps a bounded three-attempt budget on both a2a3 and a5.

A Worker holding the 48 CP-process SDMA streams is the exception and gets a single attempt: there a non-confirming reset has already blocked on the driver's 150/300-second remote-event timeout, so a retry multiplies that wait without adding a completion condition.

## Testing

- Passed all pre-commit checks; Windows skipped clang-tidy because the onboard sources require Linux/CANN headers, then the changed files passed the repository clang-tidy runner on myserver.
- Built all a2a3/a5 `host_build_graph` and `tensormap_and_ringbuffer` runtimes with CANN 9.0.0.
- Passed the non-hardware C++ unit tests, including the fatal publication, reset-budget, arena-abandon, and allocator-abandon tests.
- Passed a2a3 hardware AICore-timeout regression with DFX enabled:
  - ordinary non-SDMA worker: passed in 8.4 s and retained the non-`507015` error contract;
  - real-SDMA worker: passed in 19.2 s, including the scoped 10 s workaround, with bounded `Worker.close()`.
- Passed a2a3 hardware diagnostic-contract cases for scheduler timeout (`sub_class=S1`) and tensor wait timeout (`orch_error_code=8`).
- a5 hardware coverage is provided by the PR's `st-onboard-a5` CI job; local myserver hardware is a2a3, while the a5 runtime build and a5 C++ unit suite passed there.

## Scope

Mitigates #1425 — this bounds fatal teardown to roughly 10–30 s and keeps ordinary Workers on the fast path. It does **not** restore SDMA-Worker fault recovery to ~0.3 s: the root cause (CANN exposes no retirement fence for CP-process SDMA streams) is unresolved and remains deferred pending a CANN runtime-and-driver fix, so #1425 should stay open against that dependency.

### DIFF
--- a/docs/investigations/2026-07-a2a3-sdma-fault-teardown.md
+++ b/docs/investigations/2026-07-a2a3-sdma-fault-teardown.md
@@ -6,6 +6,63 @@
 risk and ordinary Workers are unaffected; full recovery inside an SDMA-enabled
 Worker after a fault is deferred pending a CANN runtime-and-driver fix
 
+## 2026-08 package-specific follow-up
+
+PR #1664 adds a narrower containment for the installed CANN 9.0.0 and driver
+26.0.rc1 combination. Only run streams belonging to a Worker that actually
+provisioned the SDMA workspace use stop-on-failure mode. Fatal close on such a
+Worker then waits for `max(10 seconds, configured op timeout + 5 seconds)`,
+attempts one force reset, and forgets the failed generation's host handles
+without issuing per-resource destroy/free calls.
+
+This is an empirically validated workaround, not an SDMA retirement contract.
+CANN still exposes neither a completion fence nor a portable upper bound for
+the final CP-process stream release. Ordinary a2a3 Workers and all a5 Workers
+therefore keep normal stream mode, their existing error/diagnostic behavior,
+and no added handoff delay. The hardware regression test keeps the
+package-specific workaround visible by covering real SDMA provisioning followed
+by an AICore fault and a bounded `Worker.close()`.
+
+### Reset attempts are budgeted per stream population, not globally
+
+`force_reset_device()` drains the card before resetting it and returns 0 only
+when its post-reset probe confirms a usable generation, so a second call runs
+against a settled card and can recover a poison the first could not. Ordinary
+poison therefore keeps a bounded three-attempt budget on both a2a3 and a5.
+
+A Worker holding the 48 CP-process SDMA streams is the exception and gets a
+single attempt: there a reset that does not confirm has already blocked on the
+driver's 150/300-second remote-event timeout, so a retry multiplies that wait
+without adding a completion condition.
+
+Collapsing both populations onto a single attempt regresses ordinary
+fault-injection recovery — the poisoned card stays poisoned for the next
+process that lands on it, which surfaces as unrelated tests failing with
+507018/507046 on the devices a fault-injection test just used.
+
+### Emergency shutdown broadcasts in parallel but still joins the cores
+
+The AICore exit acknowledgement is what leaves a card usable for the next
+process: only after a core confirms it stopped does the AICPU quiesce that
+core's register block, putting the dispatch register back to idle and closing
+the fast path. Returning from emergency shutdown while cores are still running
+leaves the card poisoned past the host's device reset, and the next process on
+that device fails at launch.
+
+So the fatal path signals every handshake'd core first and joins them second,
+rather than signalling and waiting one core at a time. Cores drain
+concurrently and the per-core quiesce is preserved.
+
+The join takes **one deadline for the whole group**, not one timeout per core.
+That distinction is load-bearing on the fatal path, where every core is
+typically dead: the onboard deinit timeout is 1 second, so a per-core deadline
+costs a second per unresponsive core and pushes an AICore-timeout run past a
+10-second budget, while a shared deadline caps the whole group at one second.
+
+The wait is an on-device poll of the core's `COND` register, so it costs no
+host or remote operation and is unrelated to the CANN SDMA teardown problem
+above.
+
 ## Question
 
 Can simpler keep PTO-ISA async SDMA available while avoiding the roughly
--- a/src/a2a3/platform/include/aicpu/platform_regs.h
+++ b/src/a2a3/platform/include/aicpu/platform_regs.h
@@ -139,6 +139,30 @@ inline void write_reg(uint64_t reg_base_addr, RegId reg, uint64_t value) {
  */
 void platform_init_aicore_regs(uint64_t reg_addr);
 
+/** Send the AICore exit signal without waiting for an acknowledgement. */
+void platform_signal_aicore_exit(uint64_t reg_addr);
+
+/**
+ * Absolute sys-cnt value one deinit timeout from now. Share a single deadline
+ * across a group of cores so the whole group costs one timeout rather than one
+ * per unresponsive core.
+ */
+uint64_t platform_aicore_exit_deadline();
+
+/**
+ * Wait for a signalled AICore to acknowledge exit, then quiesce its register
+ * block (dispatch register back to idle, fast path closed).
+ *
+ * Pairs with platform_signal_aicore_exit when stopping several cores: signal
+ * them all, take one deadline from platform_aicore_exit_deadline(), then finish
+ * each core against that shared deadline.
+ *
+ * @param deadline absolute sys-cnt value to give up at.
+ * @return 0 once the core acknowledges, -1 on timeout — an unresponsive core is
+ *         left for the host's device reset to clear.
+ */
+int32_t platform_finish_aicore_exit(uint64_t reg_addr, uint64_t deadline);
+
 /**
  * Deinitialize AICore registers before termination
  *
--- a/src/a2a3/platform/onboard/host/device_runner.cpp
+++ b/src/a2a3/platform/onboard/host/device_runner.cpp
@@ -22,11 +22,14 @@
 
 #include <dlfcn.h>
 
+#include <algorithm>
 #include <cassert>
+#include <chrono>
 #include <cstddef>
 #include <cstring>
 #include <iostream>
 #include <string>
+#include <thread>
 #include <vector>
 #include "acl/acl.h"
 #include "host/acl_error_log.h"
@@ -43,6 +46,7 @@
 #include "utils/elf_build_id.h"
 #include "host/host_regs.h"  // Register address retrieval
 #include "host/raii_scope_guard.h"
+#include "utils/fatal_shutdown_latch.h"
 
 // dep_gen has two shapes, one per orchestration site, and each runtime provides
 // the strong symbols for the one it uses:
@@ -496,20 +500,33 @@ int DeviceRunner::drain_execution(ActiveExecution &active) {
 void DeviceRunner::cleanup_execution(PreparedExecution &prepared, bool retire_aicore) noexcept {
     if (!prepared.resources_owned) return;
 
+    // A poisoned card cannot retire per-resource frees or stream destroys, and
+    // issuing them can block in the driver. Drop host-side ownership instead;
+    // finalize()'s force reset invalidates the whole device generation.
+    const bool abandon = device_unusable_.load(std::memory_order_acquire);
+
     // Collectors must stop before their backing arguments are released; the
     // per-run stream retires last. Each cleanup operation is idempotent.
-    finalize_collectors();
-    (void)prepared.kernel_args.finalize_device_kernel_args();
-    (void)prepared.kernel_args.finalize_runtime_args();
+    finalize_collectors(abandon);
+    if (abandon) {
+        prepared.kernel_args.abandon_after_device_failure();
+    } else {
+        (void)prepared.kernel_args.finalize_device_kernel_args();
+        (void)prepared.kernel_args.finalize_runtime_args();
+    }
     if (prepared.kernel_args.args.pmu_reg_addrs != 0) {
-        (void)mem_alloc_.free(reinterpret_cast<void *>(prepared.kernel_args.args.pmu_reg_addrs));
+        if (!abandon) {
+            (void)mem_alloc_.free(reinterpret_cast<void *>(prepared.kernel_args.args.pmu_reg_addrs));
+        }
         prepared.kernel_args.args.pmu_reg_addrs = 0;
     }
     if (prepared.kernel_args.args.regs != 0) {
-        (void)mem_alloc_.free(reinterpret_cast<void *>(prepared.kernel_args.args.regs));
+        if (!abandon) {
+            (void)mem_alloc_.free(reinterpret_cast<void *>(prepared.kernel_args.args.regs));
+        }
         prepared.kernel_args.args.regs = 0;
     }
-    if (retire_aicore && !prepared.aicore_retirement_attempted) {
+    if (retire_aicore && !abandon && !prepared.aicore_retirement_attempted) {
         prepared.aicore_retirement_attempted = true;
         (void)retire_run_aicore_stream(prepared.pipeline_slot, RunStreamSlots::CompletionStatus::Unproven);
     }
@@ -518,6 +535,37 @@ void DeviceRunner::cleanup_execution(PreparedExecution &prepared, bool retire_ai
 
 void DeviceRunner::abandon_prepared_execution(PreparedExecution &prepared) noexcept {
     cleanup_execution(prepared, /*retire_aicore=*/true);
+}
+
+int DeviceRunner::create_run_stream(void **out) {
+    if (out == nullptr) return -1;
+    *out = nullptr;
+
+    rtStream_t stream = nullptr;
+    int rc = rtStreamCreate(&stream, 0);
+    if (rc != 0) {
+        LOG_ERROR("rtStreamCreate (run stream) failed: %d", rc);
+        ACL_LOG_ERROR_DETAIL(rc);
+        return rc;
+    }
+
+    // CANN 9.0.0 + driver 26.0.rc1 needs the host to observe an SDMA-generation
+    // AICore fault before the device reaches DEV_RUNNING_DOWN; otherwise reset
+    // walks the 48 CP-process streams and blocks in a 300-second remote event.
+    // Keep this workaround strictly on Workers that actually provisioned SDMA.
+    // Ordinary a2a3 and every a5 stream retain their normal error/diagnostic
+    // contract (for example 507046 rather than the early 507015).
+    if (dma_workspace_handle_ != nullptr) {
+        aclError acl_rc = aclrtSetStreamFailureMode(stream, ACL_STOP_ON_FAILURE);
+        if (acl_rc != ACL_SUCCESS) {
+            LOG_ERROR("aclrtSetStreamFailureMode (SDMA run stream) failed: %d", static_cast<int>(acl_rc));
+            (void)rtStreamDestroy(stream);
+            return static_cast<int>(acl_rc);
+        }
+    }
+
+    *out = stream;
+    return 0;
 }
 
 int DeviceRunner::ensure_run_stream_set(unsigned slot) {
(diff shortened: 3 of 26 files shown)


ChaoZheng109:
### 审查意见 —— 合并前请处理

整体方向、机制与测试覆盖都很扎实,投资记录(investigation)写得尤其好。合并前有四点需要澄清:

**1. `Fixes #1425` 会自动关闭一个本 PR 只"收敛"而非"修复"的 issue。** #1425 要求把 SDMA Worker 的故障恢复恢复到 ~0.3s;本 PR 交付的是**有界**的 ~10–30s 拆卸,加上普通 Worker 保持快路径。根因(CANN 对 CP-process SDMA 流没有退休栅栏)仍未解决——投资记录自己也写明完整恢复"deferred pending a CANN runtime-and-driver fix"。建议把 `Fixes` 改为 `Contains/Mitigates #1425` 并保留 issue 开启(挂上残留的 CANN 依赖),或在 PR 中明确说明这只是收敛方案。

**2. 普通(非 SDMA)中毒路径的强制复位从最多 3 次降为 1 次——请确认对普通分支是刻意的。** 记录在案的单次复位理由(重试会叠加 ~300s remote-event 超时)是 SDMA 专属的;普通故障下 `force_reset_device()` 失败并无此代价,而被删掉的 3 次重试有独立理由("poison occasionally needs a drain-then-reset cycle;5/5 验证")。我追查了上层兜底:覆盖率**不会**丢(#1110 dispatcher 会在全新子进程里重跑被 poison-skip 的类),**但**在 a2a3 上,原本能靠第 2/3 次就地复位救回、留在同子进程继续的卡,现在更容易被隔离、退到"整进程重启"这条更慢的恢复路径(对 a5 无影响,a5 本就无进程内恢复)。请确认:普通分支收敛到单次复位是刻意权衡过的,并接受 a2a3 就地恢复变弱的代价;否则建议只在 SDMA 分支用单次、非 SDMA 分支保留多次重试。

**3. 针对 #1425 的核心回归测试 `test_sdma_worker_aicore_fault_teardown_is_bounded` 被 `@pytest.mark.sdma` 门控、默认不选中,而本 PR 未在 CI 中接入 `-m sdma`。** 也就是说,验证 10s sleep + 有界 close 的那个测试在常规 CI 里是**休眠的**(普通 `enable_sdma=False` 变体确实在跑)。请确认已有某个 CI job 在独占设备上执行 `-m sdma`,否则这条回归护栏永远不会运行。

**4. a2a3/a5 双树不对称。** a2a3 把 `device_unusable_` 升级为 `std::atomic<bool>`(acquire/release),a5 仍是普通 `bool`;但两者的 `can_accept_run()` 都从 `c_api_shared.cpp` 读取该标志(可能与 run() 的写入方不在同一线程)。按 codestyle 规则 10,近重复的双树应保持同步——请确认 a5 确实无需原子化,或在 a5 一并应用。此外 a5 并未获得 `publish_fatal_shutdown` 发布栅栏(它只存在于 a2a3 的 tmr 调度器);请确认这是 a5 调度模型使然、而非遗漏。

次要(可考虑):`attempt_fatal_reset_once` 是个只调用一次 callable 的模板包装,除了标注"一次"的语义外无逻辑。保持现状也可,但直接调用 `force_reset_device()` 再加一条注释是等价的。

sunkaixuan2018:
@ChaoZheng109 感谢细致的审查 —— 第 2 点直接命中了本 PR 的一个真实回归,CI 也独立印证了它。逐条回复:

**1. `Fixes #1425` → 已改为 `Mitigates`。** 你的判断正确:本 PR 交付的是有界(~10–30s)收敛,根因(CANN 对 CP-process SDMA 流没有退休栅栏)仍未解决。PR 描述已去掉自动关闭关键字,新增 "Scope" 段落写明这一点并说明 #1425 应挂着残留的 CANN 依赖保持开启。

**2. 普通路径复位 3→1 是回归,已修复 —— 并且它正是本次 CI 失败的原因。**

不是刻意权衡,是我漏掉的。你指出的"a2a3 就地恢复变弱"在这次 `st-onboard-a2a3` 上已经真实发生了。失败的 5 个 `task_timing` 用例与故意注入故障的用例是**同设备前后相邻**关系:

| 设备 | 前一个用例(故意毒化) | 后一个用例 | 结果 |
| --- | --- | --- | --- |
| 2 | `dep_pool_overflow` | `test_distinct_slots_emit_markers` | FAIL 507018 |
| 7 | `flow_control_deadlock` | `test_duplicate_slot_merges_window` | FAIL |
| 5 | `heap_ring_deadlock` | `test_hbg_distinct_slots_emit_markers` | FAIL 507046 |
| 3 | `scope_deadlock` | `test_mix_task_aggregates_across_subtasks` | FAIL |
| 6 | `require_sync_start_invalid` | `test_spmd_task_aggregates_across_threads` | FAIL |

即:单次复位未能确认干净的卡,把毒化状态留给了下一个进程。`task_timing` 本身没问题(#1681、#1675 同一 job 通过)。

修复按你的建议执行:**非 SDMA 分支恢复 3 次有界重试(a2a3 与 a5 同步),仅 SDMA 分支保留单次**。理由现在写在代码与 investigation 里:`force_reset_device()` 每次调用都会先 drain 再 reset,并以 post-reset 探测(stream create/destroy + rtMalloc/rtFree)确认,所以第二次是在已沉降的卡上运行 —— 确实存在新的完成条件,原先"重试无新完成条件"的说法只对持有 48 条 CP-process 流的 SDMA 卡成立。

**3. `-m sdma` 在 CI 里是接入了的。** `.github/workflows/ci.yml` 有独立的 "SDMA pytest (a2a3)" step:

```
SDMA_TESTS="examples tests/st -m sdma"
```

按 marker 选择且覆盖 `tests/st`,所以 `tests/st/aicore_op_timeout/` 里新增的 `test_sdma_worker_aicore_fault_teardown_is_bounded` 会被它选中,在独占设备上跑。本次 run 的日志里看不到它,是因为前一个 sweep step 先失败导致 job 中止,该 step 未执行 —— 不是护栏休眠。第 2 点修好后应能看到它执行。

**4. a5 原子化已补齐;`publish_fatal_shutdown` 的不对称确认是遗漏,但本 PR 不扩围。**

- `device_unusable_` 在 a5 已改为 `std::atomic<bool>`(acquire/release),8 处读写全部同步,与 a2a3 一致。你的分析正确:`can_accept_run()` 的读取方与 `run()` 的写入方可能不同线程。
- `publish_fatal_shutdown` 的缺失**不是** a5 调度模型使然。我核对了四棵树,`a2a3/host_build_graph`、`a5/host_build_graph`、`a5/tensormap_and_ringbuffer` 三者的 `emergency_shutdown` 与 a2a3/tmr 修改前逐字节相同,都仍走等待应答的 `platform_deinit_aicore_regs`,因此同样存在该挂起风险。

  本 PR 只改了 #1425 复现的 a2a3/tmr。扩到另外三棵树需要同时给 a5 增加 `platform_signal_aicore_exit`,并改动三个当前工作正常的调度器,而我这边没有 a5 硬件可验证。倾向于单独开 issue 跟进而不是在这个已经因行为变更而红过一次的 PR 里扩围 —— 如果你认为应该一并落地,我照办。

**次要项:** `attempt_fatal_reset_once` 已被 `attempt_fatal_reset(reset, max_attempts)` 取代,现在承载真实的重试预算逻辑,不再是空包装。

sunkaixuan2018:
@coderabbitai Thanks — all three nitpicks were reviewed against `ce65e3d2`; each was already resolved by a later commit on the branch, verified against HEAD:

1. **Consolidate duplicated stop-on-failure stream creation** — no duplication exists. `aclrtSetStreamFailureMode` / `ACL_STOP_ON_FAILURE` appear at exactly one site (`src/a2a3/platform/onboard/host/device_runner.cpp:562`), and there is no `create_stop_on_failure_stream` in `device_runner_base.cpp` to share with. The mode is applied only when the Worker actually provisioned an SDMA workspace, which is a2a3-specific, so hoisting it to the common base would misplace it.

2. **Guard the abandon path with `abandon_device_resources`** — already guarded explicitly, not relying on call order. In `finalize_common_impl`, the `chip_callable_buffers_` loop, `free_tensor(device_wall_dev_ptr_)`, and `mem_alloc_.finalize()` each sit behind `if (!abandon_device_resources)`, which is the shape your proposed diff asked for.

3. **Dead `kMaxResetAttempts` retry block in `finalize()`** — the block was removed; `grep -rn kMaxResetAttempts src/` returned nothing at the reviewed commit.

Separately, the reset budget did change in the latest push, in the opposite direction from that dead-code removal: ordinary (non-SDMA) poison now keeps a bounded three-attempt budget via `attempt_fatal_reset(reset, max_attempts)`, because `force_reset_device()` drains before resetting and confirms with a post-reset probe, so a retry does add a completion condition. Only an SDMA-provisioned card is held to a single attempt. Collapsing both onto one attempt was a real regression — it left fault-injected cards poisoned for the next process and failed five `task_timing` tests in `st-onboard-a2a3`.

sunkaixuan2018:
Follow-up: the first fix was necessary but not sufficient. Second root cause found and fixed.

**Where it stood.** Restoring the three-attempt reset budget took `st-onboard-a2a3` from 5 `task_timing` failures to 3 — real progress, but not green.

**What the remaining 3 were.** The surviving failures were on devices whose preceding test was a pool/allocator-deadlock fault injection (`dep_pool_overflow` and `heap_ring_deadlock` failed in *both* runs). Two things in the logs ruled out the reset budget as the remaining cause:

- Only 3 fatal teardowns ran in the whole job, and they belonged to the *failing* `task_timing` tests, not to their predecessors. The fault-injection tests recover in place and never take the force-reset path at all.
- Every one of those force resets confirmed clean on the **first** attempt (`aclrtResetDeviceForce(N) established a usable device generation (probe confirmed)`), so no amount of retry budget was going to matter.

**The actual cause.** This PR made emergency shutdown fire-and-forget. `platform_deinit_aicore_regs` did three things — signal exit, wait for the core to acknowledge, then quiesce its register block (dispatch register back to idle, fast path closed) — and the PR kept only the signal. The per-thread `shutdown()` path also no-ops once fatal shutdown starts, so on a fatal run *nothing* waited for the cores any more. The AICPU op returned while AICores were still running, with the fast path left open, and the card stayed poisoned past the host's device reset. The next process on that device then failed at launch with 507018/507046.

The control that isolates it: `st-onboard-a5` **passes** with the identical host-side teardown changes. a5's schedulers were untouched and kept their acknowledgement wait; only the a2a3 `tensormap_and_ringbuffer` scheduler lost it. Four sibling PRs also pass this job today, so it is not a pre-existing flake.

**The fix.** Emergency shutdown now signals every handshake'd core first, then joins them — rather than signal-and-wait one core at a time, which is what the original serial `platform_deinit_aicore_regs` loop did. Cores drain concurrently, a dead core's timeout no longer serializes behind the cores ahead of it, and each core's register block is quiesced once it confirms it stopped. `platform_finish_aicore_exit()` is split out of `platform_deinit_aicore_regs()` for this; the latter is now signal + finish and behaves exactly as before for its existing callers.

Worth being explicit that this does **not** reintroduce the #1425 hang: the acknowledgement is an on-device poll of the core's `COND` register, bounded by the variant deinit timeout. It is not a host or remote operation and has nothing to do with the CANN CP-process SDMA stream teardown that the rest of this PR contains.

The `publish_fatal_shutdown` fence is unchanged and still does its job — it keeps a healthy per-thread `shutdown()` from racing the emergency broadcast over the same cores.

sunkaixuan2018:
CI is fully green — all 18 checks pass, including both hardware jobs (`st-onboard-a2a3` 8m0s, `st-onboard-a5` 10m52s) and `ut-a2a3` / `ut-a5`.

`st-onboard-a2a3` was the job that had been red since the first push. Getting it green took three distinct fixes, each found from a different piece of evidence:

| Failures | Cause |
| --- | --- |
| 5 | Reset budget collapsed from 3 attempts to 1 on the ordinary poison path — the regression @ChaoZheng109 predicted in review point 2. |
| 3 | Emergency shutdown dropped the AICore exit acknowledgement, so the AICPU op returned with cores still running and the fast path open, leaving the card poisoned past the host device reset. |
| 1 | Restoring the join cost one deinit timeout *per core* (1s onboard), pushing an AICore-timeout run past its 10s budget. Fixed by sharing one deadline across the whole core group. |
| 1 | Rebase fallout: `CallConfig.enable_l2_swimlane` was renamed to `enable_chip_swimlane` by #1681. The test file merged cleanly but was semantically stale. |
| 0 | — |

The last run before this one finished 12-green / 3-red, where all three red jobs failed on infrastructure rather than code — `Cloning pto-isa ... timed out`, `Could not find a version that satisfies the requirement scikit-build-core` (PyPI unreachable), and a failed `Checkout target`. A single retrigger on an identical tree cleared all three, which confirms they were transient.

The branch has also been rebased onto current `main`, which required resolving 13 conflicts against #1683 (enqueue/drain split) and #1681 (layered naming). Two notes on that:

- #1683 moved per-run cleanup into `cleanup_active_run()`, so the poisoned-card guarding was re-applied there. Without it, a fatal enqueue rollback or drain would have gone back to issuing the frees and stream destroys this PR exists to avoid.
- **`device_unusable_` is now atomic on a5 upstream**, so review point 4's atomicity half is resolved by `main` rather than by this PR, and the claim has been dropped from the commit message.

Still open for your call, unchanged from the earlier comment: whether to fold the emergency-shutdown fix into the three sibling scheduler trees (`a2a3/host_build_graph`, `a5/host_build_graph`, `a5/tensormap_and_ringbuffer`), which still carry the byte-identical acknowledgement-waiting `emergency_shutdown` and would need `platform_signal_aicore_exit` / `platform_finish_aicore_exit` added to a5's `platform_regs`.

ChaoZheng109:
### 复审(对照 `ac6781f7` → `90eafb5b`)

先答你留给我的那个开放问题,然后是合并前建议改的两处。上一轮我提的四点,1/2/4a 已确认落实,第 3 点是我看错了(`ci.yml` 传 `a2a3_sdma_mode: marker`,`_st-npu.yml` 的 "SDMA pytest (a2a3)" step 就是 `-m sdma`,本次已 success,护栏一直是活的)。

---

#### T1 — 紧急停机是否同步到另外三棵树:**本 PR 不做,单开 issue 跟进**

我核对了 `upstream/main` 上 a2a3/tmr 的 `emergency_shutdown`,它原本就是串行 `platform_deinit_aicore_regs`(signal + 等应答)。所以 `a2a3/host_build_graph`、`a5/host_build_graph`、`a5/tensormap_and_ringbuffer` 三棵树停在 main 的行为上,**并没有丢失应答等待,不存在活跃缺陷**。它们缺的只是共享 deadline 和 `publish_fatal_shutdown` 栅栏。你不扩围的判断是对的,尤其在没有 a5 硬件可验证的前提下。

跟进 issue 里建议带上一条可量化的项,别只记成对称性对齐:串行版本每个无响应核付一次完整 deinit 超时(onboard 1s),整组核都死时会在 AICPU op 内部耗 N 秒 —— 这正是会撞上 STARS op-execute 超时的量级。这条值得实测后再决定要不要动那三棵树。

---

#### T5 — `shutdown()` 短路后,三处注释/文档已经说了假话

`shutdown()` 现在在 `fatal_shutdown_started_` 置位后直接 `return 0`,但下面三处仍按"无条件执行"描述它。按 `.claude/rules/doc-consistency.md` §4,这三处要和代码改动落在同一个 commit 里:

1. `src/a2a3/runtime/tensormap_and_ringbuffer/aicpu/aicpu_executor.cpp:860`
   > `// Always shutdown AICore — even if sched_ctx_.completed_ was already true.`

   这句现在是错的 —— fatal 路径下恰恰不执行。

2. `src/a2a3/runtime/tensormap_and_ringbuffer/runtime/scheduler/scheduler_cold_path.cpp:625-627`
   函数头注释仍写 "deinit AICore regs for this thread's cores (and PMU finalize if enabled)",未提 fatal 短路。顺带一提,短路也跳过了 `pmu_aicpu_finalize()` —— 整卡随后被 host force reset,实际无害,但这是个静默的行为变化,建议在这里把它作为现状事实写明,免得下一个读者当成漏写。

3. `src/a2a3/runtime/tensormap_and_ringbuffer/docs/RUNTIME_LOGIC.md:607`
   > `| shutdown(thread_idx) | per thread on exit | platform_deinit_aicore_regs for this thread's cores; PMU finalize when enabled |`

   同样需要补上 fatal 时由 `emergency_shutdown` 统一接管、本函数 no-op 这一条。

---

#### T6 — 测试的 close 上限装不下 3 次复位,会在重试真正生效时误报

`tests/st/aicore_op_timeout/test_aicore_op_timeout.py:129`

```python
close_limit = 30 if enable_sdma else (20 if st_platform == "a5" else 10)
```

这组数字是在"fatal 路径只跑 1 次复位"的那一版加进来的,复位预算恢复成 3 次后没有回头复核。算一下单次 `force_reset_device()` 的成本:

| 组成 | 本用例耗时 |
| --- | --- |
| `aclrtSynchronizeDeviceWithTimeout(stream_sync_timeout_ms)` | 4s(用例设了 `SIMPLER_STREAM_SYNC_TIMEOUT_MS=4000`) |
| `aclrtResetDeviceForce` | a5 约 11s(你自己在注释里写的实测) |
| post-reset probe | 很小 |

a5 单次已经 ≈ 15s,上限却是 20s —— **连第 2 次的开头都放不下**;a2a3 的 10s 同理,第 2 次几乎必然超。

问题不在于它今天会红(本次 run 三次 force reset 都在第 1 次就 confirm),而在于**它会在重试机制真正发挥作用的那一次判红**,并报出 `fatal teardown did not stay bounded`。那句话会把 triage 的人直接引向 #1425 的驱动停顿假设,而实际情况是重试正常工作了、只是慢。这个误报是新引入的:main 上没有 close 计时断言,唯一上界是 `timeout(60)`,余量宽得多,而且真炸也只是一条中性的 pytest 超时。

建议把上限放到能覆盖满 3 次尝试,仍远低于本用例要捕获的 150/300s 驱动停顿,护栏一点不弱:

```python
# 非 SDMA 路径最多跑 kFatalResetAttempts=3 次 force_reset_device();
# 每次 = stream-sync 预算 + 驱动复位(a5 实测 ~11s) + probe。
# 上界按满预算取,仍远低于本回归要捕获的 150/300s 驱动事件停顿。
close_limit = 30 if enable_sdma else (60 if st_platform == "a5" else 45)
```

注意非 SDMA 用例的 `@pytest.mark.timeout(60)`(:138)要一并提高,否则外层 mark 会先于断言触发,报出更难读的失败。

---

其余部分我没有阻塞项。CI 18 项全绿,含两个硬件 job;`pto_isa.pin` 固定在 `83d01313`,本 PR 未改动任何 pto-isa 头文件引用,例行确认该 commit 仍够用即可。

sunkaixuan2018:
Both actionable items fixed in `e676939de`. Thanks for catching T6 in particular — the limits were set when the fatal path ran one reset, and I restored the three-attempt budget without going back to recheck them. That is exactly the kind of cross-cutting consequence I should have caught myself.

**T1 — agreed, out of scope here.** Your reading matches mine: the three sibling trees still run `main`'s serial `platform_deinit_aicore_regs` (signal + wait), so they never lost the acknowledgement and have no active defect — they only lack the shared deadline and the `publish_fatal_shutdown` fence. I'll file the follow-up with the quantifiable framing you asked for: the serial version pays one full deinit timeout **per unresponsive core** (1 s onboard), so an all-dead-core fatal run burns N seconds inside the AICPU op, which is the same order as the STARS op-execute timeout. That number is what should decide whether those three trees get touched, and it wants a measurement rather than a symmetry argument.

**T5 — three sites corrected.**

- `aicpu_executor.cpp` — the "Always shutdown AICore" claim now states the fatal exception explicitly.
- `scheduler_cold_path.cpp` — the `shutdown()` header documents the fatal short-circuit, and I took your suggestion to record the `pmu_aicpu_finalize()` skip as a present-tense fact with its justification (the card is force-reset right after, so those counters would not survive into the next generation anyway) rather than leaving it to look like an oversight.
- `RUNTIME_LOGIC.md:607` — the table row carries the same no-op note.

**T6 — budget widened to hold the full retry.**

```python
close_limit = 30 if enable_sdma else (60 if st_platform == "a5" else 45)
```

Your arithmetic is what I used: 4 s stream-sync budget + ~11 s driver reset on a5 + a small probe ≈ 15 s per attempt, so 20 s could not even reach the second attempt. I also raised the non-SDMA `@pytest.mark.timeout` from 60 s to 180 s, since with a 60 s close ceiling plus worker setup plus the 10 s run budget the outer mark would have fired first and produced exactly the unreadable failure you flagged. The SDMA case keeps `timeout(90)` against its unchanged 30 s ceiling — one attempt plus the 10 s handoff leaves roughly 2× headroom there. The comment above the limits now derives them from `kFatalResetAttempts` instead of restating the superseded single-attempt rationale.

One extra thing surfaced while rebasing onto `b535fa210`. #1683's `release_graph_execution_buffers()` landed on the shared teardown path and calls `mem_alloc_.free()` per retained buffer, so on the fatal path it would have reintroduced exactly the device frees this PR removes. It is guarded now via a new `abandon_graph_execution_buffers()` that clears only the host-side map, matching the other `abandon_*` helpers. Guarded explicitly rather than relying on `mem_alloc_.abandon_after_device_failure()` having already emptied `ptr_size_map_`, per CodeRabbit's earlier order-independence point.

On `pto_isa.pin`: unchanged at `83d01313`, and this PR touches no pto-isa header or include path.

Independent hardware re-validation of the pre-fix tree (`90eafb5b5`, byte-identical to this one outside the four files above) on myserver, CANN 9.0.0 / driver 26.0.rc1: real SDMA plus AICore fault returned 507015 in 0.037 s, `Worker.close()` in 16.315 s, longest driver ioctl 3.25 s, no 150/300 s event timeout, and post-fault recovery ran real SDMA bit-exact. The `runtime_fatal_codes` + `task_timing` + `aicore_op_timeout` chain passed 18/18 with `aicore_op_timeout` at 9.1 s. I'll re-run CI on this commit and report.

sunkaixuan2018:
Status on `2efedc323` (rebased onto `c9ccaf653`).

**Both a5 hardware jobs now pass** — `st-onboard-a5` 8m40s and `ut-a5` 2m34s, alongside `st-onboard-a2a3` 8m30s, `ut-a2a3`, `st-pod-onboard-a2a3`, packaging, and profiling. That closes the gap I flagged earlier: the a5 branch of T6's widened budget (20s → 60s) is the tightest one, since a single a5 reset already costs ~15s, and it had never actually executed while the a5 runners were saturated. It has now.

The previous `ut-a5` red was a runner-side lock failure, not code — `[npu-lock] 错误: 无法修复共享锁权限: .../npu_device_4.lock`, with the wheel build succeeding before it. `ut (macos-latest)` reports as red in the checks list but its job conclusion is `cancelled`, not `failure`.

**`ut (ubuntu-latest)` is a genuine failure, and it is pre-existing on `main` rather than something this PR introduces.** Three cases in `tests/ut/py/test_worker/test_remote_zero_residual.py` fail with:

```
ValueError: only host_tcp transport is accepted by simpler-remote-worker
```

The interaction:

| When | What |
| --- | --- |
| 08-05 12:36 | #1692 lands `test_remote_zero_residual.py`, which passes `transport="sim"` |
| 08-05 21:07 | #1688 tightens `remote_l3_worker.py` / `worker.py` to accept only `host_tcp`, and updates `test_remote_l3_lifecycle.py` — but not the newer file #1692 had just added |

Neither PR could have caught it: each was validated against a base predating the other, and `ci.yml` triggers only on `pull_request`, never on push to `main`, so the merged combination is never revalidated. The breakage therefore surfaces on the first PR to rebase onto both, which happens to be this one.

This PR's diff touches no worker, remote, or transport file — the 26 changed files are entirely device-runner, scheduler, teardown, and their tests. I have not attempted a fix here, since the correct change is either relaxing the transport gate or updating #1692's test, and that call belongs to #1688's author rather than to a fatal-teardown PR. Flagging it for whoever owns it; happy to carry a one-line fix if you would rather it ride along.

`profiling-flags-smoke`, `packaging-matrix`, and the ubuntu sim jobs are green; the remaining macos sim jobs are still queued.

sunkaixuan2018:
T1 follow-up filed as #1710, with the measure-first framing you asked for — the N-second serial cost against the STARS op-execute timeout is the deciding number, and closing it as won't-fix is an explicit outcome if the measurement does not support touching those three trees.

That was the last outstanding item from your two reviews. Current state on `ff6964a20` (rebased onto `05b6a9471`):

| Point | Status |
| --- | --- |
| R1.1 `Fixes` → `Mitigates` | commit trailer reads `Mitigates #1425`; no auto-close keyword in the PR body |
| R1.2 non-SDMA reset budget | `kFatalResetAttempts = 3` on both a2a3 and a5; SDMA still 1 |
| R1.3 `-m sdma` guardrail | your retraction — the step is live and passing |
| R1.4 a2a3/a5 asymmetry | `publish_fatal_shutdown` fence in place; a5's `device_unusable_` is now atomic via `main` rather than this PR, so the claim is out of the commit message |
| T1 sibling trees | out of scope here, tracked in #1710 |
| T5 stale comments/docs | all three sites corrected, including the `pmu_aicpu_finalize()` skip recorded as a present-tense fact |
| T6 close budget | `30 / 60 (a5) / 45 (a2a3)` with `@pytest.mark.timeout(180)` on the non-SDMA case |

Also worth flagging separately from your review: `ut (ubuntu-latest)` was red on the previous push for a reason that was not this PR — #1692 landed a test using `transport="sim"` and #1688 then restricted simpler-remote-worker to `host_tcp`, with neither PR able to see the other. #1704 has since fixed it and this branch is rebased past it. The underlying gap is that `ci.yml` triggers only on `pull_request` and never on push to `main`, so a merged combination is never revalidated and the breakage lands on whichever unrelated PR rebases first. Probably worth its own issue if you agree.

ChaoZheng109 (dismissed):

ChaoZheng109 (approved):

PR #1666, opened 2026-08-03
Add: host_build_graph copy of the qwen3_14b_decode example

noabauma:
The 40-layer fused Qwen3-14B decode existed only for tensormap_and_ringbuffer. pto_orchestration_api.h is identical between the two runtimes and compile_orchestration() selects the include dirs of whichever runtime @scene_test names, so no C++ changes: this adds a test class and a README, and points every kernel source at the existing example's kernels/ via QWEN_KERNELS. Sharing the sources rather than copying them keeps the two runtimes' compute bit-for-bit comparable and keeps the vendored FusedInferAttentionScore tree in the repo exactly once.

The heap must hold all 40 layers' intermediates simultaneously: host_build_graph builds the whole graph before the device schedules anything, so no task has completed while the graph is being built and the heap tail never advances off 0. The 256 MiB default is exhausted in the last layers (Task Allocator Deadlock - Heap Exhausted, ~254 of 256 MiB used); 512 MiB carries it, set per case via runtime_env. The task window is not the constraint: the graph is ~10.6K tasks, inside the 16384 default. tensormap_and_ringbuffer needs no sizing at all, because each layer's scope frees its intermediates as orchestration proceeds.

Validated on a2a3 with golden checking across all 40 layers' KV writes and the hidden output, on three devices (36.8 / 36.8 / 36.9 ms device_wall).

### DIFF
--- a/examples/a2a3/host_build_graph/qwen3_14b_decode/README.md
+++ b/examples/a2a3/host_build_graph/qwen3_14b_decode/README.md
@@ -1,0 +1,206 @@
+# `qwen3_14b_decode/` — Qwen3-14B 40-layer decode (CANN fused attention), host_build_graph
+
+Self-contained SceneTestCase port of pypto-lib
+`models/qwen3/14b/decode_fwd.py` entry `decode_fwd_layers` with
+`_CHUNK_NLAYERS == 40`: **the whole Qwen3-14B decode stack as one fused
+dispatch** (hidden → hidden, no LM head), with the FP32 inter-layer residual
+carry. A simpler developer builds and runs it directly — no descent through
+pypto-lib / the JIT, no auto-built intermediate artifacts.
+
+The layer loop is a real loop in the generated orchestration
+(`for (int64_t i = 0; i < 40; i += 1)`), not an unrolled one, so a 40-layer
+chunk costs one extra literal over a 2-layer chunk rather than 20× the kernels.
+
+## Parameter regime — matches `stress_profile.py`
+
+The fixture mirrors the vLLM serving stress run (`stress_profile.py`):
+
+| Param | Value | Source |
+| ----- | ----- | ------ |
+| `BATCH` | 16 | `CONCURRENCY` (aligned with decode kernel BATCH=16) |
+| `MAX_SEQ` | 5500 | `max_model_len` (KV-pool / RoPE sizing) |
+| decode `seq_len` | 3500 | the ~3500-token prompt |
+| layers | 40 | full model (`decode_fwd_layers` N=40) |
+
+Per the lib's const-layer-0 stacked-fwd reference, every layer reuses layer-0
+weights (weights + paged KV pool are stacked ×40 along dim 0, one slice per
+layer); each layer still reads and writes its own KV pool.
+
+Footprint at this regime, bf16:
+
+| component | size |
+| --------- | ---: |
+| weights (×40) | 24.61 GiB |
+| paged KV pool (×40) | 13.44 GiB |
+| **total fixture** | **38.05 GiB** |
+
+Die HBM is 64 GiB, so it fits with ~26 GiB of headroom for ring heap and
+workspace.
+
+**This runtime needs the ring heap raised to 512 MiB**, which the case carries in
+its own `runtime_env`. On `tensormap_and_ringbuffer` each layer's intermediates
+live inside that iteration's scope and are freed at its end, so the live set does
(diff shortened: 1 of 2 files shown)


examples/a2a3/host_build_graph/qwen3_14b_decode/README.md
```diff
@@ -0,0 +1,206 @@
+# `qwen3_14b_decode/` — Qwen3-14B 40-layer decode (CANN fused attention), host_build_graph
+
+Self-contained SceneTestCase port of pypto-lib
+`models/qwen3/14b/decode_fwd.py` entry `decode_fwd_layers` with
+`_CHUNK_NLAYERS == 40`: **the whole Qwen3-14B decode stack as one fused
+dispatch** (hidden → hidden, no LM head), with the FP32 inter-layer residual
+carry. A simpler developer builds and runs it directly — no descent through
+pypto-lib / the JIT, no auto-built intermediate artifacts.
```
ChaoZheng109, line 8: This README is inherited from the `tensormap_and_ringbuffer` version, but the hbg directory ships only this test file and README — the kernels are shared via `QWEN_KERNELS`, not local. "Self-contained" and the local `kernels/` framing are misleading here.

```suggestion
`host_build_graph` SceneTestCase port of pypto-lib
`models/qwen3/14b/decode_fwd.py` entry `decode_fwd_layers` with
`_CHUNK_NLAYERS == 40`: **the whole Qwen3-14B decode stack as one fused
dispatch** (hidden → hidden, no LM head), with the FP32 inter-layer residual
carry. A simpler developer builds and runs it directly — no descent through
pypto-lib / the JIT, no auto-built intermediate artifacts.

This directory holds only this test file and README. All C++ — the generated
orchestration, the 18 AIC / 16 AIV incores and the vendored CANN attention
extern — is **shared** with the `tensormap_and_ringbuffer` example via
`QWEN_KERNELS` (`../../tensormap_and_ringbuffer/qwen3_14b_decode/kernels`)
rather than copied, keeping the two runtimes' compute bit-for-bit comparable
and the vendored `FusedInferAttentionScore` tree in the repo exactly once.
Every `kernels/…` path below therefore refers to that shared tree in the
`tensormap_and_ringbuffer` example, not to files in this directory.
```

examples/a2a3/host_build_graph/qwen3_14b_decode/README.md
```diff
@@ -0,0 +1,206 @@
...
+# PTO2_MANUAL_MAX_SEQ must be set before importing decode_fwd (read at import).
+os.environ["PTO2_MANUAL_MAX_SEQ"] = "5500"
+D = <import build/pypto-lib/models/qwen3/14b/decode_fwd.py by path>
+D._CHUNK_NLAYERS = 40          # read at trace time; rebind before the first call
+D.decode_fwd_layers(*inputs, out, config=RunConfig(
+    platform="a2a3", codegen_only=True, save_kernels=True, save_kernels_dir=OUT))
+```
+
+`codegen_only` needs no device. Then copy `OUT/orchestration/`, `OUT/kernels/`
+and `models/qwen3/14b/kernels/paged_attention_cce/` into `kernels/vendor/` here, and
```
ChaoZheng109, line 146: `here` points at this directory, but the shared kernels physically live in the `tensormap_and_ringbuffer` example. Point the copy target there so a refresh updates the one shared tree both runtimes consume.

```suggestion
`codegen_only` needs no device. Then copy `OUT/orchestration/`, `OUT/kernels/`
and `models/qwen3/14b/kernels/paged_attention_cce/` into the shared `kernels/`
tree in the `tensormap_and_ringbuffer` example (both runtimes consume it), and
```

ChaoZheng109 (commented):
The port itself looks clean — I diffed the test file against the tmr original and the 20 orchestration signatures + 37 incore `func_id`/`signature`/`core_type` are byte-identical, and the `ring_heap=512 MiB` case matches the 256 MiB `PTO2_HEAP_SIZE` default. One doc nit: this README is inherited nearly verbatim from the `tensormap_and_ringbuffer` version and still describes a local `kernels/` tree, but the hbg directory ships only the test + README and shares kernels via `QWEN_KERNELS`. So "Self-contained" (line 3) and the `kernels/… here` regen step (line 146) are misleading — they'd lead a reader to recreate a `kernels/vendor/` here, which is exactly the "in the repo exactly once" duplication the PR set out to avoid. Two suggestions below.

PR #1667, opened 2026-08-03
Add: host_build_graph copies of the paged_attention examples

noabauma:
paged_attention and paged_attention_manual_scope existed only for tensormap_and_ringbuffer. pto_orchestration_api.h is byte-identical between the two runtimes and compile_orchestration() selects the include dirs of whichever runtime @scene_test names, so the kernels and orchestration C++ are unchanged copies; the delta is the runtime= decorator, the class name, and ring sizing.

host_build_graph builds the whole graph before the device schedules anything, so nothing is reclaimed mid-orchestration and the entire graph must be resident. Case1 (65792 tasks) and Case2 (32832) therefore carry ring_task_window/ring_heap in their own CASES[*]["config"]["runtime_env"], needing no PTO2_RING_* env var.

Case3 (head_dim 256) is omitted: it fails on both runtimes, giving max_diff=nan here and max_diff=0.1255 on the unmodified tensormap_and_ringbuffer example, with a normal 16 ms device_wall and no allocator fatal. That is a pre-existing kernel numerics defect at head_dim > 128, not a property of this runtime.

Validated on a2a3 with golden checking: 12/12 cases pass across both examples.

### DIFF
--- a/examples/a2a3/host_build_graph/paged_attention/README.md
+++ b/examples/a2a3/host_build_graph/paged_attention/README.md
@@ -1,0 +1,65 @@
+# paged_attention — host_build_graph
+
+Online-softmax paged attention in bfloat16, split across AIC and AIV, on the
+`host_build_graph` runtime. The kernels and the orchestration source are the
+same ones the `tensormap_and_ringbuffer` sibling compiles.
+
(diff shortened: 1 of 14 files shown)


doraemonmj:
Not within the benchmark range

PR #1691, opened 2026-08-04
Fix: keep forked host log records intact

### LINKED ISSUE #1690: [Bug] L3 multi-rank STRACE records interleave on shared stderr and corrupt benchmark timing
### Platform

a2a3 (Ascend 910B/C hardware)

### Runtime Variant

tensormap_and_ringbuffer

### Description

L3 multi-rank benchmark timing is unreliable because forked chip-worker processes write `[STRACE]` records concurrently to the same captured `stderr` file descriptor.

`HostLogger::emit` writes one logical log record through multiple stdio calls. Its mutex is local to each `HostLogger` instance and therefore does not serialize writers from different forked processes. Records from different ranks can consequently be interleaved into the same physical line.

The STRACE parser extracts only one record from each physical line. Interleaved records are therefore lost, causing:

- Missing orchestrator or scheduler spans
- `effective_us=0.0` for valid dispatches
- Fewer reported ranks than the actual world size
- Unstable and systematically inflated median/mean latency

The MoE computation itself completes successfully and passes correctness checks. This is an instrumentation output problem, not a kernel performance regression.

### Steps to Reproduce

Use the following configuration:

```bash
source <CANN_INSTALLATION>/set_env.sh

export PYTHONPATH=<WORKSPACE>/pypto-lib
export TORCH_DEVICE_BACKEND_AUTOLOAD=0
export PYPTO_BENCH=1
export PYPTO_BENCH_WARMUP=2
export PYPTO_BENCH_ROUNDS=8
export PYPTO_BENCH_RAW=1
export DSV4_HC_PRE_IMPL=separate
export PYPTO_ROUTE_HASH_IMPL=aiv
export PTO2_RING_DEP_POOL=16384
export PTO2_RING_TASK_WINDOW=16384
export PTO2_RING_HEAP=1073741824

.venv/bin/python models/deepseek/v4-flash/moe.py \
    -p a2a3 \
    --ep 8 \
    -d 0,2,4,6,8,10,12,14 \
    --balanced-routing \
    --enable-l2-swimlane 0
```

Parallel configuration:

- Expert parallel size: 8
- World size: 8
- Participating devices: `0,2,4,6,8,10,12,14`
- Mapping: rank 0 -> device 0, rank 1 -> device 2, ..., rank 7 -> device 14
- Warmup rounds: 2
- Measured rounds: 8
- Routing: balanced
- HcPre implementation: separate
- Route hash implementation: AIV

The issue reproduces consistently in three independent runs.

### Expected Behavior

For 8 ranks and 8 measured rounds:

- All 64 `(rank, round)` cells should contain a valid `effective_us` value.
- The reported rank count should be 8.
- Median and mean latency should remain stable across repeated runs.
- Results should be comparable with the approximately 477 us reference baseline for this configuration.
- Missing STRACE spans should be reported as invalid instrumentation data rather than silently converted to `0.0`.

### Actual Behavior

All three runs complete successfully in approximately 35-44 seconds. Correctness passes with `x_next ratio_reldiff < 0.003`.

There are no runtime errors, timeouts, or device failures: no `507018`, `507014`, `507899`, `HandleTaskTimeout`, or `FATAL` records.

However, most benchmark timing samples are missing:

| Run | Reported ranks | Non-zero samples out of 64 | Minimum effective_us | Median effective_us | host_union_mean |
|---|---:|---:|---:|---:|---:|
| v3 | 7 | 14 | 476.0 us | unavailable | 2773 us |
| v4 | 7 | 13 | 476.1 us | 958.8 us | 2678 us |
| v5 | 8 | 20 | 472.8 us | 1210.6 us | 2631 us |

In v5, all eight participating ranks contain missing `0.0` cells. One participating worker has all eight measured rounds reported as `0.0`.

The minimum non-zero values from the three runs agree with the approximately 477 us reference within about 1%, indicating that kernel performance itself is not regressed. The median and mean values are corrupted by missing or incorrectly parsed spans.

### Git Commit ID

`simpler`: `9922afdb08cc6f203eaf39328661e2f2648d333d`

Related versions:

- `pypto-lib`: `fc9993fb26732bc1e718d952f65abc6b3fa2a227`
- `pypto`: `1784c635b5ea750c5666dde0fe132aa2c1d10a34`
- `pto-isa`: `83d01313d9bfc247c4b7c8bcf969d1019f0d106f`
- PTOAS: `0.48`
- `pypto-serving`: not used

### CANN Version

9.0.0, timestamp 20260428

### Driver Version

26.0.rc1

Firmware version is not reported by the board. The reported software version is 26.0.rc1.

### Host Platform

Linux (aarch64)

- OS: openEuler 24.03 SP1
- Kernel: `6.6.0-72.0.0.76.oe2403sp1.aarch64`
- Python: 3.11.6
- PyTorch: 2.8.0+cpu
- NumPy: 2.2.6
- NPU: Ascend910, Product `IT22HMDA_2_S`
- PCI Device ID: `0xD803`
- HBM: 64 GB per chip

#### Root-cause evidence

At the affected commit, `src/common/log/host_log.cpp:86-92` emits one logical record using several operations:

```cpp
fprintf(stderr, ...);
vfprintf(stderr, fmt, args);
fputc('\n', stderr);
fflush(stderr);
```

The surrounding `std::scoped_lock` protects only threads using the same `HostLogger` instance. Each forked L3 chip-worker process has its own logger and mutex, so it does not prevent cross-process writes to the shared captured `stderr` descriptor.

In `simpler_setup/tools/strace_timing.py:51-55,98-113`, `parse_spans` performs one `_STRACE_RE.search` per physical line. The greedy `attrs=.*` portion can consume the remainder of a line containing another interleaved STRACE record. The second record is not recovered.

When an orchestrator or scheduler span is lost, the benchmark reports `effective_us=0.0`. If all samples for a rank are lost, that rank is omitted from the reported rank count.

#### Suggested fixes

1. Format each host log record into a contiguous buffer and emit it with one cross-process-safe write. STRACE records written to a captured pipe should remain within `PIPE_BUF` so each record is atomic.
2. Harden `parse_spans` to detect multiple STRACE markers or malformed interleaved records instead of silently accepting partial input.
3. Validate the expected rank and round counts before calculating statistics.
4. Mark missing spans as invalid samples and report an instrumentation error; do not represent them as `0.0`.
5. Add an L3 multi-process test in which several child processes concurrently emit STRACE records to the same captured descriptor.

#### Workaround

Capture each rank's `stderr` into a separate file and parse the files independently before aggregating benchmark results. `PYPTO_BENCH_RAW=1` alone does not prevent shared-descriptor interleaving.

sunkaixuan2018:
## Summary

- format each host log record dynamically and emit it with one `write`, preventing forked chip workers from interleaving STRACE records on shared stderr
- parse every STRACE marker on a physical line instead of letting greedy `attrs` consume later records
- add deterministic forked-process logger and adjacent-record parser regressions

## Testing

- baseline regression: `test_host_log_off` fails on `main` with two child records merged onto one physical line
- `ctest -R '^test_host_log_off$'`: 8/8 passed
- `pytest tests/ut/py/test_strace_timing.py -q`: 1 passed
- `ctest -LE requires_hardware`: 79/79 passed
- pre-commit checks passed, including formatting, Ruff, cpplint, pyright, headers, and clang-tidy
- full Python UT collection was blocked by the remote environment's stale editable install/Torch visibility; the targeted Python regression passed in an isolated checkout

Fixes #1690

### DIFF
--- a/simpler_setup/tools/strace_timing.py
+++ b/simpler_setup/tools/strace_timing.py
@@ -48,16 +48,26 @@ import sys
 from collections import defaultdict
 from dataclasses import dataclass, field
 
+# The host-log record prefix `HostLogger::emit` writes ahead of every message.
+# Its func segment excludes ':' because `LOG_TIMING` passes `__FUNCTION__`, an
+# unqualified name; a qualified name containing '::' would stop this alternative
+# from matching, leaving the `[STRACE]` alternative to bound the record instead.
 _HOST_LOG_PREFIX = (
     r"\[\d{4}-\d{2}-\d{2} \d{2}:\d{2}:\d{2}\.\d{6}\]"
     r"\[T0x[0-9a-fA-F]+\]\[[A-Z]+\]\s+[^:\r\n]+:\s+"
 )
+# MULTILINE anchors the `$` alternative at every line end, so a caller passing a
+# multi-line blob rather than one line per item keeps every record but the last.
 _STRACE_RE = re.compile(
     r"\[STRACE\]\s+v=(?P<v>\d+)\s+pid=(?P<pid>\d+)\s+tid=(?P<tid>\d+)\s+"
     r"inv=(?P<inv>\d+)\s+hid=(?P<hid>[0-9a-fA-F]+)\s+depth=(?P<depth>\d+)\s+"
     r"name=(?P<name>\S+)\s+ts=(?P<ts>\d+)\s+dur=(?P<dur>\d+)(?P<attrs>.*?)"
-    rf"(?={_HOST_LOG_PREFIX}|\[STRACE\]|\r?$)"
+    rf"(?={_HOST_LOG_PREFIX}|\[STRACE\]|\r?$)",
+    re.MULTILINE,
 )
+# A record start, matched independently of whether the rest of that record
+# survived the write that emitted it.
+_STRACE_HEAD_RE = re.compile(r"\[STRACE\]\s+v=\d+")
 
 
 @dataclass
@@ -98,6 +108,16 @@ class Invocation:
         for s in self.spans:
             m.setdefault(s.name, s)
         return m
+
+
+def count_record_heads(lines):
+    """Count ``[STRACE]`` record starts, torn ones included.
+
+    Pairs with :func:`parse_spans`, which yields only records that survived
+    intact. A shortfall between the two counts is instrumentation loss, and
+    without it a torn record is indistinguishable from a real measurement.
+    """
+    return sum(len(_STRACE_HEAD_RE.findall(line)) for line in lines)
 
 
 def parse_spans(lines):
@@ -496,6 +516,13 @@ def main(argv=None):
             lines = f.readlines()
 
     spans = list(parse_spans(lines))
+    heads = count_record_heads(lines)
+    if heads > len(spans):
+        print(
+            f"warning: {heads - len(spans)} of {heads} [STRACE] records are incomplete and are "
+            "excluded from the timing below",
+            file=sys.stderr,
+        )
     invocations = group_invocations(spans)
     buckets = bucket_by_hid(invocations)
 
--- a/simpler_setup/tools/strace_timing.py
+++ b/simpler_setup/tools/strace_timing.py
@@ -48,10 +48,15 @@ import sys
 from collections import defaultdict
 from dataclasses import dataclass, field
 
+_HOST_LOG_PREFIX = (
+    r"\[\d{4}-\d{2}-\d{2} \d{2}:\d{2}:\d{2}\.\d{6}\]"
+    r"\[T0x[0-9a-fA-F]+\]\[[A-Z]+\]\s+[^:\r\n]+:\s+"
+)
 _STRACE_RE = re.compile(
     r"\[STRACE\]\s+v=(?P<v>\d+)\s+pid=(?P<pid>\d+)\s+tid=(?P<tid>\d+)\s+"
     r"inv=(?P<inv>\d+)\s+hid=(?P<hid>[0-9a-fA-F]+)\s+depth=(?P<depth>\d+)\s+"
-    r"name=(?P<name>\S+)\s+ts=(?P<ts>\d+)\s+dur=(?P<dur>\d+)(?P<attrs>.*)"
+    r"name=(?P<name>\S+)\s+ts=(?P<ts>\d+)\s+dur=(?P<dur>\d+)(?P<attrs>.*?)"
+    rf"(?={_HOST_LOG_PREFIX}|\[STRACE\]|\r?$)"
 )
 
 
@@ -96,22 +101,20 @@ class Invocation:
 
 
 def parse_spans(lines):
-    """Yield Span for every matching [STRACE] line."""
+    """Yield every complete span, including adjacent records on one line."""
     for line in lines:
-        m = _STRACE_RE.search(line)
-        if not m:
-            continue
-        yield Span(
-            pid=int(m["pid"]),
-            tid=int(m["tid"]),
-            inv=int(m["inv"]),
-            hid=m["hid"].lower(),
-            depth=int(m["depth"]),
-            name=m["name"],
-            ts=int(m["ts"]),
-            dur=int(m["dur"]),
-            attrs=m["attrs"].strip(),
-        )
+        for m in _STRACE_RE.finditer(line):
+            yield Span(
+                pid=int(m["pid"]),
+                tid=int(m["tid"]),
+                inv=int(m["inv"]),
+                hid=m["hid"].lower(),
+                depth=int(m["depth"]),
+                name=m["name"],
+                ts=int(m["ts"]),
+                dur=int(m["dur"]),
+                attrs=m["attrs"].strip(),
+            )
 
 
 def group_invocations(spans):
--- a/src/common/log/host_log.cpp
+++ b/src/common/log/host_log.cpp
@@ -15,14 +15,58 @@
 
 #include "host_log.h"
 
+#include <cerrno>
 #include <chrono>
 #include <cstdio>
 #include <cstdlib>
 #include <cstring>
 #include <ctime>
 #include <pthread.h>
+#include <sstream>
+#include <string>
+#include <vector>
+
+#include <unistd.h>
 
 using simpler::log::LogLevel;
+
+namespace {
+
+std::string format_message(const char *fmt, va_list args) {
+    va_list sizing_args;
+    va_copy(sizing_args, args);
+    const int required = vsnprintf(nullptr, 0, fmt, sizing_args);
+    va_end(sizing_args);
+    if (required < 0) {
+        return {};
+    }
+
+    std::vector<char> buffer(static_cast<size_t>(required) + 1);
+    va_list formatting_args;
+    va_copy(formatting_args, args);
+    const int written = vsnprintf(buffer.data(), buffer.size(), fmt, formatting_args);
+    va_end(formatting_args);
+    if (written < 0) {
+        return {};
+    }
+    return std::string(buffer.data(), static_cast<size_t>(written));
+}
+
+void write_stderr(const std::string &record) {
+    size_t offset = 0;
+    while (offset < record.size()) {
+        const ssize_t written = ::write(STDERR_FILENO, record.data() + offset, record.size() - offset);
+        if (written > 0) {
+            offset += static_cast<size_t>(written);
+        } else if (written < 0 && errno == EINTR) {
+            continue;
+        } else {
+            break;
+        }
+    }
+}
+
+}  // namespace
 
 HostLogger &HostLogger::get_instance() {
     static HostLogger instance;
@@ -83,13 +127,18 @@ void HostLogger::emit(const char *level_tag, const char *func, const char *fmt, 
 
     auto tid = static_cast<unsigned long>(reinterpret_cast<uintptr_t>(pthread_self()));
 
-    std::scoped_lock lock(mutex_);
-    fprintf(stderr, "[%s][T0x%lx][%s] %s: ", ts, tid, level_tag, func);
-    vfprintf(stderr, fmt, args);
+    std::ostringstream stream;
+    stream << '[' << ts << "][T0x" << std::hex << tid << "][" << level_tag << "] " << func << ": "
+           << format_message(fmt, args);
+    std::string record = stream.str();
     if (fmt[0] != '\0' && fmt[strlen(fmt) - 1] != '\n') {
-        fputc('\n', stderr);
+        record.push_back('\n');
     }
-    fflush(stderr);
+
+    std::scoped_lock lock(mutex_);
+    // STRACE records fit within PIPE_BUF, so one write keeps each record
+    // indivisible when forked workers share a captured stderr pipe.
+    write_stderr(record);
 }
 
 void HostLogger::vlog(LogLevel level, const char *func, const char *fmt, va_list args) {
--- a/src/common/log/host_log.cpp
+++ b/src/common/log/host_log.cpp
@@ -22,8 +22,6 @@
 #include <cstring>
 #include <ctime>
 #include <pthread.h>
-#include <sstream>
-#include <string>
 #include <vector>
 
 #include <unistd.h>
@@ -32,30 +30,54 @@ using simpler::log::LogLevel;
 
 namespace {
 
-std::string format_message(const char *fmt, va_list args) {
-    va_list sizing_args;
-    va_copy(sizing_args, args);
-    const int required = vsnprintf(nullptr, 0, fmt, sizing_args);
-    va_end(sizing_args);
-    if (required < 0) {
-        return {};
-    }
+// Every STRACE marker renders well inside this, so the heap fallback below
+// stays off the path a traced run pays per span.
+constexpr size_t kRecordStackCapacity = 2048;
 
-    std::vector<char> buffer(static_cast<size_t>(required) + 1);
+// Renders the timestamp/thread/level prefix, the caller's message, and an
+// optional trailing newline into `buffer`, and returns the length of the whole
+// record. A return value of `capacity` or more means `buffer` holds only a
+// truncated prefix and the caller must re-render into that many bytes.
+size_t format_record(
+    char *buffer, size_t capacity, const char *ts, unsigned long tid, const char *level_tag, const char *func,
+    const char *fmt, va_list args, bool append_newline
+) {
+    size_t length = 0;
+    auto tail = [&]() -> char * {
+        return length < capacity ? buffer + length : nullptr;
+    };
+    auto remaining = [&]() -> size_t {
+        return length < capacity ? capacity - length : 0;
+    };
+
+    const int prefix = snprintf(tail(), remaining(), "[%s][T0x%lx][%s] %s: ", ts, tid, level_tag, func);
+    if (prefix < 0) {
+        return 0;
+    }
+    length += static_cast<size_t>(prefix);
+
     va_list formatting_args;
     va_copy(formatting_args, args);
-    const int written = vsnprintf(buffer.data(), buffer.size(), fmt, formatting_args);
+    const int body = vsnprintf(tail(), remaining(), fmt, formatting_args);
     va_end(formatting_args);
-    if (written < 0) {
-        return {};
+    if (body > 0) {
+        length += static_cast<size_t>(body);
     }
-    return std::string(buffer.data(), static_cast<size_t>(written));
+
+    if (append_newline) {
+        if (length + 1 < capacity) {
+            buffer[length] = '\n';
+            buffer[length + 1] = '\0';
+        }
+        length += 1;
+    }
+    return length;
 }
 
-void write_stderr(const std::string &record) {
+void write_stderr(const char *record, size_t size) {
     size_t offset = 0;
-    while (offset < record.size()) {
-        const ssize_t written = ::write(STDERR_FILENO, record.data() + offset, record.size() - offset);
+    while (offset < size) {
+        const ssize_t written = ::write(STDERR_FILENO, record + offset, size - offset);
         if (written > 0) {
             offset += static_cast<size_t>(written);
         } else if (written < 0 && errno == EINTR) {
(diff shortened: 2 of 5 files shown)


simpler_setup/tools/strace_timing.py
```diff
@@ -51,7 +51,7 @@
 _STRACE_RE = re.compile(
     r"\[STRACE\]\s+v=(?P<v>\d+)\s+pid=(?P<pid>\d+)\s+tid=(?P<tid>\d+)\s+"
     r"inv=(?P<inv>\d+)\s+hid=(?P<hid>[0-9a-fA-F]+)\s+depth=(?P<depth>\d+)\s+"
-    r"name=(?P<name>\S+)\s+ts=(?P<ts>\d+)\s+dur=(?P<dur>\d+)(?P<attrs>.*)"
+    r"name=(?P<name>\S+)\s+ts=(?P<ts>\d+)\s+dur=(?P<dur>\d+)(?P<attrs>.*?)(?=\[STRACE\]|\r?$)"
```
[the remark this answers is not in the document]
sunkaixuan2018: Fixed in `87a75b55e`. The `attrs` match now stops at the next complete host-log prefix, and the regression test requires exact equality for `rank=0`. The targeted Linux pytest and all relevant pre-commit checks pass.

src/common/log/host_log.cpp
```diff
@@ -15,15 +15,59 @@
...
+    while (offset < record.size()) {
+        const ssize_t written = ::write(STDERR_FILENO, record.data() + offset, record.size() - offset);
+        if (written > 0) {
+            offset += static_cast<size_t>(written);
+        } else if (written < 0 && errno == EINTR) {
+            continue;
+        } else {
+            break;
+        }
+    }
```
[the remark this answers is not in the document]
sunkaixuan2018: Thanks for calling out the boundary. I am not applying a logger-wide `PIPE_BUF` cap because that would silently truncate arbitrary diagnostic/error logs and regress the existing HostLogger contract. The atomicity requirement in this issue is specifically for STRACE records: their current names are fixed literals and their only dynamic attribute buffer is bounded to 192 bytes, so those records remain well below `PIPE_BUF`. Other host logs still benefit from a single write, but do not claim the STRACE atomicity guarantee.

simpler_setup/tools/strace_timing.py
```diff
@@ -48,10 +48,15 @@
...
+_HOST_LOG_PREFIX = (
+    r"\[\d{4}-\d{2}-\d{2} \d{2}:\d{2}:\d{2}\.\d{6}\]"
+    r"\[T0x[0-9a-fA-F]+\]\[[A-Z]+\]\s+[^:\r\n]+:\s+"
+)
 _STRACE_RE = re.compile(
     r"\[STRACE\]\s+v=(?P<v>\d+)\s+pid=(?P<pid>\d+)\s+tid=(?P<tid>\d+)\s+"
     r"inv=(?P<inv>\d+)\s+hid=(?P<hid>[0-9a-fA-F]+)\s+depth=(?P<depth>\d+)\s+"
-    r"name=(?P<name>\S+)\s+ts=(?P<ts>\d+)\s+dur=(?P<dur>\d+)(?P<attrs>.*)"
+    r"name=(?P<name>\S+)\s+ts=(?P<ts>\d+)\s+dur=(?P<dur>\d+)(?P<attrs>.*?)"
+    rf"(?={_HOST_LOG_PREFIX}|\[STRACE\]|\r?$)"
```
ChaoZheng109: **Should fix.** The `\r?$` alternative only anchors at the end of the *string* — `re.MULTILINE` is not set. Since `.` never crosses `\n`, `attrs` stops at a newline, and at that position none of the three alternatives can match, so the whole record fails to match.

Effect: if a caller passes a multi-line chunk instead of one line per item, every record except the last is silently dropped. Measured against this branch:

```
input: two complete records, each on its own line, joined into one string
  old regex (main): finds record 1
  this regex:       finds only record 2   <- record 1 lost
  + re.MULTILINE:   finds both
```

In-repo this does not bite, because `main()` uses `readlines()`. But `parse_spans` is a module-level function consumers (pypto-lib) can call with a blob, and silently losing records is exactly the failure class this PR is fixing.

Adding `re.MULTILINE` to the `re.compile(...)` call is enough. I checked it leaves every other case unchanged: single line with and without trailing newline, CRLF, two records on one line, and the real `[file:line]` prefix form.

simpler_setup/tools/strace_timing.py
```diff
@@ -96,22 +101,20 @@ def by_name(self):
 
 
 def parse_spans(lines):
```
ChaoZheng109, line 123: **Should fix.** This covers the first half of suggestion (2) in #1690 — find multiple markers on a line — but not the second, detecting malformed/torn input. A record torn mid-write still yields nothing and produces **no signal at all**, which is precisely how #1690 manifested: `effective_us=0.0` with no way to distinguish instrumentation loss from a real measurement.

Suggestion: count record heads per line and compare against the number of spans actually yielded; if the totals differ, print one warning to stderr in `main()` before the table. Cheap, and it converts a silent data loss into something the operator can see.

Matching the head (`[STRACE] v=`) rather than the bare `[STRACE]` literal keeps false positives away from prose that happens to mention the marker.

src/common/log/host_log.cpp
```diff
@@ -83,13 +127,18 @@ void HostLogger::emit(const char *level_tag, const char *func, const char *fmt,
 
     auto tid = static_cast<unsigned long>(reinterpret_cast<uintptr_t>(pthread_self()));
 
-    std::scoped_lock lock(mutex_);
-    fprintf(stderr, "[%s][T0x%lx][%s] %s: ", ts, tid, level_tag, func);
-    vfprintf(stderr, fmt, args);
+    std::ostringstream stream;
```
ChaoZheng109: **Consider (performance).** The single-`write` fix is right, but `ostringstream` is an expensive way to reach it — and this PR's whole purpose is trustworthy benchmark timing.

Per record this now does: a sizing `vsnprintf`, a `std::vector<char>` heap allocation, a second `vsnprintf`, a `std::string` copy, then an `ostringstream` (allocating again) for the prefix. Measured on this branch (aarch64, `-O2`, 200k calls through `HostLogger::log` with stderr on `/dev/null`, best of 3 runs):

| variant | ns/record |
| --- | --- |
| `main` (`fprintf` x4, non-atomic) | ~1400 |
| this PR (`ostringstream`) | ~2100 |
| stack buffer + `snprintf` | ~1000 |

The atomicity does not require the extra cost. Rendering the whole record (prefix + message + newline) with `snprintf`/`vsnprintf` straight into a fixed stack buffer, falling back to a heap buffer only when it does not fit, still emits exactly one `write` — and lands ~30% *faster* than `main` instead of ~50% slower. `snprintf` returns the length it needed, so "did it fit" is just comparing that against the buffer size, and the second render only happens on the rare oversized record.

Why it matters here specifically: a span's own `dur` is sampled before its marker is emitted, but a **parent** span's `dur` includes every child's emit cost, and `PYPTO_BENCH`'s per-round wall clock includes all of it. At ~10 spans per decode invocation that is a few microseconds added to every measured round.

Mitigating: it is only paid when tracing is on (`is_enabled` short-circuits above TIMING) — but that is exactly the configuration this PR targets.

The onboard AICPU backend already uses this shape (`char buffer[2048]` + one `dlog_*` call, `src/common/platform/onboard/aicpu/device_log.cpp`), so it would also make the two backends consistent.

src/common/log/host_log.cpp
```diff
@@ -83,13 +127,18 @@ void HostLogger::emit(const char *level_tag, const char *func, const char *fmt,
...
+           << format_message(fmt, args);
+    std::string record = stream.str();
     if (fmt[0] != '\0' && fmt[strlen(fmt) - 1] != '\n') {
-        fputc('\n', stderr);
+        record.push_back('\n');
     }
-    fflush(stderr);
+
+    std::scoped_lock lock(mutex_);
+    // STRACE records fit within PIPE_BUF, so one write keeps each record
```
ChaoZheng109: **Should fix.** This holds for STRACE markers but not for `HostLogger`'s other callers, and the comment reads as an unconditional invariant.

A long `LOG_ERROR` — a diagnostic dump, say — pushes the record past `PIPE_BUF`, at which point `write_stderr`'s partial-write loop is exactly where tearing happens, silently.

Suggest wording it as the conditional it actually is: records at or below `PIPE_BUF` are indivisible, longer ones can still be torn. Per `.claude/rules/comments.md` a comment should state a fact that holds as written.

src/common/log/include/host_log.h
```diff
@@ -17,8 +17,7 @@
  * both TIMING and WARN thresholds to CANN WARN.
  */
 
-#ifndef PLATFORM_HOST_LOG_H_
-#define PLATFORM_HOST_LOG_H_
+#pragma once
```
ChaoZheng109, line 20: **Consider.** This guard swap is unrelated to the fix. The repo carries both styles (67 files use `#pragma once`, 125 use include guards), so it is not a style violation — but per `.claude/rules/discipline.md` section 2 every changed line should trace back to the request, and an unrelated hunk costs reviewer attention and invites conflicts.

The `vlog` comment update just below is the opposite case and should stay: the old text claimed the path avoided "an intermediate vsnprintf-to-buffer round-trip", which this PR makes untrue, so updating it is required doc-consistency work.

tests/ut/py/test_strace_timing.py
```diff
@@ -0,0 +1,34 @@
...
+# Please refer to the License for details. You may not use this file except in compliance with the License.
+# THIS SOFTWARE IS PROVIDED ON AN "AS IS" BASIS, WITHOUT WARRANTIES OF ANY KIND, EITHER EXPRESS OR IMPLIED,
+# INCLUDING BUT NOT LIMITED TO NON-INFRINGEMENT, MERCHANTABILITY, OR FITNESS FOR A PARTICULAR PURPOSE.
+# See LICENSE in the root of the software repository for the full text of the License.
+# -----------------------------------------------------------------------------------------------------------
+
+from simpler_setup.tools.strace_timing import parse_spans
+
+
+def _marker(pid, inv, name, attrs=""):
```
ChaoZheng109: **Consider.** This fixture is not quite the real wire shape. `LOG_TIMING` prepends `[%s:%d] ` (`__FILENAME__`, `__LINE__`) to the caller's format string (`src/common/log/include/common/unified_log.h:45`), so on stderr the marker is preceded by e.g. `[strace.h:132] ` rather than sitting flush against the `func: ` separator. I checked that the regex handles the real form correctly, so this is not a bug — but the test guards a simplified input, which weakens it as a regression barrier.

Also worth a comment near `_HOST_LOG_PREFIX`: its `[^:\r\n]+` for the func segment relies on `func` being `__FUNCTION__`, i.e. an unqualified name. If a qualified name containing `::` ever reached it, the prefix alternative would stop matching and `attrs` would silently absorb the next record's timestamp prefix — the `[STRACE]` alternative would still recover the following span, so it would degrade quietly rather than fail loudly. Worth pinning down as an assumption.

ChaoZheng109 (commented):
Reviewed the merge-base diff (`639b95fbe...87a75b55e`). Summary below, details in the inline comments.

## What the PR does

`HostLogger::emit` wrote one logical record through four stdio calls under a **process-local** mutex, so forked chip workers sharing a captured stderr interleaved partial records; the parser then took one record per line with a greedy `attrs`, losing anything that followed. This PR buffers each record and emits it with a single `write`, and teaches the parser to read every marker on a line. Root cause and mechanism both look correctly identified.

The C++ regression test is well built: 16 forked children through a shared pipe, a start-pipe barrier so they actually contend, the reader thread started *before* the children are released so a full 64K pipe cannot deadlock it, and a per-line check that no record was spliced. CI is green across all 17 checks.

## Against the suggestions in #1690

- **(1)** contiguous buffer + one atomic write — done.
- **(2)** harden `parse_spans` — half done. Multiple markers per line are handled; torn records are still dropped silently. See the inline comment on `parse_spans`.
- **(3)** validate rank/round counts and **(4)** stop reporting missing spans as `0.0` — **correctly out of scope**, and worth stating explicitly in the PR description: `effective_us` has zero occurrences in this repo, it lives in pypto-lib's benchmark. This is not a goal downgrade.
- **(5)** multi-process test — done.

## Missing documentation

"one record, one `write`, indivisible up to `PIPE_BUF`" is now a contract that multi-process consumers depend on, and nothing records it. Two places need it in this same commit (`.claude/rules/doc-consistency.md` section 4):

- **`docs/logging.md`** section *Output formats* — state the guarantee, that it rests on the single `write` rather than on `mutex_` (which each forked process holds its own copy of), and that records longer than `PIPE_BUF` are still splittable.
- **`docs/dfx/host-trace.md:25`** — "One line per span" is no longer true now that a physical line can carry several markers. That file is the marker-grammar reference, so someone reading it would write exactly the one-match-per-line parser this PR is fixing.

## Same defect class in the sim backend (not this PR's job)

`src/common/platform/sim/aicpu/device_log.cpp` still emits a record via `fprintf` + `vfprintf` + `fputc` with **no lock at all**, so sim interleaves across forked chip workers *and* across plain AICPU sim threads — `aicpu_thread_num > 1` alone is enough, no fork needed. Filed separately as #1697. Not asking for it here, but a line in the PR description would stop readers assuming the defect class is gone repo-wide. Onboard is unaffected: it buffers into `char buffer[2048]` and issues one `dlog_*` call.

## Verdict

Approve with comments. Worth resolving before merge: the `re.MULTILINE` anchor and the torn-record counter — both small, and both close the same "records vanish silently" gap this PR exists to fix. The `ostringstream` performance note is optional, but there is a concrete alternative that is faster *and* keeps the atomicity.

simpler_setup/tools/strace_timing.py
sunkaixuan2018: Fixed in dffab7e1c — added `re.MULTILINE` to the `re.compile(...)` call.

Confirmed your reproduction on this branch: two complete records joined into one string, the first was dropped silently. With `MULTILINE` both are found. I re-checked the cases you listed and they are unchanged — single line with and without a trailing newline, CRLF, two records on one physical line, and the real `[file:line]` prefix form.

`test_parse_spans_keeps_every_record_of_a_multi_line_blob` now pins this.

simpler_setup/tools/strace_timing.py
sunkaixuan2018: Fixed in dffab7e1c, taking your suggestion as described.

`count_record_heads(lines)` counts `[STRACE] v=` heads independently of whether the rest of the record survived, and `main()` prints one stderr warning before the table when that total exceeds the spans actually parsed:

    warning: 1 of 2 [STRACE] records are incomplete and are excluded from the timing below

Matching the head rather than the bare `[STRACE]` literal, as you suggested. `test_count_record_heads_sees_a_torn_record_that_parse_spans_drops` covers it.

src/common/log/host_log.cpp
sunkaixuan2018: Fixed in dffab7e1c — `ostringstream` is gone, replaced by the stack-buffer + `snprintf` shape you described, with a heap buffer allocated only when the record does not fit (`snprintf` reports the needed length, so the second render only happens on the rare oversized record).

I reproduced your benchmark independently and got the same ordering (aarch64, `-O2`, 200k calls through `HostLogger::log`, stderr on `/dev/null`, three runs each):

| variant | ns/record |
| --- | --- |
| `main` (`fprintf` x4, non-atomic) | 1401–1450 |
| previous commit (`ostringstream`) | 2428–2472 |
| this commit (stack buffer + `snprintf`) | 926–982 |

So ~34% faster than `main` rather than ~74% slower, and the single `write` is unchanged.

Two notes on the implementation:

- The record is now formatted *outside* the mutex; only the `write` is under it. That was already true in intent, but the `ostringstream` version built the string before taking the lock too, so this is not a new property — just no longer paying for it.
- `kRecordStackCapacity` is 2048, matching the AICPU backend you pointed at. The fork test in this PR writes ~2.1 KB records, so it exercises the heap fallback rather than skipping over it.

src/common/log/host_log.cpp
sunkaixuan2018: Fixed in dffab7e1c. You are right that it read as an unconditional invariant, and it is not one. New wording:

    // One write per record: a record of at most PIPE_BUF bytes reaches a shared
    // pipe indivisibly, so forked workers writing a captured stderr cannot
    // interleave inside it. A longer record — a large LOG_ERROR dump, say — has
    // no such guarantee; STRACE markers stay well below the limit.

This also answers the CodeRabbit finding on the same lines, which asked for a hard `PIPE_BUF` cap. I did not add one: truncating or rerouting every long record would regress large `LOG_ERROR` diagnostic dumps to fix a case that only arises when several forked processes share one stderr pipe *and* a single record exceeds `PIPE_BUF`. Stating the boundary honestly is the cheaper correct answer.

src/common/log/include/host_log.h
sunkaixuan2018: Keeping this one, but flagging the reason rather than just overruling you — happy to revert if you still disagree.

`.claude/rules/codestyle.md` §12 makes `#pragma once` the convention and explicitly sanctions this edit:

> Use `#pragma once` for header guards, not `#ifndef`/`#define`/`#endif`. […] Existing files are converted on sight during routine edits; a dedicated sweep is not required.

"On sight during routine edits" is exactly this case — the PR already modifies this header for the `vlog` comment you noted should stay, so the file was being touched regardless. The 67/125 split you measured is the migration in progress rather than a settled two-style equilibrium.

I read `discipline.md` §2 as aimed at drive-by reformatting and unrelated refactors; a conversion an always-loaded rule instructs you to make while editing the file is the sanctioned exception. If you would still rather keep the diff minimal, say so and I will drop the hunk.

tests/ut/py/test_strace_timing.py
sunkaixuan2018: Fixed in dffab7e1c — both parts.

The fixture is now built by a `_record()` helper that emits the real wire shape, including the `[strace.h:132] ` segment `LOG_TIMING` prepends, so the marker no longer sits flush against the `<func>: ` separator. All three parser tests share it.

And the `[^:\r\n]+` assumption is now written down at `_HOST_LOG_PREFIX`:

    # The host-log record prefix `HostLogger::emit` writes ahead of every message.
    # Its func segment excludes : because `LOG_TIMING` passes `__FUNCTION__`, an
    # unqualified name; a qualified name containing :: would stop this alternative
    # from matching, leaving the `[STRACE]` alternative to bound the record instead.

That records both the assumption and the degradation mode you described.

sunkaixuan2018:
Addressed @ChaoZheng109's review in dffab7e1c. Five of the six are fixed; one I am pushing back on with a rule citation, in that thread.

- **`re.MULTILINE`** — fixed. Reproduced the multi-line-blob drop first; both records survive now.
- **Torn-record signal** — fixed. `count_record_heads()` plus a one-line stderr warning from `main()`.
- **`ostringstream` cost** — fixed, replaced by stack buffer + `snprintf` with a heap fallback. Independently re-measured: 1401–1450 ns/record on `main`, 2428–2472 on the previous commit, 926–982 now (aarch64, `-O2`, 200k records, best of 3).
- **`PIPE_BUF` comment** — fixed, now states the boundary as the conditional it is. This also answers CodeRabbit's request for a hard cap on the same lines; I explain there why capping would regress long `LOG_ERROR` dumps.
- **Test fixture wire shape** — fixed, the fixture now carries the `[<file>:<line>] ` segment `LOG_TIMING` prepends, and the `__FUNCTION__` assumption is documented at `_HOST_LOG_PREFIX`.
- **`#pragma once`** — keeping it; `.claude/rules/codestyle.md` §12 directs converting headers "on sight during routine edits", and this PR edits the header anyway. Will drop the hunk if you would still rather keep the diff minimal.

Validation on an aarch64 box: 80/80 non-hardware C++ tests pass, the forked-pipe test passes 10/10 standalone, the 3 parser tests pass, and pre-commit is clean on all touched files.

ChaoZheng109 (approved):

PR #1694, opened 2026-08-05
Refactor: run native progress on the child loop

Crane-Liu:
## Summary

- enqueue native runs directly from the child-owned progress path and drain them from wait/finalize
- remove the per-run executor, launch signal, completion atomics, and `Launching` phase
- publish sticky launch acceptance after successful enqueue and preserve phased trace spans

## Validation

- `pip install --no-build-isolation -e .` (current-source onboard build)
- `pytest tests/ut/py/test_worker/test_host_worker.py -q` (216 passed)
- `task-submit`: a2a3 native run lifecycle (1 passed)
- `task-submit`: a2a3 worker async FIFO (3 passed)

- LCW CANN editable build and `test_native_run_execution` passed at the current head.\n- The rebased CI passed all four simulation jobs, both host UT jobs, `ut-a5`, `st-onboard-a5`, `st-pod-onboard-a2a3`, and profiling smoke; `st-onboard-a2a3` includes the host-build-graph dep_gen assertion.


ChaoWao (commented):
Reviewed at `8314b780` (base `0e3851a4`). This is the best-shaped version of B6c I've seen — the rebase onto #1714 both resolved the two red a5 checks and shrank the diff from 32 files / +1413 −1077 to **21 files / +156 −508**. Deleting the TLS transfer mechanism outright rather than replacing it with a shim, and dropping `Launching` from *both* state machines, are the right calls. `publish_acceptance` keeping `receipt.matches(identity())` as the only key preserves the plan's "acceptance only from `LaunchReceipt`" invariant, and the plan's explicit prohibition on temporary TLS park/adopt is not just respected but exceeded.

Two things I'd like addressed before merge, one of them a one-liner.

---

### 1. Must fix — `drain_execution` now runs without an attached device context

Both new drain call sites run before any `attach_current_thread`:

- `simpler_wait_run` — onboard `c_api_shared.cpp:822`, sim `:729` — no attach anywhere in the function.
- `simpler_finalize_run` — onboard `:848`, sim `:753` — the function *does* attach, but at `:859` / `:764`, i.e. **after** the drain block.

Every other lifecycle entry point attaches first (prepare `:714`, launch `:771`, poll `:807`), and the code this replaces attached explicitly: the executor lambda's first act was `attach_current_thread`, with `drain_execution` inside the `attach_rc == 0` branch.

The attach isn't decorative — `attach_current_thread` → `rtSetDevice`, and its own comment says *"CANN device context is per-thread, so every caller must attach explicitly."* `drain_execution` then does `reap_run` → `aclrtSynchronizeStreamWithTimeout` on both streams, `retire_run_aicore_stream` (stream destroy), `print_handshake_results` (device memory read), and via its RAII guard `cleanup_execution` → `rtFree`. All of those need the context.

**Why CI is green anyway:** the child progress loop is single-threaded, so the launch-time attach on the same thread covers both drains. That makes this latent, not currently-broken. But `simpler_wait_run` is public C API and its binding is `_wait_native_run` with `nb::call_guard<nb::gil_scoped_release>()` — explicitly designed to be called off the GIL — and nothing documents a same-thread requirement. `native_run_lifecycle`'s ST calls it from the prepare thread, so it can't discriminate.

Fix is the same shape as the siblings, and cheap — `rtSetDevice` on an already-attached thread is idempotent, which is why `poll` calls it unconditionally every iteration:

```cpp
int attach_rc = state->runner->attach_current_thread(state->runner->device_id());
if (attach_rc != 0) { /* record completion_rc, publish Complete, return */ }
```

In `finalize`, moving the existing attach above the drain block would do it.

### 2. Should fix — the dep_gen host-graph's same-thread invariant is now implicit, and its docs describe the deleted mechanism

`dep_gen_host_graph` keeps its capture in `static thread_local HostGraphState` (`dep_gen_host_graph.cpp:233`). The graph is built during **prepare** (`bind_callable_to_runtime_impl` → `run_host_orchestration` → `begin_capture()`, `runtime_maker.cpp:474`) and emitted during **drain** (`reap_run` → `dep_gen_host_graph_emit`, `device_runner.cpp:675`). Pre-PR those were different threads — which is exactly why `take_native_run_thread_state` / `adopt_native_run_thread_state` existed.

Deleting that transfer is correct now that prepare and drain land on one thread. But the invariant that makes it correct is unstated and unenforced, and its failure mode is quiet: `emit` returns `-3` and logs *"no capture was adopted on this thread"* while the run itself succeeds and no `deps.json` appears.

Two concrete bits:

- **`dep_gen_host_graph.h:34,39-43` (both arches) still document the deleted mechanism.** The PR edits that paragraph but only swaps "executor's thread-local state" → "progress thread's local state", so it still claims prepare *moves the graph into run-owned storage* and launch *adopts that snapshot* — neither happens now. Same for the emit error string (`dep_gen_host_graph.cpp:527`), which still refers to "the run-owned snapshot". Per `doc-consistency.md` §3 this needed a rewrite rather than a word swap.
- **`take_capture` / `adopt_capture` / `destroy_capture` remain defined and exported** in both a2a3 and a5 hbg runtimes with **zero production callers** — only `tests/ut/cpp/a2a3/test_dep_gen_host_graph.cpp` uses them. They read as live API.

Either resolution works for me: delete the now-dead capture API (preferred — removes the ambiguity entirely), or keep it and state the invariant as a present-tense fact at the emit site. What I'd avoid is shipping the current state, where the enforcement mechanism is gone but its documentation and exported surface both remain.

Worth noting the safety net is real: the hbg dep_gen ST is wired into a2a3 onboard CI with `--enable-dep-gen` (`_st-npu-a2a3.yml:142`) and asserts all 6 edges — so a genuine break would show up in `st-onboard-a2a3`.

---

### Smaller items

**3. Two deleted tests covered invariants that survive.** `test_native_run_launch_signal.cpp` goes away wholesale. Three of its five cases tested the CV handoff and are correctly gone, but two didn't:

- `StaleReceiptCannotPublishAcceptance` — `publish_acceptance` still implements exactly this gate, and it's the load-bearing half of "acceptance only from `LaunchReceipt`".
- the first half of `PublishAcceptanceStoresOnceAndKeepsNotificationSticky` — asserts the store lands with the right value.

Both port to `publish_acceptance` nearly verbatim, no threading needed. As it stands the acceptance-identity gate — the one mechanism stopping a stale run from marking a task accepted — loses its only direct test; `test_native_run_execution.cpp` covers the `LaunchReceipt` side but never the `accepted_state` store.

**4. `catch (...) {}` around both drains swallows the reason.** The rc still propagates (it stays at the `-1` initializer), so the logic is right — but pre-PR that silent catch sat at a thread boundary where an escaping exception would `terminate`; here it's a plain call in a function that already returns rc. `drain_execution` is where op-timeouts and 507018s surface, so a `LOG_ERROR` in each handler is probably the highest-value line in the file for triage.

**5. `launched` changed meaning in `finalize`, for the better — worth a line in the body.** It went from `phase != Prepared` ("launch was attempted") to `active_execution != nullptr` ("launch produced an active execution"). These differ for a `NotStarted` launch failure, and the new one is more correct: such a run never touched a stream, so `validate_runtime_impl` should get `-1` and `set_gm_sm_ptr(nullptr)` should run — which is what the new code does. A real behavior improvement hiding inside a refactor; a future bisect would otherwise attribute it to nothing.

---

### CI

The body still says *"Simulation validation was intentionally not run for this change"* — that's now moot: all four sim jobs are **green**, as are `ut-a5` and `st-onboard-a5` (both red before the rebase), `st-pod-onboard-a2a3`, `profiling-flags-smoke`, and both `ut` hosts. Only `st-onboard-a2a3` and one packaging/sim job are still in flight at the time of writing — and `st-onboard-a2a3` is the one carrying the hbg dep_gen assertion above, so it's worth waiting on specifically.

Worth updating that line in the body once it settles, so the PR record doesn't understate what was actually verified.

Crane-Liu:
@ChaoWao Addressed the B6c review:
- Attached the current thread before wait/finalize drains on both onboard and sim; attach failure is returned as the run completion result.
- Kept host-graph capture on the child progress thread, rewrote the stale snapshot/adoption docs and emit diagnostic, and retained the explicit transfer helpers for callers that deliberately transfer a capture.
- Restored direct receipt-bound acceptance coverage, including the stale-receipt rejection.
- Logged drain exceptions in wait/finalize and updated the PR validation record with the completed simulation CI coverage.

PR #1729, opened 2026-08-07
Update: cut task args over to the self-describing Tensor wire ABI

YunjiQin:
## The wire flip

#1599 froze the address-free `Tensor`; nothing dispatched one. This makes `TaskArgs`
carry it.

Before, `TaskArgs` carried the GM-address-bearing `ChipTensor`, so a sender had to know
its receiver's address space. `_rewrite_blob_host_addrs` patched addresses by **numeric
range** and mis-rewrote device pointers that happened to fall inside a registered host
range — patched by adding a `child_memory` skip whose own comment recorded the hazard.
A consumer now resolves a backing by canonical identity: map-once, exact, and no address
crosses a process boundary at all.

User guide: [`docs/buffer-abi.md`](../blob/p1-b-wire-cutover/docs/buffer-abi.md).

### One wire, not two

The wire is the mailbox blob that was already in `task_args.h`. `TaskArgs`,
`TaskArgsView`, `write_blob` and `read_blob` change **what they carry, not how many of
them exist**; `ChipTensor` survives only inside `ChipStorageTaskArgs`, which the L2 leaf
materializes into and hands to `runtime.so`.

`TaskArgsView::tensors` now runs `validate_tensor` on every element it decodes. That is
load-bearing: the element gained wire invariants a decoder can check, and since the bound
types expose their fields but not their bytes, **this is the only place a mailbox blob
becomes a `Tensor`**. There is no second way in to forget it on.

`ChipWorker` takes only `ChipStorageTaskArgs`. The blob round trip it used to sit behind
— C++ serialising bytes that Python handed straight back to C++ — is gone, along with the
`TaskArgsView` overloads that existed to unpack it.

### A copy names both of its ends

`copy_to` / `copy_from` take a `Buffer` on each side and write the two descriptors into
the control frame; the child resolves them through the same `ImportRegistry` its task
arguments go through. That replaces **both** the raw host address — meaningless across a
fork once the MAP_HOST pre-attach is gone — and the shm-name path that briefly stood in
for it, which was a second resolution rule for the one thing this ABI exists to resolve.

`ControlCopyRequest` is 184 B against 64 KiB of control room, `static_assert`-ed, and both
descriptors are validated before the child acts on either.

### Every receive path keeps its container

This is where the port was initially wrong, twice, and both are fixed here:

- A **nested next-level child** re-exports each backing to a local handle and still hands
  its orchestration function a `TaskArgs`, **scalars included**. Re-export changes what is
  inside the container, not the container. An intermediate shape of `list[Tensor]` silently
  dropped scalars — the compute-leaf path had kept them, so the two receive paths disagreed.
- A **remote L3 runner** resolves its sidecar descriptors into a `TaskArgs` too, so an
  orchestration function can forward the args it was given whichever way it was reached.
  It could not before: `_materialize_task_args` produced the chip POD while
  `_split_next_level_args` accepts only the wire type, so a remote L3 could not submit to
  its own chip children at all.

Remote buffers are allocated through `create_buffer`, so what a runner hands back carries
an identity and a descriptor instead of being a bare `SharedMemory` belonging to neither
mechanism.

### Naming

`simpler_setup.torch_interop.make_tensor_arg` → `make_chip_tensor_arg`. It returns the chip
POD where `Worker.make_tensor_arg` returns the wire `Tensor`; one name for both was the
second public meaning [rule 13](../blob/p1-b-wire-cutover/.claude/rules/codestyle.md) forbids.

`wrap_fork_inherited` callers name their backend explicitly now that #1703 stopped inferring
it from `access`. Each already knew the answer in a comment: the HeapRing backings are
MAP_SHARED, and `make_tensor_arg` follows the `shared` it already computes — at L2 the
consumer is this process, so a write reaches the owner trivially and FORK_COW's contract is
the one that would be false there.

## Verification

- [x] Build green — four arch×runtime sim trees, four onboard, plus the nanobind extension
- [x] `pyut` — 1230 passed, 6 skipped
- [x] `cpput` — 86/86. Note this was **85/86 with 30 cases never executing**: a new test
      helper tagged a descriptor `POSIX_SHM` without a body, the gate above threw from a
      worker thread, and `std::terminate` took the binary down mid-suite. Bisected to the
      first commit of this work, not pre-existing as twice claimed.
- [x] `a2a3sim` / `a5sim` full scene suites — rc=0, 0 failures on both
- [x] `examples/workers/l4/vector_add_mixed_l3` on **a2a3 silicon**, both halves:
      `local max_diff=0.000e+00`, `remote max_diff=0.000e+00`. This is the `_st-pod.yml`
      payload; a loopback daemon stood in for the second machine.
- [x] pre-commit green
- [ ] Onboard scene lanes (`--platform a2a3` / `a5` over `examples tests/st`) — not run

## Known, and deliberately not in this PR

- `IMPORT_BUFFER` maps the owner's whole shm so `entry.addr` is the whole backing's base,
  while `entry.offset` is set as if it were the exported sub-range's base. A non-zero
  `worker.remote_export(handle, offset=N)` therefore resolves N bytes short. **Pre-existing**
  — this PR reproduces the arithmetic deliberately — and untested either way: every test
  exports at offset 0.
- The remote **task** path still materializes from addresses resolved against the runner's
  own registry. Carrying `BufferDescriptor`s on the sidecar wire is P2.
- `REMOTE_SIDECAR` remains rejected at materialization, as #1599 froze it.

## Reviewing this

The single commit is large because the flip is not separable — the wire type, its decoders,
and every call site have to move together or nothing builds. The order that makes it
readable: `src/common/task_interface/task_args.h` (what the wire carries and where it is
validated), then `python/simpler/buffer.py` and `worker.py` (the receive paths), then
`remote_l3_session.py`, then the call sites.


ChaoWao:
Rebased onto `main` and folded in four doc/comment fixes. Still one commit; force-pushed as `2822c4d1`. Nothing in the mechanism changed — the file set is identical to before apart from `docs/buffer-abi.md`.

### The rebase was not mechanical — one silent bug it would have introduced

`main` gained #1739 (retire the blocking endpoint channel) and #1728 (uniformly progressable endpoints) since this branch's base, which moved tasks onto a **dedicated task frame**. Six conflicts; the one worth flagging:

In `_chip_process_loop`, `args_ptr = mailbox_addr + _OFF_TASK_ARGS_BLOB` **merged cleanly and was wrong**. Post-#1739, `mailbox_addr` is the *base* frame (controls + shutdown) while the task arrives on `task_addr`. Git had no reason to conflict there — this branch changed the line's right-hand side and `main` changed the variable's meaning. Left as-is, every chip task would have materialized whatever stale bytes sat in the base frame. Now `task_addr`.

The rest:

- `LocalMailboxEndpoint::run` / `run_with_accept` — deleted by #1739; took the deletion. The surviving `submit_progress` already writes `write_blob(blob, a)`, so the cutover is complete on the path that remains. Dropped its now-stale "matching the non-pipelined `run()` path" comment.
- `remote_endpoint.cpp` ×4 — this branch's logic, `main`'s `build_task_payload` message prefix (the function was renamed).
- `test_scheduler.cpp` ×2 — `main`'s frame selection + this branch's `read_blob` / identity-keyed decode.
- `test_remote_endpoint.cpp` — `main`'s `submit_progress` + try/catch, matching on `"local backing"`.
- `test_host_buffer_registration.py` — modify/delete; took the delete (`register_host_buffer` has 0 references left; `main` had only adapted it to the new frame protocol).

### Four fixes

1. **`docs/buffer-abi.md` was left behind, and it is published.** Its status note still said `add_tensor` takes a `ChipTensor` and that `Tensor` is deliberately *not* re-exported from `simpler.task_interface`; the scope section still said "**the dispatch wire is not connected yet**" and that the submit-time checks are unreachable. This PR's own body links that page as the User guide. Brought current, and what is still missing is now stated positively (the two other allocators; the endpoint × `address_space` check inside `materialize`).
2. `orchestrator.cpp` `reserve_slot` comment still offered `create_host_buffer` as an example allocator — 0 references left.
3. `MappedArg.buffer` labelled `FORK_SHM` as copy-on-write. It is the `MAP_SHARED` one — that distinction is why the two fork backends exist, and #1703 fixed this exact mislabel; the corrected wording is ~60 lines below in `materialize`.
4. Comments describing `ChipTensor::child_memory` now name a field called `address_space`. **Kept the Python keyword** `child_memory` — it is the name of a u8 on the remote-L3 tensor wire (`remote_wire.cpp` `encode_tensor`), so renaming the keyword would not change the contract — and added a note saying so, since a reader otherwise sees one field with two names and no reason.

### Verified locally on the rebased commit

- pyut **1223 passed, 13 skipped, 0 failed**
- cpput **90/90** (`ctest -LE requires_hardware`)
- ruff check / ruff format / clang-format clean

### Not done here (deliberate)

`materialization_map()` is rebuilt per task dispatch (cost is O(all buffers the child has ever seen), not O(args)); the chip blob is decoded twice per task; `ImportRegistry.unregister` has no caller or test; `MappedArg.buffer` hands out a writable memoryview regardless of `access`; `test_l3_host_buffer_registration.py` is still named for a removed mechanism; and `simpler_setup.Tensor` now collides with the newly re-exported `simpler.task_interface.Tensor` (rule 13). The last one is a public-API question and does not belong in a 126-file PR. Happy to open follow-ups.

One process note: CodeRabbit skipped this PR entirely ("125 files exceed the limit of 100"), so it has had no automated review.

PR #1738, opened 2026-08-07
Refactor: move comm-domain demos to scene tests

### LINKED ISSUE #1616: [Code Health] L3 comm-domain examples hand-roll a Worker and their own verification, which @scene_test already supports
### Category

### Component

Tests

### Description

Seven example tests that allocate a communication domain are plain pytest
functions that construct a `Worker(level=3, ...)` by hand, drive it, and then
compare results with hand-written tolerance checks. They total **1391 lines**.

The reason to look at them again is that **the scene-test framework already
supports everything they do** — this was checked rather than assumed, because
the obvious guess is that comm domains are what `@scene_test` cannot express.
They are not:

| Capability the demos need | Already used by |
| ------------------------- | --------------- |
| L3 `@scene_test` class | 8 classes: 5 collectives suites + `test_l3_dependency` / `test_l3_group` / `test_l3_host_buffer_registration` |
| Python orchestration function | `CALLABLE["orchestration"]`, invoked at `simpler_setup/scene_test.py:1402` |
| `orch.allocate_domain(...)` with `CommBufferSpec` | `tests/st/worker/collectives/_helpers.py:117` and `:186`, inside `allreduce_orch_fn` — the same call shape as the demos |
| **Per-rank** goldens | `tests/st/worker/collectives/all_to_all/test_all_to_all.py:103` — `compute_golden` writes a different expected tensor per rank via `out_{r}` |
| `device_count`, `num_sub_workers`, `enable_sdma` | `CASES[*]["config"]`; `enable_sdma` since #1609 |

So the split in the tree is not between "L3 with domains" and "L3 without" — the
collectives suites are L3 *with* domains and are ordinary scene tests. The seven
demos are simply older than the framework support.

What being plain functions costs them:

- **Verification is hand-rolled.** Each file computes a max-abs-diff, ANDs it
  into a local `ok`, and `run()` returns `0`/`1` for `assert run(...) == 0`
  (3–4 such sites per file). There is no golden tensor and nothing checks that
  the comparison itself is still meaningful — a mis-edited tolerance passes
  silently, where `compute_golden` would not.
- **No case parametrization**, so `--rounds` / `--case` and multi-shape coverage
  are unavailable.
- **Device allocation is not the dispatcher's**, so these run in the Resource
  phase as standalone subprocesses rather than sharing the pooled L2/L3 Worker.
- **a2a3 and a5 copies drift independently.** The pairs differ by only 8–21
  lines (`async_notify_demo` 11, `deferred_notify_demo` 8,
  `sdma_async_completion_demo` 21), and `platforms` is a `CASES` field — so the
  duplication exists only because the hand-rolled form has nowhere to put it.

**Counter-argument, stated deliberately:** these are *examples*, and a demo that
reads top-to-bottom without framework indirection has real value for someone
learning the API. The same argument is what closed golden vectorisation in
[`docs/investigations/2026-07-qwen-scene-test-406s-decomposition.md`](../blob/main/docs/investigations/2026-07-qwen-scene-test-406s-decomposition.md)
("a reference implementation whose readability is the point"). This issue is
therefore a *decision request* as much as a cleanup: converting all seven is
mechanical, but whether examples should read like `tests/st/` is a judgement the
maintainers should make before anyone starts.

### Location

The seven files, all under `examples/`:

```text
a2a3/tensormap_and_ringbuffer/async_notify_demo/test_async_notify_demo.py           179
a2a3/tensormap_and_ringbuffer/deferred_notify_demo/test_deferred_notify_demo.py     191
a2a3/tensormap_and_ringbuffer/sdma_async_completion_demo/test_*.py                  209
a5/tensormap_and_ringbuffer/async_notify_demo/test_async_notify_demo.py             176
a5/tensormap_and_ringbuffer/deferred_notify_demo/test_deferred_notify_demo.py       191
a5/tensormap_and_ringbuffer/sdma_async_completion_demo/test_*.py                    212
a5/tensormap_and_ringbuffer/urma_deferred_completion_demo/test_*.py                 233
```

Hand-rolled Worker, e.g.
`examples/a2a3/tensormap_and_ringbuffer/sdma_async_completion_demo/test_sdma_async_completion_demo.py:128`.

The template to follow: `tests/st/worker/collectives/` — `_helpers.py` (265
lines, shared orch functions) plus thin per-case classes (`all_to_all` is 131
lines and covers several rank counts).

Two `tests/ut/py/test_worker/` files also use `allocate_domain`
(`test_dynamic_alloc_hw.py`, `test_dynamic_alloc_sim.py`). Those are unit tests
and are **out of scope** — they should stay plain functions.

### Proposed Fix

If the decision is to convert, the shape is already proven by the collectives
suites:

1. Factor each demo's `orch_fn` into a shared helper module, as
   `tests/st/worker/collectives/_helpers.py` does.
2. Replace each `run()` with a `@scene_test(level=3, runtime=...)` class whose
   `CASES` carry `device_count` / `num_sub_workers` / `enable_sdma` and whose
   `platforms` list absorbs the a2a3-vs-a5 split, collapsing the near-duplicate
   pairs.
3. Turn each hand-rolled tolerance check into `compute_golden`, per-rank via the
   `out_{r}` pattern.
4. Keep `main()` so `python test_*.py -p <platform>` still works, as the
   converted `prefetch_async_demo` does.

`sdma_async_completion_demo` is the natural first one: it already carries
`@pytest.mark.sdma` from #1609, so its capability and ordering are expressible
today and only the body needs moving.

Doing it one file at a time is safe — the two forms coexist in the tree already.

### Priority

---

Related: #1604 (scene-test CI cost — different root cause, same subsystem).
Found while scoping the follow-up named in #1609's commit message; that message
called the conversion "a larger change", which is accurate about size but
implied a framework limitation that does not exist.

doraemonmj:
## Summary
- move duplicated A2/A3 and A5 notification/deferred-completion demos into shared communication-domain scene tests
- convert architecture-local SDMA and URMA demos to `SceneTestCase` while preserving their platform and workspace gates
- keep SDMA marker selection while allowing A5 to use its communication-domain workspace instead of Worker-global provisioning
- reuse scene-test rehosted `Buffer` handles when staging communication windows and building output tensor references
- align async/deferred child signatures with their submitted `CoreTaskArgs` and derive the deferred producer shape from the mailbox argument
- update example, capability, and profiling documentation for the new coverage paths

## Testing
- pre-commit hooks passed
- scene-test Python unit suite (`53 passed`)
- A5 SDMA scene remains selected by `-m sdma` (`1 test collected`)
- A3 `task-submit`: shared async-notify and deferred-notify scene tests (`2 passed`)
- `a2a3sim`: deferred-notify scene test (`1 passed`)
- previous PR head passed GitHub CI, including A2/A3 and A5 onboard jobs; amended-head CI is running

Fixes #1616


examples/a5/tensormap_and_ringbuffer/sdma_async_completion_demo/test_sdma_async_completion_demo.py
```diff
@@ -7,219 +7,133 @@
...
-@pytest.mark.runtime("tensormap_and_ringbuffer")
-@pytest.mark.device_count(2)
 @pytest.mark.skipif(
     _urma_workspace_enabled(),
     reason="SDMA demo requires the default SDMA backend; unset SIMPLER_ENABLE_PTO_URMA_WORKSPACE and rebuild simpler.",
 )
-def test_sdma_async_completion_demo(st_device_ids, st_platform) -> None:
-    assert run(st_platform, [int(d) for d in st_device_ids]) == 0
-
+@scene_test(level=3, runtime="tensormap_and_ringbuffer")
```
ChaoZheng109, line 78: **Must fix — preserve A5's per-domain SDMA provisioning.** Applying `@scene_test` changes the semantics of the existing `@pytest.mark.sdma`: the L3 scene-test fixture sets `enable_sdma=True` for any marked class (`conftest.py:1488`). A5 deliberately reports `dma_workspace_supported_mask() == 0`, so `Worker.init()` rejects that Worker-global workspace request before `orch.allocate_domain()` can provision A5's per-domain SDMA workspace. The current PR run landed on the X64 A5 runner and excluded `-m 'not sdma'`; an ARM64 A5 run will select this class. Please separate the selection/ordering marker from the Worker-global opt-in, or otherwise ensure this A5 class creates its Worker with `enable_sdma=False`, and cover the ARM path.

docs/capability-survey.md
```diff
@@ -219,7 +219,7 @@ values yield `PTO2_ERROR_ASYNC_COMPLETION_INVALID`.
 
 | Engine | a2a3 | a5 | Status |
 | ------ | ---- | -- | ------ |
-| COUNTER (default) | registered | registered | **Shipped** — `async_notify_demo` runs onboard on both arches and `deferred_notify_demo` runs in sim on both, through the `st-onboard-*` / `st-sim-*` jobs in `ci.yml`. Routed by `@pytest.mark.platforms`, no `skipif` |
+| COUNTER (default) | registered | registered | **Shipped** — `tests/st/worker/comm_domain/async_notify` runs onboard on both architectures; `tests/st/worker/comm_domain/deferred_notify` runs in sim on both and onboard on a2a3, through the `st-onboard-*` / `st-sim-*` jobs in `ci.yml`. Routed by `CASES[*]["platforms"]`, no `skipif` |
```
ChaoZheng109, line 222: **Please update the remaining stale names.** This row correctly establishes the new `comm_domain/async_notify` and `comm_domain/deferred_notify` paths, but `docs/dfx/core-swimlane-profiling.md:420` still lists the removed `async_notify_demo` and `deferred_notify_demo` names and calls them demos. Please update that reference to the new test names/paths as part of this move.

ChaoZheng109 (commented):
Two actionable findings from the comm-domain scene-test review.

tests/st/worker/comm_domain/async_notify/test_async_notify.py
```diff
@@ -0,0 +1,111 @@
...
+                        "signature": [D.IN, D.OUT, D.OUT, D.IN],
+                    }
+                    for func_id, source in enumerate(
+                        [
+                            "kernels/aiv/kernel_producer_notify.cpp",
+                            "kernels/aiv/kernel_consumer.cpp",
+                            "kernels/aiv/kernel_notify_wait.cpp",
+                        ]
+                    )
+                ],
```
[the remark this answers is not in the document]
doraemonmj: Fixed — both CALLABLE specs now declare a separate signature for each func_id, matching the producer, consumer, and notify-wait CoreTaskArgs payloads.

docs/capability-survey.md
doraemonmj: Fixed — the Core swimlane documentation now calls these communication-domain tests and references tests/st/worker/comm_domain/async_notify and tests/st/worker/comm_domain/deferred_notify.

examples/a5/tensormap_and_ringbuffer/sdma_async_completion_demo/test_sdma_async_completion_demo.py
doraemonmj: Fixed — the A5 class keeps the sdma marker for CI selection and ordering but sets worker_workspace=False. Both pytest fixture and standalone SceneTestCase Worker creation honor this opt-out, so Worker-global provisioning stays disabled and orch.allocate_domain() can provision the A5 per-domain SDMA workspace.

ChaoZheng109 (approved):

PR #1750, opened 2026-08-09
Add: enable direct-chip depth-two admission

Crane-Liu:
## Summary

- let the direct-chip `ChipRunLane` follow the runtime `PipelineContract`: compatible runs admit one active run plus one prepared successor
- keep depth-one admission when the runtime cannot prepare concurrently or native preparation reports an incompatible successor
- backpressure the third direct submission before selecting a slot, minting a generation, or starting native preparation
- let a successor handle advance its FIFO predecessor during `done()` and bounded `wait()` calls

This is W1c on top of #1748. It does not add a Python-side FIFO or change endpoint, RequestSession, profiling, or cross-repository behavior.

## Testing

- `ctest --test-dir tests/ut/cpp/build-w1c -LE requires_hardware --output-on-failure` (91/91 passed)
- `ctest --test-dir tests/ut/cpp/build-w1c -R "^test_chip_run_lane$" --output-on-failure`
- `clang-format --dry-run --Werror src/common/worker/chip_run_lane.cpp tests/ut/cpp/hierarchical/test_chip_run_lane.cpp`
- `.venv/bin/python -m pyright python/simpler/worker.py`

### DIFF
--- a/python/bindings/task_interface.cpp
+++ b/python/bindings/task_interface.cpp
@@ -2168,7 +2168,8 @@ NB_MODULE(_task_interface, m) {
             },
             nb::arg("callable_id"), nb::arg("args"), nb::arg("config"), nb::call_guard<nb::gil_scoped_release>(),
             "Submit materialized task args to the chip native-run lane without a pipeline lease and return the live "
-            "run. The lane admits at capacity one: this call drains its predecessor before admitting."
+            "run. The lane follows the runtime PipelineContract: compatible runs admit one active plus one prepared "
+            "successor; otherwise this call drains its predecessor before admitting."
         )
         .def(
             "_prepare_native_run_materialized",
--- a/python/simpler/worker.py
+++ b/python/simpler/worker.py
@@ -9668,8 +9668,7 @@ class Worker:
         Dispatch:
           - L2: ``callable`` is a ``CallableHandle`` returned by
             ``Worker.register(chip_callable)``. Routes to the private slot
-            carried by the handle. The current L2 backend remains blocking, so
-            the returned handle is already complete.
+            carried by the handle and returns a live completion handle.
           - L3+: ``callable`` is a Python orch fn invoked with the
             ``Orchestrator`` handle. Graph construction completes synchronously;
             device completion is reported by the returned handle.
@@ -9907,11 +9906,10 @@ class Worker:
     def _submit_l2_locked(self, callable_id: int, args, cfg: CallConfig) -> RunHandle:
         """Admit one direct-chip run and return a handle to it while it runs.
 
-        The lane is the sole admission authority and takes runs at capacity one:
-        submitting drains the predecessor first, so a second submit blocks here
-        rather than overlapping device work. What the handle adds is that the
-        *first* submit no longer waits — the caller owns the completion fence
-        through :meth:`RunHandle.wait`.
+        The lane is the sole admission authority. Its runtime contract permits
+        one active plus one prepared compatible run; otherwise submission drains
+        the predecessor and retains depth-one behavior. The caller owns the
+        completion fence through :meth:`RunHandle.wait`.
         """
         assert self._chip_worker is not None
         chip_args = self._materialize_l2_args(args)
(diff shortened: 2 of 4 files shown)


ChaoWao (approved):
## Review — W1c direct depth-two

Approve. I verified the claims independently rather than reading the PR body, and also ran the two suites the body does not mention.

### Stated vs real goal

The PR body tracks `implementation-plan.md` §W1b-W1c almost verbatim — "one active plus one prepared successor", "third submit backpressures before preparation or frame mutation", "A5 and incompatible TMR remain depth one" — and the code does that. **No goal downgrade**, which is the failure mode I check for first. The plan's negative constraint also holds: `git diff` touches no Python FIFO, phase table, progress-owner election or poison state; the Python delta is two docstrings.

### Mechanism, as I read it independently

Admission changes from *always drain, then admit* to *drain only if the front cannot host a successor*. The `while` loop is the backpressure: it spins the drain until either the FIFO is empty or the single front run is successor-compatible, and only then mints a generation and picks a slot. That ordering is what makes the "before preparation or frame mutation" clause literally true — on the third submit, `g_prepare_count` is asserted unchanged while the submitter thread is parked, which is a real concurrency assertion, not a linear-case one.

The second change is `progress()`: a non-front target now advances the front first. That is required, because with two runs admitted a caller can hold a handle to the successor and `done()`/`wait()` on it while the predecessor still owns the device. Recursion is bounded — FIFO is capped at 2, and the recursive call targets the front, which takes the non-recursive path.

### What I verified, not assumed

| Claim | How checked | Result |
|---|---|---|
| 91/91 C++ UT | built + `ctest -LE requires_hardware` at PR head | **91/91** |
| depth-two engages on a2a3 TMR | `concurrent_native_prepare_supported_impl` returns `1` in `a2a3/.../tensormap_and_ringbuffer`, `0` in sim | real, not a silent fallback |
| `block_on_front()`'s "only the front can be LAUNCHED" invariant survives | `launch_front()` only ever launches `fifo.front()` | **holds** |
| recursion terminates | `./test_chip_run_lane` under `timeout 120` | 17/17, 0 ms |

**Two suites the PR body does not claim, which I ran because this PR changes admission underneath them:**

- **Python UT — 1286 passed / 13 skipped**, matching the W1b baseline. Worth stating explicitly: `_submit_l2_locked` is unchanged, so every Python-side W1b test now exercises the *depth-two* lane and still passes.
- **a2a3 onboard sweep under `task-submit` — 56 passed + 24 passed / 2 skipped, 0 failures**, with 122 TMR cases on real silicon. Same numbers as the W1b baseline.

### One thing I checked and cleared, recorded so nobody re-derives it

The direct path picks a slot but never writes `state_->generations[slot_id]`, unlike the leased path (`:385`). That looks like a missing update. It is not: direct submits pass `pipeline_leased=false` → `admit_pipeline_generation=false`, so `prepare_native_run_on_slot` deliberately skips `pipeline_generations_.is_admissible`. The direct generation is a private monotonic counter, not a lease the filter arbitrates. Pre-existing W1b behavior, correctly left alone.

### Issues

**Consider (non-blocking, no change requested):**

The `while` loop calls `drain_front()` while holding `state_->mu`, and `drain_front` blocks on `wait_native_run`. So a third submit parks the lane mutex for the length of a device run — which is exactly what backpressure should do, but it also blocks `done()`/`wait()`/`close()` on *other* handles for that period. This is pre-existing (W1b's `submit` drained under the same lock), so it is not a regression, and the new test proves the intended blocking works. Flagging only because depth-two makes the window twice as likely to be hit: with two runs in flight there is now a handle that a caller may plausibly be polling while a third submit holds the lock.

**Not raised as issues, deliberately:** no missing-test finding (all three acceptance clauses are covered, including the fallback and the no-concurrent-prepare cases); no doc drift (both docstrings and the binding docstring were updated in the same commit, per `doc-consistency.md` §4).

### Verdict

**Approve.** The scope matches the plan exactly, the negative constraints hold, and the acceptance row — active-plus-prepared, third-submit pre-prepare backpressure, runtime fallback — is covered by tests that assert the mechanism rather than the outcome.

Disclosure: I authored W1b (#1748), which this stacks on, so I reviewed the parts that touch my code against the invariants I wrote rather than against my memory of them.

PR #1760, opened 2026-08-10
CI: run A5 SDMA tests on x86

doraemonmj:
## Summary

- Run the full A5 onboard scene-test corpus, including SDMA, on x86_64 and ARM64 runners.
- Remove the A5 x86_64 `not sdma` marker filter while preserving the A2/A3 SDMA isolation policy.
- Update CI and testing documentation to match the runner capability.

## Testing

- [x] Pre-commit hooks passed.
- [x] Workflow YAML and Markdown lint passed.
- [ ] Hardware tests not run locally; the updated A5 CI job exercises the full corpus.

### DIFF
--- a/.claude/skills/test-all-device/SKILL.md
+++ b/.claude/skills/test-all-device/SKILL.md
@@ -18,11 +18,20 @@ Detection / isolation procedures referenced below live in
    `task-submit`, let it pick via `--device auto --device-num <range size>`.
 5. **Read the marker selector out of the same job** you took the timeout from.
    `st-onboard-a2a3` is two pytest passes, not one: the sweep deselects
-   `-m "not sdma"` and a later step runs `-m sdma`. `st-onboard-a5` has no
-   marker filter. Take the expression from `ci.yml` rather than assuming, so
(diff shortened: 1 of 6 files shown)


ChaoZheng109:
The A5 marker-selection wording is still inconsistent in two testing skills:

- `.claude/skills/test-all-device/SKILL.md` says `st-onboard-a5` has no marker filter.
- `.claude/skills/test-runtime-device/SKILL.md` says the same.

After this PR, the A5 workflow still uses `-m "not pod"`; it only stops excluding `sdma`. Saying there is no marker filter can make local/agent reproductions run the pod tests in the ordinary single-machine A5 sweep.

Could we update both skills to say something like: `st-onboard-a5 uses -m "not pod" and does not exclude sdma`?

doraemonmj:
@ChaoZheng109 Addressed — both testing skills now state that `st-onboard-a5` uses `-m "not pod"` without excluding `sdma`, and include matching A5 command examples.

ChaoZheng109 (approved):

PR #1766, opened 2026-08-10
[Performance] A2/A3/A5 host_build_graph: init-on-write the tensormap entry pool

SergioMartin86:
# Human Summary

The entire 8mb long tensormap is being re-set at every launch. This PR removes this memset in favor of initializing tensor entries just-in-time, as (and if) needed.

Applies to both A2/A3 and A5

# AI Summary

## PR notes — `host_build_graph`: init-on-write the tensormap entry pool

- **Branch:** `hbg-tensormap-init-on-write` (commit `e06a7ad6`, off `upstream/main` `cf0fbc06`)
- **Diff:** 4 files, +72 / −42 — `pto_tensormap.cpp` + `pto_tensormap.h`, mirrored across **a2a3 and a5**
- **Arch/runtime:** `host_build_graph` on **both a2a3 and a5** (see [a5 parity](#a5-parity))
- **Composes with** `hbg-skip-redundant-orch-init` and `hbg-ready-queue-8192` (different files; any merge order).

## TL;DR

`PTO2TensorMap::init_data_from_layout` zeroed the **entire 65536-entry pool** on
every bind — an **8 MB memset** (`PTO2TensorMapEntry` is 128 B) plus a full-pool
`bucket_index = -1` pass and a redundant free-list memset — all `O(pool_size)`,
independent of how many tensors the run registers. That is the bulk of the
orchestrator init. Initializing each entry lazily (init-on-write) makes it
`O(live entries)`: **orch_reinit drops ~920 → ~85 µs (−91 %)**, host bind wall
**−22 %**, device time unchanged, golden-clean. The pool size is unchanged, so
there is **no overflow risk**.

## Why it was expensive

`init_data_from_layout` did three `O(pool_size)` things every bind:

1. `memset(entry_pool, 0, pool_size * 128)` — **8 MB** at the default pool size.
2. a full-pool loop setting `entry_pool[i].bucket_index = -1`.
3. `memset(free_entry_list, 0, pool_size * 8)`.

None depend on the run's actual tensor count. For paged_attention the whole
orchestrator re-init (dominated by this) was ~920 µs of the host bind wall.

## Why init-on-write is safe

Only *live* entries are ever read:

- **buckets[] start empty** (`nullptr`), so a lookup chain never reaches an
  unallocated slot.
- **Allocation is a bump/free-list**: `new_entry()` hands out `entry_pool[next_
  entry_idx++]` (or a recycled slot). Slots at index ≥ `next_entry_idx` are never
  handed out and never linked, so nothing reads them.
- The only full-pool readers are **debug-only** (`print_stats` / `valid_count`,
  `LOG_DEBUG`, no runtime callers).

So the change:

- **Skips all three `O(pool_size)` inits.**
- **`new_entry()` initializes each fresh bump-allocated slot** to the exact clean
  "unlinked" state `free_entry()` already leaves a recycled slot in
  (`bucket_index = -1`, all four link pointers null). The `insert` path overwrites
  the remaining fields — exactly as it already does for a recycled slot — so a
  fresh slot is **indistinguishable from a reused one**. Recycled slots already
  work with stale non-link fields, which is what proves this correct.
- **Bounds the two debug scans to `next_entry_idx`** (only initialized slots).
- **Leaves the pool size at 65536** — the 8 MB is still *reserved*, just no longer
  *touched* — so unlike shrinking the pool (whose overflow is an `always_assert`
  hard-crash), there is no occupancy risk.

## Measurement

paged_attention, a2a3 onboard, median of 50, **same-device back-to-back A/B**:

| Phase | base | init-on-write | Δ |
| --- | --: | --: | --: |
| **orch_reinit** | 920.5 | **84.7** | **−836 µs (−90.8 %)** |
| arena_build | 2296.1 | 1506.4 | −790 µs (see note) |
| arena_upload | 1066.5 | 1037.2 | unchanged |
| device_wall | 72.8 | 72.8 | unchanged |
| **host wall** | **6346.6** | **4927.5** | **−1419 µs (−22.4 %)** |

Note: this A/B is off `cf0fbc06`, where the orchestrator is initialized twice per
bind (once in arena-build, once as orch_reinit). The change makes the tensormap
init cheap in *both*, so `arena_build` drops too. On top of
`hbg-skip-redundant-orch-init` (which removes the arena-build init) only
`orch_reinit` remains, and this change takes that from ~920 → ~85 µs.

## Correctness

Golden-clean across the `host_build_graph` suite (bgemm, matmul, vector,
paged_attention, and all four `graph_execution` variants). `graph_execution` was
additionally run 5× (5/5) after one transient device-fault force-reset that did
not recur — a contention artifact, not a result mismatch. The tensormap is
load-bearing for dependency discovery, so a correctness bug would surface as a
wrong result or a hang; none observed.

## Reproduce

```bash
python tests/st/a2a3/host_build_graph/paged_attention/test_paged_attention.py \
    -p a2a3 -d <dev> --rounds 50 --skip-golden
python -m simpler_setup.tools.strace_timing <captured-stderr> --tree   # orch_reinit span

cmake --build build/cache/a2a3/onboard/host_build_graph/host  -j"$(nproc)"
cmake --build build/cache/a2a3/onboard/host_build_graph/aicpu -j"$(nproc)"
```


src/a2a3/runtime/host_build_graph/runtime/shared/pto_tensormap.cpp
```diff
@@ -75,35 +75,36 @@ PTO2TensorMapLayout PTO2TensorMap::reserve_layout_default(DeviceArena &arena, in
...
+    // Init-on-write: the entry pool is NOT pre-zeroed here. new_entry() puts each
+    // slot into its clean "unlinked" state (bucket_index == -1, link pointers
+    // null) the first time it is bump-allocated -- the same state free_entry()
+    // leaves a recycled slot in -- so only the current_used() live entries are
+    // ever touched, avoiding the O(pool_size) memset (8 MB at the default pool
+    // size). buckets[] start empty and unallocated slots are never reached via a
+    // bucket chain, so nothing reads an uninitialized slot; the two debug scans
+    // (print_stats / valid_count) are bounded to next_entry_idx. free_entry_list
+    // is a stack sized by free_num, meaningful only after frees, so it too needs
+    // no init.
```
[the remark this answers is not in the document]
SergioMartin86: Thanks for the review. Addressed in 2a5e2d6a.

1. init_data_from_layout doc (lines 78–107) — fixed. You're right on all three points. The doc now states it addresses the arena regions directly via arena.region_ptr and therefore runs before wire_arena_pointers (my previous wording said "after", which was wrong), corrects the complexity to O(num_buckets + task_window_size) (it also resets the per-task entry heads), and calls out the task_entry_heads reset explicitly.

3. new_entry() contract comment (lines 408–426) — fixed. The old // new_entry only allocates memory, does not assign attributes was stale after this change. Updated to: new_entry allocates a slot and initializes only its linkage (bucket_index + the four link pointers) to the clean unlinked state; insert() assigns the tensor attributes and producer_task_id.

2. a5 init_data_from_layout reordering — not applicable, skipping. I checked the a5 implementation: it has no task_entry_head_epochs and no task_epoch field, and its task_entry_heads reset already runs after the pool-cursor reset — the same order as a2a3. So there's no cleanup to move and nothing to realign there.

On keeping the two runtimes in sync more broadly: this PR is deliberately scoped to a2a3 host_build_graph, validated on a2a3 silicon, matching how the sibling HBG optimizations were landed (a2a3 first, a5 as a separate change). The a5 tensormap still does the full pool memset; porting init-on-write to a5 is a tracked follow-up that needs a5-silicon validation, which is why it isn't bundled here rather than shipped unvalidated.

Both changes are comment-only, so behavior is unchanged from the golden-validated build (whole host_build_graph suite green, graph_execution 5/5).

src/a2a3/runtime/host_build_graph/runtime/shared/pto_tensormap.cpp
```diff
@@ -75,35 +75,38 @@ PTO2TensorMapLayout PTO2TensorMap::reserve_layout_default(DeviceArena &arena, in
...
-    }
-
-    // free_entry_list: zeroed (was calloc'd before); contents become meaningful
-    // only after entries are freed back, so the body of the array stays as 0.
-    memset(free_list_arena, 0, static_cast<size_t>(pool_size) * sizeof(PTO2TensorMapEntry *));
+    // Init-on-write: the entry pool is NOT pre-zeroed here. new_entry() puts each
+    // slot into its clean "unlinked" state (bucket_index == -1, link pointers
+    // null) the first time it is bump-allocated -- the same state free_entry()
+    // leaves a recycled slot in -- so only the current_used() live entries are
+    // ever touched, avoiding the O(pool_size) memset (8 MB at the default pool
```
ChaoZheng109: **Nit — comment rule (`comments.md` §1): the "avoiding the memset" clause is change-rationale**

> `... so only the current_used() live entries are ever touched, avoiding the O(pool_size) memset (8 MB at the default pool size).`

"avoiding the … memset" only makes sense relative to the prior code — per `.claude/rules/comments.md`, a comment states a present-tense invariant, not a justification for the change or a reference to what was removed. The perf numbers belong in the commit message / PR body, not the comment. The surrounding invariants ("nothing reads an uninitialized slot", "free_entry_list is a stack … needs no init") are the keepers. Suggested: drop just the rationale clause —

```cpp
// Init-on-write: the entry pool is not pre-zeroed. new_entry() puts each
// bump-allocated slot into the clean "unlinked" state (bucket_index == -1,
// link pointers null) -- the same state free_entry() leaves a recycled slot
// in -- so only current_used() live entries are ever touched. buckets[] start
// empty and unallocated slots are never reached via a bucket chain, so
// nothing reads an uninitialized slot; the two debug scans (print_stats /
// valid_count) are bounded to next_entry_idx. free_entry_list is a stack
// sized by free_num, meaningful only after frees, so it needs no init.
```

ChaoZheng109 (commented):
Comment-rule nit — see inline.

src/a2a3/runtime/host_build_graph/runtime/shared/pto_tensormap.cpp
SergioMartin86: Good catch — fixed. Applied your suggested wording verbatim: dropped just the avoiding the O(pool_size) memset (8 MB …) rationale clause and kept the surrounding invariants (nothing reads an uninitialized slot, free_entry_list is a stack … needs no init). The comment now states only present-tense facts; the perf numbers stay in the commit message where they belong.

While in here I also mirrored the change to the a5 host_build_graph sibling in the same commit, per the arch-parity convention — a2a3 and a5 are near-duplicate trees with no divergence here (verified: identical entry pool and new_entry / free_entry / insert→link_entry structure). a5's init_data_from_layout drops the same three O(pool_size) inits, new_entry() puts each fresh slot into the clean unlinked state a5's free_entry leaves (with insert overwriting producer_task_id, exactly as on a2a3), and the two debug scans are bounded to next_entry_idx. The a5 header ends up byte-identical to a2a3's. a5 is compile-verified only (host lib links clean; the dev box is a2a3 silicon, so no a5 perf numbers); all pre-commit hooks pass on all four files.

Commit is force-pushed as a single commit.

ChaoZheng109 (approved):

PR #1807, opened 2026-08-12
Fix A2A3 repeated-run AICore stream regression

### LINKED ISSUE #1791: [Performance] A2A3 fresh AICore stream per run regresses repeated HCA/SWA Effective latency by 3.8–4.4%
## Background

In pypto-lib, repeated same-image DeepSeek-V4 Flash HCA/SWA execution on
`a2a3` with the `tensormap_and_ringbuffer` runtime has a performance
regression at Simpler PR #1540: median `effective_us` increases by 3.81–4.36%.

Repro cases:

- `models/deepseek/v4-flash/decode_attention_hca.py`
- `models/deepseek/v4-flash/decode_attention_swa.py`

Both entrypoints and their helpers are public at pypto-lib
`1f48761cb5be21a8cf0611fad10d513ba40b0863` on `main`. The test also applies
the public pypto-lib commit `c0751b7582618f7ccef29385236b1964cb480d42`,
which only migrates 17 `valid_shapes` keyword arguments to `valid_shape`.

Reproduction environment:

| Component | Version |
| --- | --- |
| pypto-lib | `1f48761cb5be21a8cf0611fad10d513ba40b0863` + compatibility patch `c0751b7582618f7ccef29385236b1964cb480d42` (`main`) |
| PyPTO | `2e59c01ccc34e586e19b49ecaa1a94b2e21db459` (`main`) |
| Simpler baseline | `8e00319e28cd2ea9f7f7f795a2b1e182956b91bd` (detached) |
| Simpler regressing commit | `e51c3aeecfaf982137b6adc67ca13749b0c95717` (detached) |
| PTOAS | `0.54` |
| PTO ISA | `83d01313d9bfc247c4b7c8bcf969d1019f0d106f` |
| CANN | `9.0.0` |
| Driver / Ascend HAL | `26.0.rc1` / `7.35.23` |
| Python | `3.10.19` |

Controlled A/B note: PyPTO `2e59c01c` records Simpler `dccb8379`. To isolate
PR #1540, the runtime checkout was intentionally overridden with the PR parent
and PR commit above. PTOAS and PTO ISA remained fixed. A separate pin-consistent
PyPTO runtime-bundle A/B independently showed the same regression direction and
magnitude.

Diagnosis: **simpler** — the direct Simpler parent/commit A/B reproduces the
regression while pypto-lib, PyPTO, PTOAS, PTO ISA, model inputs, and benchmark
settings remain fixed.

### Platform

`a2a3 (Ascend 910B/C hardware)`

### Runtime Variant

`tensormap_and_ringbuffer`

### Summary

Simpler PR #1540 changes the A2/A3 runner from retaining an AICore stream for
repeated runs of the same image to creating and retiring a fresh AICore stream
for every run. A direct parent/commit A/B on DeepSeek-V4 Flash HCA and SWA shows
median `effective_us` regressions of 4.36% and 3.81%, respectively.

Each endpoint used 5 warmups followed by 100 measured iterations. All four
runs passed `kv_cache` and `x_out` validation.

### Git Commit ID

`e51c3aeecfaf982137b6adc67ca13749b0c95717`

### CANN Version

`9.0.0`

### Driver Version

`26.0.rc1` (Ascend HAL `7.35.23`)

### Host Platform

`Linux (aarch64)` — Huawei Cloud EulerOS 2.0, kernel
`5.10.0-182.0.0.95.r1941_123.hce2.aarch64`

### Reproduction

Use isolated checkouts and a separate virtual environment for each Simpler
endpoint.

```bash
git -C <pypto-lib-checkout> checkout \
  1f48761cb5be21a8cf0611fad10d513ba40b0863

git -C <pypto-lib-checkout> show --format= \
  c0751b7582618f7ccef29385236b1964cb480d42 \
  -- models/deepseek/v4-flash/hc_pre.py | \
  git -C <pypto-lib-checkout> apply -

git -C <pypto-checkout> checkout \
  2e59c01ccc34e586e19b49ecaa1a94b2e21db459
git -C <pypto-checkout> submodule update --init --recursive

git -C <pypto-checkout>/runtime checkout <runtime-revision>
git -C <pto-isa-checkout> checkout \
  83d01313d9bfc247c4b7c8bcf969d1019f0d106f

source <CANN-root>/set_env.sh
export PTO_ISA_ROOT=<pto-isa-checkout>

python -m pip install --no-build-isolation -e <pypto-checkout>
python -m pip install --no-build-isolation -e <pypto-checkout>/runtime

task-submit --ptoas 0.54 --device 4 --timeout 0 --max-time 1800 --run '
cd <pypto-lib-checkout>
export PYPTO_BENCH=1
export PYPTO_BENCH_WARMUP=5
export PYPTO_BENCH_ROUNDS=100
export PYPTO_BENCH_RAW=1
export PTO2_RING_TASK_WINDOW=262144
export PTO2_RING_DEP_POOL=262144
export PTO2_RING_HEAP=2147483648
export PYPTO_RUNTIME_LOG=error

python models/deepseek/v4-flash/decode_attention_hca.py \
  -p a2a3 -d "$TASK_DEVICE" \
  --start-pos 8192 --enable-l2-swimlane 0

python models/deepseek/v4-flash/decode_attention_swa.py \
  -p a2a3 -d "$TASK_DEVICE" \
  --start-pos 8192 --enable-l2-swimlane 0
'
```

Run once with baseline runtime
`8e00319e28cd2ea9f7f7f795a2b1e182956b91bd`, then repeat with
`e51c3aeecfaf982137b6adc67ca13749b0c95717`.

### Expected Performance

With parent commit `8e00319e28cd2ea9f7f7f795a2b1e182956b91bd`:

| Case | Median `effective_us` |
| --- | ---: |
| HCA | 259.2 us |
| SWA | 244.0 us |

Repeated same-image execution should retain this steady-state latency after
warmup while preserving correctness when different images alternate.

### Actual Performance

With PR #1540 commit `e51c3aeecfaf982137b6adc67ca13749b0c95717`:

| Case | Median `effective_us` | Delta |
| --- | ---: | ---: |
| HCA | 270.5 us | +11.3 us / +4.36% |
| SWA | 253.3 us | +9.3 us / +3.81% |

Test window: 2026-08-11 19:05:39–19:10:05 UTC+8. One 100-round batch was
run per endpoint/case. A separate PyPTO runtime-bundle boundary A/B produced
HCA +12.5 us and SWA +8.7 us, independently corroborating the result.

### Profiling Data (Optional)

Before PR #1540, `ensure_run_stream_set(slot, aicore_image_hash)` returns
immediately when the retained AICore stream carries the same image hash. The
5 warmups and 100 measured rounds therefore share one AICore stream.

After PR #1540:

- `RunStreamSlots::acquire()` always creates a fresh AICore stream.
- The stream is retired after every completed run.
- The 5-warmup/100-round benchmark therefore uses 105 fresh AICore streams.

The reported `effective_us` metric is the on-device merged orch/sched window.
Host-side `rtStreamCreate` occurs before that device window, and
`rtStreamDestroy` occurs after device completion and timing readback. The
9–11 us regression is therefore not direct host stream API latency.

Generated AICore binaries are unchanged across the direct A/B:

| Case | Byte-identical child binaries | Regression per child |
| --- | ---: | ---: |
| HCA | 35/35 | 0.323 us |
| SWA | 28/28 | 0.332 us |

The strongest current explanation is a device-side cold-start effect: using a
fresh stream for every same-image run prevents retention of prior
instruction-fetch/I-cache/stream-affinity state. The commit boundary and stream
lifecycle change are proven; PMU `icache_req`/`icache_miss` data or a
stream-only toggle is still needed to confirm the exact hardware mechanism.

### Additional Context

The fresh-stream policy was introduced for correctness when different images
are published to the same GM code address. A plain restoration of the old
per-slot image-hash cache may reintroduce stale-instruction risk for an
`A -> B -> A` sequence across depth-two slots.

A safer remediation direction is runner-wide, generation-aware same-image
reuse: reuse only for consecutive identical images when no overlapping code
range has been republished, and recreate the stream when image identity or code
address generation changes. Stable non-overlapping code banks or a supported
AICore instruction-cache invalidation operation are alternatives.

Acceptance should cover repeated same-image performance, alternating-image
correctness, depth-two slot alternation, and stream-creation counts.

Source inspection of Simpler `main` at
`c5a587400721ea98f0a00321569088a5321539b9` shows that
`RunStreamSlots::acquire()` still unconditionally creates a fresh AICore stream.
The numerical A/B above was performed at the PR parent/commit, not at that
latest-main commit.

Related: #1717 discusses hiding per-run device launch latency and persistent
execution for `host_build_graph`; this report concerns repeated same-image
device Effective regression in `tensormap_and_ringbuffer`.

Crane-Liu:
## Summary

- retain proven-complete per-slot A2A3 AICore streams while no new AICore code has been published
- mark every retained slot stream stale only after a successful AICore H2D upload; content-hash dedup hits do not invalidate streams
- recreate stale streams on the next acquire, while preserving conservative destruction for unproven runs and failed-destroy retry ownership
- cover repeated reuse, deduplicated registration, concurrently resident callables, cross-slot `A -> B -> A` address-recycle publication, retirement gating, and exact stream-creation counts
- update `docs/task-flow.md` with the actual content-hash allocation and publication invariant

Fixes #1791

## Validation

- changed-file pre-commit: all hooks passed (`clang-format`, `clang-tidy`, `cpplint`, `ruff`, `pyright`, `markdownlint`)
- `RunStreamSlots` C++ unit tests: 19/19 passed with GCC 15 on LCW
- A2A3 architecture probe: passed on `task-submit` (`task_20260814_024716_37579914368`)
- GitHub CI on `a9ec96bb2`: 18/18 checks completed with no failures (17 success, expected `deploy` skip), including `ut-a2a3`, `ut-a5`, `st-onboard-a2a3`, `st-onboard-a5`, `st-pod-onboard-a2a3`, both A2A3/A5 simulation matrices, packaging, profiling, and pre-commit

An additional LCW mixed-venv targeted run (`task_20260814_025329_67147427068`) reached its 600-second pytest session timeout in `native_run_lifecycle::test_run` before any assertion result. It is not treated as product evidence; the clean GitHub `st-onboard-a2a3` run above passed the full suite on this exact head.

The stream-creation assertions validate the mechanism. The issue-specific HCA/SWA latency benchmark pins older PyPTO/PTO-ISA revisions, so no mixed-revision latency number is reported as comparable performance evidence.

### DIFF
--- a/docs/task-flow.md
+++ b/docs/task-flow.md
@@ -290,14 +290,16 @@ that window needs the admission layer to gate dispatch on `pool.owns(lease)`
 before handing work down, which is whole-run admission's job, not this
 layer's.
 
-AICore streams are outside that lease. The instruction cache belongs to the
-cores, every slot publishes its image to the same GM code address, and the
-platform offers no cache invalidation for code replaced there. Creating a
-stream is the only operation known to leave a core free of the previous image's
-instructions; selecting an already-existing one is not. So each run creates its
-own AICore stream and retires it on every exit path, and no record of which
-image a stream last ran is load-bearing. The AICPU stream carries no such state
-and stays with its slot.
+AICore streams are outside that lease. Registered callables keep content-hash
+deduplicated GM allocations: simultaneously resident code images occupy
+different allocations, while unregister frees an allocation that a later
+registration may reuse. A dedup miss is the only path that publishes new
+AICore instruction bytes. After that H2D copy succeeds, every retained slot
+stream is marked stale; its next acquire destroys and recreates the AICore
+stream before launch. Without a new publication, completed slot streams remain
+reusable even when two resident callables alternate. Unproven completion still
+destroys the stream conservatively. The AICPU stream carries no instruction
+cache state and stays with its slot.
 
 #### Whole-run FIFO admission
 
--- a/python/bindings/task_interface.cpp
+++ b/python/bindings/task_interface.cpp
@@ -2305,9 +2305,9 @@ NB_MODULE(_task_interface, m) {
         .def_prop_ro(
             "run_stream_set_create_count", &ChipWorker::run_stream_set_create_count,
             "Number of AICore run streams the bound runner has created. The AICPU "
-            "stream belongs to a pipeline slot for the worker's lifetime, while each "
-            "run creates and retires its own AICore stream, so this advances once per "
-            "run; platforms whose runs use the persistent bootstrap pair report 0."
+            "stream belongs to a pipeline slot for the worker's lifetime. A completed "
+            "AICore stream is reused until a new code upload makes it stale. Platforms "
+            "using the persistent bootstrap pair report 0."
         )
         .def_prop_ro(
             "committed_device_memory", &ChipWorker::committed_device_memory,
--- a/python/simpler/worker.py
+++ b/python/simpler/worker.py
@@ -11221,10 +11221,10 @@ class Worker:
     def run_stream_set_create_count(self) -> int:
         """L2 only: number of AICore run streams the runner has created.
 
-        AICPU streams belong to pipeline slots for the worker's lifetime, while
-        each run creates and retires its own AICore stream, so this advances
-        once per run. Returns 0 on non-L2 workers and on platforms whose runs
-        use the persistent bootstrap stream pair (simulation, a5).
+        AICPU streams belong to pipeline slots for the worker's lifetime. A
+        completed AICore stream is reused until a new code upload makes it
+        stale. Returns 0 on non-L2 workers and on platforms whose runs use the
+        persistent bootstrap stream pair (simulation, a5).
         """
         if self.level != 2 or self._chip_worker is None:
             return 0
--- a/src/a2a3/platform/onboard/host/device_runner.cpp
+++ b/src/a2a3/platform/onboard/host/device_runner.cpp
@@ -464,10 +464,9 @@ int DeviceRunner::drain_execution(ActiveExecution &active) {
         return rc;
     }
 
-    // On a successful device drain, failure to retire this run's AICore stream
-    // is the run's error. Mark the attempt first so cleanup does not immediately
-    // retry a failed destroy; the retained handle keeps the slot unusable and
-    // lets finalize retry it later.
+    // A proven-complete stream is reusable until a code publication marks it
+    // stale. Publish retirement so cleanup does not replace it with an
+    // unproven state after the device result has already been established.
     prepared.aicore_retirement_attempted = true;
     rc = retire_run_aicore_stream(pipeline_slot, RunStreamSlots::CompletionStatus::Complete);
     if (rc != 0) return rc;
--- a/src/a2a3/platform/onboard/host/device_runner.h
+++ b/src/a2a3/platform/onboard/host/device_runner.h
@@ -105,6 +105,7 @@ public:
     bool can_accept_run() const override { return !device_unusable_.load(std::memory_order_acquire); }
     int provision_native_run_resources(uint32_t pipeline_slot) override;
     int abandon_native_run_resources(uint32_t pipeline_slot) override;
+    void mark_run_streams_stale() override { run_stream_slots_.mark_all_stale(); }
 
     // Map/unmap a device buffer into host address space via
     // halHostRegister(DEV_SVM_MAP_HOST) / halHostUnregister. The returned host
@@ -225,14 +226,10 @@ private:
     // One slot per in-flight run the pipeline contract can declare.
     static constexpr unsigned kRunStreamSetCount = PTO_PIPELINE_MAX_DEPTH;
 
-    // AICPU streams belong to slots for the worker's lifetime. AICore streams
-    // belong to a single run: the platform offers no instruction-cache
-    // invalidation for code replaced at a reused GM address, and creating the
-    // stream is the only operation known to leave a core free of the previous
-    // image's instructions. Selecting an existing stream is not.
-    // Stream lifetimes live in RunStreamSlots so the slot state machine —
-    // fresh AICore stream per run, handle kept when a destroy fails — is
-    // testable without a device.
+    // AICPU streams belong to slots for the worker's lifetime. A completed
+    // AICore stream remains reusable until a new code upload marks every slot
+    // stale. Stream lifetimes live in RunStreamSlots so publication and
+    // failed-destroy states are testable without a device.
     RunStreamSlots run_stream_slots_{
         [this](void **out) {
             return create_run_stream(out);
--- a/src/common/platform/include/host/run_stream_slots.h
+++ b/src/common/platform/include/host/run_stream_slots.h
@@ -24,11 +24,9 @@
  * Per-slot ownership of the two streams a run submits on.
  *
  * The AICPU stream carries no instruction-cache state, so it belongs to its
- * slot for the runner's lifetime. The AICore stream belongs to a single run:
- * the platform offers no instruction-cache invalidation for code replaced at a
- * reused GM address, and creating the stream is the only operation known to
- * leave a core free of the previous image's instructions — selecting an
- * existing one is not.
+ * slot for the runner's lifetime. A completed AICore stream remains reusable
+ * until new AICore code is published. Publication marks every slot stale; its
+ * next acquire destroys the old AICore stream before creating a replacement.
  *
  * A destroy that fails keeps its handle. Such a stream may still hold the
  * previous image's instructions, so the slot must refuse the next run rather
@@ -54,11 +52,7 @@ public:
         create_(std::move(create)),
         destroy_(std::move(destroy)) {}
 
-    /**
-     * Ready `slot` for a run: its AICPU stream on first use, and always a fresh
-     * AICore stream. Fails when the slot still holds an AICore stream a prior
-     * run could not retire.
-     */
+    /** Ready `slot` for a run, reusing its AICore stream unless stale. */
     int acquire(unsigned slot) {
         if (slot >= slots_.size()) return -1;
         Slot &s = slots_[slot];
@@ -70,16 +64,35 @@ public:
                 return rc;
             }
         }
-        if (s.aicore != nullptr) return -1;
+        if (s.aicore != nullptr) {
+            if (s.submitted || !s.complete) return -1;
+            if (!s.stale) {
+                s.submitted = false;
+                s.complete = false;
+                return 0;
+            }
+            int rc = destroy_(s.aicore);
+            if (rc != 0) return rc;
+            s.aicore = nullptr;
+        }
         int rc = create_(&s.aicore);
         if (rc != 0) {
             s.aicore = nullptr;
             return rc;
         }
         created_count_.fetch_add(1, std::memory_order_relaxed);
+        s.stale = false;
         s.submitted = false;
         s.complete = false;
         return 0;
+    }
+
+    /** Mark every retained AICore stream stale after code publication. */
+    void mark_all_stale() {
+        for (Slot &s : slots_) {
+            std::lock_guard<std::mutex> lock(s.mutex);
+            s.stale = true;
+        }
     }
 
     /** Make the prepared stream pair visible to non-blocking poll. */
@@ -116,9 +129,7 @@ public:
     }
 
     /**
-     * Retire `slot`'s AICore stream. The handle survives a failed destroy.
-     * Only a successful device query or stream synchronization may pass
-     * Complete; launch rollback and abandoned preparation pass Unproven.
+     * Complete retains the stream for reuse. Unproven destroys it.
      */
     int retire_aicore(unsigned slot, CompletionStatus completion_status) {
         if (slot >= slots_.size()) return -1;
@@ -130,6 +141,7 @@ public:
         // later failing sync, so the sync error remains authoritative.
         s.submitted = false;
         s.complete = completion_status == CompletionStatus::Complete;
+        if (completion_status == CompletionStatus::Complete) return 0;
         if (s.aicore == nullptr) return 0;
         int rc = destroy_(s.aicore);
         if (rc != 0) return rc;
@@ -144,6 +156,7 @@ public:
             std::lock_guard<std::mutex> lock(s.mutex);
             s.submitted = false;
             s.complete = false;
+            s.stale = false;
             for (void **stream : {&s.aicpu, &s.aicore}) {
                 if (*stream == nullptr) continue;
                 int rc = destroy_(*stream);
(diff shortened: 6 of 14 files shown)


ChaoWao (commented):
Reviewed at `a15632e3` against merge-base `50c06606`. The state machine is clean and the UT coverage around it is genuinely good, so this is not a rewrite request — but I think the **reuse key is the wrong one**, and picking the right one makes the patch both smaller and safer.

## The invalidation trigger has a single chokepoint

I audited every `RT_MEMCPY_HOST_TO_DEVICE` on the onboard host path:

| site | writes |
| --- | --- |
| `device_runner_base.cpp:692` (`upload_chip_callable_buffer`) | **AICore child code — the only one** |
| `host_regs.cpp:210` | MMIO register address table |
| `device_runner_helpers.cpp:76` / `:105` | runtime args / kernel args |
| `device_runner_base.cpp:146` (`copy_to_device`) | tensor payload |
| `ensure_binaries_loaded` | AICPU SO dispatcher bootstrap, not AICore code |

So the instruction bytes in GM change in exactly one place, and only when the content-hash dedup **misses** (a dedup hit returns before the `rtMemcpy`, so no code is written).

That means the question a slot needs to answer is *"has any code been published since this stream was created?"* — not *"which image did this slot run last time?"* Those are different predicates, and keying on the second one is wrong in **both** directions.

## Under-invalidates: `A -> B -> A` across two slots

`bound_image` is per-slot, but the instruction cache belongs to the cores, which both pipeline slots dispatch to. `upload_chip_callable_buffer` allocates via `mem_alloc_.alloc` → `rtMalloc`, and `unregister_callable` → `release_chip_callable_buffer` → `rtFree` releases the block, so an address is recyclable:

1. register A → run on **slot 0** → `S0` retained, `bound_image = H_A`, `complete`
2. unregister A → `rtFree(X)`
3. register B → `rtMalloc` returns `X` → B's children occupy A's old code range
4. run B on **slot 1** → slot 1 creates its own stream → cores fetch B at `X`
5. unregister B; register A again → `rtMalloc` returns `X`; A's bytes are back at `X`
6. run A on **slot 0** → `bound_image == H_A` → **stream reused, no fresh-stream flush**

Slot 0 never observed step 4 because it happened on the other slot. Under the exact hardware premise #1540 was built on, step 6 can execute B's cached lines. This is the case #1791 warned about in *Additional Context*:

> A plain restoration of the old per-slot image-hash cache **may reintroduce stale-instruction risk** for an `A -> B -> A` sequence across depth-two slots.

## Over-invalidates: two resident images alternating

When both callables stay registered, they occupy two distinct `rtMalloc` blocks and **no code byte changes between runs** — yet every transition destroys and recreates. The PR's own test pins that cost:

```python
# test_alternating_code_images_never_execute_stale_instructions
assert st_worker.run_stream_set_create_count == before + 1   # every run
```

So an alternating-resident-image workload pays the full #1540 penalty and recovers nothing. Worth checking against real serving traffic, where prefill and decode are separate callables on one worker — the dsv4 HCA/SWA benchmark in #1791 is a single callable in a loop, so it is the one shape this keying does fix.

## Suggested shape

One bool per slot, using the vocabulary already in the tree (`stale instructions` is the wording in this file's own test name and in `docs/task-flow.md:293`):

```cpp
// RunStreamSlots::Slot
bool stale{false};

void mark_all_stale() {
    for (Slot &s : slots_) {
        std::lock_guard<std::mutex> lock(s.mutex);
        s.stale = true;
    }
}

int acquire(unsigned slot) {
    ...
    if (s.aicore != nullptr) {
        if (!s.complete) return -1;
        if (!s.stale) return 0;               // no publication since creation: reuse
        rc = destroy_(s.aicore); if (rc != 0) return rc;
        s.aicore = nullptr;
    }
    rc = create_(&s.aicore); ...
    s.stale = false;
}
```

Hook it after the successful `rtMemcpy` in `upload_chip_callable_buffer` (not on the dedup-hit path). `run_stream_slots_` lives on the a2a3 `DeviceRunner` while upload lives on the base, so a default-no-op virtual on the base that a2a3 overrides — the shape `abandon_native_run_resources` already uses at `device_runner_base.h:524`.

Two notes on why this needs no extra machinery:

- **In-flight runs need no coordination.** A resident callable holds a refcount and the allocator only hands out free blocks, so a publication can never overwrite the code of an executing image. A newly-set `stale` flag only affects the next `acquire`.
- **No counter or per-slot snapshot needed.** Publication is rare and already does a full H2D plus stream sync, so taking the two slot mutexes is free, and those mutexes are the existing serialization point against `acquire`/`retire`/`poll`. (I'd also avoid the word "generation" here — this repo already uses it for the pipeline lease `{slot_id, generation}` and the `ChipWorker` high-water mark.)

This also removes most of the current diff: `acquire(slot)` keeps its signature and stays in `provision_native_run_resources`, so the `callable_aicore_image_hash()` accessor, the `ensure_aicpu` / `ensure_aicpu_locked` split, the move of `acquire` into `prepare_execution`, and the `aicore_image_hash == 0` failure path all become unnecessary.

## Other findings

**Must fix**

1. **`docs/task-flow.md:293-301` states the invariant this PR reverses and is not updated** — *"each run creates its own AICore stream ... and no record of which image a stream last ran is load-bearing."* After this PR that record **is** load-bearing. Same-commit fix per `.claude/rules/doc-consistency.md` §1/§4. While rewriting: the paragraph's claim that "every slot publishes its image to the same GM code address" does not match `upload_chip_callable_buffer`, which `rtMalloc`s per `content_hash` with refcounting — simultaneously-registered callables sit at *different* addresses, and an address is only reused after an unregister frees it.

2. **No test covers the hazard the policy weakens.** Both `test_alternating_code_images_never_execute_stale_instructions` and the new `test_depth_two_slots_rebuild_on_independent_image_transitions` register add and sub **concurrently** and only unregister in `finally`. Two live leases means two distinct blocks, so no code address is ever recycled and neither test can fail from stale instructions regardless of policy. The docstring correction from "at one reused GM code address" to "on one pipeline slot" is honest, but the PR description still lists A/B/A/B as coverage for this risk. The nearest existing coverage, `dynamic_register/test_dynamic_register.py::test_unregister_last_handle_allows_reprepare_same_hashid`, re-registers *identical* content and is safe by construction. A test that walks the six steps above would close it.

3. **All five self-hosted checks are red** — `ut-a2a3`, `ut-a5`, `st-onboard-a2a3`, `st-onboard-a5`, `st-pod-onboard-a2a3`, each annotated *"The job was not acquired by Runner of type self-hosted even after multiple attempts."* Infra starvation, not attributable to this diff — but it means no hardware validation ran in CI for a change that relaxes a hardware correctness invariant. Worth a re-run before merge.

4. **No performance evidence.** #1791's acceptance list leads with "repeated same-image performance", and the description concedes *"I did not report mixed-revision latency numbers as comparable performance evidence."* The stream-count assertions prove the mechanism, not that the 3.81–4.36% is recovered. Happy to run the dsv4-flash HCA/SWA A/B if that helps.

**Should fix**

5. **The reuse gate `s.complete` is weaker than the `aicore == nullptr` gate it replaces.** `poll()` sets `complete = true` from a progress thread on a successful device query, before the executor drains. A second `acquire` on a slot whose run has not been retired used to be hard-refused; it now succeeds and hands back the in-flight stream. Consider a flag set only by `retire_aicore(Complete)` and cleared by `acquire`/`mark_submitted`.

6. **`retire_aicore(Complete)` turns a benign no-op into an error.** The old code fell through `if (s.aicore == nullptr) return 0;`. It now returns −1 when the handle is gone, and `drain_execution` propagates that as the run's error after a *successful* device drain (reachable in principle via `abandon_all()` on the device-reset path). Either keep it benign or document why absence must be an error.

7. **`aicore_image_hash == 0` is a hard prepare-time failure but is not validated at registration.** `record_device_orch_callable` / `record_host_orch_callable` check `chip_buffer_hash` and `chip_dev` but not this field, so a registration path that omits it registers cleanly and then fails every run with *"active callable has no AICore image identity"*, far from the cause. (Moot if the `stale` shape is adopted.)

8. **`test_depth_two_slots_own_separate_resources` gained assertions that cannot fail** — `assert after_slot0 in (stream_sets, stream_sets + 1)` and `assert after_slot1 in (after_slot0, after_slot0 + 1)` accept both branches. If slot warmth is order-dependent under the shared `st_worker`, warm the slots explicitly and assert exact counts.

**Consider**

9. The description should surface the goal narrowing: #1791 recommended runner-wide, publication-aware reuse; this implements per-slot content-hash. Even with the above fixed, stating why is worth a line, and a `docs/investigations/` entry next to `2026-06-aicore-cold-start-warmup.md` would help — the underlying mechanism is still unproven in both directions (#1791: *"PMU `icache_req`/`icache_miss` data or a stream-only toggle is still needed"*).

10. `_run_registered` and `_run_registered_with_lease` each inline `(base+1)*(base+2)`, duplicating `compute_golden`. Parameterizing `compute_golden` on `subtract` would avoid the third copy.

Crane-Liu:
@ChaoWao Addressed in `a9ec96bb2`; thanks for identifying the publication chokepoint.

The reuse policy is now runner-wide and publication-aware rather than keyed by a per-slot image hash:

- `upload_chip_callable_buffer()` calls a default-no-op virtual hook only after a successful AICore H2D copy. The content-hash dedup return remains before the hook.
- A2A3 overrides the hook with `RunStreamSlots::mark_all_stale()`. Each slot now has one `stale` bit; `acquire(slot)` reuses only a formally retired, non-stale stream and recreates a stale one.
- Removed the callable image accessor, `ensure_aicpu` split, prepare-time image lookup, and zero-hash failure path. Acquisition remains in `provision_native_run_resources()`.
- `poll()` completion alone cannot enable reuse because `submitted` remains set until `retire_aicore(Complete)`. Complete retirement without a handle is benign.
- Updated `docs/task-flow.md` to describe content-hash allocations, dedup hits, address recycling, and the successful H2D publication invariant.

Coverage now includes repeated no-publication reuse, deduplicated registration without invalidation, resident A/B alternation without recreation, the cross-slot unregister/register `A -> B -> A` publication hazard, publication during an active run, retire gating, exact depth-two counts, and failed-destroy retry ownership. The golden calculation is shared instead of duplicated.

Validation on the new head: changed-file pre-commit passed all hooks and the `RunStreamSlots` C++ suite passed 19/19 with GCC 15. A2A3 architecture probe `task_20260814_024716_37579914368` passed. GitHub CI completed 18/18 checks with no failures (17 success, expected deploy skip), including A2A3/A5 UT, A2A3/A5 onboard ST, pod A2A3 ST, and both simulation matrices. An additional mixed-venv LCW run `task_20260814_025329_67147427068` timed out in `native_run_lifecycle::test_run` before producing an assertion result, so I am not treating it as product evidence; the clean GitHub A2A3 onboard suite passed this exact head.

I have not reported a mixed-revision latency number as performance evidence. Your offer to run the canonical dsv4-flash HCA/SWA A/B on this head would close that remaining evidence gap without conflating PyPTO/PTO-ISA revisions.

PR #1826, opened 2026-08-13
Fix: honor ASCEND_RT_VISIBLE_DEVICES at comm_hccl MemSetAccess sites

lterrac:
## Problem

Under `ASCEND_RT_VISIBLE_DEVICES` isolation, `comm_alloc_domain_windows` fails on every card pair except an identity mapping:

```
domain_alloc_via_fabric: [comm_hccl.cpp:1165] [comm rank 0] alloc_domain: create local Fabric window -> 507899
```

`507899` is `ACL_ERROR_RT_DRV_INTERNAL_ERROR` — the driver was handed a device id in a numbering it does not use. Cards `0,1` hide it because there logical == physical, which is also why serving has never hit it: DeepSeek V4 takes all 8 cards.

This is the class #1600 fixed, at a site it did not cover — neither arch's `comm_hccl.cpp` included `common/acl_hal_device.h`.

## Root cause — the two `location.id` fields want different id spaces

`aclrtMemAccessDesc::location.id` is consumed in the driver-visible space and must translate, even though `aclrtMemSetAccess` is an ACL entry point. Its sibling `aclrtPhysicalMemProp::location.id` must **not** — `aclrtMallocPhysical` rejects a driver-visible id outright.

Measured by walking the window sequence one call at a time, in one process. The grant matters: the translated id of the bound device has to fall outside the logical range, or the same value is legal in both spaces and no outcome can tell them apart — an earlier revision of this section probed `1,2`, where it cannot. Under `ASCEND_RT_VISIBLE_DEVICES=2,3` bound to logical `0` (= card 2), logical space `[0,1]`:

```
aclrtMemGetAllocationGranularity  prop.location.id=0 (bound logical) -> 0        OK
aclrtMallocPhysical               prop.location.id=0 (bound logical) -> 0        OK, 2 GiB appears on card 2
aclrtMemGetAllocationGranularity  prop.location.id=2 (translated)    -> 107001   rejected
aclrtMallocPhysical               prop.location.id=2 (translated)    -> 107001   rejected
aclrtMallocPhysical               prop.location.id=1 (other logical) -> 507899   rejected
aclrtReserveMemAddress                                               -> 0        OK
aclrtMapMem                                                          -> 0        OK
aclrtMemSetAccess                 desc.location.id=0 (logical)       -> 507899   FAILS
aclrtMemSetAccess                 desc.location.id=2 (translated)    -> 0        OK
```

`107001` is `ACL_ERROR_RT_INVALID_DEVICEID`, and that row is what fixes the field's id space: under a driver-visible reading, `2` *is* the bound card and the call would have succeeded. The card the memory lands on was read off npu-smi's HBM column during a 2 GiB hold, not inferred.

With the variable unset and the card bound directly, the whole sequence completes — measured on card 1 when this was first traced, and identical for `ACL_HBM_MEM_HUGE` (fabric) and `ACL_HBM_MEM_NORMAL` (ipc). Today's re-measurement above covers `ACL_HBM_MEM_HUGE` only; the id spaces are a property of the fields, not of `memAttr`.

The asymmetry is **per-field, not per-API**, so `acl_hal_device.h` now records both halves. A symmetric "fix" of the memory property fails loudly rather than silently allocating elsewhere — `107001` for the translated id, `507899` for any other logical id — because the field must name the *bound* logical device, not merely a logical one.

## Change

Three sites carry the id into `aclrtMemSetAccess`:

- a2a3 `reserve_and_map_vmm_window` — the funnel for every local and imported window, fabric and IPC
- a5 `alloc_windows_via_ipc` and `domain_alloc_via_ipc` — their descriptors are reused for the peer mappings, so the translation rides along

No CMake change: both onboard host targets already carry `src/common/platform/include`.

## Test

`tests/st/vis_isolation` gains a comm-domain case next to the chip-init one, so both device-id layers live in one module. Chip init never maps device memory, so it cannot catch this. The new case runs on all granted cards rather than a single logical id, and reuses the existing ascending-list rule and skip-on-identity guard — under an identity grant the two id spaces hold the same number and the test would pass without proving anything.

## Verification

**a2a3, on 910B2** — A/B on the same granted cards in the same job, only the runtime differing:

| Cards | Mapping | Stock | This change |
|---|---|---|---|
| visible `1,2`, logical 0,1 | non-identity | both ranks abort, `create local Fabric window -> 507899` | both complete byte-exact |
| visible `0,1`, logical 0,1 | identity | passes | passes (no regression) |

The identity case matters: the broker hands out the lowest free cards, so a two-card request usually lands on exactly the mapping that proves nothing.

**a5** — no a5 hardware available to me; it rides on `_st-npu-a5.yml`. The change there is the same pattern as a2a3, applied to the two inlined copies.

## Related, not fixed here

The peer-access-status confusion already recorded in comments at a5 `comm_hccl.cpp:553-561` and `:972-977` (#1018) is the same id-space split at a different call (`aclrtDevicePeerAccessStatus` / `aclrtDeviceEnablePeerAccess` take a *peer* id). Left untouched rather than widening this diff.

### DIFF
--- a/src/a2a3/platform/onboard/host/comm_hccl.cpp
+++ b/src/a2a3/platform/onboard/host/comm_hccl.cpp
@@ -24,6 +24,7 @@
 #include "platform_comm/comm_context.h"
 #include "pto_runtime_c_api.h"
 
+#include "common/acl_hal_device.h"
 #include "common/unified_log.h"
 #include "host/file_marker_handshake.h"
 
@@ -268,10 +269,14 @@ static aclError reserve_and_map_vmm_window(
         return status;
     }
 
+    // aclrtMemAccessDesc::location.id is consumed in the driver-visible space, unlike
+    // aclrtPhysicalMemProp::location.id above, which takes the ACL-logical id. Under
+    // ASCEND_RT_VISIBLE_DEVICES the logical id here names the wrong card and aclrtMemSetAccess
+    // fails with 507899 (ACL_ERROR_RT_DRV_INTERNAL_ERROR). See common/acl_hal_device.h.
     aclrtMemAccessDesc access_desc{};
     access_desc.flags = ACL_RT_MEM_ACCESS_FLAGS_READWRITE;
     access_desc.location.type = ACL_MEM_LOCATION_TYPE_DEVICE;
-    access_desc.location.id = static_cast<uint32_t>(device_id);
+    access_desc.location.id = static_cast<uint32_t>(pto::acl_to_hal_device_id(device_id));
     status = aclrtMemSetAccess(base, size, &access_desc, 1);
     if (status != ACL_SUCCESS) {
         aclrtUnmapMem(base);
--- a/src/a2a3/platform/onboard/host/comm_hccl.cpp
+++ b/src/a2a3/platform/onboard/host/comm_hccl.cpp
@@ -270,8 +270,8 @@ static aclError reserve_and_map_vmm_window(
     }
 
     // aclrtMemAccessDesc::location.id is consumed in the driver-visible space, unlike
-    // aclrtPhysicalMemProp::location.id above, which takes the ACL-logical id. Under
-    // ASCEND_RT_VISIBLE_DEVICES the logical id here names the wrong card and aclrtMemSetAccess
+    // aclrtPhysicalMemProp::location.id in its caller create_local_vmm_window, which takes the ACL-logical
+    // id. Under ASCEND_RT_VISIBLE_DEVICES the logical id here names the wrong card and aclrtMemSetAccess
     // fails with 507899 (ACL_ERROR_RT_DRV_INTERNAL_ERROR). See common/acl_hal_device.h.
     aclrtMemAccessDesc access_desc{};
     access_desc.flags = ACL_RT_MEM_ACCESS_FLAGS_READWRITE;
--- a/src/common/platform/include/common/acl_hal_device.h
+++ b/src/common/platform/include/common/acl_hal_device.h
@@ -33,6 +33,16 @@ namespace pto {
  *     function: `hal_fn(acl_to_hal_device_id(device_id))`.
  *   - ACL/rt entry points (`aclrtSetDevice`, `rtMalloc`, `rtMemcpy`, stream
  *     APIs, ...) take the logical id and MUST NOT.
+ *   - One ACL exception, measured on CANN 9.0.0 / a2a3:
+ *     `aclrtMemAccessDesc::location.id` reaches the driver untranslated and MUST
+ *     translate, even though `aclrtMemSetAccess` is an ACL entry point. Its
+ *     sibling `aclrtPhysicalMemProp::location.id` does NOT — `aclrtMallocPhysical`
+ *     rejects a driver-visible id with 107001. Under isolation, binding logical
+ *     0 of `ASCEND_RT_VISIBLE_DEVICES=1,2`:
+ *       aclrtMallocPhysical  prop.location.id=0 -> 0        prop.location.id=2 -> 107001
+ *       aclrtMemSetAccess    desc.location.id=0 -> 507899   desc.location.id=1 -> 0
+ *     The asymmetry is per-field, not per-API, so neither half generalises to the
+ *     other.
  *
  * `ASCEND_RT_VISIBLE_DEVICES` renumbers the logical space to `0..N-1` while the
  * HAL keeps indexing the driver-visible space; a direct HAL call made with the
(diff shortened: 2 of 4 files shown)


src/common/platform/include/common/acl_hal_device.h
```diff
@@ -33,6 +33,16 @@ namespace pto {
...
  *   - ACL/rt entry points (`aclrtSetDevice`, `rtMalloc`, `rtMemcpy`, stream
  *     APIs, ...) take the logical id and MUST NOT.
+ *   - One ACL exception, measured on CANN 9.0.0 / a2a3:
+ *     `aclrtMemAccessDesc::location.id` reaches the driver untranslated and MUST
+ *     translate, even though `aclrtMemSetAccess` is an ACL entry point. Its
+ *     sibling `aclrtPhysicalMemProp::location.id` does NOT — `aclrtMallocPhysical`
+ *     rejects a driver-visible id with 107001. Under isolation, binding logical
+ *     0 of `ASCEND_RT_VISIBLE_DEVICES=1,2`:
+ *       aclrtMallocPhysical  prop.location.id=0 -> 0        prop.location.id=2 -> 107001
+ *       aclrtMemSetAccess    desc.location.id=0 -> 507899   desc.location.id=1 -> 0
```
ChaoZheng109: The two rows in this table use different driver-visible ids for the same logical id, and I can't tell which one is the typo:

    Under isolation, binding logical 0 of ASCEND_RT_VISIBLE_DEVICES=1,2:
      aclrtMallocPhysical  prop.location.id=0 -> 0       prop.location.id=2 -> 107001
      aclrtMemSetAccess    desc.location.id=0 -> 507899  desc.location.id=1 -> 0

For logical 0 under `1,2` the driver-visible id is 1, which is what the MemSetAccess row uses. The MallocPhysical row uses 2 — that is the driver-visible id of logical 1, and it is also out of range of the logical space [0,1]. So 107001 there is equally consistent with "invalid device id", which means that half of the row does not distinguish the two id spaces at all.

The left half does: `prop.location.id=0 -> 0` succeeds even though card 0 is not in the granted set, which is real evidence that the field is logical. The conclusion looks right to me — it's specifically the second column I can't reconcile.

This matters more than a normal doc nit because this table is the only stated evidence for the per-field-not-per-API asymmetry, and it's what will stop the next reader from "fixing" `prop.location.id` symmetrically and silently allocating on the wrong card. Could you confirm whether 2 is what was actually measured, and if so which logical id it was bound to?

The same figure appears in the PR description.

src/a2a3/platform/onboard/host/comm_hccl.cpp
```diff
@@ -268,10 +269,14 @@ static aclError reserve_and_map_vmm_window(
         return status;
     }
 
+    // aclrtMemAccessDesc::location.id is consumed in the driver-visible space, unlike
+    // aclrtPhysicalMemProp::location.id above, which takes the ACL-logical id. Under
```
ChaoZheng109: `"unlike aclrtPhysicalMemProp::location.id above"` — in this file the prop is not above: it lives in `create_local_vmm_window` at line 301, which is this function's caller and is defined below it.

Looks like this block was copied from the a5 file, where the wording is accurate (prop at 460, desc at 492).

Suggest: "unlike `aclrtPhysicalMemProp::location.id` in its caller `create_local_vmm_window`, which takes the ACL-logical id".

src/common/platform/include/common/acl_hal_device.h
lterrac: You're right, and to answer the question directly: yes, `2` is what was measured, bound to logical `0`. So the row does not distinguish the two spaces — and it is a bit worse than ambiguous, because `107001` is literally `ACL_ERROR_RT_INVALID_DEVICEID`, which makes "invalid id" the natural reading rather than a competing one.

Under a `1,2` grant nothing can discriminate: the translated id of logical `0` is `1`, and `1` is itself a valid logical id there, so every value is legal in both spaces.

Re-measured on `ASCEND_RT_VISIBLE_DEVICES=2,3` bound to logical `0` (card 2), where the translated id of the bound device falls outside the logical range `[0,1]`:

```
aclrtMallocPhysical  prop.location.id=0 (bound logical) -> 0        2 GiB appears on card 2
                     prop.location.id=2 (translated)    -> 107001
                     prop.location.id=1 (other logical) -> 507899
aclrtMemSetAccess    desc.location.id=0 (logical)       -> 507899
                     desc.location.id=2 (translated)    -> 0
```

Now the `107001` row carries the argument: under a driver-visible reading, `2` *is* the bound card, so the call would have succeeded. `aclrtMemGetAllocationGranularity` rejects the same value identically, so the check is not specific to the allocation call, and the card the memory landed on was read off npu-smi's HBM column during a 2 GiB hold rather than inferred.

One thing your question made me test, and I had it wrong: I would have said a symmetric translation of the memory property allocates on the wrong card silently. It does not — `507899` for any other logical id — so it fails loudly. The header now says that instead, since "silent mis-allocation" would have been exactly the wrong warning to leave for the next reader.

Fixed in f6b511af, PR description included.

src/a2a3/platform/onboard/host/comm_hccl.cpp
lterrac: Correct — copied from the a5 file, where `prop` (460) does precede `desc` (492). In this file it is in the caller `create_local_vmm_window`, defined below. Used your wording in f6b511af.

ChaoZheng109 (approved):

PR #1840, opened 2026-08-14
Fix: support scope stats in host-built graphs

### LINKED ISSUE #1801: [Bug] host_build_graph: --enable-scope-stats produces empty scope_stats.jsonl (inert on hbg)
### Platform

a2a3 (Ascend 910B/C hardware)

### Runtime Variant

host_build_graph

### Description

On `host_build_graph` (hbg), `--enable-scope-stats` produces an empty (no-scope) `scope_stats.jsonl`. The capability is effectively **inert** on hbg.

Root cause: scope-stats is an orchestrator-driven measurement (per-scope ring/heap/tensormap occupancy at `begin_scope`/`end_scope`). In hbg the orchestrator runs on the **host**, where:
- `scope_stats_begin` / `scope_stats_end` / `scope_stats_on_fatal` / `scope_stats_set_pending_site` resolve to the weak **no-op stubs** in `host_orch_compat_stubs.cpp`, and the enable gate `is_scope_stats_enabled()` is a weak-false stub, so the orchestrator records nothing.
- There is **no device-side (scheduler) driver** for scope-stats in hbg — the on-device driver (`scope_stats_set_ring_capacity` / `set_tensormap_capacity` / `aicpu_flush_buffers`) is called only from the `tensormap_and_ringbuffer` AICPU executor; hbg's `aicpu_executor.cpp` includes the header but makes none of those calls.

The host `device_runner` still flips the DFX bit and drains device shared memory when the flag is passed, but the arena stays empty → empty JSONL. There is no hbg scope_stats scene test (only tmr has one), consistent with this being unimplemented for hbg. Documented as tmr-only in #995.

This is the host-orch weak-stub class of hbg DFX gaps tracked under #1706.

### Steps to Reproduce

```markdown
1. Run any hbg example with scope-stats enabled:
   pytest tests/st/a2a3/host_build_graph/paged_attention --platform a2a3 -d <dev> --enable-scope-stats
2. Inspect <output_prefix>/scope_stats.jsonl — it is empty / contains no scope records.
3. Contrast with tensormap_and_ringbuffer, where the same flag yields populated per-scope peaks.
```

### Expected Behavior

`--enable-scope-stats` on hbg emits per-scope ring/heap/tensormap occupancy peaks to `scope_stats.jsonl` (consumable by `simpler_setup.tools.scope_stats_plot`), matching `tensormap_and_ringbuffer`.

### Actual Behavior

`scope_stats.jsonl` is empty; nothing is ever recorded (host recorder is a no-op stub; device has no scheduler-side driver).

Key locations:
- Host no-op stubs: `src/a2a3/runtime/host_build_graph/host/host_orch_compat_stubs.cpp:39-46`
- Orchestrator (host) call sites: `src/a2a3/runtime/host_build_graph/runtime/orchestrator_core/pto_orchestrator.cpp:939-945,971-977`
- Device driver present only in tmr: `src/a2a3/runtime/tensormap_and_ringbuffer/aicpu/aicpu_executor.cpp:652,657,697,719`

Suggested fix: since scopes are fully known at host graph-build time for hbg, implement a **host-side** scope-stats collector (replace the no-op stubs with a real host recorder feeding the existing `scope_stats.jsonl` exporter).

### Git Commit ID

50c0660646217f1b56172a81990279bd066ed971

### Host Platform

Linux (aarch64)

### Additional Context

Related: #1706, #995. Sibling hbg DFX gaps filed separately (arg-dump selective/hybrid empty; chip-swimlane L4 no-op).

vegetabledoww:
## Summary

Fixes `--enable-scope-stats` for the `host_build_graph` (HBG) runtime. HBG now captures orchestration scope boundaries on the host and writes populated version 6 `scope_stats.jsonl` output, while the existing `tensormap_and_ringbuffer` (T&R) device-side collection path remains unchanged.

Fixes #1801.

## Root cause

HBG executes orchestration while constructing the graph on the host. Its scope-stat hooks previously resolved to weak no-op compatibility stubs, so no scope records were produced. The runner still enabled and drained the device collector, but HBG has no device-side scheduler driving that collector; the resulting JSONL file was therefore empty.

## Changes

- Add a thread-local host recorder for HBG scope begin/end events, source locations, resource counters, heap wraps, and fatal state.
- Replace the a2a3 and a5 HBG no-op scope-stat hooks with host-capture implementations.
- Start capture before the outer orchestration scope and emit the captured records during runner teardown.
- Skip the unused device scope-stat flag, shared-memory allocation, and drain thread for HBG.
- Reuse the shared JSONL serializer so HBG and T&R produce the same version 6 schema and remain compatible with `simpler_setup.tools.scope_stats_plot`.
- Preserve the existing AICPU streaming collector for T&R.
- Add HBG scene coverage for a2a3 and a5, host-capture unit tests, and updated scope-stats documentation.

HBG represents a whole-graph execution with one polling ring and no dependency-list pool. Its output therefore populates ring index 0 and reports `dep_pool_max[0]` as zero.

## Validation

- C++ scope-stats unit tests: 4/4 passed.
- a2a3sim HBG scope-stats scene passed.
- a5sim HBG scope-stats scene passed.
- Existing a2a3sim and a5sim T&R scope-stats regressions passed.
- a5sim regression matrix passed: 37 resource cases, HBG L2 15/15, and T&R L2 40/40.
- a2a3 hardware HBG scope-stats scene passed: 1 test.
- a2a3 hardware T&R scope-stats regression passed: 1 test.
- Editable builds passed for a2a3/a5 onboard and simulator runtimes.
- Full pre-commit suite passed: clang-format, clang-tidy, cpplint, markdownlint, ruff, and pyright.
- `git diff --check` passed.

a5 hardware was not available, so a5 validation is simulator-only.

### DIFF
--- a/docs/dfx/scope-stats.md
+++ b/docs/dfx/scope-stats.md
@@ -7,9 +7,9 @@ high-water mark. When a model runs out of task windows, heap, or
 tensormap / dependency-list entries, the failure tells you *which* resource is exhausted
 but not *where*; scope stats gives you the where.
 
-It is a diagnostic-only, opt-in feature for the
-**tensormap_and_ringbuffer (T&R)** runtime. When disabled (the default)
-it costs a single bool load per probe.
+It is a diagnostic-only, opt-in feature for both
+**tensormap_and_ringbuffer (T&R)** and **host_build_graph (HBG)**. When
+disabled (the default) it costs a single bool load per probe.
 
 This guide also covers the background behind the feature, the T&R
 resource/ring_depth/scope model, and the data flow behind the HTML report.
@@ -21,7 +21,7 @@ resulting `scope_stats/scope_stats.jsonl` into an HTML report.
 
 ### Step 1 — Run with `--enable-scope-stats`
 
-Pass the flag to any T&R example or scene test:
+Pass the flag to a T&R or HBG example or scene test:
 
 ```bash
 CASE=...
@@ -29,9 +29,9 @@ NAME=...
 python "tests/st/${CASE}/test_${NAME}.py" -p a2a3 -d 0 --enable-scope-stats
 ```
 
-The flag is bit 4 of `enable_profiling_flag`; on a T&R run it turns on
-per-scope peak tracking. On other runtimes the flag is accepted but
-produces no records.
+The flag is bit 4 of `enable_profiling_flag` on the device-orchestrated
+T&R path. HBG consumes the same flag before graph construction and records
+the host orchestrator's scope boundaries directly.
 
 ### Step 2 — Locate the output
 
@@ -68,6 +68,8 @@ PY
 The three arrays are indexed by `ring` (`0..3`) and should match the effective
 runtime configuration. Per-sample `ring` values show which scope-depth rings
 were actually touched by the run; they are scope records, not task counts.
+HBG is whole-graph-resident with one polling ring and no dependency-list pool,
+so only index 0 is populated and `dep_pool_max[0]` is zero.
 
 ### Step 3 — Visualize with `scope_stats_plot.py`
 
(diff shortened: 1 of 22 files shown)


ChaoWao:
Need to evaluate that why hbg needs scope states. It does not need task/dep/heap ring in the end.

PR #1892, opened 2026-08-19
CI: centralize bounded APT installation

doraemonmj:
## Problem

`packaging-matrix` could remain in **Set up C++ compiler** until the job's
60-minute limit when an Ubuntu package mirror stalled. The mirror instability
predates this change; #1805 did not create it, but its packaging-tool setup
introduced unbounded `apt-get update` and `apt-get install` calls that exposed
the job to it.

Of the OS packages installed there, only the host compiler is required.
`ccache` is an optimization and the packaging matrix already supports running
without it. `ninja` can be installed with the Python build dependencies instead
of requiring another system-package install.

## Changes

- Add a shared `apt-install` composite action for required and optional
  packages. Every update, install, and dpkg recovery operation has a wall-clock
  bound; interrupted state is cleaned up before retrying.
- Reuse the runner's cached package indexes by default in packaging,
  pre-commit, and no-hardware UT. If a required install fails, refresh the
  system indexes once and retry the install.
- Refresh only the Ubuntu Toolchain PPA source in `setup-gcc-15`, preserving
  the runner's existing Ubuntu indexes. The PPA update and required install
  each receive one retry.
- Treat `ccache` as optional in packaging: an APT or Homebrew failure emits a
  warning and the matrix continues without compiler caching.
- Install `ninja>=1.11` with the venv and update the packaging documentation and
  verification hint accordingly.
- Add regression coverage for the shared action and workflow wiring, including
  Bash 3.2 compatibility used by macOS runners.

## Bounds

| Operation | Policy |
| --- | --- |
| APT connection | 20-second HTTP/HTTPS timeout, no internal APT retry |
| Regular required install | 240 seconds; on failure, one 120-second system-index refresh and one final install |
| GCC PPA refresh | Up to two 60-second attempts, limited to the PPA `.sources` file |
| GCC required install | Up to two 240-second attempts |
| dpkg recovery | 60 seconds, then a 15-second kill grace period |

These are per-operation bounds. No job-level `timeout-minutes` value is changed.

## Testing

- All pre-commit hooks passed locally and in CI.
- Focused APT-action regression tests passed on Ubuntu and macOS.
- Packaging and no-hardware UT passed on both Ubuntu and macOS in
  [CI run 32325726639](https://github.com/hw-native-sys/simpler/actions/runs/32325726639).

### DIFF
--- a/.github/actions/setup-gcc-15/action.yml
+++ b/.github/actions/setup-gcc-15/action.yml
@@ -16,7 +16,8 @@ inputs:
 runs:
   using: composite
   steps:
-    - name: Install GCC 15 and build tools
+    - name: Prepare GCC 15 and build tools
+      id: packages
       if: runner.os == 'Linux'
       shell: bash
       env:
@@ -88,14 +89,42 @@ runs:
               sudo tee /etc/apt/sources.list.d/ubuntu-toolchain-r-test.sources >/dev/null
           fi
 
-          apt_options=(
-            -o Acquire::Retries=3
-            -o Acquire::http::Timeout=20
-            -o Acquire::https::Timeout=20
-          )
-          sudo apt-get "${apt_options[@]}" update
-          sudo apt-get "${apt_options[@]}" install -y "${packages[@]}"
+          if [[ "$needs_gcc_ppa" == true ]]; then
+            echo "update-mode=source-file" >> "$GITHUB_OUTPUT"
+            echo "source-file=/etc/apt/sources.list.d/ubuntu-toolchain-r-test.sources" >> "$GITHUB_OUTPUT"
+            echo "update-attempts=2" >> "$GITHUB_OUTPUT"
+            echo "install-attempts=2" >> "$GITHUB_OUTPUT"
+          else
+            echo "update-mode=none" >> "$GITHUB_OUTPUT"
+            echo "source-file=" >> "$GITHUB_OUTPUT"
+            echo "update-attempts=1" >> "$GITHUB_OUTPUT"
+            echo "install-attempts=1" >> "$GITHUB_OUTPUT"
+          fi
+          echo "required-packages=${packages[*]}" >> "$GITHUB_OUTPUT"
         fi
+
+    - name: Install GCC 15 and build tools
+      if: runner.os == 'Linux' && steps.packages.outputs.required-packages != ''
+      uses: ./.github/actions/apt-install
+      with:
+        required-packages: ${{ steps.packages.outputs.required-packages }}
+        update-mode: ${{ steps.packages.outputs.update-mode }}
+        source-file: ${{ steps.packages.outputs.source-file }}
+        update-attempts: ${{ steps.packages.outputs.update-attempts }}
+        update-timeout-seconds: "60"
+        install-timeout-seconds: "240"
+        install-attempts: ${{ steps.packages.outputs.install-attempts }}
+
+    - name: Verify GCC 15 and build tools
+      if: runner.os == 'Linux'
+      shell: bash
+      run: |
+        set -euo pipefail
+
+        is_version_15() {
+          command -v "$1" >/dev/null 2>&1 &&
+            [[ "$("$1" -dumpversion | cut -d. -f1)" == 15 ]]
+        }
 
         command -v gcc-15
         command -v g++-15
--- a/.github/workflows/_packaging.yml
+++ b/.github/workflows/_packaging.yml
@@ -38,26 +38,40 @@ jobs:
           persist-credentials: false
 
       - name: Set up C++ compiler
+        id: packaging-tools
         if: inputs.setup_variant == 'github'
         run: |
+          # ninja arrives from PyPI with the venv, so the only package a
+          # packaging build cannot proceed without is a host compiler. ccache
+          # merely warms the matrix's later modes and the matrix step already
+          # runs without it, so a package manager that cannot serve it is a
+          # warning, not a failure.
           if [[ "${{ runner.os }}" == "Linux" ]]; then
-            missing_packages=()
-            command -v ninja >/dev/null 2>&1 || missing_packages+=(ninja-build)
-            command -v ccache >/dev/null 2>&1 || missing_packages+=(ccache)
+            required_packages=()
             if ! command -v gcc >/dev/null 2>&1 || ! command -v g++ >/dev/null 2>&1; then
-              missing_packages+=(g++)
+              required_packages+=(g++)
             fi
-            if (( ${#missing_packages[@]} )); then
-              sudo apt-get update
-              sudo apt-get install -y "${missing_packages[@]}"
-            fi
+            optional_packages=()
+            command -v ccache >/dev/null 2>&1 || optional_packages+=(ccache)
+            echo "required-packages=${required_packages[*]}" >> "$GITHUB_OUTPUT"
+            echo "optional-packages=${optional_packages[*]}" >> "$GITHUB_OUTPUT"
           else
-            command -v ninja >/dev/null 2>&1 || brew install ninja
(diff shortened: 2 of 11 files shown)


.github/actions/setup-gcc-15/action.yml
```diff
@@ -88,13 +88,18 @@ runs:
...
-            -o Acquire::http::Timeout=20
-            -o Acquire::https::Timeout=20
+            -o Acquire::Retries=1
+            -o Acquire::http::Timeout=15
+            -o Acquire::https::Timeout=15
           )
-          sudo apt-get "${apt_options[@]}" update
-          sudo apt-get "${apt_options[@]}" install -y "${packages[@]}"
+          sudo timeout 60 apt-get "${apt_options[@]}" update
+          sudo timeout 120 apt-get "${apt_options[@]}" install -y "${packages[@]}"
```
ChaoZheng109: **`timeout 120` here carries much less headroom than the Bounds table implies, and `Retries=1` lands on the one apt call whose stall mode this PR is *not* fixing.**

The "healthy runs take 9-15s, so the caps carry 4-8x headroom" figure is measured on the packaging / pre-commit / ut steps. This call is not comparable to those:

- Its packages are `gcc-15 g++-15`, which pull in `cpp-15`, `libgcc-15-dev` and `libstdc++-15-dev` — roughly 70-90 MB of .debs, not an index refresh plus a couple of small tools.
- They come from the **ubuntu-toolchain-r PPA** (`ppa.launchpadcontent.net`), which is neither `azure.archive.ubuntu.com` nor the public archive — i.e. not the mirrorlist fallback path described in the PR body. Launchpad throughput from Azure egress is commonly 1-5 MB/s, so download plus unpack puts the real headroom nearer 1.5-2x than 4-8x.

Lowering `Acquire::Retries` 3 -> 1 on the same call compounds it: this is the most download-heavy apt invocation left in the tree, and it is now strictly less resilient than before the PR, for a stall mode the PR did not set out to address.

The blast radius is also wider than packaging's — `setup-gcc-15` is consumed by `_st-sim-a2a3.yml`, `_st-sim-a5.yml`, `_profiling-flags-smoke.yml` and `sanitizers.yml`. If the bound trips, all of those turn red rather than slow, and each one becomes a triage item.

Suggestion: give this site its own bound (e.g. `timeout 240` on the install), and/or keep `Acquire::Retries=3` here since the PPA is not the host the mirrorlist retries protect against. Either way, please note in the Bounds table that this row's headroom is measured against a different workload from the other three.

.github/workflows/_packaging.yml
```diff
@@ -72,7 +98,7 @@ jobs:
       - name: Install build + runtime deps in venv
         uses: ./.github/actions/setup-venv
         with:
-          packages: "scikit-build-core>=0.10.0 nanobind>=2.0.0 cmake>=3.15 pytest>=6.0"
+          packages: "scikit-build-core>=0.10.0 nanobind>=2.0.0 cmake>=3.15 ninja>=1.11 pytest>=6.0"
```
ChaoZheng109, line 89: **`ninja` is added to the CI venv but not to the documented local install line, so `verify_packaging.sh` now behaves differently in CI and locally.**

`docs/python-packaging.md:126` and `:161` both document the one-time install as

```
pip install 'scikit-build-core>=0.10.0' 'nanobind>=2.0.0' 'cmake>=3.15' 'pytest>=6.0' 'torch>=2.3'
```

and `tools/verify_packaging.sh`'s own failure hint prints that same line. After this change CI's packaging venv has ninja while a developer following the docs does not — so locally the script takes the `ninja executable: not found; CMake will select another generator` branch and CI does not. Given that the point of `verify_packaging.sh` is for the five install paths to reproduce locally exactly as CI runs them, generator selection is the one thing that should not diverge.

Per `.claude/rules/doc-consistency.md` §1/§4 this belongs in the same commit: add `ninja` to both doc lines and to the script's error hint, or state in `docs/ci.md` why the divergence is acceptable.

Related, on the Note in the PR description: I'd keep `setup-venv` ungated — identical generator selection across lanes is worth more than saving one wheel on the self-cpu lane. But then `docs/ci.md:253` ("cpu runner contract: dnf-installed `cmake ninja-build gcc-c++ ...`") should say that the venv's PyPI ninja shadows the dnf-provided one in the packaging job, otherwise that contract starts to lie.

ChaoZheng109 (commented):
Two Should-fix items from a full review of this PR. The diagnosis and the overall approach look right to me — putting the bound on the apt call rather than the workflow step is the judgement that makes the required/optional split possible, the "every apt invocation" claim holds under an exhaustive sweep of `.github/` at HEAD, and both packaging jobs are green so the PyPI-ninja path is exercised end to end. These two are the parts I'd want addressed or answered before merge.

.github/actions/setup-gcc-15/action.yml
doraemonmj: Agreed on both counts — the headroom figure was measured on the wrong workload and the retry reduction landed on the one call it doesn't help. Fixed in e153ed89.

`setup-gcc-15` now has its own bound and keeps the original acquire settings:

```yaml
apt_options=(
  -o Acquire::Retries=3
  -o Acquire::http::Timeout=20
  -o Acquire::https::Timeout=20
)
sudo timeout 60  apt-get "${apt_options[@]}" update
sudo timeout 240 apt-get "${apt_options[@]}" install -y "${packages[@]}"
```

So this site is unchanged from `main` except for the `timeout` wrapper, which is the part that was actually missing. Your framing of *why* retries should stay at 3 is the one I'd missed: the mirrorlist has no second host to offer for a single-host PPA, so `Retries=1` there costs resilience and buys nothing — the reasoning that motivated the reduction only applies to the mirrorlist sites, which keep `Retries=1` / `Timeout=15`. That's now in a comment above the `apt_options` block so the next reader doesn't re-derive it, and the divergence is spelled out in `docs/ci.md` too.

The Bounds table row is `5 min (60 + 240)`, with a paragraph under it stating that this row is measured against a different workload from the other three — 70-90 MB of compiler .debs off launchpad versus an index refresh plus small tools — and naming the four consumers so the blast radius is on the page rather than in the review thread.

The three workflow sites are unchanged: `timeout 60` / `timeout 120`, `Retries=1`, `Timeout=15`. Those are the ones the 4-8x figure was actually measured on.

.github/workflows/_packaging.yml
doraemonmj: Right, and this is the divergence that matters most — generator selection is exactly what `verify_packaging.sh` exists to keep identical. Fixed in e153ed89.

`ninja>=1.11` added to:

- `docs/python-packaging.md:126` and `:161` — the two one-time install lines
- `tools/verify_packaging.sh:145` — the failure hint, which printed the same line
- `docs/getting-started.md:111` — the daily-dev setup, which isn't in your list but is the same divergence and would have left the most common local path on the other branch

I left `ninja` out of the script's `required` tuple deliberately: the `ninja executable: not found; CMake will select another generator` branch is a real fallback for someone who hasn't re-run the install line, and promoting it to a hard failure trades a working build for an error. The docs now steer everyone onto the ninja path, which closes the divergence without removing the fallback.

On the Note: agreed, `setup-venv` stays ungated. `docs/ci.md`'s cpu runner contract now says so explicitly —

> `setup-venv` installs a PyPI `ninja` into the venv on every lane, which shadows the dnf-provided `ninja-build` on `PATH` whenever the venv is active — deliberately, so CMake's generator selection is identical on this lane and on the github-hosted ones. Same tool, different provenance; the dnf package stays in the contract for builds that run outside the venv.

The last clause is why `ninja-build` stays in the dnf list rather than being dropped: not every build on that lane runs inside the venv.

ChaoZheng109 (approved):

Issue #1832, opened 2026-08-14
[Bug] Daily full-scene run #31722222789 exposes manual PA regressions and sim lane poisoning


doraemonmj:
## Summary

[Daily Full Scene Tests run #31722222789](https://github.com/hw-native-sys/simpler/actions/runs/31722222789) at `9a6a785934c3747d29547c67ee8c5849cd5b2c72` failed in 5 of 7 jobs.

The run contains **26 pytest failure records**, but only **20 are direct failures**. The remaining **6 are cascades after an a2a3sim run lane was poisoned by a scheduler timeout**.

This is not caused by Daily using a different execution environment from Per-PR. Daily and Per-PR reuse the same `_st-*` workflows and platforms. The important difference is selection:

- Per-PR runs with `manual_mode=exclude`.
- Daily runs with `manual_mode=include`, i.e. the Per-PR scene set **plus manual cases**.
- The pytest node is usually reported as `Class::test_run`, while that method loops through multiple `CASES`. Therefore, the same class passing in Per-PR only proves that its non-manual cases passed.

For comparison, the corresponding scene jobs passed in [PR run #31697418830](https://github.com/hw-native-sys/simpler/actions/runs/31697418830). The Daily failures are all reached through manual cases; they do not contradict the Per-PR result.

## Failure classification

| Category | Direct failure records | Affected coverage | Observed cause |
|---|---:|---|---|
| Host Build Graph task-ring exhaustion | 2 | a2a3 and a5 onboard | `Task Allocator Deadlock - Task Ring Full`, `active=16383/16384`; these HBG cases build roughly 64K tasks before scheduling |
| Paged Attention golden mismatch | 12 | a2a3 and a5 onboard | 11 large BF16 `Case3` workloads fail with `max_diff` around 1.05–1.08; new A5 SPMD zero-query case fails with `max_diff=inf` |
| SPMD PA compile failures | 4 | a2a3/a5 sim, macOS and Ubuntu | Clang rejects `unsigned long` to `uint32_t` narrowing; Linux also calls nonexistent `setEntryOffset` instead of `setentryOffset` |
| HighPerf PA scheduler stall | 2 | a2a3sim on macOS and Ubuntu | Manual `b4_h32_kv8_s512_bs128_fp16` stalls with `SCHEDULER_TIMEOUT`, `completed=0/1`, `stuck_task_id=0` |
[...]

PR #1918, opened 2026-08-20
Fix: prevent poisoned chip-run lane reuse

doraemonmj:
## Summary

- Associate a poisoned chip-run lane with the run that caused the poison, while preserving errors from earlier independent runs.
- Retire pooled Workers after poll/finalize lane failures, poison markers, or sticky device-context errors so later tests cannot reuse a terminal lane.
- Align native-run error expectations, classifier coverage, and runtime troubleshooting documentation.

## Failure and result

A terminal native-run failure poisoned the pooled chip-run lane, but the failed Worker could be returned to the L2 pool. Later tests then failed with the inherited poison instead of running. The lane now reports poison against the responsible run, and pytest retires the Worker for poll/finalize failures, poison markers, and known sticky device errors.

## Scope

This change addresses only the poisoned-lane cascade tracked by #1832. The HighPerf paged-attention simulator behavior remains unchanged and is outside this PR.

## Testing

- Python unit tests: 1677 passed, 13 skipped, 14 deselected.
- C++ unit tests: 114 passed.
- Pre-commit hooks passed for all changed files.

Refs #1832

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -1255,19 +1255,17 @@ def pytest_sessionfinish(session, exitstatus):  # noqa: ARG001
 # physical device. The Worker records actual SDMA dispatch attempts; merely
 # registering an unused SDMA callable does not rotate the pool.
 #
-# The driver-side guard
-# (``DeviceRunner::run`` fail-fast on ``device_unusable_``) keeps that cascade
-# from wedging, but the Worker stays poisoned.
-#
-# On a5 the poison survives close()+device-reset for the life of the process
-# (an in-process Worker.init rebuild fails with rtStreamCreate 507899; a
-# force-reset is unsafe on this shared box), so there is no in-process
-# recovery — only a fresh worker process gets a clean device. The native status
-# query tells the ``st_worker`` finalizer when to drop a poisoned or
-# SDMA-exposed pooled Worker; the hook below also stashes setup/call exceptions
-# as a diagnostic fallback. The L2 branch rebuilds where the arch allows it and otherwise
-# skips the remaining tests for that runtime (``_l2_poisoned``). Golden
-# mismatches / ordinary assertions leave a non-SDMA Worker reusable.
+# The driver-side guard (``DeviceRunner::run`` fail-fast on
+# ``device_unusable_``) keeps that cascade from wedging. Fatal onboard teardown
+# force-resets the exclusively allocated card with ``aclrtResetDeviceForce``
+# and probes the new device generation before confirming recovery. A successful
+# reset lets the next Worker initialize in the same process; an unconfirmed
+# reset or failed rebuild marks the runtime in ``_l2_poisoned`` and falls back
+# to a fresh-process retry. For runtime failures, the hook below preserves the
+# setup/call exception text and the ``st_worker`` finalizer uses the retirement
+# classifier below to decide whether the pooled Worker can be reused. Exception
+# wording is therefore part of that decision input, not merely diagnostic.
+# Golden mismatches and ordinary assertions leave a non-SDMA Worker reusable.
 # ---------------------------------------------------------------------------
 
 # Plain attribute name on the test item — NOT pytest.StashKey()/item.stash,
@@ -1284,18 +1282,17 @@ def pytest_runtest_makereport(item, call):
     yield
     # A fixture that depends on st_worker may dispatch during setup. Preserve
     # that error before st_worker's teardown finalizer decides whether its
-    # Worker can return to the session pool. Native lifecycle status below is
-    # authoritative; this text classification remains a diagnostic fallback.
+    # Worker can return to the session pool; its retirement decision consumes
+    # this exception text through the classifier below.
     if call.when in {"setup", "call"} and call.excinfo is not None:
         setattr(item, _ST_CALL_EXCINFO, call.excinfo)
 
 
-# Only these error codes mean the device/ACL context is sticky-errored — i.e.
-# the worker is poisoned and must be rebuilt/skipped. Matching on the bare
-# "<op> failed with code" prefix would also catch ordinary kernel/run failures
-# (every non-zero rc is wrapped that way), turning a normal failing test into a
-# spurious worker rebuild and, downstream, a misleading runtime-wide skip. So
-# we extract the trailing <N> and match only the known poison codes:
+# Code-bearing native/init failures retire a Worker only when the code identifies
+# a sticky device/ACL context. Matching every "<op> failed with code" message
+# would also catch ordinary run failures and turn them into spurious rebuilds.
+# This code allowlist is separate from the retirement markers below, which
+# identify a non-reusable Worker independently of the originating status code:
 #   207001 ACL_ERROR_RT_MEMORY_ALLOCATION  (the CI cascade trigger)
 #   507000 ACL_ERROR_RT_INTERNAL_ERROR  (a5 op-timeout at AICPU stream sync)
 #   507015 ACL_ERROR_RT_AICORE_EXCEPTION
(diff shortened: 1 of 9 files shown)


conftest.py
```diff
@@ -1258,6 +1258,14 @@ def pytest_runtest_makereport(item, call):
...
     "device reset was not confirmed, so this process quarantined the device",
     "cleanup could not confirm device reset, so this process quarantined the device",
+    # A finalize failure poisons the chip run lane (ChipRunLaneState::poison_with),
+    # and a poisoned lane rejects every later submit on that Worker. The first
+    # marker is what the run that poisoned it reports, the second what every
+    # subsequent run on the same Worker reports. Neither carries a code the regex
+    # above can read: the verb is finalize_native_run, and the code is the device
+    # runtime status (-100 for a scheduler timeout), not an ACL error code.
+    "finalize_native_run failed with code",
+    "chip run lane is poisoned",
```
ChaoZheng109: These two markers are right, and the comment's diagnosis ("neither carries a code the regex above can read") is correct for the `finalize_native_run` path. But the reason is more general than the comment says, and it leaves two adjacent holes open.

**The regex on line 1255 no longer matches *any* message `ChipWorker` produces.** The leading `\b` cannot match `*_native_run`, because `_` is a word character, so there is no word boundary between `_` and `run`:

```
NO MATCH   | prepare_native_run  failed with code 507018 (run_id=0 slot=0)
NO MATCH   | launch_native_run   failed with code 507018 (run_id=0 slot=0)
NO MATCH   | poll_native_run     failed with code 507018 (run_id=0 slot=0)
NO MATCH   | finalize_native_run failed with code -100   (run_id=0 slot=0)
MATCH 507899 | simpler_init failed with code 507899
```

And `run failed with code` / `run_prepared failed with code` / `prepare_callable failed with code` have **no producer left** anywhere in `src/` or `python/` — the only surviving references are this regex and a `pytest.raises(match=...)` in `tests/st/aicore_op_timeout/test_aicore_op_timeout.py:110`, which still passes only because `match` uses `re.search` *without* `\b` and therefore matches the `...native_run failed with code N` substring. The four real messages are `chip_worker.cpp:677/744/776/836`. So `_DEVICE_POISON_CODES` is effectively dead surface today: `simpler_init` is the one verb it still reaches.

Two consequences for this fix:

**1. `poll_native_run` also poisons the lane and also escapes the classifier.** `progress()` (`chip_run_lane.cpp:176-183`) calls `poison_with` unconditionally when `poll_native_run` throws, exactly like `finish()` does for finalize. Its message matches neither the regex nor either new marker, so the *first* failing test keeps the poisoned Worker pooled and the *next* one eats a false failure before the `chip run lane is poisoned` marker recycles it. Cascade drops from N to 1, not to 0.

**2. `prepare_native_run` / `launch_native_run` failures are not reachable by the marker approach at all.** Neither poisons the lane — `launch_front()` (`:123-131`) just marks that run `TERMINAL` and pops it, and `fail_launch()` (`:100-103`) only poisons if the subsequent finalize throws. So no later test ever sees `chip run lane is poisoned`, the markers never fire, and the regex misses the message too. If the *device context* is poisoned (507018 / 507899 / 207001), the Worker stays pooled forever and every later test re-trips the same failure. That one is pre-existing, not introduced here — but it is the reason the code allowlist is not doing the job it was written to do.

Both close with a one-line change to the verb list:

```python
_DEVICE_ERROR_CODE_RE = re.compile(
    r"(?:(?:prepare|launch|poll|finalize)_native_run|simpler_init) failed with code (-?\d+)\b")
```

That keeps the two layers cleanly separated: the codes classify the *originating* failure (device-context poisoning), the markers classify what every *subsequent* reuse sees (lane poisoning).

If you'd rather keep that out of this PR's scope — fair, hole 2 predates it — could you at least note in the comment here that the regex verb list has drifted from `chip_worker.cpp`? As written, the comment reads as though the allowlist above is still working, and the next reader will believe it.

ChaoZheng109 (commented):
Reviewed the poison-classification half of this PR against the C++ lane semantics in `src/common/worker/chip_run_lane.cpp` and `chip_worker.cpp`. The two markers are correctly placed and do close the exact path #1832 hit. Left three inline notes: the regex they sit next to has drifted away from every message `ChipWorker` actually produces, which leaves two adjacent poison paths still unclassified. Details inline.

tests/ut/py/test_device_poison_detection.py
```diff
@@ -0,0 +1,59 @@
...
+    return mod
+
+
+@pytest.mark.parametrize(
+    "msg",
+    [
+        f"finalize_native_run failed with code -100 {_LANE_POISON_TAIL}",
+        f"chip run lane is poisoned: finalize_native_run failed with code -100 {_LANE_POISON_TAIL}",
+        "chip run lane is poisoned by an unknown native failure",
+        "run_prepared failed with code 507018",
```
ChaoZheng109: `run_prepared failed with code 507018` is not a message this codebase can produce — grep finds no producer in `src/` or `python/`. The real one is `prepare_native_run failed with code 507018 (run_id=... slot=...)`, from `chip_worker.cpp:677`.

So this case asserts `True` via a string that only the (drifted) regex recognises, and the message that would actually arrive here asserts `False`. Same for the golden-path pair on lines 39-40, which are real — those two are the ones carrying the test's weight.

Suggest replacing this with the real spelling; it turns a passing-by-coincidence case into one that would have caught the drift.

tests/ut/py/test_device_poison_detection.py
```diff
@@ -0,0 +1,59 @@
...
+)
+def test_poisoning_messages_recycle_the_worker(msg):
+    assert _load_root_conftest()._is_device_runtime_error_msg(msg)
+
+
+@pytest.mark.parametrize(
+    "msg",
+    [
+        "Golden mismatch on 'out': max_diff=0.0625, rtol=0.005, atol=0.02",
+        "run failed with code 42",
```
ChaoZheng109: Two things stacked here.

First, `run failed with code 42` is also a format with no producer left (see the note on line 42).

Second, and more important: this codifies "an originating native failure whose code is outside the allowlist keeps the Worker pooled" as *intended* behaviour. But lane poisoning in `chip_run_lane.cpp` is entirely code-independent — `poison_with` is called on any throw from poll/wait/finalize regardless of `rc`. So this assertion locks in the residual one-test cascade described on the conftest thread, and anyone fixing the verb list later has to argue this green test away first.

If the allowlist is a deliberate cost trade-off (rebuilding a Worker is expensive, so only known-fatal codes should trigger it), that rationale is worth a comment right here — it is not obvious from the outside, and it is the part a future reader will need in order to decide whether they may touch it.

doraemonmj: Fixed in `052883c9`. The negative case now uses a real ordinary `prepare_native_run ... code 42` failure, which does not poison the lane. Poll/finalize failures of any code moved to the positive retirement cases because those phases leave the Worker lane non-reusable. The adjacent comments now explicitly distinguish sticky device-context codes from code-independent lane-terminal markers.

tests/ut/py/test_device_poison_detection.py
doraemonmj: Fixed in `052883c9`. The synthetic `run_prepared` case was replaced with the real `prepare_native_run` spelling, and coverage now includes the real launch, poll, and finalize forms. The remaining stale synthetic native-run message in the resource-failure test was updated as well.

conftest.py
doraemonmj: Fixed in `052883c9`. `_DEVICE_ERROR_CODE_RE` now matches the four real `prepare/launch/poll/finalize_native_run` spellings plus `simpler_init`. `poll_native_run` and `finalize_native_run` are also unconditional Worker-retirement markers because either failure poisons the lane regardless of status code; prepare/launch still use the sticky-device code allowlist. The classifier tests cover the real messages and arbitrary poll/finalize codes.

ChaoZheng109 (approved):

PR #1923, opened 2026-08-20
Optimize: pack HBG Graph replay data into one arena block

sunkaixuan2018:
## Summary

Reduce host-side allocation and H2D-copy overhead for host-built Graph replay data.

- Carry replay boundary tensors and scalars in the outer `GRAPH` task payload, leaving `GraphSubmission` as a fixed-size Definition reference plus device replay state.
- Compute and validate the complete `[Definitions][Submissions]` layout, including every size, alignment, and offset, before acquiring device storage.
- Store one grow-only GraphBlock on the arena bank selected for the run. Reuse its allocation when capacity is sufficient and grow it only when required; the block is not Worker-owned.
- Address all internal objects as `GraphBlock base + offset` and overwrite the complete live range with exactly one H2D copy on every run.
- Release or abandon the GraphBlock with its arena bank on normal or fatal teardown for both onboard and simulator runners.

This deliberately does not add content-based copy skipping, Task Window packing, queue sizing, heap sizing, or environment-variable-driven allocation.

## Motivation

Host `rtMalloc` is expensive, and one large `rtMemcpy` is materially cheaper than many small copies. The previous path retained each Definition separately and allocated/uploaded every submission independently. Qwen3 performs 40 Graph replays, producing 41 allocation/H2D operations per run.

The new path computes the whole GraphBlock first, performs at most one allocation/grow operation for the selected arena bank, and then performs one H2D upload.

## Performance

Measured with `GraphExecutionBatch16Seq3500` from the Qwen3-14B decode HBG example on A2A3. Baseline and PR were rebuilt from adjacent runs against current `main` (`93adc386`).

| Metric | main | this PR | Change |
|---|---:|---:|---:|
| Graph upload operations | 41 | 1 | -40 calls |
| Graph upload bytes | 232,312 B | 131,832 B | -43.25% |
| `graph_upload` host span | 1,381.044 us | 243.623 us | -82.36% |
| Device wall | 39.920 ms | 39.925 ms | no regression observed |

The direct `graph_upload` span is the relevant result; end-to-end device execution is expected to remain unchanged because this PR only changes host preparation and upload.

## Validation

- Rebased cleanly onto current `main` (`93adc386`).
- Rebuilt A2A3 and A5 runtime/bindings after the rebase.
- C++ no-hardware suite: 107/107 passed.
- Graph execution simulator coverage passed on both A2A3 and A5:
  - `record_then_replay_1d`
  - `record_then_replay_2d` (multiple Definitions and submissions)
- A2A3 Qwen3 `GraphExecutionBatch16Seq3500` correctness and profiling passed.
- A three-round A2A3 Qwen3 run passed and covered arena-bank rotation followed by reuse of an already provisioned bank.

A5 onboard hardware was not available on the A2A3 validation host; A5 compilation, C++ tests, and A5 simulator coverage passed.

### DIFF
--- a/src/a2a3/platform/sim/host/device_runner.cpp
+++ b/src/a2a3/platform/sim/host/device_runner.cpp
@@ -802,20 +802,21 @@ int DeviceRunner::finalize() {
     release_callable_state();
 
     unload_executor_binaries();
-    release_graph_definition_buffers();
 
-    // Release the three per-Worker pooled arenas. Must precede mem_alloc_.finalize()
+    // Release the arena-bank regions. Must precede mem_alloc_.finalize()
     // so the arenas free through the still-live allocator, not after it.
     for (auto &bank : arena_banks_) {
         bank->gm_heap.release();
         bank->gm_sm.release();
         bank->runtime_pool.release();
+        bank->graph_block.release();
     }
     clear_temporary_buffer();
     for (auto &bank : arena_banks_) {
         bank->cached_gm_heap_size = 0;
         bank->cached_gm_sm_size = 0;
         bank->cached_runtime_arena_size = 0;
+        bank->cached_graph_block_size = 0;
     }
     prebuilt_runtime_arena_cache_valid_ = false;
     prebuilt_runtime_arena_cache_key_.clear();
--- a/src/a2a3/runtime/host_build_graph/host/runtime_maker.cpp
+++ b/src/a2a3/runtime/host_build_graph/host/runtime_maker.cpp
@@ -417,55 +417,54 @@ static bool relocate_host_orch_image(
     return ok;
 }
 
-bool upload_graph_submissions(
-    Runtime *runtime, const HostApi *api, GraphHostState &graph_state, uint64_t &uploaded_bytes
-) {
+bool upload_graph_block(const HostApi *api, GraphHostState &graph_state, uint64_t &uploaded_bytes) {
     uploaded_bytes = 0;
     const size_t count = graph_host_upload_count(graph_state);
-    // Pass 1: upload each distinct Definition once as a shared device object
-    // ([GraphDefinitionHeader][Definition image]) keyed by content identity.
-    // Submissions reference the object's GM address, so this pass completing
-    // before any submission is uploaded is what makes the reference safe —
-    // the device boots only after both passes.
-    GraphHostDefinitionList definitions = graph_host_definitions(graph_state);
-    struct UploadedDefinition {
-        void *device_object;               // GM address; host must not dereference
-        const GraphDefinition *host_view;  // the host-side image the object was built from
+    struct PackedDefinition {
+        const GraphHostDefinition *entry;
+        const GraphDefinition *definition;
+        size_t offset;
     };
-    std::unordered_map<uint64_t, UploadedDefinition> definition_objects;
+    struct PackedSubmission {
+        PTO2TaskSlotState *outer_slot;
+        const GraphSubmission *submission;
+        size_t definition_index;
+        size_t offset;
+    };
+    auto reserve_region = [](size_t bytes, size_t alignment, size_t *cursor, size_t *offset) -> bool {
+        if (cursor == nullptr || offset == nullptr || alignment == 0 || (alignment & (alignment - 1)) != 0 ||
+            *cursor > SIZE_MAX - (alignment - 1)) {
+            return false;
+        }
+        const size_t aligned = (*cursor + alignment - 1) & ~(alignment - 1);
+        if (bytes > SIZE_MAX - aligned) return false;
+        *offset = aligned;
+        *cursor = aligned + bytes;
+        return true;
+    };
+
+    // Compute and validate the complete layout before asking the arena bank to
+    // allocate. Definitions lead the block; fixed-size submissions follow and
+    // refer to their Definition by device_base + offset.
+    GraphHostDefinitionList definitions = graph_host_definitions(graph_state);
+    std::vector<PackedDefinition> packed_definitions;
+    packed_definitions.reserve(definitions.entries.size());
+    std::unordered_map<uint64_t, size_t> definition_indices;
+    size_t block_bytes = 0;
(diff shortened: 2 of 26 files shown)


ChaoWao (commented):
The GraphBlock half of this looks right to me, and it goes further than the batching
I had staged locally — one grow-only block on the arena bank with `base + offset`
addressing and reuse-when-large-enough is the shape the "one `rtMalloc`, one
`rtMemcpy`" constraint asks for, where mine only collapsed the per-submission
allocations into a single fresh one per run. I'll drop mine.

One problem in the other half, though: **`graph_submit_outer` writes the boundary
args into `payload.tensors` / `payload.scalars` with no bound check, and those arrays
are four times smaller than what `GraphTaskArgs` can legally hold.**

```cpp
// pto_orchestrator.cpp, this PR
payload.tensor_count = args.tensor_count();
for (int32_t i = 0; i < args.tensor_count(); ++i)
    payload.tensors[i] = args.tensor(i).ref();
std::memcpy(payload.scalars, args.scalar_data(), args.scalar_count() * sizeof(uint64_t));
```

| | capacity | where |
| --- | --- | --- |
| `GraphTaskArgs` (the source) | **128** tensors / **64** scalars | `pto_types.h:689-690` |
| `PTO2TaskPayload.tensors` / `.scalars` (the destination) | **32** / **16** | `CORE_MAX_TENSOR_ARGS` / `CORE_MAX_SCALAR_ARGS` |

The only cap on the path in between is a `debug_assert(args.tensor_count() <=
GRAPH_MAX_TENSOR_ARGS)` in `rt_submit_graph_impl`, i.e. 128.

`examples/a2a3/host_build_graph/deepseek_v4_flash_decode` exceeds both. Measured on
a2a3 by logging the counts at the write site and running the case's host path only
(`SIMPLER_SKIP_DEVICE_RUN=1`), on this PR's head:

```
[PROBE] graph boundary: tensors=118 (payload holds 32)  scalars=31 (payload holds 16)
[PROBE] would write tensors[0..117] into a 32-slot array: bytes [640, 15744)
        against sizeof(payload)=4864, i.e. 10880 bytes past the payload
        (2.24 following payloads at stride 4864)
```

40 occurrences per run — 2 ranks x 20 submissions. `tensors` starts at offset 640 and
`scalars` at 4736, so `tensors[32..117]` runs straight over the scalar region and then
10,880 bytes past the end of a 4,864-byte payload, into the following entries of the
payload array. The counts come from the case's own orchestration
(`decode_fwd_graph.cpp:8788-8919`: 113 `add_input` + 5 `add_inout`, and 31 scalars
across 8 `add_scalar` calls), so they are static, not workload-dependent. My probe
returned early instead of performing the write, so nothing above is an observed
corruption — it is the extent of the write the code would make.

`main` also states the invariant this breaks, in the comment on the type being read
here:

> Boundary arguments of a Graph. Sized independently of CoreTaskArgs because **a Graph
> boundary never becomes a task payload** — `graph_reset_outer_payload` zeroes the outer
> GRAPH task's `tensor_count`, and the boundary is read host-side only […] Widening it
> costs orchestration stack, **not `PTO2TaskPayload`** […]
> — `pto_types.h:692-696`

That paragraph would need to change with this PR either way; right now the tree would
carry a comment that contradicts its own code, and `GRAPH_MAX_TENSOR_ARGS = 128`
becomes a number no caller can actually use.

Four things in fairness, because I do not think this is a regression of a green test:

- **dsv4 does not pass on `main` either.** Its device path fails with
  `sched_error_code=5 INVALID_ARGS` ("an orchestration API rejected its arguments").
  So the change does not turn a passing case red; it turns an already-failing case
  from a clean rejection into an out-of-bounds host write whose bytes then ship H2D.
- **It is not silent-wrong-results.** `graph_execution_localize` does gain
  `tensor_count > MAX_TENSOR_ARGS → return nullptr`, so the replay is refused. But
  that check runs on device, after the host has already written past the array, so the
  reported failure point and the actual fault are one step apart.
- **The validation set cannot see it.** qwen3-14b's boundary is 26 tensors / 0 scalars
  and fits in 32/16, and `record_then_replay_1d/2d` are smaller still. dsv4 is
  `manual: True`, so the default `pytest examples tests/st` sweep excludes it — CI will
  stay green on this indefinitely.
- **The benefit and the bug share a root.** The 232,312 -> 131,832 B (-43%) saving comes
  from deleting the duplicate variable-size arg image, and that is exactly the
  mechanism a >32-arg boundary needs. So it is a real design tension, not an oversight
  to patch with a clamp: bounding the boundary at `MAX_TENSOR_ARGS` would cap every
  Graph at 32 boundary args and exclude dsv4 outright.

What I have locally for this is a pooled `GraphBoundaryImage` — a POD
`{total_bytes, tensor_count, scalar_count, off_tensors, off_scalars}` header in a
host-written slot of a boundary pool, addressed `base + index x stride` with the index
being the outer task's submission order, bounded by `GRAPH_MAX_*` rather than
`MAX_TENSOR_ARGS`. It keeps the args out of `GraphSubmission` (so the fixed-size
submission object this PR introduces stays fixed-size) without putting them in a
32-slot array, and the pool's base and stride can live in the same GraphBlock this PR
adds. Happy to rebase it onto this and hand it over, or to just supply the bound check
and a failing test if you would rather keep the scope here as-is.

Reproduction, if useful:

```bash
# log args.tensor_count()/scalar_count() at the write site in graph_submit_outer, then:
SIMPLER_SKIP_DEVICE_RUN=1 SIMPLER_LOG_LEVEL=TIMING task-submit --device auto --device-num 2 --run \
  "python examples/a2a3/host_build_graph/deepseek_v4_flash_decode/test_deepseek_v4_flash_decode.py \
     -p a2a3 --manual only -d \$TASK_DEVICE \
     --case TestDeepseekV4FlashDecodeHostBuildGraph:: --runtime host_build_graph --level 3"
```

Note that `SIMPLER_SKIP_DEVICE_RUN` is presence-based, and that a `skip_golden` case
reports PASSED under it having executed nothing — so read the probe output, not the
exit status.

PR #1925, opened 2026-08-20
Fix: support existing 64x256x64 paged attention Case3

doraemonmj:
## Summary

- Support the paged-attention `Case3` tuple `(q_tile, head_dim, block_size) = (64, 256, 64)` in the 11 affected TMR variants and the standalone A5 host-build-graph kernel copies.
- Restore `Case3` in the three A2/A3 host-build-graph wrappers that share the corrected TMR kernels.
- Preserve the existing small-shape and 128-wide dispatch paths, including the batched two-dimensional query layout.
- Keep the task ABI unchanged.

## Scope

These scene kernels dispatch the shapes exercised by their cases; they are not a general paged-attention shape dispatcher. This PR does not claim support for every `head_dim=256` combination. In particular, no current case uses `q_tile=16, head_dim=256`, and the existing fallback behavior for unsupported shapes is unchanged.

## Root cause

The `Case3` workloads use `head_dim=256`, while the affected kernels dispatched the QK, PV, and online-update operations to templates whose head-dimension axis was fixed at 128. As a result, QK reduced only the first 128 elements, and PV/online-update produced or normalized only the first 128 output elements.

The A5 `host_build_graph/paged_attention` scene carries independent kernel copies rather than referencing `TMR_CASE`, so those copies require the same dispatch fix for its existing `Case3`.

Three A2/A3 host-build-graph wrappers omitted `Case3` when they were introduced because these shared kernels failed golden at 256 width. They now reuse the corrected kernels, so the matching case definitions are restored.

## Fix

Select the existing small/128-wide templates or the new 256-wide template instantiations from the runtime tensor shapes:

- QK uses the query tensor's head dimension for the reduction width.
- PV and online-update use the output tensor's head dimension for the output width.
- Other dispatch behavior and the kernel argument layout remain unchanged.

The restored batched HBG `Case3` uses a case-local 1 GiB graph heap. Its 256-wide graph exhausts the 512 MiB setting used by the existing 128-wide cases.

## Case3 resource budget

- QK L0A/L0B: 32 + 32 KiB.
- PV L0C `AccTile`: 64 KiB.
- Online-update UB: approximately 130 KiB.

These budgets are specific to the supported 256-wide Case3 and are one reason this change should not be read as allowing arbitrary wider shapes.

## Case3 validation

| Platform | Area | Variant | Case | Result |
| A2/A3 | example | `paged_attention` | existing `Case3` | Passed on A2 onboard |
| A2/A3 | example | `paged_attention_manual_scope` | existing `Case3` | Passed on A2 onboard |
| A2/A3 | example | `paged_attention_unroll_manual_scope` | existing `Case3` | Passed on A2 onboard |
| A2/A3 | ST | `batch_paged_attention` | existing `Case3` | Passed on A2 onboard |
| A2/A3 | ST | `paged_attention_unroll` | existing `Case3` | Passed on A2 onboard |
| A2/A3 | ST | `paged_attention_unroll_4dims` | existing `Case3` | Passed on A2 onboard |
| A2/A3 | HBG example | `paged_attention_unroll_manual_scope` | restored `Case3` | Passed on A2 onboard |
| A2/A3 | HBG ST | `batch_paged_attention` | restored `Case3`, 1 GiB heap | Passed on A2 onboard |
| A2/A3 | HBG ST | `paged_attention_unroll` | restored `Case3` | Passed on A2 onboard |
| A5 | example | `paged_attention` | existing `Case3` | Passed on A5 onboard |
| A5 | example | `paged_attention_unroll_manual_scope` | existing `Case3` | Passed on A5 onboard |
| A5 | ST | `batch_paged_attention` | existing `Case3` | Passed on A5 onboard |
| A5 | ST | `paged_attention_unroll` | existing `Case3` | Passed on A5 onboard |
| A5 | ST | `paged_attention_unroll_4dims` | existing `Case3` | Passed on A5 onboard |
| A5 | HBG ST | `paged_attention` | existing `Case3` | Requires A5 onboard; A5sim `SmallCase1` compile/smoke passed |

Per-PR CI excludes these manual onboard-only `Case3` cases, so the green CI checks the existing small/128-wide paths rather than the 256-wide runtime result.

## Additional checks

- All pre-commit hooks passed after rebasing onto the latest `main`.
- All three restored A2/A3 HBG `Case3` workloads passed standalone golden validation inside an allocated `task-submit` job.

Part of #1832.

### DIFF
--- a/examples/a2a3/host_build_graph/paged_attention_unroll_manual_scope/test_paged_attention_unroll_manual_scope.py
+++ b/examples/a2a3/host_build_graph/paged_attention_unroll_manual_scope/test_paged_attention_unroll_manual_scope.py
@@ -62,8 +62,6 @@ class TestPagedAttentionUnrollManualScopeHostBuildGraph(SceneTestCase):
         ],
     }
 
-    # The shared head_dim=256 kernels fail golden on both runtimes, so Case3
-    # is not a valid cross-runtime benchmark.
     CASES = [
         {
             "name": "Case1",
@@ -88,6 +86,21 @@ class TestPagedAttentionUnrollManualScopeHostBuildGraph(SceneTestCase):
                 "num_heads": 64,
                 "kv_head_num": 1,
                 "head_dim": 128,
+                "block_size": 64,
+                "context_len": 8192,
+                "max_model_len": 32768,
+                "dtype": "bfloat16",
+            },
+        },
+        {
+            "name": "Case3",
+            "platforms": ["a2a3"],
+            "manual": True,
+            "params": {
+                "batch": 64,
+                "num_heads": 64,
+                "kv_head_num": 1,
+                "head_dim": 256,
                 "block_size": 64,
                 "context_len": 8192,
                 "max_model_len": 32768,
--- a/examples/a2a3/tensormap_and_ringbuffer/paged_attention/kernels/aic/aic_pv_matmul.cpp
+++ b/examples/a2a3/tensormap_and_ringbuffer/paged_attention/kernels/aic/aic_pv_matmul.cpp
@@ -10,9 +10,10 @@
  */
 // PV Matmul Kernel: pij(M, K) @ vj(K, N) -> oi_new(M, N)
 //
-// Supports two tile configurations via runtime dispatch:
+// Supports three tile configurations via runtime dispatch:
 //   Case1: (16, 128) @ (128, 128) -> (16, 128)
 //   Case2: (64,  64) @ ( 64, 128) -> (64, 128)
+//   Case3: (64,  64) @ ( 64, 256) -> (64, 256)
 //
 // pij is bfloat16 (converted from fp32 in softmax_prepare via TCVT).
 // vj is stored as (K, N) = (block_size, head_dim) in row-major (ND) layout.
@@ -103,11 +104,14 @@ extern "C" __aicore__ void kernel_entry(__gm__ int64_t *args) {
     __gm__ ChipTensor *vj = reinterpret_cast<__gm__ ChipTensor *>(args[1]);
     __gm__ ChipTensor *oi_new = reinterpret_cast<__gm__ ChipTensor *>(args[2]);
     uint64_t q_tile_size = static_cast<uint64_t>(pij->shapes[0]);
+    uint64_t head_dim = static_cast<uint64_t>(oi_new->shapes[1]);
 
     if (q_tile_size == 16 && pij->shapes[1] <= 16) {
         pv_matmul_impl<16, 16, 16>(pij, vj, oi_new);
     } else if (q_tile_size == 16) {
         pv_matmul_impl<16, 128, 128>(pij, vj, oi_new);
+    } else if (head_dim == 256) {
+        pv_matmul_impl<64, 64, 256>(pij, vj, oi_new);
     } else {
         pv_matmul_impl<64, 64, 128>(pij, vj, oi_new);
     }
--- a/examples/a2a3/tensormap_and_ringbuffer/paged_attention/kernels/aic/aic_qk_matmul.cpp
+++ b/examples/a2a3/tensormap_and_ringbuffer/paged_attention/kernels/aic/aic_qk_matmul.cpp
@@ -10,9 +10,10 @@
  */
 // QK Matmul Kernel: qi(M, K) @ kj.T(K, N) -> sij(M, N)
 //
-// Supports two tile configurations via runtime dispatch:
+// Supports three tile configurations via runtime dispatch:
 //   Case1: (16, 128) @ (128, 128).T -> (16, 128)
 //   Case2: (64, 128) @ (128,  64).T -> (64,  64)
+//   Case3: (64, 256) @ (256,  64).T -> (64,  64)
 //
 // kj is stored as (N, K) = (block_size, head_dim) in row-major memory.
 // This is equivalent to (K, N) in column-major (DN) layout.
@@ -104,11 +105,14 @@ extern "C" __aicore__ void kernel_entry(__gm__ int64_t *args) {
     __gm__ ChipTensor *kj = reinterpret_cast<__gm__ ChipTensor *>(args[1]);
     __gm__ ChipTensor *sij = reinterpret_cast<__gm__ ChipTensor *>(args[2]);
     uint64_t q_tile_size = static_cast<uint64_t>(qi->shapes[0]);
+    uint64_t head_dim = static_cast<uint64_t>(qi->shapes[1]);
 
-    if (q_tile_size == 16 && qi->shapes[1] <= 16) {
+    if (q_tile_size == 16 && head_dim <= 16) {
         qk_matmul_impl<16, 16, 16>(qi, kj, sij);
     } else if (q_tile_size == 16) {
         qk_matmul_impl<16, 128, 128>(qi, kj, sij);
+    } else if (head_dim == 256) {
+        qk_matmul_impl<64, 256, 64>(qi, kj, sij);
     } else {
         qk_matmul_impl<64, 128, 64>(qi, kj, sij);
     }
--- a/examples/a2a3/tensormap_and_ringbuffer/paged_attention_manual_scope/kernels/aic/aic_pv_matmul.cpp
+++ b/examples/a2a3/tensormap_and_ringbuffer/paged_attention_manual_scope/kernels/aic/aic_pv_matmul.cpp
@@ -10,9 +10,10 @@
  */
 // PV Matmul Kernel: pij(M, K) @ vj(K, N) -> oi_new(M, N)
 //
-// Supports two tile configurations via runtime dispatch:
+// Supports three tile configurations via runtime dispatch:
 //   Case1: (16, 128) @ (128, 128) -> (16, 128)
 //   Case2: (64,  64) @ ( 64, 128) -> (64, 128)
+//   Case3: (64,  64) @ ( 64, 256) -> (64, 256)
 //
 // pij is bfloat16 (converted from fp32 in softmax_prepare via TCVT).
 // vj is stored as (K, N) = (block_size, head_dim) in row-major (ND) layout.
@@ -103,11 +104,14 @@ extern "C" __aicore__ void kernel_entry(__gm__ int64_t *args) {
     __gm__ ChipTensor *vj = reinterpret_cast<__gm__ ChipTensor *>(args[1]);
     __gm__ ChipTensor *oi_new = reinterpret_cast<__gm__ ChipTensor *>(args[2]);
     uint64_t q_tile_size = static_cast<uint64_t>(pij->shapes[0]);
+    uint64_t head_dim = static_cast<uint64_t>(oi_new->shapes[1]);
 
     if (q_tile_size == 16 && pij->shapes[1] <= 16) {
         pv_matmul_impl<16, 16, 16>(pij, vj, oi_new);
     } else if (q_tile_size == 16) {
         pv_matmul_impl<16, 128, 128>(pij, vj, oi_new);
+    } else if (head_dim == 256) {
+        pv_matmul_impl<64, 64, 256>(pij, vj, oi_new);
     } else {
         pv_matmul_impl<64, 64, 128>(pij, vj, oi_new);
     }
--- a/examples/a5/tensormap_and_ringbuffer/paged_attention/kernels/aic/aic_qk_matmul.cpp
+++ b/examples/a5/tensormap_and_ringbuffer/paged_attention/kernels/aic/aic_qk_matmul.cpp
@@ -10,9 +10,10 @@
  */
 // QK Matmul Kernel: qi(M, K) @ kj.T(K, N) -> sij(M, N)
 //
-// Supports two tile configurations via runtime dispatch:
+// Supports three tile configurations via runtime dispatch:
 //   Case1: (16, 128) @ (128, 128).T -> (16, 128)
 //   Case2: (64, 128) @ (128,  64).T -> (64,  64)
+//   Case3: (64, 256) @ (256,  64).T -> (64,  64)
 //
 // kj is stored as (N, K) = (block_size, head_dim) in row-major memory.
 // This is equivalent to (K, N) in column-major (DN) layout.
@@ -104,12 +105,14 @@ extern "C" __aicore__ void kernel_entry(__gm__ int64_t *args) {
     __gm__ ChipTensor *kj = reinterpret_cast<__gm__ ChipTensor *>(args[1]);
     __gm__ ChipTensor *sij = reinterpret_cast<__gm__ ChipTensor *>(args[2]);
     uint64_t q_tile_size = static_cast<uint64_t>(qi->shapes[0]);
-    // args[4] = head_dim (128), args[5] = block_size
+    uint64_t head_dim = static_cast<uint64_t>(qi->shapes[1]);
 
-    if (q_tile_size == 16 && qi->shapes[1] <= 16) {
+    if (q_tile_size == 16 && head_dim <= 16) {
         qk_matmul_impl<16, 16, 16>(qi, kj, sij);
     } else if (q_tile_size == 16) {
         qk_matmul_impl<16, 128, 128>(qi, kj, sij);
+    } else if (head_dim == 256) {
+        qk_matmul_impl<64, 256, 64>(qi, kj, sij);
     } else {
         qk_matmul_impl<64, 128, 64>(qi, kj, sij);
     }
--- a/tests/st/a2a3/tensormap_and_ringbuffer/paged_attention_unroll/kernels/aiv/aiv_online_update.cpp
+++ b/tests/st/a2a3/tensormap_and_ringbuffer/paged_attention_unroll/kernels/aiv/aiv_online_update.cpp
@@ -10,9 +10,10 @@
  */
 // Online Softmax Update + Normalize Kernel (AIV)
 //
-// Operates on full tiles where M=q_tile_size, N=head_dim (128):
+// Operates on full tiles where M=q_tile_size and N=head_dim:
 //   Case1: oi/oi_new are (16, 128), mij/lij/mi/li are 16-element vectors
 //   Case2: oi/oi_new are (64, 128), mij/lij/mi/li are 64-element vectors
+//   Case3: oi/oi_new are (64, 256), mij/lij/mi/li are 64-element vectors
 //
 // Scalar layout strategy using TRESHAPE (zero-copy UB reshape):
 //   Scalars loaded as DN ColMajor (M, 1) for TROWEXPANDMUL/TROWEXPANDDIV.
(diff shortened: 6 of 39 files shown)


examples/a2a3/tensormap_and_ringbuffer/paged_attention/kernels/aic/aic_qk_matmul.cpp
```diff
@@ -104,11 +105,14 @@ extern "C" __aicore__ void kernel_entry(__gm__ int64_t *args) {
     __gm__ ChipTensor *kj = reinterpret_cast<__gm__ ChipTensor *>(args[1]);
     __gm__ ChipTensor *sij = reinterpret_cast<__gm__ ChipTensor *>(args[2]);
     uint64_t q_tile_size = static_cast<uint64_t>(qi->shapes[0]);
+    uint64_t head_dim = static_cast<uint64_t>(qi->shapes[1]);
 
-    if (q_tile_size == 16 && qi->shapes[1] <= 16) {
+    if (q_tile_size == 16 && head_dim <= 16) {
         qk_matmul_impl<16, 16, 16>(qi, kj, sij);
     } else if (q_tile_size == 16) {
```
ChaoZheng109, line 112: **The `q_tile_size == 16` arm still shadows the `head_dim` check.**

```cpp
if (q_tile_size == 16 && head_dim <= 16) { ...<16,16,16> }
else if (q_tile_size == 16)              { ...<16,128,128> }   // <- head_dim never consulted
else if (head_dim == 256)                { ...<64,256,64> }
else                                     { ...<64,128,64> }
```

A future `head_dim=256` workload with `num_heads=16` (hence `q_tile = min(num_heads, 128) = 16`) takes this arm and reproduces *exactly the bug this PR fixes*, just in the 16-row corner. No current case hits it, so it isn't a live defect — but the hazard is preserved in all 33 files. Consider keying this arm on `head_dim` as well.

examples/a2a3/tensormap_and_ringbuffer/paged_attention/kernels/aic/aic_qk_matmul.cpp
```diff
@@ -104,11 +105,14 @@ extern "C" __aicore__ void kernel_entry(__gm__ int64_t *args) {
...
         qk_matmul_impl<16, 128, 128>(qi, kj, sij);
+    } else if (head_dim == 256) {
+        qk_matmul_impl<64, 256, 64>(qi, kj, sij);
     } else {
         qk_matmul_impl<64, 128, 64>(qi, kj, sij);
```
ChaoZheng109, line 117: **No unsupported-shape trap — this is the actual root cause.**

This `else` silently *assumes* `head_dim == 128` and `block_size == 64` (the `N` in `<64, 128, 64>`). A `head_dim=192/512` workload, or a `head_dim=256, block_size=128` one, would again compute a wrong answer with no diagnostic — the same silent-wrong-answer failure mode that produced #1832, whose only symptom was `max_diff ~= 1.0` in a Daily run months later.

Adding a third arm makes today's cases correct without changing that property. A `TRAP`/assert on the fall-through, or a small explicit `(q_tile, head_dim, block_size)` table, would convert a silent wrong answer into a loud failure. Out of scope for this PR, but worth a follow-up issue.

examples/a2a3/tensormap_and_ringbuffer/paged_attention/kernels/aic/aic_pv_matmul.cpp
```diff
@@ -10,9 +10,10 @@
  */
 // PV Matmul Kernel: pij(M, K) @ vj(K, N) -> oi_new(M, N)
 //
-// Supports two tile configurations via runtime dispatch:
+// Supports three tile configurations via runtime dispatch:
 //   Case1: (16, 128) @ (128, 128) -> (16, 128)
 //   Case2: (64,  64) @ ( 64, 128) -> (64, 128)
+//   Case3: (64,  64) @ ( 64, 256) -> (64, 256)
```
ChaoZheng109, line 16: **The buffer budget is the load-bearing constraint for this Case3 and is nowhere recorded.**

`head_dim=256` works because the numbers happen to fit:

| Kernel | Tile | Case2 (128) | Case3 (256) |
| :-- | :-- | --: | --: |
| QK | L0A / L0B | 16 + 16 KiB | 32 + 32 KiB |
| PV | L0C `AccTile` (M*N fp32) | 32 KiB | **64 KiB** |
| online-update | UB (`2 * kDataBytes` + 9 scalar DN tiles) | ~66 KiB | **~130 KiB** |

`head_dim=512` would put the `AccTile` at 128 KiB and the online-update UB at ~258 KiB, so it is *not* a free extension of the same pattern — even though the online-update `TASSIGN` layout is parametric in `kDataBytes` and would appear to scale on its own.

Suggest recording those two numbers in the PR body (or in this header comment next to the Case3 line). The next person adding an arm needs them and would otherwise re-derive them from a chip hang.

examples/a5/tensormap_and_ringbuffer/paged_attention/kernels/aic/aic_qk_matmul.cpp
```diff
@@ -104,12 +105,14 @@ extern "C" __aicore__ void kernel_entry(__gm__ int64_t *args) {
     __gm__ ChipTensor *kj = reinterpret_cast<__gm__ ChipTensor *>(args[1]);
     __gm__ ChipTensor *sij = reinterpret_cast<__gm__ ChipTensor *>(args[2]);
     uint64_t q_tile_size = static_cast<uint64_t>(qi->shapes[0]);
-    // args[4] = head_dim (128), args[5] = block_size
+    uint64_t head_dim = static_cast<uint64_t>(qi->shapes[1]);
```
ChaoZheng109, line 108: Minor: this line replaces `// args[4] = head_dim (128), args[5] = block_size`. Dropping it is right under `.claude/rules/comments.md` — it was stale about the value *and* the kernel no longer reads those slots.

But the orchestration still **passes** `args[4]`/`args[5]`, so those ABI slots now have no documentation at the consumer end. If they are genuinely dead, removing them from the orchestration side too (separate PR — this one keeps the ABI unchanged by design) would be the cleaner end state.

ChaoZheng109 (commented):
Reviewed the full 33-file diff against merge-base `93adc386`. **The core fix is correct and applied with real uniformity** — I checked every file individually and all 33 read `head_dim` from the correct tensor and the correct axis (`qi->shapes[1]` for QK, `oi_new->shapes[1]` for PV/online-update, `shapes[3]` in the 4-dims variants) and pass the correct template arguments. Buffer budgets clear on both archs (QK L0A/L0B 32+32 KiB, PV L0C `AccTile` 64 KiB, online-update UB ~130 KiB — the UB layout is parametric in `kDataBytes`, which is why no `TASSIGN` offset needed editing). Scoping to the 11 TMR variants is honest and matches #1832's failure list exactly.

Inline comments cover four points on the dispatch itself. Two more findings are on files **outside this diff**, so they cannot be anchored inline:

---

### 1. `tests/st/a5/host_build_graph/paged_attention/` still breaks its live `head_dim=256` `Case3` — the a5 half of #1832 is not met

That directory does **not** reference `TMR_CASE`; it carries its own three kernel copies, and they still read:

```cpp
// tests/st/a5/host_build_graph/paged_attention/kernels/aic/aic_qk_matmul.cpp:106-114
uint64_t q_tile_size = static_cast<uint64_t>(qi->shapes[0]);
// args[4] = head_dim (128), args[5] = block_size
...
} else {
    qk_matmul_impl<64, 128, 64>(qi, kj, sij);   // Case3 lands here
}
```

Meanwhile `tests/st/a5/host_build_graph/paged_attention/test_paged_attention.py:106-119` defines:

```python
{
    "name": "Case3",
    "platforms": ["a5"],
    "manual": True,
    "params": {"batch": 64, "num_heads": 64, "kv_head_num": 1, "head_dim": 256, ...},
}
```

`num_heads=64` gives `q_tile = min(num_heads, 128) = 64`, so it takes exactly the broken arm. It likely did not show up as a 12th golden mismatch in run #31722222789 only because `Case1` in the same class dies first on `Task Ring Full`, and the class loops through `CASES` under a single `test_run` node — the reporting shape #1832 itself calls out.

Since #1832's acceptance criterion says "a2a3 **and** a5", either extend the fix to those three files (identical three-line change) or state the deferral explicitly in the PR body. It shouldn't land silently.

For the record, a5 HBG `batch_paged_attention` and `paged_attention_unroll` *do* reference `TMR_CASE`, so they inherit the fix for free. The gap is `paged_attention` only.

### 2. Three files now carry a comment this PR makes false

```python
# The shared head_dim=256 kernels fail golden on both runtimes, so Case3
# is not a valid cross-runtime benchmark.
```

- `tests/st/a2a3/host_build_graph/batch_paged_attention/test_batch_paged_attention.py:65`
- `tests/st/a2a3/host_build_graph/paged_attention_unroll/test_paged_attention_unroll.py:65`
- `examples/a2a3/host_build_graph/paged_attention_unroll_manual_scope/test_paged_attention_unroll_manual_scope.py:65`

(introduced by #1786, `0194fde7`). All three point at `TMR_CASE` kernels that this PR fixes, so the stated reason for dropping `Case3` no longer holds. `.claude/rules/doc-consistency.md` §1/§3 puts that fix in the same commit. Better still, reinstate the `Case3` entries — re-enabling that cross-runtime benchmark is the payoff of this fix.

---

### One note on the green CI

All 15 checks pass, but **none of them executes a `head_dim=256` case**: per-PR CI runs `manual_mode=exclude`, and `Case3` is `"manual": True` *and* onboard-only (no sim platform in its `platforms` list). The green tick proves the 16/128 paths did not regress; the fix itself is attested only by the manual onboard runs in your table. Worth saying so explicitly in the PR body so a later reader doesn't misread it.

examples/a2a3/tensormap_and_ringbuffer/paged_attention/kernels/aic/aic_qk_matmul.cpp
doraemonmj: Thanks. I agree this combination would not be handled by the current arm ordering. No existing scene case uses `(q_tile, head_dim) = (16, 256)`, and these kernels are case-shaped rather than a general shape dispatcher. I narrowed the PR title and description to the existing `(64, 256, 64)` Case3 and am intentionally leaving this future combination out of scope.

examples/a2a3/tensormap_and_ringbuffer/paged_attention/kernels/aic/aic_qk_matmul.cpp
doraemonmj: Agreed that the fallback is not a general unsupported-shape guard. This PR is limited to the existing scene cases and keeps the existing dispatch behavior unchanged; the PR description now states that explicitly. We do not plan a follow-up generalization unless a supported case requires a new shape.

examples/a2a3/tensormap_and_ringbuffer/paged_attention/kernels/aic/aic_pv_matmul.cpp
doraemonmj: Added the Case3 resource budgets to the PR description: QK L0A/L0B 32 + 32 KiB, PV L0C `AccTile` 64 KiB, and online-update UB approximately 130 KiB. The scope section also makes clear that these numbers do not imply support for arbitrary wider shapes.

examples/a5/tensormap_and_ringbuffer/paged_attention/kernels/aic/aic_qk_matmul.cpp
doraemonmj: Thanks. I checked the current orchestration: `params_qk` contains only the `qi` input, `kj` input, and `sij` output; it does not pass scalar `head_dim` or `block_size` slots. The `args[4]`/`args[5]` comment was therefore legacy text rather than documentation of live ABI slots. The stale comment remains removed, and no orchestration ABI change is needed.

tests/st/a2a3/tensormap_and_ringbuffer/paged_attention_unroll_4dims/kernels/aic/aic_qk_matmul.cpp
```diff
@@ -126,9 +127,12 @@ extern "C" __aicore__ void kernel_entry(__gm__ int64_t *args) {
 
     // qi is a 4D view (batch, q_len, num_heads_tile, head_dim); decode fixes batch=q_len=1.
     uint64_t q_tile_size = static_cast<uint64_t>(qi->shapes[2]);
+    uint64_t head_dim = static_cast<uint64_t>(qi->shapes[3]);
 
     if (q_tile_size == 16) {
         qk_matmul_n_impl<16, 128, 128>(qi, key_cache, block_table_t, sij_buf, n_blocks, bt_offset);
+    } else if (head_dim == 256) {
+        qk_matmul_n_impl<64, 256, 64>(qi, key_cache, block_table_t, sij_buf, n_blocks, bt_offset);
```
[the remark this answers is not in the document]
doraemonmj: Thanks. The (16, 256) QK tuple is not used by any existing case. The PR title and scope are limited to the existing (64, 256, 64) Case3, so coordinated support or rejection for this future shape is intentionally out of scope.

tests/st/a2a3/tensormap_and_ringbuffer/paged_attention_unroll/kernels/aiv/aiv_online_update.cpp
```diff
@@ -246,10 +247,12 @@ extern "C" __aicore__ void kernel_entry(__gm__ int64_t *args) {
     uint64_t is_first = static_cast<uint64_t>(args[7]);
     uint64_t is_last = static_cast<uint64_t>(args[8]);
     uint64_t q_tile_size = static_cast<uint64_t>(mij->shapes[0]);
-    // args[10] = head_dim (128)
+    uint64_t head_dim = static_cast<uint64_t>(oi_new->shapes[1]);
 
     if (q_tile_size == 16) {
         online_update_impl<16, 128>(mij, lij, oi_new, mi, li, oi, is_first, is_last, dst);
+    } else if (head_dim == 256) {
+        online_update_impl<64, 256>(mij, lij, oi_new, mi, li, oi, is_first, is_last, dst);
```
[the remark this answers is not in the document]
doraemonmj: Thanks. No existing scene case reaches the (16, 256) online-update path. This PR supports only the existing (64, 256, 64) Case3 and intentionally preserves dispatch behavior for other unsupported tuples.

examples/a2a3/tensormap_and_ringbuffer/paged_attention_manual_scope/kernels/aic/aic_pv_matmul.cpp
```diff
@@ -103,11 +104,14 @@ extern "C" __aicore__ void kernel_entry(__gm__ int64_t *args) {
...
     __gm__ ChipTensor *oi_new = reinterpret_cast<__gm__ ChipTensor *>(args[2]);
     uint64_t q_tile_size = static_cast<uint64_t>(pij->shapes[0]);
+    uint64_t head_dim = static_cast<uint64_t>(oi_new->shapes[1]);
 
     if (q_tile_size == 16 && pij->shapes[1] <= 16) {
         pv_matmul_impl<16, 16, 16>(pij, vj, oi_new);
     } else if (q_tile_size == 16) {
         pv_matmul_impl<16, 128, 128>(pij, vj, oi_new);
+    } else if (head_dim == 256) {
+        pv_matmul_impl<64, 64, 256>(pij, vj, oi_new);
```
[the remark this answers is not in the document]
doraemonmj: Thanks. This is the same unsupported future tuple discussed above. No existing case uses (q_tile_size, head_dim) = (16, 256); this PR is explicitly scoped to the existing (64, 256, 64) Case3 and intentionally leaves other scene shapes unchanged.

examples/a2a3/host_build_graph/paged_attention_unroll_manual_scope/test_paged_attention_unroll_manual_scope.py
```diff
@@ -62,8 +62,6 @@ class TestPagedAttentionUnrollManualScopeHostBuildGraph(SceneTestCase):
         ],
     }
 
-    # The shared head_dim=256 kernels fail golden on both runtimes, so Case3
-    # is not a valid cross-runtime benchmark.
     CASES = [
```
ChaoZheng109, line 65: **Removing this comment leaves `Case3`'s absence unexplained — and the blocker it cited is exactly what this PR removes.**

The deleted lines were:

```python
# The shared head_dim=256 kernels fail golden on both runtimes, so Case3
# is not a valid cross-runtime benchmark.
```

I checked the wiring: all three of these HBG scenes resolve their kernels through `TMR_CASE`, e.g.

```python
"source": f"{TMR_CASE}/kernels/aic/aic_qk_matmul.cpp",
```

— i.e. the very files this PR fixes. So deleting the comment is correct as far as it goes: the sentence is no longer true, and keeping a false comment is worse than keeping none (`.claude/rules/comments.md`).

But the *state* it documented has not changed. None of the three scenes defines a `Case3`, and after this PR nothing records why. A reader who opens `test_paged_attention_unroll.py` next month cannot distinguish "`Case3` was deliberately dropped here" from "`Case3` was never added here" — the only remaining evidence is this PR's diff. That is the case `.claude/rules/doc-consistency.md` §3 is about: the text goes when its referent is gone, and here the referent (the deliberately-absent `Case3`) is still present.

Affects — identical hunk, same anchor line in each:

- `examples/a2a3/host_build_graph/paged_attention_unroll_manual_scope/test_paged_attention_unroll_manual_scope.py#L65` (this comment)
- `tests/st/a2a3/host_build_graph/batch_paged_attention/test_batch_paged_attention.py#L65`
- `tests/st/a2a3/host_build_graph/paged_attention_unroll/test_paged_attention_unroll.py#L65`

Two ways to close it; either is fine:

1. **Restore `Case3` in the three scenes.** The stated blocker is gone, and the kernels these scenes run are the ones you validated onboard — this is what the deleted comment implies should happen once the kernels are fixed.
2. **Keep the scope as declared and replace the comment with the reason that holds now** — e.g. that the HBG variants are intentionally held to the 128-wide cross-runtime benchmark set — so the omission stays documented.

Given the PR explicitly states "Do not add or change any test-case definition or parameter", (2) is the consistent choice here, with (1) as a follow-up. Deleting the comment without either is the one outcome that loses information.

doraemonmj: Fixed — restored Case3 in all three A2/A3 HBG wrappers. The batched HBG case uses a case-local 1 GiB graph heap because its 256-wide graph exhausts 512 MiB. All three restored Case3 workloads passed onboard golden validation inside task-submit.

ChaoZheng109 (approved):

PR #1973, opened 2026-08-24
Fix: prefer staging views for host tensor access

### LINKED ISSUE #1848: [Performance] hbg: host-view register/unregister of every entry tensor costs ~256 ms/run unregistering 38 GiB the orchestrator never reads
### Platform

a2a3 (Ascend 910B/C hardware)

### Runtime Variant

host_build_graph

### Summary

`init_runtime_impl` installs a host view for **every** entry tensor, then tears them all down after
orchestration. On `qwen3_14b_decode` that is 20 mappings covering **38.05 GiB**, and the teardown
loop alone costs **~256 ms per run** — roughly 5.7x the whole device execution (`device_wall`
~45 ms).

The host orchestrator needs host views only to read small control tensors (`seq_lens`,
`block_table`, `slot_mapping`) to shape the graph. It never reads the multi-GiB weight or KV bytes,
yet those are registered and unregistered every run.

`HostTensorAccessor::close()` does no copying — it is purely
`halHostUnregister` per mapping:

```cpp
void HostTensorAccessor::close() noexcept {
    for (void *dev_ptr : impl_->mappings) {
        impl_->api->unregister_device_memory_from_host(dev_ptr);   // -> halHostUnregister
    }
```

Measured attributes: `count=20 bytes=40860162880`, `dur=255982626` ns. That is 12.8 ms per call, or
159.6 GB/s. Since the 20 tensors are wildly unequal in size (11 weights ~24.6 GiB, 2 KV tensors
13.44 GiB, the rest a few KB), a fixed per-call cost is implausible for a 64-byte buffer — the cost
is almost certainly **O(mapped bytes)**, i.e. page-table / IOMMU teardown. 38 GiB at 4 KiB pages is
~10M pages in 256 ms, or ~25.6 ns/page. I did not separate per-call from per-byte experimentally;
the size-asymmetry argument is the basis for that reading.

Two observations suggest most of this work is avoidable:

1. **A fallback host view is always available on this path.** `runtime_maker.cpp:849` calls
   `tensor_access.add(dev_ptr, size, host_ptr)` with the original host buffer, and `add()` prefers
   `halHostRegister` unconditionally, only falling back if registration fails.
2. **The mapping buys nothing for writes.** `mirrored` is `host_view != dev_base`, so a successful
   registration (which returns a distinct host VA) still sets `mirrored = true`, and
   `HostTensorAccessor::write` does `memcpy` **plus** `copy_to_device` — exactly what the fallback
   path does.

So registering only the tensors the orchestrator may actually read (or preferring the fallback
whenever one exists) should remove both the registration cost and this entire ~256 ms.

One design constraint to respect rather than a blanket equivalence: for a **pure OUT** tensor,
staging is skipped (`bind.args` reports `staged=19` against `count=20`), so its fallback host buffer
does not hold the device bytes. Reads of a not-yet-written OUT tensor are meaningless, but the
mapped and fallback paths are not strictly interchangeable for that one tensor. Separately, a
device-resident (`is_device_memory()`) tensor takes the pass-through branch and never reaches
`add()` at all, so it has no host view — which is why any change here must keep control tensors
host-side.

Related: #1841 (same host-orch H2D phase, different root cause — device-side
`device_malloc` / `acquire_graph_execution_buffer` dominating), #1531 (same
`halHostRegister` API, but a support failure on 64 KiB-page hosts), #1717 (hiding per-run launch
latency behind orchestration rather than reducing it), #1706 (hbg tracking).

### Git Commit ID

288e18035e714e2ff9effd768d24a1b2d553e5ee

### CANN Version

9.0.0 (npu-smi 26.0.rc1)

### Host Platform

Linux (aarch64)

### Reproduction

```bash
# NOTE: the per-phase numbers below come from local host-side STRACE sub-spans added to
# src/{a2a3,a5}/runtime/host_build_graph/host/runtime_maker.cpp. That instrumentation is NOT yet
# upstream, so stock HEAD cannot attribute this phase - it only shows the aggregate
# simpler_run.bind. The six existing millisecond-resolution LOG_INFO("TIMING: ...") lines do not
# cover host_view_close, and are suppressed at the default log level anyway.

.claude/skills/onboard-arch-precheck/check.sh a2a3 || exit 1
  "python -m pytest examples/a2a3/host_build_graph/qwen3_14b_decode \
   --platform a2a3 --device \$TASK_DEVICE --manual include --rounds 4 -s -q"

# On stock HEAD the visible symptom is the aggregate: simpler_run.bind ~1.0 s per round against
# simpler_run.runner_run.device_wall ~45 ms.
```

### Expected Performance

Host-view setup/teardown should be proportional to what the orchestrator actually reads — the few
KB of control tensors — so a sub-millisecond contribution per run, not a fixed ~256 ms tax scaling
with total fixture size.

### Actual Performance

`simpler_run.bind.host_view_close` = **255.98 ms** per run (`count=20`, `bytes=40860162880`),
stable at 250-268 ms across rounds.

For context, the same runs give this steady-state breakdown of `simpler_run.bind` (~1.02-1.27 s
total, fully attributed to within 0.3 ms):

| phase | steady-state | bytes |
| --- | --- | --- |
| `bind.args` (19 tensors staged H2D + 20 registrations) | 744-1020 ms | 40,859,999,040 |
| **`bind.host_view_close`** | **250-268 ms** | 40,860,162,880 (20 mappings) |
| `bind.host_orch` (host graph build, 47 outer tasks) | 0.93-1.49 ms | — |
| `bind.graph_upload` (40 Graph submissions) | 0.78-1.14 ms | 5,310,080 |
| `bind.arena_h2d` | 0.34-0.62 ms | 2,242,048 |
| `bind.runtime_init` | 0.14-0.46 ms | — |
| `bind.sm_h2d` | 0.061-0.079 ms | 233,863 |
| `bind.relocate` | ~0.0005 ms | — |
| `bind.arena_build` | ~0.004 ms | 11,515,648 |
| `bind.static_arena` | ~0.001 ms | — |

So `host_view_close` is the second-largest term in the whole bind, behind only the 38 GiB staging
itself, and both dwarf the actual control-plane work (~2.3-3.2 ms combined).

### Additional Context

The registered byte count (40,860,162,880) exceeds the staged byte count (40,859,999,040) by exactly
163,840 — the pure-OUT tensor, which is registered but correctly not staged.

All 20 registrations succeed on this host, so the `halHostRegister` SVM path is fully active here
(contrast #1531, where it is rejected on 64 KiB-page hosts). A fix should therefore not assume
registration is normally failing.

TaoZQY:
## Summary
- Prefer existing caller buffers when host graph orchestration accesses staged entry tensors, removing per-tensor host registrations from the normal bind path.
- Keep the optional platform-mapping path as an explicit capability escape hatch; no current runtime-maker caller reaches it. Mapping addresses may differ from device addresses, so access always uses the returned host VA.
- Leave pure output tensors unregistered because they have no valid pre-execution bytes and a5 has no device-mapping capability; both attempted reads and writes fail closed.
- Read the shared BGEMM configuration through `get_tensor_data()` so HBG never dereferences a raw device address. This is required by the tensor-access contract, not a drive-by change.
- Keep a2a3 and a5 behavior aligned. a5 onboard already used caller-buffer fallbacks because it has no `halHostRegister` path.

## Performance
- Qwen3-14B HBG reports `host_view_close count=0 bytes=0` across four onboard rounds.
- `host_view_close` is 20.12–24.73 us instead of about 255.98 ms in the issue reproduction.
- The old `args` phase included 20 registrations plus 19 H2D copies; the current path retains the required H2D work but removes all 20 registrations.

## Testing
- [x] Pre-commit hooks passed
- [x] C++ no-hardware suite: 119/119 passed
- [x] Host tensor access regression: 1/1 passed, including distinct returned mapping VA and fallback write-back
- [x] a2a3sim HBG benchmark_bgemm: 1/1 passed
- [x] a5sim HBG benchmark_bgemm: 1/1 passed
- [x] Earlier a2a3 onboard CI regressions: paged_attention + benchmark_bgemm, 2 passed (`task_20260823_234452_297495916244`)
- [x] Earlier a2a3 onboard Qwen3-14B HBG, 4 rounds: 1 passed (`task_20260823_195845_151406924`)

Fixes #1848

### DIFF
--- a/src/a2a3/runtime/host_build_graph/host/host_tensor_access.cpp
+++ b/src/a2a3/runtime/host_build_graph/host/host_tensor_access.cpp
@@ -26,9 +26,9 @@ struct HostTensorRegion {
     uint64_t dev_base;
     uint64_t size;
     unsigned char *host_view;
-    // The host view is a copy of the device bytes rather than the device bytes
-    // themselves, so a write reaches the device only once pushed back.
-    bool mirrored;
+    // A caller-buffer fallback needs a push-back. A platform mapping writes
+    // the device allocation directly, even when its host VA differs.
+    bool needs_push_back;
 };
 
 // One entry per tensor staged for the run being orchestrated. A run stages a
@@ -72,23 +72,23 @@ bool HostTensorAccessor::add(uint64_t dev_base, uint64_t size, void *fallback_ho
     if (impl_->api == nullptr || dev_base == 0 || size == 0) {
         return false;
     }
-    // Both push_backs below run after the registration succeeds, so reserve
-    // first: a reallocation that threw there would leak the mapping.
     impl_->regions.reserve(impl_->regions.size() + 1);
-    impl_->mappings.reserve(impl_->mappings.size() + 1);
-    void *host_view = impl_->api->register_device_memory_to_host(reinterpret_cast<void *>(dev_base), size);
-    if (host_view != nullptr) {
-        impl_->mappings.push_back(reinterpret_cast<void *>(dev_base));
-        impl_->mapped_bytes += size;
-    } else {
-        host_view = fallback_host_view;
+    const bool needs_push_back = fallback_host_view != nullptr;
+    void *host_view = fallback_host_view;
+    if (host_view == nullptr) {
+        // Reserve before registration: a later reallocation that threw would
+        // leak the mapping.
+        impl_->mappings.reserve(impl_->mappings.size() + 1);
+        host_view = impl_->api->register_device_memory_to_host(reinterpret_cast<void *>(dev_base), size);
+        if (host_view != nullptr) {
+            impl_->mappings.push_back(reinterpret_cast<void *>(dev_base));
+            impl_->mapped_bytes += size;
+        }
     }
     if (host_view == nullptr) {
         return false;
     }
-    impl_->regions.push_back(
-        {dev_base, size, static_cast<unsigned char *>(host_view), reinterpret_cast<uintptr_t>(host_view) != dev_base}
-    );
+    impl_->regions.push_back({dev_base, size, static_cast<unsigned char *>(host_view), needs_push_back});
     return true;
 }
 
@@ -110,7 +110,7 @@ bool HostTensorAccessor::write(uint64_t dev_addr, const void *src, uint64_t byte
     }
     unsigned char *dst = region->host_view + offset;
     memcpy(dst, src, bytes);
-    if (!region->mirrored) {
+    if (!region->needs_push_back) {
         return true;
     }
     return impl_->api->copy_to_device(reinterpret_cast<void *>(dev_addr), dst, static_cast<size_t>(bytes)) == 0;
--- a/src/a2a3/runtime/host_build_graph/host/runtime_maker.cpp
+++ b/src/a2a3/runtime/host_build_graph/host/runtime_maker.cpp
@@ -1070,16 +1070,13 @@ extern "C" int bind_callable_to_runtime_impl(
         LOG_DEBUG("  ChipTensor %d: %zu bytes at %p", i, size, dev_ptr);
 
         // host_build_graph runs the orchestrator on the host, which may read
-        // control tensors (e.g. paged_attention's context_lens/block_table) via
-        // get_tensor_data to shape the graph. Give it a host view of this
-        // buffer: the device buffer itself where the platform can map it into
-        // the host address space (released in validate_runtime_impl before
-        // device_free), otherwise the staging copy, which holds the same bytes
-        // for the whole orchestration window and whose writes are pushed back
-        // to the device. A tensor with neither is not host-accessible, so the
-        // prepare fails here rather than the orchestrator dereferencing a
-        // device address.
-        if (!tensor_access.add(reinterpret_cast<uint64_t>(dev_ptr), size, host_ptr)) {
+        // staged control tensors (e.g. paged_attention's context_lens and
+        // block_table) via get_tensor_data to shape the graph. A pure output
+        // has no valid readable bytes before execution, and a5 cannot map it;
+        // exposing its caller buffer would therefore make reads unsafe. Leave
+        // it unregistered so both get_tensor_data and set_tensor_data fail
+        // closed during orchestration.
+        if (!is_pure_output && !tensor_access.add(reinterpret_cast<uint64_t>(dev_ptr), size, host_ptr)) {
             LOG_ERROR("host-orch: no host view for tensor %d (dev_ptr %p, %zu bytes)", i, dev_ptr, size);
             return PTO_RUNTIME_ERR_INTERNAL;
         }
--- a/src/a2a3/runtime/host_build_graph/runtime/host_tensor_access.h
+++ b/src/a2a3/runtime/host_build_graph/runtime/host_tensor_access.h
@@ -20,25 +20,32 @@
  * that capability is resolved, so the orchestrator core never dereferences a
  * device address itself.
  *
- * A region is registered per staged tensor with the host address serving it:
+ * The current bind path registers one region per staged tensor, backed by the
+ * caller's host tensor buffer:
  *
- *   - `host_view == dev_base`  — the device buffer is mapped into the host
- *     address space (a2a3 `halHostRegister(DEV_SVM_MAP_HOST)`; sim, where a
- *     device pointer is already a host pointer). Reads and writes land on the
- *     device bytes directly.
- *   - `host_view != dev_base`  — no host mapping exists. The region is served
- *     from the staging buffer holding the same bytes, and a write is pushed
- *     back through the device-copy hook so the device observes it.
+ *   - A read observes that caller buffer.
+ *   - A write mutates that caller buffer, then uses the device-copy hook so the
+ *     device observes it too. This is visible even for an `IN` argument if its
+ *     host orchestration calls `set_tensor_data`.
+ *
+ * `add` also retains a null-fallback platform path: it asks the platform for a
+ * host-readable mapping whose address may equal or differ from `dev_base`, and
+ * always accesses the returned address. The current runtime-maker path cannot
+ * reach it: staged tensors always have the caller buffer, while pure outputs
+ * are deliberately left unregistered. The path remains as an explicit
+ * platform-capability escape hatch in `add` and is covered directly by unit
+ * tests; no current production caller reaches it.
  *
  * An address no registered region covers is a failure, never a raw
- * dereference: only tensors the runtime staged have a host view at all, so a
- * GM-heap tensor or a pass-through child-memory buffer resolves to nothing.
+ * dereference. Pure outputs, GM-heap tensors and pass-through child-memory
+ * buffers have no region, so both reads and writes resolve to nothing.
  *
- * Registrations are owned by one orchestration run — the window between
- * staging and the first dispatched task. A mirror is a copy, and nothing has
- * executed yet to make it stale; once tasks run, a stale mirror would be
- * indistinguishable from live device memory. `HostTensorAccessor` bounds that
- * window and releases its mappings on every exit path.
+ * Regions and any optional mappings are owned by one orchestration run — the
+ * window between staging and the first dispatched task. A caller-buffer view
+ * holds the staged bytes, and nothing has executed yet to make it stale; once
+ * tasks run, that view would be indistinguishable from live device memory.
+ * `HostTensorAccessor` bounds the window and releases its mappings on every
+ * exit path.
  *
  * The read/write pair carries weak fallbacks in the runtime translation unit
  * (`orchestrator_core/runtime_core.cpp`) that dereference `dev_addr` directly,
@@ -55,8 +62,8 @@
 struct HostApi;  // common/host_api.h — fwd-declared so this header stays out of platform includes
 
 /**
- * The registered regions of one orchestration run, and the mappings that run
- * installed to serve them.
+ * The registered regions of one orchestration run, and any optional mappings
+ * that run installed to serve them.
  *
  * One accessor per run, mutated only by the thread running that run's
  * orchestration. The region and mapping tables are plain vectors with no lock,
@@ -87,9 +94,11 @@ public:
     HostTensorAccessor &operator=(const HostTensorAccessor &) = delete;
 
     /**
-     * Register `[dev_base, dev_base + size)`, preferring a host mapping of the
-     * device buffer and falling back to `fallback_host_view` (the staging copy)
-     * when the platform cannot map it.
+     * Register `[dev_base, dev_base + size)`, using `fallback_host_view` (the
+     * caller's host tensor buffer) when available and asking the platform for a
+     * host mapping otherwise. The current runtime-maker always supplies the
+     * fallback for staged tensors and skips pure outputs, so its bind path does
+     * not install mappings.
      *
      * @return false for an empty region, a null `api`, or when neither a
      *         mapping nor a fallback view is available.
--- a/tests/ut/cpp/a2a3/test_hbg_tensor_access.cpp
+++ b/tests/ut/cpp/a2a3/test_hbg_tensor_access.cpp
@@ -13,11 +13,9 @@
  * Host-view resolution for the host orchestrator's tensor reads and writes,
  * and the per-run ownership of the mappings that serve them.
  *
- * The mirror path (a platform that cannot map device memory into the host
- * address space) has no reachable call site on a2a3, whose SVM map always
- * succeeds, so these tests are the only place it executes. `g_registered_view`
- * is what the fake `register_device_memory_to_host` hands back: null models a
- * platform with no mapping, so `add` falls back to the staging view.
+ * The fallback path serves staged tensors without mapping their device
+ * allocations. `g_registered_view` is what the fake
+ * `register_device_memory_to_host` hands back when no fallback is available.
  */
 
 #include <gtest/gtest.h>
@@ -46,6 +44,7 @@ struct CopyCall {
 std::vector<CopyCall> g_copies;
 std::vector<void *> g_unregistered;
 void *g_registered_view = nullptr;
+int g_register_count = 0;
 int g_copy_result = 0;
 
 int record_copy(void *, void *dev_ptr, const void *host_ptr, size_t size) {
@@ -53,7 +52,10 @@ int record_copy(void *, void *dev_ptr, const void *host_ptr, size_t size) {
     return g_copy_result;
 }
 
-void *record_register(void *, void *, size_t) { return g_registered_view; }
+void *record_register(void *, void *, size_t) {
+    ++g_register_count;
+    return g_registered_view;
+}
 
 void record_unregister(void *, void *dev_ptr) { g_unregistered.push_back(dev_ptr); }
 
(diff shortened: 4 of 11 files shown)


src/a2a3/runtime/host_build_graph/host/runtime_maker.cpp
```diff
@@ -990,13 +990,11 @@ extern "C" int bind_callable_to_runtime_impl(
...
-        // device_free), otherwise the staging copy, which holds the same bytes
-        // for the whole orchestration window and whose writes are pushed back
-        // to the device. A tensor with neither is not host-accessible, so the
-        // prepare fails here rather than the orchestrator dereferencing a
-        // device address.
+        // buffer: the staging copy, which holds the same bytes for the whole
+        // orchestration window and whose writes are pushed back to the device,
+        // or a device mapping when no staging view is available. Without
+        // either, prepare fails here rather than the orchestrator dereferencing
+        // a device address.
```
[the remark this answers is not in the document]
TaoZQY: Fixed in a7114c1b. Pure OUT tensors now pass a null staging fallback, so `HostTensorAccessor` retains the device mapping for their unstaged bytes. The regression test verifies registration, mapping accounting, and cleanup when no fallback exists. I also changed the shared BGEMM config read to use `get_tensor_data()` instead of directly dereferencing a device address; the two previously failing a2a3 onboard cases now pass together.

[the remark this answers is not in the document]
TaoZQY: Follow-up after the a5 onboard result: a5 cannot create this device mapping, so passing a null fallback made every pure-OUT bind fail with `no host view`. The final fix in 2f3989e6 leaves pure outputs unregistered instead. They contain no readable bytes before execution, and an attempted `get_tensor_data()` still fails closed through the existing unregistered-span path, without requiring unsupported mapping support. The a2a3/a5 sim regressions and accessor fail-closed test pass; the refreshed onboard CI is running.

src/a2a3/runtime/host_build_graph/host/host_tensor_access.cpp
```diff
@@ -72,16 +72,17 @@ bool HostTensorAccessor::add(uint64_t dev_base, uint64_t size, void *fallback_ho
...
     impl_->regions.reserve(impl_->regions.size() + 1);
-    impl_->mappings.reserve(impl_->mappings.size() + 1);
-    void *host_view = impl_->api->register_device_memory_to_host(reinterpret_cast<void *>(dev_base), size);
-    if (host_view != nullptr) {
-        impl_->mappings.push_back(reinterpret_cast<void *>(dev_base));
-        impl_->mapped_bytes += size;
-    } else {
-        host_view = fallback_host_view;
+    void *host_view = fallback_host_view;
+    if (host_view == nullptr) {
```
ChaoZheng109, line 78: After this change the `fallback_host_view == nullptr` branch is unreachable from production code -- not merely rare. The chain:

1. `HostTensorAccessor::add()` has exactly one caller per arch: `runtime_maker.cpp:995` (a5: `:1011`).
2. That call site is guarded by `!is_pure_output`.
3. Under `!is_pure_output`, `api->copy_to_device(dev_ptr, host_ptr, size)` has already run a few lines above.
4. `copy_to_device` hard-rejects a null `host_ptr`: `src/common/platform/sim/host/c_api_shared.cpp:82` and `src/common/platform/onboard/host/c_api_shared.cpp:96` both `return PTO_RUNTIME_ERR_INTERNAL`.
5. On a non-zero rc the call site does `device_free(dev_ptr); return PTO_RUNTIME_ERR_INTERNAL;` -- `add()` is never reached.

So `register_device_memory_to_host` / `unregister_device_memory_from_host`, the dlsym'd HAL loading in `device_runner.cpp:926-971`, the `mappings` / `mapped_bytes` bookkeeping, `close()`'s unregister loop, and the `host_view_close` DFX phase are all retained for a branch only the unit tests reach.

A one-line alternative keeps the branch meaningful and removes the write-side regression noted on `runtime_maker.cpp:995`:

```cpp
tensor_access.add(reinterpret_cast<uint64_t>(dev_ptr), size,
                  is_pure_output ? nullptr : host_ptr);
```

A pure OUT tensor is exactly the case this branch was designed for: it has device memory, no meaningful host content, but may still need to be written by orchestration. Cost, using the figures in #1848: 40,860,162,880 - 40,859,999,040 = 163,840 B (160 KiB), roughly 40 pages, ~1 us -- still 99.9996% of the mapped bytes removed.

If you would rather keep the current shape, please say so in the header: the branch is unreachable in production and retained only as a platform-capability escape hatch. Otherwise the next reader will assume it still runs.

src/a2a3/runtime/host_build_graph/runtime/host_tensor_access.h
```diff
@@ -22,13 +22,14 @@
...
- *     address space (a2a3 `halHostRegister(DEV_SVM_MAP_HOST)`; sim, where a
- *     device pointer is already a host pointer). Reads and writes land on the
- *     device bytes directly.
- *   - `host_view != dev_base`  — no host mapping exists. The region is served
- *     from the staging buffer holding the same bytes, and a write is pushed
- *     back through the device-copy hook so the device observes it.
+ *   - `host_view != dev_base`  — the region uses the available staging buffer,
+ *     and a write is pushed back through the device-copy hook so the device
+ *     observes it.
+ *   - `host_view == dev_base`  — without a staging buffer, the device buffer is
```
ChaoZheng109: This paragraph is rewritten here, so it seems the right place to ask: what is the evidence that a2a3 `halHostRegister(DEV_SVM_MAP_HOST)` yields `host_view == dev_base`? The available sources disagree.

- **This header** states that a2a3 maps such that `host_view == dev_base`.
- **#1848** states the opposite: "a successful registration (which returns a distinct host VA) still sets `mirrored = true`".
- **`src/common/platform/include/common/host_api.h:31`** warns "The returned VA **may differ from dev_ptr**, so callers must use it, not dev_ptr, for host access".
- **The driver header** (`ascend_hal_base.h`, `halHostRegister`) names the parameters `src_ptr` -> `dst_ptr` and documents the out parameter as "the address of the **allocated** dst memory pointer", i.e. the contract produces a new address.

This is load-bearing for this PR, because it decides what the `bgemm_orch.cpp` change actually is:

- If `host_view == dev_base`, then `data_as<int64_t>()` worked on a2a3 **because of** this mapping, and removing the mapping would dereference an address that is no longer host-readable -- the bgemm change is a required crash fix.
- If the returned VA is distinct, the device address was host-readable independently and the bgemm change is hardening only.

Cutting against the "distinct VA" reading: `examples/a2a3/host_build_graph/benchmark_bgemm` is `level=2`, runs in `st-onboard-a2a3`, and is green on main while dereferencing `t.buffer.addr` directly.

The consequence reaches past bgemm. Under the first reading, any orchestration source reused by hbg that dereferences a device address host-side now faults. I swept the tree: the remaining `data_as<void>()` uses (the paged_attention family's `query_ptr` / `kc_ptr` / `out_ptr`) only pass the address to `make_tensor_external` and never dereference it, so they are safe -- but nothing documents or enforces that invariant.

A one-line `LOG_INFO` comparing `dev_ptr` against `host_va` inside `register_device_memory_to_host` on a2a3 would settle it, and the answer belongs in this paragraph.

src/a2a3/runtime/host_build_graph/host/host_tensor_access.cpp
```diff
@@ -72,16 +72,17 @@ bool HostTensorAccessor::add(uint64_t dev_base, uint64_t size, void *fallback_ho
...
-    // first: a reallocation that threw there would leak the mapping.
     impl_->regions.reserve(impl_->regions.size() + 1);
-    impl_->mappings.reserve(impl_->mappings.size() + 1);
-    void *host_view = impl_->api->register_device_memory_to_host(reinterpret_cast<void *>(dev_base), size);
-    if (host_view != nullptr) {
-        impl_->mappings.push_back(reinterpret_cast<void *>(dev_base));
-        impl_->mapped_bytes += size;
-    } else {
-        host_view = fallback_host_view;
+    void *host_view = fallback_host_view;
```
ChaoZheng109, line 77: `mirrored` is computed below as `host_view != dev_base`, so letting the fallback win here also changes where orchestration **writes** land. Note that `fallback_host_view` is the caller's own host tensor buffer (`runtime_maker.cpp`'s `host_ptr`, the source `copy_to_device` reads from), not a runtime-owned staging allocation.

- Before: the write went to the mapped VA, or straight to the device bytes, and never touched the caller's buffer.
- After: `HostTensorAccessor::write` does a `memcpy` into the caller's buffer, then `copy_to_device`.

This holds regardless of how the question on `host_tensor_access.h:28` resolves; only the old destination differs between the two readings.

Consequence: for a tensor declared `D.IN`, `needs_copy_back` is false, so a `set_tensor_data` during orchestration mutates the caller's input array in place and nothing overwrites it afterwards -- a silent mutation of user-owned memory. No in-repo case hits this today (`available_aicore_counts`, the only hbg case using `set_tensor_data`, is `[INOUT, INOUT]`), which is why CI is green.

Worth stating as a present-tense invariant in the header: on a fallback region, a write lands in the caller's host buffer as well as on the device.

src/a2a3/runtime/host_build_graph/host/runtime_maker.cpp
```diff
@@ -988,16 +988,11 @@ extern "C" int bind_callable_to_runtime_impl(
...
-        // for the whole orchestration window and whose writes are pushed back
-        // to the device. A tensor with neither is not host-accessible, so the
-        // prepare fails here rather than the orchestrator dereferencing a
-        // device address.
-        if (!tensor_access.add(reinterpret_cast<uint64_t>(dev_ptr), size, host_ptr)) {
+        // staged control tensors (e.g. paged_attention's context_lens and
+        // block_table) via get_tensor_data to shape the graph. Pure outputs
+        // have no readable bytes before execution, so they remain unregistered
+        // and any attempted read fails closed.
+        if (!is_pure_output && !tensor_access.add(reinterpret_cast<uint64_t>(dev_ptr), size, host_ptr)) {
```
ChaoZheng109, line 1079: The comment above says a pure output stays unregistered so "any attempted **read** fails closed", but skipping `add()` drops the region entirely, so **writes** fail closed too: `set_tensor_data` on a pure OUT tensor now reaches `report_fatal` in `runtime_core.cpp:371`.

That path worked before this change -- through the mapping on a2a3, and through the staging view on a5 onboard, which has always taken the fallback path. It is a capability regression, and neither the comment nor the PR description mentions it. Either state it here, or adopt the `is_pure_output ? nullptr : host_ptr` form suggested on `host_tensor_access.cpp:77`, which restores the write path for exactly this case at roughly 1 us.

One thing this does improve, and which is worth keeping: `runtime_core.cpp`'s error text ("only tensors the runtime staged are readable") is now accurate. Before, pure outputs were registered but never staged, so the message was slightly wrong.

src/a2a3/runtime/host_build_graph/runtime/host_tensor_access.h
```diff
@@ -87,9 +88,8 @@ class HostTensorAccessor {
     HostTensorAccessor &operator=(const HostTensorAccessor &) = delete;
 
     /**
-     * Register `[dev_base, dev_base + size)`, preferring a host mapping of the
-     * device buffer and falling back to `fallback_host_view` (the staging copy)
-     * when the platform cannot map it.
+     * Register `[dev_base, dev_base + size)`, using `fallback_host_view` (the
+     * staging copy) when available and mapping the device buffer otherwise.
```
ChaoZheng109: Anchoring here because the affected files are outside this diff. The DFX docs still describe the pre-change behaviour, and `.claude/rules/doc-consistency.md` SS1/SS3 asks for that fix to land in the same commit.

- `docs/dfx/hbg-bind-phases.md:28` -- describes `args` as "staging the caller's host tensors **and mapping them for host access**". The mapping no longer happens.
- `docs/dfx/hbg-bind-phases.md:33`, `:38`, `:50`, `:347` -- the definition of `host_view_close` and the surrounding discussion.
- `docs/dfx/hbg-bind-phases.md:334` -- the table row `host_view_close (excluded) | 0.25 s / 40.9 GB | 0.28 s / 45.8 GB`, now permanently ~0 at `count=0`.
- `docs/investigations/2026-08-hbg-graph-block-decomposition.md:187` -- "312 ms for `host_view_close`".

Related, on the Performance section: the PR reports the teardown side (`host_view_close` at 20-24 us) but not the registration side. #1848 expected both to go -- its `bind.args` was 744-1020 ms covering "19 tensors staged H2D **+ 20 registrations**". With `count=0` those 20 `halHostRegister` calls are gone as well; that delta is the other half of the win and is worth quoting.

tests/ut/cpp/a2a3/test_hbg_tensor_access.cpp
```diff
@@ -121,8 +148,8 @@ TEST_F(HostTensorAccessTest, MirroredWriteReportsCopyFailure) {
 }
 
 // The fail-closed contract: an address outside every registered region — a
-// GM-heap tensor the orchestrator created, or a pass-through child-memory
-// buffer — resolves to nothing instead of being dereferenced.
+// pure output, GM-heap tensor the orchestrator created, or a pass-through
```
ChaoZheng109: This comment now claims a pure output is covered by the fail-closed test, but the test never passes through `runtime_maker.cpp`'s `!is_pure_output` gate -- it drives `HostTensorAccessor` directly, so what it actually covers is "an address in no registered region". The gate that decides a pure output is never registered has no coverage at all.

Either restore this line to what it described before, or add a case that exercises the gate.

The `HostTensorAccessor`-level tests added in this PR are otherwise a good addition -- `FallbackViewAvoidsDeviceMapping` and the `mapping_count` / `mapped_bytes` assertions on `MissingFallbackUsesDeviceMapping` pin the new preference order precisely. The gap is one level up.

ChaoZheng109 (commented):
Reviewed the merge-base diff (`66ba5c4a`...`e42b92d4`). The direction is right and the change is well scoped: 8 files / +94 -72, the a2a3 and a5 diffs are byte-identical apart from line offsets, and CI is fully green.

Two things I verified that strengthen the case and are worth adding to the description:

- **a5 onboard has always run this way.** `register_device_memory_to_host` is overridden only by a2a3 onboard (`src/a2a3/platform/onboard/host/device_runner.cpp:926`); a5 onboard uses the base default (`src/common/platform/onboard/host/device_runner_base.h:160`, `return nullptr`) and has therefore always taken the fallback path. This PR aligns a2a3 with behaviour a5 already ships. sim is a third case -- `src/common/platform/sim/host/device_runner_base.h:203` returns `dev_ptr` identity, so it took the mapped path.
- **The `bgemm_orch.cpp` change is not a drive-by.** `examples/a2a3/host_build_graph/benchmark_bgemm` reuses that source and runs in `st-onboard-a2a3` (`pytest examples tests/st --exclude-level 4`, and the case is `level=2`), so it is load-bearing -- but its exact status depends on the question left inline on `host_tensor_access.h:28`.

Six inline comments below, in severity order. The dead-branch question and the `halHostRegister` address-semantics question are the two I would like answered before merge. The write-semantics and pure-output-write notes are comment/invariant fixes. Doc sync is required by `.claude/rules/doc-consistency.md` SS1/SS3. The last one is a test comment that overstates its coverage.

Everything below applies equally to the a5 copies of the same files.

src/a2a3/runtime/host_build_graph/host/host_tensor_access.cpp
ChaoZheng109: **Correction -- withdrawing the alternative I suggested above.** I had not read the resolved threads on this PR before posting. You already implemented exactly `is_pure_output ? nullptr : host_ptr` in a7114c1b, and a5 onboard rejected it: with no mapping capability there, both views end up null, `add()` returns false, and every pure-OUT bind fails with `no host view`. 2f3989e6 (leaving pure outputs unregistered) is the right call, and my suggestion would have reintroduced the exact failure you already fixed. Apologies for re-litigating a settled point.

What survives is the observation itself, and the a5 result sharpens rather than weakens it. The mapping branch is unreachable on a2a3 by the five-step chain above, **and** it cannot be served on a5 at all, since `register_device_memory_to_host` there is the unoverridden base default returning `nullptr` (`src/common/platform/onboard/host/device_runner_base.h:160`). Your a7114c1b experiment is the empirical proof of the second half.

So the remaining question is narrower than I framed it. It is not "which of two shapes" -- routing pure outputs into that branch is closed. It is simply: given the branch cannot be reached on a2a3 and cannot be served on a5, should the header record that, or should the plumbing (`register_device_memory_to_host` / `unregister_device_memory_from_host`, the dlsym'd HAL loading, the `mappings` / `mapped_bytes` bookkeeping, `close()`'s unregister loop, the `host_view_close` DFX phase) be retired? Either answer is defensible; what should not survive the merge is a reader assuming that code still runs on some platform.

src/a2a3/runtime/host_build_graph/host/runtime_maker.cpp
ChaoZheng109: **Correction:** strike the "adopt the `is_pure_output ? nullptr : host_ptr` form" suggestion from the comment above. a7114c1b tried exactly that and a5 onboard rejected it -- details in the reply on `host_tensor_access.cpp:77`. Leaving pure outputs unregistered, as 2f3989e6 does, is correct.

The documentation point stands on its own and is unaffected: the comment says a pure output stays unregistered so reads fail closed, but writes fail closed too (`set_tensor_data` reaches `report_fatal` in `runtime_core.cpp:371`). Given the a5 constraint that is a deliberate and unavoidable consequence, which is exactly why it is worth stating here rather than leaving for the next reader to discover from a fatal at runtime.

src/a2a3/runtime/host_build_graph/host/host_tensor_access.cpp
TaoZQY: Fixed in 4f72be84 — the header now records that the null-fallback branch has no current production caller and is retained as an explicit platform-capability escape hatch. The DFX docs also state that the current bind path installs no mappings.

src/a2a3/runtime/host_build_graph/runtime/host_tensor_access.h
TaoZQY: Fixed in 4f72be84 — the accessor no longer infers mapping semantics from address equality. Push-back is selected by view origin: caller-buffer fallbacks copy to device, while platform mappings write directly even when the returned host VA differs. The updated unit test uses a deliberately distinct mapping VA. BGEMM still uses the tensor-access API because the HostApi contract never permits dereferencing the raw device address.

src/a2a3/runtime/host_build_graph/host/host_tensor_access.cpp
TaoZQY: Fixed in 4f72be84 — the header now explicitly states that fallback writes mutate the caller buffer and then copy to device, including for an IN argument. The renamed FallbackWriteMutatesCallerBufferAndPushesToDevice test pins that behavior.

src/a2a3/runtime/host_build_graph/host/runtime_maker.cpp
TaoZQY: Fixed in 4f72be84 — both runtime-maker copies and the PR description now state that pure outputs remain unregistered so both reads and writes fail closed, with the a5 no-mapping constraint documented as the reason.

src/a2a3/runtime/host_build_graph/runtime/host_tensor_access.h
TaoZQY: Fixed in 4f72be84 — both DFX documents and the reporting-tool comment now distinguish the historical mapping path from the current count=0 path. The docs also record that the old args phase included 20 registrations plus 19 H2D copies, while the current path removes the registration side.

tests/ut/cpp/a2a3/test_hbg_tensor_access.cpp
TaoZQY: Fixed in 4f72be84 — the test comment now claims only what the accessor-level test covers: an address outside every registered region. It no longer claims coverage of the runtime-maker pure-output gate.

ChaoZheng109 (approved):

Issue #1950, opened 2026-08-21
[Code Health] Nothing records whether a run paid torch_npu's device autoload


ChaoWao:
### Category

Robustness (potential edge-case failure)

### Component

Tests

### Description

`torch` >= 2.5 runs every registered `torch.backends` entry point on `import torch`.
On this box that entry point exists and belongs to `torch_npu`, verified without
importing anything:

```console
$ python3 -c "import importlib.metadata as md; [print(e.name,'->',e.value,'from',e.dist.name) for e in md.entry_points().select(group='torch.backends')]"
torch_npu -> torch_npu:_autoload from torch_npu
$ python3 -c "import torch, importlib.metadata as md; print(torch.__version__, md.version('torch_npu'))"
2.7.1+cpu 2.7.1
```

So **every `import torch` in this repo runs `torch_npu._autoload()`** unless
`TORCH_DEVICE_BACKEND_AUTOLOAD=0` is set, and that autoload acquires an NPU device.
The repo imports `torch` in 21+ example/test files plus the golden modules and the
scene-test harness itself.

Two things make this worth tracking:

**Nothing in this repo uses torch on the NPU.** There is no `import torch_npu`, no
`.npu()`, and no `device="npu"` anywhere in the tree — `torch` is used purely for
CPU golden math (positive control: 21 files under `examples/` + `tests/` call CPU
torch ops). Whatever the autoload costs, it buys this repo nothing.

**The condition is unobservable after the fact.** No code here reads
`TORCH_DEVICE_BACKEND_AUTOLOAD`, so no log line records whether a given run paid
for the autoload. Two logs therefore cannot be compared for "same ruler" without
out-of-band knowledge of how each was launched.

That is not hypothetical. During the `host_build_graph` bind-phase measurement work
one comparison arm lacked `TORCH_DEVICE_BACKEND_AUTOLOAD=0`, paid for `torch_npu`
grabbing a device at import, and the difference was credited to the branch under
test — a plausible-looking wrong number. The trap is currently only *documented*
(`docs/dfx/hbg-bind-phases.md`), and the bind-phase recipe works around it by
echoing its own command into the log, which covers only runs launched that way.

Scope note: the harm here is to measurement, not to any product output. No test
result is wrong because of this; a timing number can be.

### Location

- `simpler_setup/scene_test.py:180`, `:319`, `:370`, `:1209` — the harness's own `import torch`
- `simpler_setup/torch_interop.py:45`
[...]

PR #1977, opened 2026-08-24
Support: record torch backend autoload state

Crane-Liu:
## Summary

- Record the effective torch backend autoload configuration and observed `torch`/`torch_npu` module state before the first SceneTest dispatch.
- Preserve the raw environment value as bounded JSON with an explicit truncation marker.
- Emit one TIMING record per SceneTest interpreter across L2 and L3 execution.
- Show every distinct autoload state in the bind-phase report and warn when a log does not carry the record.
- Document the measurement invocation and the distinction between configuration intent and observed module state.

## Scope

This PR covers SceneTest measurement and reporting. Other entry points that import torch directly remain tracked by #1950.

## Testing

- Pre-commit hooks
- Focused autoload and bind-report tests: 17 passed
- Python unit tests: 1968 passed, 15 skipped
- Full a2a3sim SceneTest resource and L2 runtime phases
- Full a5sim SceneTest resource and L2 runtime phases
- SceneTest smoke with backend autoload disabled
- Queued real-environment smoke with backend autoload at its default

Refs #1950

### DIFF
--- a/docs/dfx/hbg-bind-phases.md
+++ b/docs/dfx/hbg-bind-phases.md
@@ -101,10 +101,19 @@ Four switches and one flag make the measurement, and each is load-bearing:
 | Switch | Why |
 | ------ | --- |
 | `SIMPLER_HBG_BIND_BREAKDOWN_ENABLE=1` | emits the `bind phase=` lines at all |
-| `SIMPLER_LOG_LEVEL=TIMING` | the level they are emitted at |
-| `TORCH_DEVICE_BACKEND_AUTOLOAD=0` | otherwise `torch_npu` grabs a device on import, which a host-only measurement does not want — and nothing in the log records whether it was set |
+| `--log-level timing` | enables the TIMING records used by the report; this is the SceneTest default |
+| `TORCH_DEVICE_BACKEND_AUTOLOAD=0` | keeps CPU golden imports from loading `torch_npu`; the `torch_backend_autoload` timing record confirms the effective setting and observed module state |
 | `SIMPLER_SKIP_DEVICE_RUN=1` | returns at `simpler_launch_run`, so the host path is measured without a working device run |
 | `--skip-golden` | with the device skipped the outputs a golden check compares are never produced, so a checking case such as qwen otherwise fails at validation with the whole measurement already complete in the log. On dsv4 it also skips the fixture upload its driver does by default, which is 42.6 GiB per rank of H2D you are not measuring |
+
+SceneTest emits one `torch_backend_autoload` record per interpreter after
+torch-dependent argument preparation and before the first dispatch. `effective`
+reports the environment's dispatch-time intent using torch's current private
+autoload predicate; `torch_npu_loaded` reports the observed module state and is
+the authoritative field if the environment changed after torch was imported.
+`raw` is the JSON-encoded environment value (`null` when unset); values longer
+than 64 characters carry their first 64 characters and
+`raw_truncated=true`, keeping the record single-line and bounded.
 
 **`SIMPLER_SKIP_DEVICE_RUN` is presence-based.** `SIMPLER_SKIP_DEVICE_RUN=0` still
 skips; `unset` it. It is a temporary handle from the dsv4 bring-up and is deleted
@@ -191,10 +200,14 @@ its failure modes have already produced wrong answers on this box.
 **Both arms must be the same ruler, and the log is the only witness you get.** A
 baseline missing `TORCH_DEVICE_BACKEND_AUTOLOAD=0` produced a wrong number once:
 it alone paid for `torch_npu` grabbing a device on import, and the difference was
-attributed to the branch. Nothing in this repo reads that variable — it is
-`torch_npu`'s own — so no log line records whether it was set. The recipe
-therefore echoes the command it is about to run, verbatim, as the log's first
-line, and `hbg_bind_phases` prints that line above the table:
+attributed to the branch. Compare the `torch_backend_autoload` timing record in
+both logs; the measurement recipe produces
+`setting=0 raw="0" raw_truncated=false effective=disabled torch_imported=true torch_npu_loaded=false`.
+`hbg_bind_phases` prints every distinct record above its table and warns when the
+log does not carry one.
+
+The recipe also echoes the command it is about to run, verbatim, as the log's
+first line, and `hbg_bind_phases` prints that line above the table:
 
 ```bash
 diff <(head -1 base.log) <(head -1 measure.log)   # must differ only in the commit
--- a/simpler_setup/scene_test.py
+++ b/simpler_setup/scene_test.py
@@ -24,6 +24,7 @@ from __future__ import annotations
 
 import gc
 import inspect
+import json
 import logging
 import os
 import platform as host_platform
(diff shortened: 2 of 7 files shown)


docs/dfx/hbg-bind-phases.md
```diff
@@ -179,10 +182,12 @@ its failure modes have already produced wrong answers on this box.
 **Both arms must be the same ruler, and the log is the only witness you get.** A
 baseline missing `TORCH_DEVICE_BACKEND_AUTOLOAD=0` produced a wrong number once:
 it alone paid for `torch_npu` grabbing a device on import, and the difference was
-attributed to the branch. Nothing in this repo reads that variable — it is
-`torch_npu`'s own — so no log line records whether it was set. The recipe
-therefore echoes the command it is about to run, verbatim, as the log's first
-line, and `hbg_bind_phases` prints that line above the table:
+attributed to the branch. Compare the `torch_backend_autoload` timing record in
```
ChaoZheng109, line 203: **The log now carries the fact, but the report still doesn't show it — the audit loop isn't closed.**

#1950's stated goal is that a number be "auditable from its own log instead of from memory", and it explicitly faults the existing workaround for "cover[ing] only runs launched that way". The new record has the dual weakness: it's *in the log, but not in the report*. This line asks the human to `diff` two logs by hand, which is strictly weaker than what the repo already does for the command stamp.

`simpler_setup/tools/hbg_bind_phases.py` already has the channel for this: `parse_stamp()` prints the `[stamp]` line above the table, and warns explicitly when it's missing (`:131-135`). A `parse_torch_autoload()` in the same shape — printed next to the stamp, warning when absent — is ~15 lines and makes the record impossible to miss.

Suggest doing it in this PR. Otherwise the failure mode the issue set out to kill just changes form, from "audited from memory" to "audited by manual diff".

simpler_setup/scene_test.py
```diff
@@ -103,6 +105,33 @@ def _effective_diagnostic_options(
     return _DiagnosticOptions(0, 0, 0, False, False, False)
 
 
+@cache
```
ChaoZheng109, line 110: **The pid-keying has no reachable path in the current tree, and the test covering it cannot fail.**

`@cache` keyed on pid differs from a plain module-level bool only when a process emits the record and *then* forks a child that also calls this function. I couldn't find such a path:

- L3's chip/sub children are forked by eager `Worker.init()` **before** `generate_args` runs (see the comment at `scene_test.py:1775`), and they never enter `_run_and_validate_l2/_l3`, so they never emit.
- Multi-device per-case parallelism goes through `simpler_setup/parallel_scheduler.py`, which uses `subprocess.Popen` — fresh interpreters whose cache is empty regardless.

So the PR description's "including forked processes" currently claims more than the code can demonstrate. Two ways out, either is fine:

- name a real path that reaches it, and record it as a comment here so it doesn't read as speculative; or
- simplify to a module-level flag and drop the claim from the PR body.

simpler_setup/scene_test.py
```diff
@@ -103,6 +105,33 @@ def _effective_diagnostic_options(
...
+def _log_torch_backend_autoload(_process_id: int) -> None:
```
ChaoZheng109: `_process_id`'s leading underscore reads as "unused", but it's the sole cache key and therefore load-bearing. The docstring doesn't state that invariant either, so the next reader can reasonably delete the parameter as dead weight and silently turn "once per process" into "once per interpreter".

Suggest renaming it and stating the invariant as a present-tense fact, per `.claude/rules/comments.md`:

```python
@cache
def _log_torch_backend_autoload(process_id: int) -> None:
    \"\"\"Record torch backend autoload configuration and observed module state.

    Keyed by pid: a process that inherits this cache across a fork misses on
    its own pid and emits its own record.
    \"\"\"
```

While you're here — one more sentence worth adding: `effective` is sampled at dispatch time, whereas torch reads the variable at `import torch` time, so the two can disagree if anything mutates `os.environ` in between (nothing in the tree does today). Stating that `effective` is the *intent* and `torch_npu_loaded` is the *observed fact*, and that the latter wins on conflict, would keep a future reader from trusting the wrong field.

ChaoZheng109 (commented):
Reviewed against #1950. Everything the issue makes a hard requirement is met, and one of those is easy to "fix" by mistake, so let me state it up front:

- **`setting=invalid` → `effective=disabled` is correct, not a typo.** torch gates on `os.getenv("TORCH_DEVICE_BACKEND_AUTOLOAD", "1") == "1"`, so any value other than `1`/unset disables autoload. `!= "0"` would have been the bug. Please don't change this.
- No behavior change, no new gate, and the harness does not set `=0` itself — exactly the "deliberately not proposed here" boundary the issue drew.
- Sampling `sys.modules` after torch has actually been imported is the right moment to take the reading.

Three inline comments below, plus one scope question that needs an author decision.

### Scope: does this PR close #1950, or only part of it?

#1950's problem statement covers "every `import torch` in this repo", and its Location list names `python/simpler/env_manager.py` as "the existing env-reading surface ... no recording of effective values". This PR records only inside `SceneTestCase._run_and_validate_l2/_l3`, so any entry point that does not go through the SceneTest harness — direct `Worker` use, the pypto-lib / pypto-serving cross-repo benchmarks — still has no record. Those are precisely the comparisons where a "same ruler" mistake is easiest to make and hardest to catch.

The narrowing is defensible: the issue's Component is `Tests` and its scope note says the harm is to measurement, and the bind-phase recipe does go through SceneTest. But it is currently unstated. Please pick one:

1. **Cover it in this PR** — also record at the `Worker` / `env_manager` level, or
2. **Deliberately leave it out** — say so in the PR body with the reason, and change `Fixes #1950` to `Refs #1950` so the issue stays open for the remaining surface.

Either is fine. What I'd like to avoid is landing as `Fixes` while a named part of the issue is untouched.

Related, non-blocking: the record only fires for cases that reach dispatch, so a run that dies during compile, during collection, or inside `generate_args` produces no record at all — even though the bind-phase workflow explicitly expects partial runs to still yield a usable host picture.

tests/ut/py/test_scene_test_torch_autoload.py
```diff
@@ -0,0 +1,181 @@
...
+
+    scene_test._log_torch_backend_autoload_once()
+    scene_test._log_torch_backend_autoload_once()
+
+    assert len(_messages(caplog)) == 1
+
+
+def test_record_is_emitted_again_in_a_forked_process(monkeypatch, caplog):
+    process_id = [11]
+    monkeypatch.setattr(scene_test.os, "getpid", lambda: process_id[0])
```
ChaoZheng109: This is the test backing the "emitted again in a forked process" claim, but it fakes `os.getpid`, so it passes whether or not any real fork path reaches `_log_torch_backend_autoload_once`.

See my comment on `simpler_setup/scene_test.py:108`: if the pid-keying stays, this test doesn't show that it works; if it goes, this test goes with it.

docs/dfx/hbg-bind-phases.md
Crane-Liu: Fixed - `hbg_bind_phases` now prints each distinct torch autoload record next to the run stamp and warns when the record is absent. Regression tests cover both paths.

simpler_setup/scene_test.py
Crane-Liu: Fixed - the recorder now uses a zero-argument once-per-interpreter cache, matching the reachable SceneTest execution model.

simpler_setup/scene_test.py
Crane-Liu: Fixed - the pid cache key is gone. The documentation now distinguishes dispatch-time configuration intent from observed module state and makes the latter authoritative on disagreement.

tests/ut/py/test_scene_test_torch_autoload.py
Crane-Liu: Fixed - the synthetic pid test was removed; coverage now asserts reachable once-per-interpreter behavior and complete L2/L3 lifecycle ordering.

ChaoZheng109 (approved):

PR #1982, opened 2026-08-24
Fix: size HBG ready queues from reachable tasks

### LINKED ISSUE #1920: [Code Health] hbg: ready queues are sized by a fixed constant, and one is reachable by more tasks than it has slots
### Category

### Component

AICPU Scheduler

### Description

All nine of the scheduler's ready queues are reserved at the same fixed
`PTO2_READY_QUEUE_SIZE` (8192) regardless of how many tasks can route to each, and
on `examples/a2a3/host_build_graph/qwen3_14b_decode` one of them is reachable by
more tasks than it has slots.

`enqueue_ready` routes every task to exactly one queue — DUMMY shape or a failed
predicate to the dummy queue, a sync-start requirement to the per-shape sync queue,
everything else to the per-shape ready queue — and a task occupies one slot at a
time, so the count of tasks that can reach a queue is an upper bound on its peak
occupancy. A `false` push then latches `PTO2_ERROR_READY_QUEUE_OVERFLOW` and fails
the run. That is deliberate and correct as a response (silently dropping the task
would stall the run instead), but it means **queue capacity is a correctness bound,
not a tuning knob** — there is no retry path.

The bound is exceeded on qwen3-14b decode, and the counts that show it are already
printed by `main`:

```
$ SIMPLER_HBG_BIND_BREAKDOWN_ENABLE=1 SIMPLER_LOG_LEVEL=TIMING \
    python examples/a2a3/host_build_graph/qwen3_14b_decode/test_qwen3_14b_decode.py \
      -p a2a3 -d $TASK_DEVICE --manual include --rounds 3
...
host-orch phase=record_node       total_ns=148731 count=277
host-orch phase=graph_submit      total_ns=200812 count=40
```

The run replays one 277-node Definition 40 times, so 40 x 277 = 11,080 nodes enter
the outer task set on top of the host's own tasks. Even that loose upper bound
already exceeds 8192; instrumenting the routing rule per queue puts
`ready_queue[AIC]`'s reachable population at **9,400 against its 8,192 slots**.

**What keeps this closed today is drain rate, not the bound.** The peak occupancy
the run actually reaches is 101 of 8192, so the queue never fills and nothing
reports the gap — the capacity is simply never compared against the population that
can arrive. Any workload that widens the DAG's antichain (a larger batch, a branchier
graph, more layers at the same nodes per layer, or a slower consumer) closes the
distance, and the failure mode when it does is a latched fatal error rather than
back-pressure.

Both architecture trees carry the same constant, so this is not arch-specific.

### Location

On `93adc386`:

- `src/a2a3/runtime/host_build_graph/runtime/pto_runtime2_types.h:104` —
  `#define PTO2_READY_QUEUE_SIZE 8192`
- `src/a5/runtime/host_build_graph/runtime/pto_runtime2_types.h:103` — same constant
- `src/a2a3/runtime/host_build_graph/runtime/shared/pto_runtime2_init.cpp:99-116` —
  all nine queues reserved at that constant, before orchestration knows the task set
- `src/a2a3/runtime/host_build_graph/runtime/scheduler/pto_scheduler.h:548-570` —
  `enqueue_ready`'s routing and the `if (!pushed) latch_ready_queue_overflow();`
- `src/a2a3/runtime/host_build_graph/runtime/scheduler/pto_scheduler.h:527` —
  `latch_ready_queue_overflow`, a single CAS into `sched_error_code` with no retry

### Proposed Fix

Derive each queue's capacity from the population that can reach it, and raise the
reservation ceiling to admit the largest.

The population is computable at bind time: walk the shared-memory task window for
the host's own tasks, then per Graph submission add the node histogram of the
Definition it references — counted once per distinct Definition and applied per
submission, so a run replaying one Definition 40 times pays one 277-node walk. A
predicated task counts toward both the dummy queue and its shape's queue, since the
predicate is evaluated on device.

Costs, measured:

- the arena reservation grows 11,515,584 -> 16,824,064 bytes for a ceiling of 32768
  across nine queues, allocated once per worker and reused across runs
- `arena_h2d` grows 632 -> 720 bytes; the +88 is the per-queue capacities travelling
  in the layout, which since #1894 / #1895 is their only channel to the device
- the population pass itself costs 0.0021 ms on qwen's 47 tasks and 0.0076 ms on
  DeepSeek-V4 FLASH decode's 1,131, i.e. ~6.6 ns a task

Note that this bounds the exposure rather than removing it: a population past
whatever ceiling is chosen still has to clamp, and the clamp should report at bind
time instead of letting the run fail later. Removing it for *any* population needs a
full-queue push the caller can recover from, and both shapes of that are currently
blocked — spinning until the push succeeds deadlocks at `aicpu_thread_num=1`, where
the pushing thread is the only consumer, and parking the task for retry from the
thread's own dispatch loop needs a thread identity that none of the four calling
chains carries.

A tighter bound would shrink capacity without touching the push path at all: the
DAG's antichain width rather than the reachable count, which the Definition's fanin
CSR already carries.

I have this implemented and measured on a branch and can open the PR.

Related: #1706

### Priority

TaoZQY:
## Summary

- Size each of the nine HBG scheduler ready queues from the number of tasks
  that can reach it during bind, instead of giving every queue the same
  logical capacity of 8192.
- Raise the arena reservation ceiling to 32768 slots per queue so the known
  Qwen3 decode workload (9400 reachable AIC tasks) is covered.
- Reject a graph during host bind with structured
  `SIMPLER_ERROR_READY_QUEUE_OVERFLOW` diagnostics when any single queue's
  reachable population exceeds that ceiling.
- Apply the same implementation to a2a3 and a5, with regression tests and
  updated error-code documentation.

## Motivation

`enqueue_ready` has no retry or back-pressure path: if a queue is full, the
run fails. Queue capacity is therefore a correctness bound rather than only a
performance tuning value.

The Qwen3-14B decode `GraphExecutionBatch16Seq3500` workload has 9400 tasks
that can reach `ready_queue[AIC]`, exceeding the previous fixed logical
capacity of 8192. Its measured peak occupancy is currently much lower, so the
failure is latent, but a wider DAG or a slower consumer can expose it.

## Implementation

- Count the host task window once at bind time.
- Count each distinct Graph Definition once, cache its per-queue population,
  and add it for every submission that references the definition.
- Count predicated tasks toward both the dummy queue and their shape queue,
  because the predicate is evaluated on device.
- Round each logical queue capacity up to a power of two (minimum 2) and pass
  the nine derived capacities to the device through `SchedulerLayout`.
- Keep the physical reservation bounded at 32768 slots per queue. Populations
  above that limit fail before device execution and report the same error via
  the return status, shared scheduler status, logs, and DFX path.
- Keep population derivation host-only so it is not compiled into the AICPU
  runtime, and remove the redundant/confusable reservation-capacity field.

## Memory and performance trade-off

The physical reservation for the nine queues grows from
`9 x 8192 x 24 = 1,769,472` bytes (~1.7 MB) to
`9 x 32768 x 24 = 7,077,888` bytes (~7.1 MB) per reusable worker arena.
Logical capacities remain demand-derived, so small graphs seed only the slots
they can use even though the maximum reservation is larger.

Same-device a2a3 onboard A/B for `paged_attention/small1`, 50 rounds through
`task-submit`:

| Metric (median) | `main` (`d90c574e`) | This PR (`2c92429a`) | Change |
| --- | ---: | ---: | ---: |
| Host wall | 783.8 us | 554.6 us | -29.2% |
| Device wall | 392.9 us | 163.8 us | -58.3% |
| Bind control plane | 28 us | 27 us | no regression |

The small-graph improvement comes from reducing its logical queue capacity
from 8192 to 2, which removes unnecessary device-side slot seeding.

## Validation

- C++ no-hardware unit tests: 122/122 passed.
- a2a3 HBG simulation: 15 passed, 7 skipped.
- a5 HBG simulation: 12 passed.
- `paged_attention/small1` a2a3 onboard golden test passed through
  `task-submit` (`task_20260826_195127_9682646898`).
- Qwen3-14B decode `GraphExecutionBatch16Seq3500` onboard golden test passed
  3/3 through `task-submit` (`task_20260824_003633_278330810026`).
- Pre-commit hooks and all required GitHub CI checks passed.

Fixes #1920

### DIFF
--- a/src/a2a3/runtime/host_build_graph/host/runtime_maker.cpp
+++ b/src/a2a3/runtime/host_build_graph/host/runtime_maker.cpp
@@ -61,6 +61,7 @@
 #include "../runtime/graph_host_state.h"
 #include "../runtime/host_phase_trace.h"
 #include "../runtime/orchestrator.h"
+#include "../runtime/ready_queue_sizing.h"
 #include "graph_recorder_pool.h"
 #include "../runtime/runtime_core.h"
 #include "../runtime/shared_memory.h"
@@ -373,7 +374,10 @@ struct DefinitionUploads {
 // claimed, so this pass writes the headers, copies in whatever did not fit, and
 // issues a single H2D of the used prefix. The device initial classify then replaces
 // each task's graph_context with an execution constructed in its own heap.
-bool bind_graph_definitions(const HostApi *api, GraphHostState &graph_state, DefinitionUploads *uploads) {
+bool bind_graph_definitions(
+    const HostApi *api, GraphHostState &graph_state, DefinitionUploads *uploads,
+    ReadyQueuePopulations *ready_queue_populations
+) {
     *uploads = DefinitionUploads{};
     const size_t count = graph_host_upload_count(graph_state);
     GraphHostDefinitionList definitions = graph_host_definitions(graph_state);
@@ -384,6 +388,8 @@ bool bind_graph_definitions(const HostApi *api, GraphHostState &graph_state, Def
         size_t object_offset;   // of the object's header, from the block base
         size_t image_bytes;     // the Definition image alone
         const std::byte *copy;  // the image to copy in, or nullptr when built in place
+        ReadyQueuePopulations ready_queue_populations;
+        bool populations_ready{false};
     };
     std::unordered_map<uint64_t, PackedDefinition> packed;
     // Objects the recorders built already occupy the arena's used prefix at the
@@ -393,12 +399,12 @@ bool bind_graph_definitions(const HostApi *api, GraphHostState &graph_state, Def
     for (const GraphHostDefinition &entry : definitions.entries) {
         if (entry.bytes < sizeof(GraphDefinition)) continue;
         if (entry.spill == nullptr) {
-            packed.emplace(entry.full_key, PackedDefinition{entry.object_offset, entry.bytes, nullptr});
+            packed.emplace(entry.full_key, PackedDefinition{entry.object_offset, entry.bytes, nullptr, {}, false});
             continue;
         }
         const size_t object_offset = block_bytes;
         block_bytes += align_up(sizeof(GraphDefinitionHeader) + entry.bytes);
-        packed.emplace(entry.full_key, PackedDefinition{object_offset, entry.bytes, entry.spill});
+        packed.emplace(entry.full_key, PackedDefinition{object_offset, entry.bytes, entry.spill, {}, false});
         uploads->spilled++;
     }
 
@@ -490,6 +496,22 @@ bool bind_graph_definitions(const HostApi *api, GraphHostState &graph_state, Def
             LOG_ERROR("host-orch: Graph runtime storage address is misaligned");
             return false;
         }
+        PackedDefinition &packed_definition = object_it->second;
+        if (!packed_definition.populations_ready) {
+            const GraphNodeDefinition *nodes =
+                graph_definition_array<GraphNodeDefinition>(*definition, definition->off_nodes, definition->task_count);
+            if (nodes == nullptr) {
+                LOG_ERROR("host-orch: invalid Graph Definition node array");
+                return false;
+            }
+            for (uint32_t i = 0; i < definition->task_count; ++i) {
+                packed_definition.ready_queue_populations.add_task(
+                    ActiveMask(nodes[i].active_mask), TaskAttrs(nodes[i].task_attrs), TaskKind::GRAPH_NODE
+                );
+            }
+            packed_definition.populations_ready = true;
+        }
+        ready_queue_populations->add(packed_definition.ready_queue_populations);
         upload->outer_slot->graph_context = reinterpret_cast<GraphDefinition *>(
             reinterpret_cast<uintptr_t>(block) + object_it->second.object_offset + sizeof(GraphDefinitionHeader)
         );
@@ -666,12 +688,26 @@ int32_t run_host_orchestration(
     // After the span closes: the reduction walks a few hundred records and emits
     // five markers, which must not be charged to the bind it measures.
 
+    // total_tasks sizes the bounded per-segment H2D copies below; a value outside
+    // [0, task_capacity] would make those copies read/write out of bounds.
+    if (total_tasks < 0 || static_cast<uint64_t>(total_tasks) > task_capacity) {
+        LOG_ERROR("host-orch: total_tasks %d out of range [0, %" PRIu64 "]", total_tasks, task_capacity);
+        return PTO_RUNTIME_ERR_INTERNAL;
+    }
+
+    ReadyQueuePopulations ready_queue_populations{};
+    SharedMemoryTaskHeader &tasks = host_sm_handle.header->tasks;
+    for (int32_t task_id = 0; task_id < total_tasks; ++task_id) {
+        const ChipTaskSlotState &slot = tasks.get_slot_state_by_task_id(task_id);
+        ready_queue_populations.add_task(slot.active_mask, slot.task_attrs, slot.task_kind);
+    }
+
     // Upload each distinct Definition as its own retained device object and bind
     // every outer Graph task to it. Per-invocation data already lives in that
     // task's payload regions and is copied with the shared-memory image below.
     const int64_t t_graph_ns = bind_phase_begin();
     DefinitionUploads definition_uploads{};
-    if (!bind_graph_definitions(api, *graph_state, &definition_uploads)) {
+    if (!bind_graph_definitions(api, *graph_state, &definition_uploads, &ready_queue_populations)) {
         return PTO_RUNTIME_ERR_INTERNAL;
     }
     {
@@ -689,12 +725,22 @@ int32_t run_host_orchestration(
         record_bind_phase(HostPhaseKind::BindGraphUpload, t_graph_ns, attrs, definition_uploads.bytes);
     }
 
-    // total_tasks sizes the bounded per-segment H2D copies below; a value outside
-    // [0, task_capacity] would make those copies read/write out of bounds.
-    if (total_tasks < 0 || static_cast<uint64_t>(total_tasks) > task_capacity) {
-        LOG_ERROR("host-orch: total_tasks %d out of range [0, %" PRIu64 "]", total_tasks, task_capacity);
-        return PTO_RUNTIME_ERR_INTERNAL;
+    ReadyQueueCapacities ready_queue_capacities{};
+    const int32_t ready_queue_status =
+        derive_ready_queue_capacities(ready_queue_populations, *host_sm_handle.header, &ready_queue_capacities);
+    if (ready_queue_status != 0) {
+        LOG_ERROR(
+            "host-orch: ready queue reachable population exceeds %" PRIu64 " (ready=%" PRIu64 "/%" PRIu64 "/%" PRIu64
+            ", sync=%" PRIu64 "/%" PRIu64 "/%" PRIu64 ", dummy=%" PRIu64 ", graph=%" PRIu64 "/%" PRIu64 ")",
+            READY_QUEUE_CAPACITY_LIMIT, ready_queue_populations.ready[0], ready_queue_populations.ready[1],
+            ready_queue_populations.ready[2], ready_queue_populations.ready_sync[0],
+            ready_queue_populations.ready_sync[1], ready_queue_populations.ready_sync[2], ready_queue_populations.dummy,
+            ready_queue_populations.graph_ready, ready_queue_populations.graph_prepare
+        );
+        LOG_RUNTIME_FAILURE(SIMPLER_ERROR_NONE, SIMPLER_ERROR_READY_QUEUE_OVERFLOW, ready_queue_status);
+        return ready_queue_status;
     }
+    rt->prebuilt_layout.sched.capacities = ready_queue_capacities;
     host_phase_trace_note_submitted(static_cast<uint64_t>(total_tasks));
 
     // The count travels inside the header the restack copies wholesale, which is
--- a/src/a2a3/runtime/host_build_graph/runtime/scheduler/scheduler.h
+++ b/src/a2a3/runtime/host_build_graph/runtime/scheduler/scheduler.h
@@ -438,6 +438,14 @@ void ready_queue_init_data_from_layout(ChipReadyQueue *queue, uint64_t capacity)
 void ready_queue_wire_arena_pointers(ChipReadyQueue *queue, DeviceArena &arena, size_t slots_off);
 void ready_queue_destroy(ChipReadyQueue *queue);
 
+struct ReadyQueueCapacities {
+    uint64_t ready[NUM_RESOURCE_SHAPES]{};
+    uint64_t ready_sync[NUM_RESOURCE_SHAPES]{};
+    uint64_t dummy{0};
+    uint64_t graph_ready{0};
+    uint64_t graph_prepare{0};
+};
+
 /**
  * Statistics returned by mixed-task completion processing
  */
@@ -461,7 +469,7 @@ struct SchedulerLayout {
     size_t off_graph_prepare_queue_slots;
     size_t off_early_dispatch_queue_slots[NUM_RESOURCE_SHAPES];
     size_t off_early_sync_start_queue_slots;
-    uint64_t ready_queue_capacity;
+    ReadyQueueCapacities capacities;
 };
 
 /**
@@ -558,8 +566,8 @@ struct SchedulerState {
             }
         }
         // Every ready / sync / dummy / graph task routes to exactly one queue. A
-        // false push means that queue's peak concurrent occupancy exceeded
-        // CHIP_READY_QUEUE_SIZE — a capacity mis-sizing, not a normal condition.
+        // false push means that queue's peak concurrent occupancy exceeded its
+        // bind-time capacity — a capacity mis-sizing, not a normal condition.
         // Silently dropping the task would stall the run, so latch a named error
         // (surfaces as READY_QUEUE_OVERFLOW rather than an anonymous
         // forward-progress timeout). The graph_ready push is checked identically
--- a/src/a2a3/runtime/host_build_graph/runtime/shared/runtime_init.cpp
+++ b/src/a2a3/runtime/host_build_graph/runtime/shared/runtime_init.cpp
@@ -79,24 +79,30 @@ void SchedulerState::TaskHeaderView::destroy() { tasks = nullptr; }
 
 SchedulerLayout SchedulerState::reserve_layout(DeviceArena &arena) {
     SchedulerLayout layout{};
-    layout.ready_queue_capacity = CHIP_READY_QUEUE_SIZE;
+    for (int i = 0; i < NUM_RESOURCE_SHAPES; ++i) {
+        layout.capacities.ready[i] = READY_QUEUE_CAPACITY_LIMIT;
+        layout.capacities.ready_sync[i] = READY_QUEUE_CAPACITY_LIMIT;
+    }
+    layout.capacities.dummy = READY_QUEUE_CAPACITY_LIMIT;
+    layout.capacities.graph_ready = READY_QUEUE_CAPACITY_LIMIT;
+    layout.capacities.graph_prepare = READY_QUEUE_CAPACITY_LIMIT;
 
-    // Fixed-capacity early-dispatch queues first, then the CHIP_READY_QUEUE_SIZE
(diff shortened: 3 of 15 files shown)


src/a2a3/runtime/host_build_graph/host/runtime_maker.cpp
```diff
@@ -668,12 +707,19 @@ int32_t run_host_orchestration(
...
+    if (!ready_queue_populations.derive_capacities(&ready_queue_capacities)) {
+        LOG_ERROR(
+            "host-orch: ready queue reachable population exceeds %" PRIu64 " (ready=%" PRIu64 "/%" PRIu64 "/%" PRIu64
+            ", sync=%" PRIu64 "/%" PRIu64 "/%" PRIu64 ", dummy=%" PRIu64 ", graph=%" PRIu64 "/%" PRIu64 ")",
+            READY_QUEUE_CAPACITY_LIMIT, ready_queue_populations.ready[0], ready_queue_populations.ready[1],
+            ready_queue_populations.ready[2], ready_queue_populations.ready_sync[0],
+            ready_queue_populations.ready_sync[1], ready_queue_populations.ready_sync[2], ready_queue_populations.dummy,
+            ready_queue_populations.graph_ready, ready_queue_populations.graph_prepare
+        );
+        return runtime_status_from_error_codes(SIMPLER_ERROR_NONE, SIMPLER_ERROR_READY_QUEUE_OVERFLOW);
```
ChaoZheng109: **`READY_QUEUE_OVERFLOW` now has two origins, and the docs only describe one.**

`docs/troubleshooting/device-error-codes.md:111` files 104 under *Scheduler errors (100+)*:

| Code | Name | Whose bug, usually |
| ---- | ---- | ------------------ |
| 104 | READY_QUEUE_OVERFLOW | runtime-internal / config |

This line makes the same code reachable from **host bind**, before any device code runs, with a different cause (the graph exceeds the reservation ceiling) and a different remedy. This PR changes 0 files under `docs/`.

Per `.claude/rules/doc-consistency.md` §1/§4 that update belongs in this commit. Two things worth writing down:

1. Code 104 now has a host-side origin; the triage path differs from the device-side one.
2. The bind-time rejection is a **new user-visible failure mode** and therefore a support contract: *"single-queue reachable population must be <= 32768; larger graphs are rejected at bind."* Right now that contract exists only as a constant and a `LOG_ERROR` format string.

src/a2a3/runtime/host_build_graph/host/runtime_maker.cpp
```diff
@@ -668,12 +707,19 @@ int32_t run_host_orchestration(
...
     }
 
-    // total_tasks sizes the bounded per-segment H2D copies below; a value outside
-    // [0, task_capacity] would make those copies read/write out of bounds.
-    if (total_tasks < 0 || static_cast<uint64_t>(total_tasks) > task_capacity) {
-        LOG_ERROR("host-orch: total_tasks %d out of range [0, %" PRIu64 "]", total_tasks, task_capacity);
-        return PTO_RUNTIME_ERR_INTERNAL;
+    ReadyQueueCapacities ready_queue_capacities{};
+    if (!ready_queue_populations.derive_capacities(&ready_queue_capacities)) {
+        LOG_ERROR(
```
ChaoZheng109, line 732: **This failure does not report through the same path as its sibling.**

The other early return in this function goes through the structured reporter:

```cpp
LOG_RUNTIME_FAILURE(orch_error, SIMPLER_ERROR_NONE, status);
return status;
```

This branch does not call it, and never stores the code into `sm_header->sched_error_code`. So the returned status says `READY_QUEUE_OVERFLOW` while every consumer that reads `sched_error_code` afterwards -- `read_runtime_status`, `LOG_RUNTIME_FAILURE`'s own output, DFX -- sees `SIMPLER_ERROR_NONE`. Whoever triages this cannot reconcile the two, and the mismatch will read like a runtime bug rather than a rejected bind.

To be clear, the `LOG_ERROR` text itself is good: printing the limit plus all nine populations is exactly what a reader needs. The gap is only the structured path.

tests/ut/cpp/common/test_hbg_ready_queue_seed.cpp
```diff
@@ -132,3 +132,48 @@ TEST(HbgReadyQueueSeed, ReseedRecoversAPartiallyUsedRegion) {
...
+    EXPECT_EQ(capacities.graph_prepare, 8);
+}
+
+TEST(HbgReadyQueueSizing, RejectsPopulationPastReservationLimit) {
+    ReadyQueuePopulations populations{};
+    populations.add_task(ActiveMask(SUBTASK_MASK_AIC), TaskAttrs{}, TaskKind::KERNEL, READY_QUEUE_CAPACITY_LIMIT + 1);
+
+    ReadyQueueCapacities capacities{};
+    EXPECT_FALSE(populations.derive_capacities(&capacities));
+}
```
ChaoZheng109, line 176: **Nothing covers the path this PR exists for.**

All three new tests stop at `derive_capacities`. The chain that actually delivers the behaviour is:

```
derive_capacities
  -> rt->prebuilt_layout.sched.capacities
  -> arena_h2d
  -> init_data_from_layout(layout.capacities.ready[i])   // NOT layout.ready_queue_capacity
```

If that last step were changed back to `layout.ready_queue_capacity`, every test in this file still passes, the behaviour silently reverts to pre-PR, and CI stays green. Since the two fields now sit side by side in `SchedulerLayout` with near-identical names (see the comment on `scheduler.h`), that is a realistic edit for someone to make.

A test that builds a `SchedulerLayout` with *distinct* per-queue capacities, runs `init_data_from_layout`, and asserts each queue's resulting `capacity` field would close it.

Two smaller gaps while you are here:
- `add()` saturating merge -- two populations of 20000 each must make `derive_capacities` return false. Only the single-call `add_task` overflow is covered today.
- the bind-time rejection's returned status code.

src/a2a3/runtime/host_build_graph/runtime/scheduler/scheduler.h
```diff
@@ -462,6 +482,7 @@ struct SchedulerLayout {
     size_t off_early_dispatch_queue_slots[NUM_RESOURCE_SHAPES];
     size_t off_early_sync_start_queue_slots;
     uint64_t ready_queue_capacity;
+    ReadyQueueCapacities capacities;
```
ChaoZheng109, line 472: **Two fields one word apart, with unrelated meanings -- and the older one may now be dead.**

`SchedulerLayout` now carries:

- `ready_queue_capacity` -- the arena **reservation** ceiling
- `capacities` -- the per-queue **logical** capacity

A rename of the former (`ready_queue_reservation`?) would help, but it is worth checking whether it needs to exist at all: after this PR `reserve_layout` assigns it `READY_QUEUE_CAPACITY_LIMIT` unconditionally, and its only readers are the `ready_queue_reserve_layout` calls immediately below. It carries no information the constant does not.

Suggest deleting the field and passing `READY_QUEUE_CAPACITY_LIMIT` directly at the reserve sites -- that also removes the confusable pair outright, and removes the field that the missing test above would otherwise have to guard against.

src/a2a3/runtime/host_build_graph/runtime/shared/runtime_init.cpp
```diff
@@ -59,6 +59,77 @@ void ready_queue_destroy(ChipReadyQueue *queue) {
...
+uint64_t capacity_for_population(uint64_t population) {
+    uint64_t capacity = 2;
+    while (capacity < population)
+        capacity <<= 1;
+    return capacity;
+}
+
+}  // namespace
+
+void ReadyQueuePopulations::add_task(ActiveMask active_mask, TaskAttrs task_attrs, TaskKind task_kind, uint64_t count) {
```
ChaoZheng109: **Host-only bind logic, compiled into the AICPU `.so`.**

`ReadyQueuePopulations` runs only at bind: its callers are `run_host_orchestration` and `bind_graph_definitions`, both in `host/runtime_maker.cpp`. But it lands in a translation unit that both targets build:

```python
# build_config.py
"aicpu": { "source_dirs": ["aicpu", "runtime", "orchestration", ...] },
```

and the platform CMakeLists globs with `file(GLOB_RECURSE "${SRC_DIR}/*.cpp")`, so `runtime/shared/runtime_init.cpp` is compiled into the AICPU kernel too, where none of this can ever execute.

The established pattern for exactly this is already in the same runtime -- a header under `runtime/`, the implementation under `host/`, which appears only in the `host` target's `source_dirs`. `host_tensor_access`, `host_phase_trace` and `dep_gen_host_graph` are all built that way.

Suggested split:

- **`ReadyQueueCapacities` stays** in `scheduler.h`. It is a member of `SchedulerLayout`, travels in the arena H2D, and is read by device-side `init_data_from_layout`.
- **`ReadyQueuePopulations`** plus `add_task` / `add` / `derive_capacities` and the two anonymous-namespace helpers move to `runtime/ready_queue_sizing.h` + `host/ready_queue_sizing.cpp`.

`build_config.py` needs no change -- `host/` is already listed. a5 moves in the same commit per `codestyle.md` §10.

Side benefit: on the host side this code is then free to use the STL per `codestyle.md` §8, which matters if the counting is ever replaced by something needing scratch buffers.

tests/ut/cpp/common/test_hbg_ready_queue_seed.cpp
```diff
@@ -132,3 +132,48 @@ TEST(HbgReadyQueueSeed, ReseedRecoversAPartiallyUsedRegion) {
...
     EXPECT_EQ(queue.pop(), fake_slot_state(0));
 }
+
+TEST(HbgReadyQueueSizing, ReservationLimitCoversQwenAicPopulation) {
+    constexpr uint64_t QWEN_AIC_REACHABLE_TASKS = 9400;
+    DeviceArena arena;
+
+    const SchedulerLayout layout = SchedulerState::reserve_layout(arena);
+
+    EXPECT_GE(layout.ready_queue_capacity, QWEN_AIC_REACHABLE_TASKS);
```
ChaoZheng109: **This asserts one constant against another.**

`reserve_layout` sets `layout.ready_queue_capacity = READY_QUEUE_CAPACITY_LIMIT` unconditionally, so the assertion reduces to `EXPECT_GE(32768, 9400)`. The `DeviceArena` and the `reserve_layout` call above contribute nothing to it, which makes it read like a scenario test while exercising no behaviour.

Two ways to make it earn its place, depending on the intent:

- *Pin the supported bound against a known workload* -- `static_assert(READY_QUEUE_CAPACITY_LIMIT >= QWEN_AIC_REACHABLE_TASKS)` states the same fact at zero runtime cost, and reads as the invariant it is.
- *Test `reserve_layout`* -- assert what it computes: that the nine slot regions are reserved at the ceiling, and that the returned offsets are distinct and correctly ordered.

Related: `9400` is a property of one workload's shape. If it stays in the test, it is worth a comment naming where it came from, so the next person knows whether to update it.

src/a2a3/runtime/host_build_graph/runtime/shared/runtime_init.cpp
```diff
@@ -115,14 +193,14 @@ bool SchedulerState::init_data_from_layout(const SchedulerLayout &layout, Device
     }
 
     for (int i = 0; i < NUM_RESOURCE_SHAPES; i++) {
-        ready_queue_init_data_from_layout(&sched->ready_queues[i], layout.ready_queue_capacity);
+        ready_queue_init_data_from_layout(&sched->ready_queues[i], layout.capacities.ready[i]);
```
ChaoZheng109, line 124: **This is where the PR's runtime effect actually lands, and the Performance check does not measure this axis.**

`seed_slots()` loops over `capacity`, not over the reservation:

```cpp
void seed_slots() { for (uint64_t i = 0; i < capacity; i++) { ... } }
```

so a derived capacity changes device-side boot work on every bind. `paged_attention` (measured peak occupancy **1** in #1762) goes from seeding 8192 slots to 2 -- a real win this PR does not claim. qwen's `ready[AIC]` goes from 8192 to 16384 -- a small regression the PR does not surface either.

The Performance check reports qwen **device latency** (39.03 -> 39.18 ms). That is downstream of all of this and would not move in either direction. The axis that does move is host bind + device boot, and the case that shows it is a *small* graph -- which is exactly where #1762 measured its win taking this constant the other way:

> 65536 -> 8192, paged_attention, a2a3 onboard, median of 50 rounds:
> host wall 8627 -> 5037 us (-41.6%), arena_upload 2667 -> 360 us (-86%), arena_build 2623 -> 1490 us (-43%)

Suggest an A/B on the bind path with `paged_attention` included.

Related, for the PR description: raising the ceiling takes the queue reservation from ~1.7 MB to ~7.1 MB (9 x 32768 x 24 B). That number is in #1920 but not in the Summary here, and a reader of "derive every capacity from the tasks that can reach it" would reasonably assume per-demand sizing *reduced* memory rather than quadrupling it. Worth stating outright that the derived capacities shrink seed work while the reservation grows.

ChaoZheng109 (commented):
Review of the implementation, taking the counting method from #1920 as given -- everything below is independent of how the population is computed.

Verified and *not* commented on, since it is all correct: the routing in `add_task` matches `push_ready_routed` exactly, including counting a predicated task toward both `dummy` and its shape queue (the right conservative treatment, since the predicate resolves on device); the per-Definition histogram is walked once and cached by `full_key`; the saturating add uses `LIMIT + 1` as a sentinel so `uint64` cannot wrap; `capacity_for_population(32768)` lands exactly on the reservation; hoisting the `total_tasks` range check ahead of the new scan is necessary rather than incidental; and a2a3/a5 are line-for-line mirrors.

Seven comments below. The first three I would treat as blocking (error-code contract vs. docs, structured error reporting, and no coverage of the derive -> H2D -> init path); the other four are worth resolving before merge.

tests/ut/cpp/common/test_hbg_ready_queue_seed.cpp
TaoZQY: Fixed in 2c92429a — added coverage for distinct per-queue capacities through SchedulerState::init_data_from_layout, 20000+20000 saturating merge rejection, and bind-time returned/shared status.

src/a2a3/runtime/host_build_graph/runtime/shared/runtime_init.cpp
TaoZQY: Fixed in 2c92429a — moved ReadyQueuePopulations and all sizing helpers to runtime/ready_queue_sizing.h plus host/ready_queue_sizing.cpp, mirrored for a2a3 and a5, so AICPU no longer compiles the host-only logic.

src/a2a3/runtime/host_build_graph/runtime/scheduler/scheduler.h
TaoZQY: Fixed in 2c92429a — removed SchedulerLayout::ready_queue_capacity and pass READY_QUEUE_CAPACITY_LIMIT directly to all nine reservation sites.

src/a2a3/runtime/host_build_graph/runtime/shared/runtime_init.cpp
TaoZQY: Fixed in 2c92429a — updated the PR with the 1.7 MB to 7.1 MB reservation trade-off and task-submit paged_attention A/B. Final same-NPU 50-round medians: host wall 783.8 to 554.6 us and device wall 392.9 to 163.8 us; isolated bind control-plane min/median 26/28 to 25/27 us.

tests/ut/cpp/common/test_hbg_ready_queue_seed.cpp
TaoZQY: Fixed in 2c92429a — replaced the runtime constant comparison with a static_assert and documented that 9400 comes from qwen3_14b_decode GraphExecutionBatch16Seq3500.

src/a2a3/runtime/host_build_graph/host/runtime_maker.cpp
TaoZQY: Fixed in 2c92429a — documented both host-bind and device-scheduler origins of error 104, the <=32768 single-queue support contract, and the distinct remedies.

src/a2a3/runtime/host_build_graph/host/runtime_maker.cpp
TaoZQY: Fixed in 2c92429a — bind rejection now stores sched_error_code, returns the same -104 status, and goes through LOG_RUNTIME_FAILURE; the structured state is covered by a unit test.

ChaoZheng109 (approved):

PR #2010, opened 2026-08-25
CI: improve macOS packaging cache reuse

doraemonmj:
## Summary

- resolve the platform-default `gcc`/`g++` paths once when neither `CC` nor `CXX` is supplied
- preserve complete and partial caller compiler overrides
- keep all five packaging modes on one compiler identity so nested runtime objects retain stable ccache keys
- add a focused workflow contract test for default and caller-supplied compiler selection

## Why

The packaging matrix invokes the same builds through pip and direct CMake modes. On macOS, the unversioned `gcc`/`g++` commands are Apple Clang drivers, but different textual compiler identities prevented ccache from recognizing equivalent nested runtime compilations. Resolving one absolute compiler pair before the matrix keeps the platform-default toolchain while making those cache keys stable.

The code intentionally uses the unversioned platform default rather than the separately installed versioned Homebrew GCC. Explicit `CC`/`CXX` selections remain authoritative.

## Validation

| Runner | Comparable baseline | Clean-head result | Difference |
| --- | ---: | ---: | ---: |
| Ubuntu | 218s | 222s | +4s, effectively flat |
| macOS | 422s | 261s | -161s (-38%) |

- full pre-commit suite passed
- focused compiler-identity contract test passed
- all five packaging modes remain enabled, including the full direct-CMake target
- clean-head jobs: [Ubuntu](https://github.com/doraemonmj/simpler_wc/actions/runs/32802099807/job/97664932411), [macOS](https://github.com/doraemonmj/simpler_wc/actions/runs/32802099807/job/97664932351)
- benchmark history and rejected experiments: [personal-fork PR #15](https://github.com/doraemonmj/simpler_wc/pull/15)

## Compatibility

This does not pin a compiler version across runner-image updates. It pins the resolved compiler paths only for the duration of one packaging job, which is the scope needed for cross-mode ccache reuse.

### DIFF
--- a/.github/workflows/_packaging.yml
+++ b/.github/workflows/_packaging.yml
@@ -91,6 +91,20 @@ jobs:
       - name: Run packaging matrix
         run: |
           . .venv/bin/activate
+          if [[ -z "${CC:-}" && -z "${CXX:-}" ]]; then
+            # Use the unversioned platform-default compilers intentionally (Apple
+            # Clang on current macOS runners), not versioned Homebrew GCC. Absolute
+            # paths keep the compiler identity stable across packaging modes.
+            if CC="$(command -v gcc)" && CXX="$(command -v g++)"; then
+              export CC CXX
+              echo "Resolved platform compilers: CC=$CC CXX=$CXX"
+              "$CC" --version
+              "$CXX" --version
+            else
+              echo "::error::gcc/g++ not found on PATH; runner provisioning is incomplete"
+              exit 1
+            fi
+          fi
           if [[ "${{ runner.os }}" == "macOS" ]]; then
             export MACOSX_DEPLOYMENT_TARGET
             MACOSX_DEPLOYMENT_TARGET="$(sw_vers -productVersion | cut -d. -f1,2)"
--- a/tests/ut/py/test_packaging_compiler_identity.py
+++ b/tests/ut/py/test_packaging_compiler_identity.py
@@ -1,0 +1,91 @@
+# Copyright (c) PyPTO Contributors.
+# This program is free software, you can redistribute it and/or modify it under the terms and conditions of
+# CANN Open Software License Agreement Version 2.0 (the "License").
+# Please refer to the License for details. You may not use this file except in compliance with the License.
+# THIS SOFTWARE IS PROVIDED ON AN "AS IS" BASIS, WITHOUT WARRANTIES OF ANY KIND, EITHER EXPRESS OR IMPLIED,
+# INCLUDING BUT NOT LIMITED TO NON-INFRINGEMENT, MERCHANTABILITY, OR FITNESS FOR A PARTICULAR PURPOSE.
+# See LICENSE in the root of the software repository for the full text of the License.
+# -----------------------------------------------------------------------------------------------------------
+
+import os
+import subprocess
+from pathlib import Path
+from typing import Any
+
+import yaml
+
+PROJECT_ROOT = Path(__file__).resolve().parents[3]
+VENV_ACTIVATE = ". .venv/bin/activate\n"
+MACOS_GUARD = 'if [[ "${{ runner.os }}" == "macOS" ]]'
+
+
+def _packaging_script() -> str:
+    workflow: dict[str, Any] = yaml.safe_load((PROJECT_ROOT / ".github" / "workflows" / "_packaging.yml").read_text())
+    matches = [
+        step["run"]
+        for job in workflow["jobs"].values()
+        for step in job["steps"]
+        if step.get("name") == "Run packaging matrix"
+    ]
+    assert len(matches) == 1
+    return matches[0]
+
+
+def _compiler_setup() -> str:
+    script = _packaging_script()
+    _, activate, remainder = script.partition(VENV_ACTIVATE)
+    assert activate, "packaging step no longer has the expected venv activation"
+    setup, guard, _ = remainder.partition(MACOS_GUARD)
+    assert guard, "packaging step no longer has the expected compiler setup before the macOS guard"
+    return setup
+
+
+def _run_setup(script: str, env: dict[str, str], *, check: bool = True) -> subprocess.CompletedProcess[str]:
+    command = f'{script}\nprintf "%s\\n%s\\n" "${{CC-unset}}" "${{CXX-unset}}"'
+    return subprocess.run(
+        ["/bin/bash", "-euo", "pipefail", "-c", command],
+        check=check,
+        capture_output=True,
+        env=env,
+        text=True,
+    )
+
+
+def test_packaging_modes_share_one_compiler_identity_without_overriding_callers(tmp_path: Path):
+    bin_dir = tmp_path / "bin"
+    bin_dir.mkdir()
+    for compiler in ("gcc", "g++"):
+        path = bin_dir / compiler
+        path.write_text('#!/bin/sh\nprintf "%s %s\\n" "$0" "$*"\n')
+        path.chmod(0o755)
+
+    env = os.environ.copy()
+    env.pop("CC", None)
+    env.pop("CXX", None)
+    env["PATH"] = f"{bin_dir}:{env['PATH']}"
+    script = _compiler_setup()
+
+    result = _run_setup(script, env)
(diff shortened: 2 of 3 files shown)


.github/workflows/_packaging.yml
```diff
@@ -91,6 +91,14 @@ jobs:
       - name: Run packaging matrix
         run: |
           . .venv/bin/activate
+          if [[ -z "${CC:-}" && -z "${CXX:-}" ]]; then
+            # Use the unversioned platform-default compilers intentionally (Apple
+            # Clang on current macOS runners), not versioned Homebrew GCC. Absolute
+            # paths keep the compiler identity stable across packaging modes.
+            CC="$(command -v gcc)"
+            CXX="$(command -v g++)"
```
ChaoZheng109: Under `bash -e` (the Actions default shell) a failing `command -v` aborts the step with a bare `exit 1` and no diagnostic. Confirmed locally:

```
$ env -u CC -u CXX PATH=<dir with bash but no gcc> /bin/bash -e -o pipefail -c '<this snippet>'
exit=1      # and nothing at all on stdout/stderr
```

That matters for the self-hosted lane. `ci-self-cpu.yml:91` calls this same workflow with `setup_variant: self-cpu`, and the `Set up C++ compiler` step is gated on `inputs.setup_variant == 'github'` — so on that runner nothing guarantees an unversioned `gcc`/`g++` on PATH. Before this change CMake searched `cc` / `gcc` / `clang` itself and the job proceeded; now it hard-fails with nothing in the log pointing at the cause.

Low probability (an aarch64 Linux box almost certainly has `gcc`), but the failure mode is undiagnosable, which is what makes it worth a line. Either fail loudly, or fall back to the old behaviour:

```bash
if [[ -z "${CC:-}" && -z "${CXX:-}" ]]; then
  if CC="$(command -v gcc)" && CXX="$(command -v g++)"; then
    export CC CXX
  else
    echo "::notice::unversioned gcc/g++ not on PATH; leaving compiler selection to CMake"
    unset CC CXX
  fi
fi
```

tests/ut/py/test_packaging_compiler_identity.py
```diff
@@ -0,0 +1,70 @@
...
+        for step in job["steps"]
+        if step.get("name") == "Run packaging matrix"
+    ]
+    assert len(matches) == 1
+    return matches[0]
+
+
+def _compiler_setup() -> str:
+    script = _packaging_script()
+    return script.split(". .venv/bin/activate\n", 1)[1].split('if [[ "${{ runner.os }}" == "macOS" ]]', 1)[0]
```
ChaoZheng109: `str.split(sep, 1)[0]` returns the *whole* string when the separator is absent. So if the macOS block is ever reordered above the compiler block, `_compiler_setup()` silently returns the entire step body — which `_run_setup` then hands to `subprocess.run(..., check=True)`, `bash tools/verify_packaging.sh` included.

In practice bash dies on `${{ runner.os }}` (bad substitution) before reaching that line, so the 5-mode build won't actually launch. But the test then fails with an opaque bash error instead of "the packaging step no longer has the shape this test assumes", which is the thing you'd want it to say.

Cheap to make explicit:

```python
ACTIVATE = ". .venv/bin/activate\n"
MACOS_GUARD = 'if [[ "${{ runner.os }}" == "macOS" ]]'


def _compiler_setup() -> str:
    script = _packaging_script()
    assert ACTIVATE in script and MACOS_GUARD in script, "packaging step no longer has the expected shape"
    return script.split(ACTIVATE, 1)[1].split(MACOS_GUARD, 1)[0]
```

.github/workflows/_packaging.yml
```diff
@@ -91,6 +91,14 @@ jobs:
       - name: Run packaging matrix
         run: |
           . .venv/bin/activate
+          if [[ -z "${CC:-}" && -z "${CXX:-}" ]]; then
+            # Use the unversioned platform-default compilers intentionally (Apple
+            # Clang on current macOS runners), not versioned Homebrew GCC. Absolute
```
ChaoZheng109, line 96: This comment makes the Homebrew GCC in the `Set up C++ compiler` step provably unused for this job, so the obvious follow-up is to delete it. I checked before suggesting that, and the evidence cuts both ways — hence your call, not a request.

Confirming it really is unused: `Gxx15Toolchain` (`g++-15`) is only instantiated at `kernel_compiler.py:142`, for sim **kernel** incore compilation. `verify_packaging.sh` builds runtimes and then only imports and runs `--help` — it never compiles a kernel. Runtime builds go through `GxxToolchain(prefer_g15=bool(self._sanitizers))` (`runtime_compiler.py:199,210,237,248`), and packaging runs no sanitizer, so `prefer_g15=False` → plain `g++`. Nothing in this job reaches `g++-15`, before or after your change.

But removing `brew install gcc@15 || brew install gcc` buys approximately nothing. On this PR's macOS run, `Set up C++ compiler` took **6s** (11:02:13 → 11:02:19), i.e. `gcc-15` and `ccache` are already preinstalled on the runner image and both `brew install`s are skipped by their own guards. So this is a readability trade-off, not a time one: keeping it costs 0s but leaves an install the job's own comment says it deliberately does not use; dropping it removes a fallback that protects nothing here, since the job now explicitly selects `/usr/bin/gcc`. Fine as a follow-up, or not at all.

One thing worth confirming while you're in here, because the comment asserts it: on the macOS runner does `command -v gcc` actually resolve to `/usr/bin/gcc` (Apple Clang) rather than a Homebrew shim? Homebrew's gcc formula normally installs only versioned `gcc-15` and no unversioned `gcc`, so the comment should be correct — but I could not read it out of the job log, and if Homebrew's bin ever precedes `/usr/bin` with an unversioned `gcc`, this line would quietly select real GCC to build the macOS extension modules.

ChaoZheng109 (commented):
Three inline notes. The mechanism and the win both hold up — for what it's worth I checked the gain against upstream rather than the fork: macOS packaging on other recent PRs runs 8m33s / 8m45s / 8m48s, this PR's run is 6m29s (~-26%), and Ubuntu came out at 3m33s against 4m21s / 4m31s / 4m48s, so it is not just flat there either.

Only the first two are suggestions. The third is a decision for you, and the numbers I found argue against the cleanup it looks like it implies, so please read that one before acting on it.

.github/workflows/_packaging.yml
ChaoZheng109: Correcting my own comment above — the premise is wrong, and the suggestion is weaker than I made it sound.

`ci-self-cpu.yml` carries a provisioning contract for the cpu runner in its header, which I missed by reading only the job body:

> Provisioning contract for the cpu runner (out-of-repo, dnf): every cpu agent needs the same set — cmake ninja-build **gcc-c++** clang-tools-extra graphviz gtest-devel python3-devel

`gcc-c++` pulls in `gcc` on RHEL/HCE, so `/usr/bin/gcc` and `/usr/bin/g++` are both contractually present on that runner. My "nothing guarantees an unversioned `gcc`/`g++` on PATH" is false — the workflow doesn't guarantee it, but the out-of-repo provisioning does, and it says so in the same file I was citing.

What survives is only the failure shape: `set -e` plus a failing `command -v` still exits the step with no output whatsoever, so if provisioning ever drifts the log gives you nothing to work from. That makes the guard a cheap robustness measure rather than the near-miss I framed it as, and it's entirely reasonable to skip it here.

And to be explicit, since it's worth not leaving ambiguous: none of this touches macOS. macOS only runs through `ci.yml`'s `packaging-matrix` with `setup_variant: github`, where `Set up C++ compiler` does execute (6s on this PR), and `ci-self-cpu.yml` notes that T2 (macOS) is intentionally absent from that lane.

.github/workflows/_packaging.yml
doraemonmj: Addressed the verification side of this. The default-selection path now logs the resolved CC/CXX paths and runs both compilers with --version, with contract-test coverage for those diagnostics. I am keeping the Homebrew GCC setup unchanged: as your timing shows, removing it has no meaningful job-time benefit, so I would rather leave that cleanup out of this PR. The new macOS packaging run is queued; its log will confirm whether the defaults resolve to /usr/bin/gcc and /usr/bin/g++ as expected.

ChaoZheng109 (approved):

PR #2041, opened 2026-08-27
Refactor: keep Qwen decode parameters device-resident

doraemonmj:
## Summary

- move the A2/A3 and A5 Qwen3-14B TMR/HBG cases to standalone L2 Worker drivers with thin pytest wrappers
- allocate all 20 entry tensors once, stream the 19 valid inputs to device once, and reuse the same buffers and TaskArgs across rounds
- keep correctness explicit by copying back only out, k_cache, and v_cache after the measured rounds; --skip-golden still uploads a valid fixture but performs no D2H
- add compile-only CI warm-up, lifecycle unit coverage, and documentation for the standalone/device-resident behavior

## Memory and lifecycle

The fixture generator yields one parameter at a time. The upload loop drops its host reference immediately after synchronous copy_to, before requesting the next parameter, so adjacent large weights do not overlap on host. Device buffers remain owned by the Worker until Worker.close().

## Validation

- all pre-commit hooks passed, including Ruff, Pyright, YAML, and Markdownlint
- 17 focused Python unit tests passed
- 4 manual Qwen pytest wrappers collected
- A2/A3 TMR: 2-round --skip-golden device smoke passed
- A2/A3 HBG: 2-round --skip-golden device smoke passed
- A2/A3 HBG full torch correctness passed for out and the complete K/V caches

A5 received the same driver and static coverage; local hardware validation was limited to A2/A3.


examples/a2a3/tensormap_and_ringbuffer/qwen3_14b_decode/main.py
```diff
@@ -0,0 +1,752 @@
...
+
+def _build_task_args(buffers: dict[str, Buffer], n_layers: int, signature: list) -> TaskArgs:
+    specs = param_specs(n_layers)
+    if len(specs) != len(signature):
+        raise ValueError(f"Qwen entry has {len(specs)} tensors but its orchestration signature has {len(signature)}")
+    args = TaskArgs()
+    for spec, direction in zip(specs, signature, strict=True):
+        tag = _DIRECTION_TAGS[direction]
+        args.add_tensor(buffers[spec.name].tensor(spec.shape, getattr(DataType, spec.dtype)), tag)
+    return args
```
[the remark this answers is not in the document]
doraemonmj: Fixed in 6f0e0d99. Both Qwen drivers now use ordinary zip(specs, signature); the explicit equal-length validation remains immediately before iteration, preserving the mismatch guard on Python 3.9.

examples/a2a3/host_build_graph/qwen3_14b_decode/README.md
```diff
@@ -38,12 +40,14 @@ configuration.
...
 
 ```bash
-.claude/skills/onboard-arch-precheck/check.sh a2a3 || exit 1
-  ".venv/bin/python -m pytest \
-  examples/a2a3/host_build_graph/qwen3_14b_decode \
-  --platform a2a3 --device \$TASK_DEVICE --manual include"
+  '.claude/skills/onboard-arch-precheck/check.sh a2a3 && \
+   .venv/bin/python examples/a2a3/host_build_graph/qwen3_14b_decode/main.py \
+   -p a2a3 -d "$TASK_DEVICE" --rounds 100 --skip-golden'
```
ChaoZheng109: **This block cannot run the benchmark, and it also locks a device before the arch gate.** Two separate defects.

### 1. Single quotes make the command multi-line, and `task-submit`'s task file is line-based

Inside `'...'` bash does not process `\`+newline, so the string handed to `--run` keeps its literal newlines. The previous `"..."` form collapsed to one line — that difference is the regression:

```console
$ printf '%s' 'AAA && \
   BBB' | cat -A
AAA && \$
   BBB
$ printf '%s' "AAA \
   BBB" | cat -A
AAA    BBB
```

`task-submit` writes the value as `COMMAND=$cmd` into a `key=value` task file (`task-submit:718-728`), and `task-daemon:81-93` reads it back with `while IFS='=' read -r key value`. Replaying that write+parse with this exact command:

```text
COMMAND=[.claude/skills/onboard-arch-precheck/check.sh a2a3 && \]
DEVICE=[3]
MAX_TIME=[1800]
```

The command is truncated at the first newline and the two python lines are silently dropped (they match no key and fall through the `case`). `DEVICE` / `MAX_TIME` come after, so **the task is still scheduled and still locks a card**. On a matching a2a3 box the remainder behaves like this:

```console
$ bash -c 'true && \'
bash: \: command not found
exit=127
```

So: precheck passes, `\` is then executed as a command, the task exits 127 with an error that points nowhere near the cause, and **the qwen benchmark never runs**. On a mismatched box `&&` short-circuits and it looks like the gate worked — but only because the gate is all that survived the truncation.

### 2. The precheck now runs after the device lock

`--device auto` queues, waits for a free card and takes `npu-lock` before `COMMAND` executes. `check.sh` needs no device at all (it reads `npu-smi info -t board` plus CANN's `platform_config/*.ini`) and costs ~600 ms cold, ~5 ms cached. Gating inside `--run` means paying the full queue wait and occupying a card on a shared box to reach a refusal that could have been instant.

This contradicts the skill it invokes — `onboard-arch-precheck`'s frontmatter says "refuse mismatched `--platform` onboard hardware test invocations **BEFORE any device is locked**", and its own How-to-use section shows the inverse of this block:

```bash
# Before task-submit (gate BEFORE the device lock):
.claude/skills/onboard-arch-precheck/check.sh "$ARCH" || exit 1
task-submit --device auto --device-num 1 --run "pytest ... --platform $ARCH ..."
```

`.claude/rules/running-onboard.md` (Pre-flight) states the same rule. Secondary effects: a refusal now registers as a failed task in `task-submit --list` rather than a clean pre-flight exit, and `check.sh`'s distinct `0` / `1` (cannot detect) / `2` (mismatch) exit codes get flattened into one non-zero.

### Suggested fix

```bash
.claude/skills/onboard-arch-precheck/check.sh a2a3 || exit 1
  ".venv/bin/python examples/a2a3/host_build_graph/qwen3_14b_decode/main.py \
   -p a2a3 -d \$TASK_DEVICE --rounds 100 --skip-golden"
```

Double quotes so the continuations collapse to a single line, and `\$TASK_DEVICE` so the outer shell does not expand it — `task-daemon:529` keys on the literal `$TASK_DEVICE` in `COMMAND` to decide not to append its own `--device`.

`examples/a5/host_build_graph/qwen3_14b_decode/README.md` carries the same block and needs the same fix.

examples/a5/host_build_graph/qwen3_14b_decode/README.md
```diff
@@ -17,12 +18,14 @@ temporary memory stays constant as the layer count grows.
...
 
 ```bash
-.claude/skills/onboard-arch-precheck/check.sh a5 || exit 1
-  ".venv/bin/python -m pytest \
-  examples/a5/host_build_graph/qwen3_14b_decode \
-  --platform a5 --device \$TASK_DEVICE --manual include"
+  '.claude/skills/onboard-arch-precheck/check.sh a5 && \
+   .venv/bin/python examples/a5/host_build_graph/qwen3_14b_decode/main.py \
+   -p a5 -d "$TASK_DEVICE" --rounds 100 --skip-golden'
```
ChaoZheng109: Same two defects as `examples/a2a3/host_build_graph/qwen3_14b_decode/README.md:43-46` — see that comment for the evidence.

1. The single-quoted string keeps literal newlines, `task-submit` stores it as `COMMAND=` in a line-based task file and `task-daemon` truncates it at the first newline, so only `check.sh a5 && \` is executed. The task still locks a card (`DEVICE=` parses fine), then exits 127 on `\: command not found`, and the benchmark never runs.
2. The precheck runs after the device lock instead of before it.

```bash
.claude/skills/onboard-arch-precheck/check.sh a5 || exit 1
  ".venv/bin/python examples/a5/host_build_graph/qwen3_14b_decode/main.py \
   -p a5 -d \$TASK_DEVICE --rounds 100 --skip-golden"
```

ChaoZheng109 (commented):
## Follow-up: the two HBG README run recipes are broken

Both `host_build_graph` READMEs moved `onboard-arch-precheck` inside `task-submit --run` and switched the outer quoting from `"` to `'` in the same edit. I checked this against `task-submit` / `task-daemon` on the dev box and reproduced the parse, and the result is worse than a style problem: **as written, neither command ever runs the benchmark.**

- The `'...'` form keeps literal `\`+newline, so `--run` receives a multi-line string. `task-submit` stores it as `COMMAND=` in a line-oriented `key=value` task file and `task-daemon` parses that with `while IFS='=' read -r key value`, truncating the value at the first newline. `DEVICE=` / `MAX_TIME=` still parse, so the task is scheduled and **a card is locked**, but only `check.sh <arch> && \` survives — which exits 127 (`\: command not found`) on a matching box.
- Independently, gating inside `--run` acquires the device lock before a check that needs no device, contradicting `onboard-arch-precheck`'s own documented usage and `.claude/rules/running-onboard.md`.

Details and a suggested fix inline. This is new in this PR and was not raised in the earlier review round; everything else from that round (the `SKILL.md` command, the benchmark-rounds assertion, the arg-tag duplication, the uninitialized `out` invariant, and the dropped swimlane diagnostics) I re-checked against the current head and all five are addressed.

examples/a2a3/host_build_graph/qwen3_14b_decode/README.md
ChaoZheng109: One-click version of the fix — verified end to end (outer shell → `task-submit` task file → `task-daemon` parse → exec): the command stays on one line, so `COMMAND=` is no longer truncated, and the literal `$TASK_DEVICE` survives for `task-daemon:529`.

```suggestion
.claude/skills/onboard-arch-precheck/check.sh a2a3 || exit 1
task-submit --device auto --device-num 1 --run \
  ".venv/bin/python examples/a2a3/host_build_graph/qwen3_14b_decode/main.py \
   -p a2a3 -d \$TASK_DEVICE --rounds 100 --skip-golden"
```

examples/a5/host_build_graph/qwen3_14b_decode/README.md
ChaoZheng109: Same one-click fix for a5:

```suggestion
.claude/skills/onboard-arch-precheck/check.sh a5 || exit 1
task-submit --device auto --device-num 1 --run \
  ".venv/bin/python examples/a5/host_build_graph/qwen3_14b_decode/main.py \
   -p a5 -d \$TASK_DEVICE --rounds 100 --skip-golden"
```

examples/a2a3/host_build_graph/qwen3_14b_decode/README.md
doraemonmj: Fixed the multiline `COMMAND=` truncation in a8242c1e. The `--run` payload now uses double quotes, so the backslash-newline pairs collapse before submission, and `\$TASK_DEVICE` remains literal until the allocated job runs.

I intentionally kept `onboard-arch-precheck` inside the allocated job. This shared host’s top-level policy requires every npu-smi/DCMI query—including this precheck—to run only after acquiring a device allocation, and explicitly overrides repository guidance that places it before `task-submit`. Thus the quoting defect is fixed, while the ordering remains host-compliant.

examples/a5/host_build_graph/qwen3_14b_decode/README.md
doraemonmj: Fixed the same multiline parsing defect in a8242c1e: outer double quotes now collapse the line continuations, and escaped `\$TASK_DEVICE` expands only inside the allocated job. The precheck remains inside the allocation for the shared-host DCMI policy explained in the A2/A3 thread.

ChaoZheng109 (approved):

PR #2057, opened 2026-08-28
Fix: preserve explicit dependency flags in dep_gen replay

### LINKED ISSUE #1827: [Code Health] dep_gen: DepGenRecord carries no per-dep DepFlags, so explicit add_dep_wait edges are mis-recorded in deps.json
### Category

Observability / DFX accuracy (tech debt)

### Component

`dep_gen` — `DepGenRecord` capture + `dep_gen_replay` (tensormap_and_ringbuffer)

### Description

PR #1806 (part of #1375) added orthogonal `DepFlags { WAIT, RETAIN }` per fanin edge and
threaded them into `deps.json` and the dep_gen differential gate. Creator and tensormap
(modifier) edge flags are re-derivable at replay time (`arg_types` + `owner_task_id` +
tensormap), so those are recorded accurately.

**Explicit-dep flags are not.** `DepGenRecord` records `explicit_deps[]` as bare
`PTO2TaskId::raw` values with no per-dep `DepFlags`. So the replay hard-codes every explicit
edge to `DEP_WAIT | DEP_RETAIN`. Consequences:

- A dep added via the ordering-only `CoreTaskArgsWithDeps::add_dep_wait()` API (a genuine
  `["wait"]` edge at runtime) is recorded in `deps.json` as `["wait","retain"]`.
- The dual-pass differential gate cannot catch an explicit-kind divergence, because both the
  oracle and the annotated pass read the same hard-coded constant.

This is functionally harmless (runtime behaviour is correct; only the DFX artifact is
imprecise) and is currently documented as a known replay limitation in `docs/dfx/dep-gen.md`,
but it means `deps.json` cannot be trusted for the WAIT/RETAIN classification of explicit
edges.

Related, and better handled inside the follow-up reduction PR rather than here: `dep_gen`
reconstructs edges by re-running `compute_task_fanin` (construction only), so it does not model
the planned device-side transitive reduction. Once reduction lands, `deps.json` flags will be
the pre-reduction values unless the replay also models the reduction. (Tracking that with the
reduction work, noted here for context.)

### Location

- `src/common/platform/include/common/dep_gen.h` — `struct DepGenRecord` (has
  `explicit_deps[]`, no kinds; `static_assert(sizeof(DepGenRecord) == 4672)` locks the wire
  layout)
- `src/{a2a3,a5}/runtime/tensormap_and_ringbuffer/host/dep_gen_replay.cpp` — hard-codes
  explicit edges to `DEP_WAIT | DEP_RETAIN`
- `src/common/platform/shared/host/dep_gen_collector.cpp` — capture path
- `docs/dfx/dep-gen.md` — documents the limitation

### Proposed Fix

Add `explicit_dep_kinds[DEP_GEN_MAX_EXPLICIT_DEPS]` (one `uint8_t` `DepFlags` per explicit dep)
to `DepGenRecord` (and the overflow-record chain), populate it at capture, read it in
`dep_gen_replay` instead of the constant, and make the differential gate explicit-kind-aware.
Update the `sizeof(DepGenRecord)` static_assert + `docs/dfx/dep-gen.md`, and drop the
"explicit edges always recorded as wait+retain" caveat. Priority is low — only the DFX
artifact accuracy for `add_dep_wait` edges is affected, not runtime correctness.

zmnobug:
## Summary

- Capture per-dependency WAIT/RETAIN flags in inline and overflow dep_gen records.
- Replay explicit edges with their recorded flags while preserving runtime OR-dedup semantics with creator dependencies, and reject records containing unknown flag bits.
- Preserve DFX buffer ownership after a ready entry is published when the push gate times out. `DeviceProfilerEngine` is shared by PMU, scope stats, L2 phase/task profiling, dep_gen, and args dump; this independent correctness fix was uncovered while investigating #1827.
- Cover inline, single-overflow, multi-overflow, wait-only, and explicit/creator dedup cases on A2/A3 and A5, and update the DFX documentation.

## Differential Gate Scope

Explicit dependencies are captured input shared by the oracle and annotated replay passes, so the differential gate cannot independently validate their recorded kind bytes. The A2/A3 and A5 dep_gen chain scene tests validate the capture/replay round trip across inline and overflow records.

`DepGenRecord` changes from 4672 to 4736 bytes. Host and device runtime artifacts must be rebuilt together rather than mixed with partial incremental outputs.

## Testing

- C++ unit tests: 125/125 passed, including `test_dep_gen_replay`, `test_dep_gen_collector_aicpu`, and `test_profiler_device_engine`.
- A2/A3 simulation dep_gen scene tests with manual cases: 2 passed.
- A5 simulation dep_gen scene tests with manual cases: 2 passed.
- A2/A3 onboard dep_gen scene tests with manual cases: 2 passed.
- `pre-commit run --from-ref upstream/main --to-ref HEAD`.
- A5 onboard was not run locally because this host has A2/A3 silicon; PR CI covers A5 hardware.

Fixes #1827

### DIFF
--- a/docs/dfx/dep-gen.md
+++ b/docs/dfx/dep-gen.md
@@ -73,6 +73,10 @@ inputs to each submit are captured and the graph is reconstructed afterwards.
   This is the guarantee against silent shotgun modifications — anyone
   who changes `compute_task_fanin` semantics will trip the gate
   immediately and know to update the annotated mirror.
+  Explicit deps are a common captured input to both passes, so this gate
+  cannot independently validate their recorded kind bytes. The a2a3/a5
+  dep_gen chain scene tests validate that capture/replay round-trip across
+  inline and overflow records.
 - **Output.** `<output_prefix>/deps.json` — strided-Tensor schema with
   `tasks[]`, `tensors[]`, and tensor-annotated `edges[]` (see §4).
 
@@ -218,7 +222,7 @@ Each edge is `{pred, succ}` plus annotation. Fields:
 | `pred`, `succ` | uint64 (string) | always | `TaskId::raw` of producer and consumer |
 | `arg` | int32 | always | Consumer's arg-slot index; `-1` for `explicit` source |
 | `source` | string | always | `explicit` (from `explicit_deps[]`), `creator` (`owner_task_id` retention), or `tensormap` (overlap lookup hit) |
-| `flags` | string array | always | Subset of `["wait", "retain"]` — the edge's `DepFlags`. `wait` = ordering (readiness); `retain` = producer lifetime held until the consumer releases. `creator` edges are `["wait","retain"]`; `tensormap` edges `["wait"]`. `explicit` edges are **always recorded as `["wait","retain"]`**: the `DepGenRecord` does not carry per-dep kinds, so a replayed explicit dep cannot distinguish the ordering-only `CoreTaskArgsWithDeps::add_dep_wait()` API from the default. The differential gate is unaffected (both passes read the same constant). At runtime an `add_dep_wait()` edge is genuinely `["wait"]`; that distinction is a known replay limitation, not written to `deps.json`. |
+| `flags` | string array | `tensormap_and_ringbuffer` | Subset of `["wait", "retain"]` — the edge's `DepFlags`. `wait` = ordering (readiness); `retain` = producer lifetime held until the consumer releases. `creator` edges are `["wait","retain"]`; `tensormap` edges `["wait"]`. `explicit` edges start with the per-dependency kind captured at submit time, so `CoreTaskArgsWithDeps::add_dep_wait()` emits `["wait"]` while the default dependency kind emits `["wait","retain"]`. When the same producer is also the creator of an input, replay matches runtime fanin dedup by OR-accumulating the creator's flags into the explicit edge. The host_build_graph writer does not currently emit this field. |
 | `overlap` | string | `source=tensormap` | `covered` (producer slice fully contains consumer slice) or `other` |
 | `tensor_id` | uint64 (string) | not `explicit` | Identity of the underlying tensor; cross-references `tensors[]` |
 | `consumer_dtype` | string | not `explicit` | Element type the consumer reads as |
@@ -365,13 +369,14 @@ slots that overlay normal record slots in the buffer.
 | Submit `explicit_dep_count` | Chain shape | Per-submit slots used |
 | --------------------------- | ----------- | --------------------- |
 | `0 ≤ dc ≤ 64` | base only — fast path, no chain bookkeeping | 1 |
-| `65 ≤ dc ≤ 646` | base (64 deps) + 1 overflow (up to 582 deps) | 2 |
-| `647 ≤ dc ≤ 1228` | base + 2 overflow | 3 |
-| general | base + `⌈(dc − 64) / 582⌉` overflow | `1 + ⌈(dc − 64) / 582⌉` |
+| `65 ≤ dc ≤ 588` | base (64 deps) + 1 overflow (up to 524 deps) | 2 |
+| `589 ≤ dc ≤ 1112` | base + 2 overflow | 3 |
+| general | base + `⌈(dc − 64) / 524⌉` overflow | `1 + ⌈(dc − 64) / 524⌉` |
 
-**Wire format.** An overflow record reinterprets the same 4672-byte slot
-as `{ task_id (8) + flags (4) + dep_count (2) + _reserved (2) + deps[582] }` —
-no tensor blobs (those live on the base record). Replay distinguishes
+**Wire format.** An overflow record reinterprets the same 4736-byte slot
+as `{ task_id (8) + flags (4) + dep_count (2) + _reserved (2) +
+deps[524] + kinds[524] }`; `kinds[]` is a byte-for-byte parallel array of
+`DepFlags`. There are no tensor blobs (those live on the base record). Replay distinguishes
 the two views by `flags & DEP_GEN_FLAG_OVERFLOW`. Chain records always
 share the base's `task_id`, and the last one sets
 `DEP_GEN_FLAG_LAST_OVERFLOW` so replay knows the chain is complete
@@ -387,7 +392,7 @@ base via `task_id`.
 
 **Truncation tail.** A submit whose chain exceeds the buffer's slot
 budget (`PLATFORM_DEP_GEN_RECORDS_PER_BUFFER = 1024` slots → roughly
-`64 + 1023 × 582 = 595450` deps max in the best case) is logged via
+`64 + 1023 × 524 = 536116` deps max in the best case) is logged via
 `LOG_ERROR` and truncated to the largest dc that fits. Runtime
 correctness is unaffected — `CoreTaskArgs::set_dependencies` keeps the full dep
 list; only the dep_gen replay graph loses the tail.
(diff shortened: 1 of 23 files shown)


doraemonmj (changes requested):
Reviewed against merge-base `89ba2a3f`. Verified the wire-layout arithmetic by hand, diffed the a2a3 and a5 trees against each other for parity, and traced each item back to #1827. Two blocking items, both small; no technical objection to the approach.

## What this gets right

Not just "it works" — a few things here required actually understanding the subsystem rather than following the issue's Proposed Fix:

- **The layout arithmetic holds together.** 4736 B, `tensors` at offset 640, and `DEP_GEN_OVERFLOW_DEPS_PER_RECORD = 524` are mutually consistent (`16 + 9N <= 4736`), and 640 preserves the original intent that each 128 B tensor blob covers exactly two cache lines instead of straddling three. The `static_assert` set was updated in step rather than left to rot.
- **The new `is_trivially_copyable_v` / `is_standard_layout_v` guards close a pre-existing gap** on these host/device wire structs — that is required by the codestyle rule for device-copied structs and was missing before this PR.
- **The OR-dedup handling is the substantive part.** Once an explicit edge can be `wait`-only, `emit_creator`'s `!first` early-return silently drops the creator's `retain`. Introducing `explicit_edge_index` to fold the creator flags back into the explicit edge keeps replay aligned with `FaninBuilder`'s collapse semantics — and it is documented in `docs/dfx/dep-gen.md` rather than left as folklore. I checked the `.at()` on that map: it is reachable only from the `first == false` branch, and `annot_preds` is cleared per record with STEP 1 as its only writer, so it cannot throw today.
- **a2a3/a5 are strictly symmetric.** I diffed the two subtrees' patches against each other; the only differences are `block_num` vs `core_num` and line offsets.
- **The tests are real.** The scene-test cases are `manual` only on `a2a3sim`, so the new flag assertions genuinely execute in `st-onboard-a2a3` and `st-onboard-a5`. Pinning the UT at `kDepCount = 64 + 524 + 1` anchors the new constant, and the trailing creator+explicit probe covers the dedup path that the `explicit_edge_index` change exists for.

## Blocking

**1. `docs/dfx/global-backpressure-design.md` now contradicts the code.**

The "Timeout backstop" section still states:

> When it expires the writer breaks to its single failure exit (`return false`), and the caller (`switch_buffer`) accounts the dropped records.

After this change that is only true for the **pop** side (`wait_for_ready_queue_space` / `wait_for_free_queue_entry`). On the **push** side the timeout is now discarded and nothing is accounted as dropped. Per the repo's doc-consistency rule the fix lands in the same commit as the code change. Splitting the paragraph to distinguish the two sides is enough.

**2. `return -5` collides with an existing code in the same function.**

`dep_gen_replay.cpp` already returns `-5` for `write_deps_json()` failure (near the end of `dep_gen_replay_emit_deps_json`). The new invalid-flags path reuses it. Callers in all four `device_runner.cpp` sites only log `dep_gen replay failed (%d)`, and the header documents codes as "see source", so `-5` becomes genuinely ambiguous between "a record carried a corrupt flags byte" and "deps.json could not be written". `-7` in both the a2a3 and a5 copies.

## Non-blocking, but worth addressing

**3. The DFX-engine fix deserves more than one bullet.** The commit message does carry `Preserve host ownership when a published DFX buffer hits a gate timeout`, and the rationale is in this PR's conversation — so this is not an undeclared change. The gap is scope visibility: `DeviceProfilerEngine` is shared by pmu, scope_stats, l2 phase, l2 AicoreTask, dep_gen and args_dump, so this is a behavior change across six subsystems riding in a PR whose title and body are entirely about dep_gen. The PR body's Summary omits it, and once this squash-merges the conversation context does not follow into git history. A short paragraph in the body — what the bug was, that it affects every DFX module, and that it is independent of #1827 — costs little and is what a future bisect will actually find. Splitting it into its own PR would be cleaner still, since it is independently testable and independently valuable.

**4. #1827's third item is dropped silently.** The issue also asks to "make the differential gate explicit-kind-aware". Both `oracle_preds` and `annot_preds` are still seeded from the same `kinds_data[i]` in STEP 1, so an explicit-kind divergence remains undetectable by construction — the shared hard-coded constant was replaced by a shared recorded byte. I think this is genuinely infeasible in the current design: the oracle is `compute_task_fanin`, which does not process explicit deps at all, so there is no independent second source to compare against. That is a fine answer, but it should be stated, because `Fixes #1827` will close the issue and take the item with it. Either say so in the body (and ideally in a comment next to STEP 1), or leave a follow-up issue.

**5. Rejecting the whole trace over one bad byte is inconsistent with the surrounding code.** The same function handles the same class of input — values read back from device shared memory that could be corrupt — by clamping and continuing:

```cpp
// ...we clamp on read too so a corrupted record never drives an OOB read
if (base_dc > DEP_GEN_MAX_EXPLICIT_DEPS) { LOG_ERROR(...); base_dc = DEP_GEN_MAX_EXPLICIT_DEPS; }
```

The new `raw_kind` check instead tears down both tensormaps and discards the entire `deps.json`. Unknown bits cannot cause an OOB read, and this is a DFX artifact rather than a correctness path, so `LOG_ERROR` plus `raw_kind &= kKnownDepFlags` (or falling back to `WAIT|RETAIN`) would match the established posture and still surface the anomaly.

**6. The push-gate timeout is now entirely unobservable.** A 30 s expiry means "the host is gone" — previously it at least showed up as an inflated dropped count; now there is no counter, no log, no sticky flag. Worth an `on_push_gate_timeout` trait hook or a one-shot log (not per-iteration — the AICPU hot-path logging rule applies).

## Consider

- The scene test maps alternating kinds back to producers via `sorted(barrier_explicit_edges.items())`, i.e. it assumes `TaskId::raw` ordering equals submission order. True today for a single ring, but a change in id allocation would surface as a confusing "flags mismatch" rather than an obvious failure. Deriving the producer sequence from the order of `deps["tasks"]` would be sturdier.
- No test covers the new rejection path. A host-side case feeding a record with `explicit_dep_kinds[0] = 0xFF` is cheap. (If item 5 is adopted, assert the clamped flags instead.)
- `reinterpret_cast<const uint8_t *>(args.explicit_dep_kinds_data())` relies on `enum DepFlags : uint8_t`. A `static_assert(sizeof(DepFlags) == 1)` next to the cast beats making the reader go look it up.
- The `WAIT|RETAIN` default is now spelled in three places (`Arg::explicit_dep_kind()`, both orchestrator call sites, and the UT's bare `3u`). A named constant would help; at minimum the UT could use `DEP_WAIT | DEP_RETAIN` instead of `3u`.
- The `flags` row in the `deps.json` schema table is marked `always`, but `host_build_graph`'s writer (`dep_gen_host_graph.cpp`) emits no `flags` field at all. Pre-existing, not introduced here — but since that row is already being rewritten, noting "tensormap_and_ringbuffer only" is nearly free.

## Notes

- `DepGenRecord`'s layout changed with no version field on the host/device boundary, and `static_assert` only guards one side at compile time. A partially-rebuilt tree (host runtime on 4672, AICPU on 4736) would produce a plausible-looking but garbage `deps.json` with nothing to detect it. Not a defect in this PR — the format never had a version — but worth a line in the body telling reviewers to do a full rebuild rather than an incremental one.
- Capacity impact is real but bounded: overflow slots drop 582 -> 524 (-10%), so a submit with e.g. `dc = 600` now needs 3 slots where 2 sufficed. The theoretical per-submit ceiling falls from ~595k to ~536k deps, which `CHIP_DEP_LIST_POOL_SIZE` (16384) puts far out of reach anyway. Record memory grows 1.37% and only when `--enable-dep-gen` is set, since `DepGenCollector::init()` is gated on it. The audit doc's added sentence is an accurate characterization.
- The scene test now makes half the barrier's deps ordering-only, which is a real runtime semantic change and not just an assertion change. Safe here because the producers write the external `ext_X` rather than a runtime-owned output, so there is no lifetime to lose — worth keeping in mind if that kernel is ever reworked to use owned outputs.
- `pto_isa.pin` is untouched and no pto-isa header references changed, so no pin bump is implied.

Items 1 and 2 are the only things holding this up; both are a few lines. The core change is well-built and I have no objection to the approach.

ChaoZheng109 (approved):

PR #2063, opened 2026-08-29
Feature: add A5 HBG Ready routing

zhusy54:
## Why

This is the second increment in the six-PR A5 `host_build_graph` resident-scheduler stack. It builds on the scheduler contracts merged in #2056 and the Host/AICore boundary fix in #2062.

The contracts define the shared GM layout and graph access model; this PR adds the dependency-to-Ready layer that future single-lane dispatch and production cutover use. It does **not** switch the production execution path: the existing AICPU scheduler remains active until the later cutover increment.

### Bootstrap and wake resolution

- Classify bootstrap tasks as immediately Ready, already completed, or waiting on the first executable producer.
- Register consumers on producer wake lists and reclassify/migrate them as producers complete.
- Close and drain wake lists without losing a waiter when completion races with registration.
- Validate task IDs, fanin ordering, task shape, predicates, and payload materialization before publication.

### Ready ownership and routing

- Add resolver-owned, per-core-type AIC/AIV FIFO Ready inboxes.
- Keep published work and owner-only pending work in separate banks; only the owner can promote the pending bank after the public bank drains.
- Publish sharded Ready-directory hints and use local-first, shard-constrained peer stealing.
- Make the successful CAS on the public inbox head the only task-ownership linearization point. Directory bits are hints and may be stale-positive; they never grant ownership.
- Preserve FIFO order within one Resolver and one core type. No global FIFO ordering across Resolvers is introduced.

### Dispatch and completion integration

- Add Ready claim and generation-tagged dispatch-slot fill helpers for the later single-lane scheduler loop.
- Resolve a completed producer by closing its wake list, rerouting its waiters, and publishing newly Ready AIC/AIV batches to the completing Resolver's inbox.
- Count each successful `scheduler_resolve_completion` exactly once in `resolve_count`; Ready-to-kernel latency remains owned by executor start because it measures a later lifecycle stage.

## Correctness and scope

- Concurrent consumers can claim each task at most once, including during stealing.
- A thief cannot observe or promote an owner's pending bank, and older pending work stays ahead of newly arriving work.
- Shared scheduler helpers follow #2062's `__aicore__`-only annotation contract; no CUDA-style `__host__` compatibility layer is reintroduced.
- Unit-test graph descriptors, payloads, and fanins share one aligned contiguous image so `SelfRelativePtr` fanin references remain valid on all hosts.
- Product code remains under `src/a5/runtime/host_build_graph/`; this PR does not change A2/A3, common HBG runtime layout, public APIs, wire contracts, environment variables, or feature gates.

## Reviewer guide

- Verify wake registration versus producer completion cannot lose or duplicate a consumer.
- Treat the public-head CAS, not the Ready directory, as the ownership point.
- Check owner-only pending-bank promotion and per-owner/per-type FIFO preservation.
- Confirm error paths do not publish partial Ready or dispatch state, and that this increment does not alter the production scheduler entry point.

## Testing

- [x] Targeted Ready unit tests: 16/16 passed.
- [x] ASan/UBSan Ready unit tests: 16/16 passed.
- [x] Full no-hardware C++ unit-test suite: 124/124 passed.
- [x] Pre-commit on changed files: all hooks passed.
- [x] GitHub CI: all applicable Ubuntu/macOS packaging, unit, simulation, onboard, profiling-smoke, and pre-commit checks passed.

### DIFF
--- a/src/a5/runtime/host_build_graph/runtime/scheduler/scheduler_ready.h
+++ b/src/a5/runtime/host_build_graph/runtime/scheduler/scheduler_ready.h
@@ -1,0 +1,1138 @@
+/*
+ * Copyright (c) PyPTO Contributors.
+ * This program is free software, you can redistribute it and/or modify it under the terms and conditions of
+ * CANN Open Software License Agreement Version 2.0 (the "License").
+ * Please refer to the License for details. You may not use this file except in compliance with the License.
+ * THIS SOFTWARE IS PROVIDED ON AN "AS IS" BASIS, WITHOUT WARRANTIES OF ANY KIND, EITHER EXPRESS OR IMPLIED,
+ * INCLUDING BUT NOT LIMITED TO NON-INFRINGEMENT, MERCHANTABILITY, OR FITNESS FOR A PARTICULAR PURPOSE.
+ * See LICENSE in the root of the software repository for the full text of the License.
+ * -----------------------------------------------------------------------------------------------------------
+ */
+
+#pragma once
+
+#include "scheduler_types.h"
+#include "scheduler_memory.h"
+#include "scheduler_graph.h"
+
+enum class SchedulerRouteResult : uint64_t {
+    READY = 0,
+    READY_TO_ENQUEUE = 1,
+    WAITING = 2,
+    COMPLETED = 3,
+    ERROR = 4,
+};
+
+enum class SchedulerPredicateResult : uint8_t {
+    PASS = 0,
+    FAIL = 1,
+    MALFORMED = 2,
+};
+
+struct SchedulerWakeStats {
+    uint64_t fanin_state_load_count{0};
+    uint64_t wake_register_count{0};
+    uint64_t wake_cas_retry_count{0};
+    uint64_t wake_closed_retry_count{0};
+    uint64_t wake_migrate_count{0};
+    uint64_t wake_close_count{0};
+};
+
+struct SchedulerReadyStats {
+    uint64_t enqueue_count{0};
+    uint64_t batch_count{0};
+    uint64_t pop_count{0};
+    uint64_t steal_count{0};
+    uint64_t cas_retry_count{0};
+    uint64_t link_wait_count{0};
+    uint64_t link_wait_max{0};
+};
+
+struct SchedulerCompletionStats {
+    uint64_t enqueue_count{0};
+    uint64_t resolve_count{0};
+    uint64_t ready_to_kernel_cycles{0};
+    uint64_t ready_to_kernel_max_cycles{0};
+};
+
+struct SchedulerReadyBatch {
+    int64_t head{SCHEDULER_INBOX_EMPTY};
+    int64_t tail{SCHEDULER_INBOX_EMPTY};
+    uint64_t count{0};
+};
+
+inline __aicore__ void scheduler_ready_batch_reset(SchedulerReadyBatch *batch) {
+    batch->head = SCHEDULER_INBOX_EMPTY;
+    batch->tail = SCHEDULER_INBOX_EMPTY;
+    batch->count = 0;
+}
+
+inline __aicore__ uint64_t scheduler_ready_pending_pack(int64_t head, int64_t tail) {
+    const uint64_t packed_head = static_cast<uint32_t>(static_cast<int32_t>(head));
+    const uint64_t packed_tail = static_cast<uint32_t>(static_cast<int32_t>(tail));
+    return packed_head | (packed_tail << 32);
+}
+
+inline __aicore__ int64_t scheduler_ready_pending_head(uint64_t packed) {
+    return static_cast<int64_t>(static_cast<int32_t>(static_cast<uint32_t>(packed)));
+}
+
+inline __aicore__ int64_t scheduler_ready_pending_tail(uint64_t packed) {
+    return static_cast<int64_t>(static_cast<int32_t>(static_cast<uint32_t>(packed >> 32)));
+}
+
+inline __aicore__ bool scheduler_ready_pending_endpoint_fits(int64_t task_id) {
+    return task_id >= SCHEDULER_INBOX_EMPTY && task_id <= INT32_MAX;
+}
+
+inline __aicore__ void scheduler_ready_owner_init(__gm__ SchedulerReadyOwnerState *owner_state) {
+    if (owner_state == nullptr) return;
+    for (uint32_t type = 0; type < SCHEDULER_CORE_TYPE_COUNT; ++type) {
+        scheduler_gm_store(owner_state->queues[type].pending_endpoints, SCHEDULER_READY_PENDING_EMPTY);
+        scheduler_gm_store(owner_state->queues[type].advertised, UINT64_C(0));
+    }
+    scheduler_cache_barrier();
+}
+
+struct SchedulerReadyClaim {
+    int64_t task_id{SCHEDULER_TASK_ID_INVALID};
+    uint64_t inbox_index{UINT64_MAX};
+    SchedulerReadySource source{SchedulerReadySource::LOCAL};
+    uint64_t claim_start_cycles{0};
+    uint64_t claim_end_cycles{0};
+};
+
+struct SchedulerFreeSlotClaim {
+    uint64_t worker_id{UINT64_MAX};
+    uint32_t slot_index{UINT32_MAX};
+    uint32_t generation{0};
+};
+
+struct SchedulerCompletionServiceTiming {
+    uint64_t scan_cycles{0};
+    uint64_t consume_cycles{0};
+    uint64_t resolve_cycles{0};
+    uint64_t ready_publish_cycles{0};
+    uint64_t refill_cycles{0};
+    uint64_t finalize_cycles{0};
+};
+
+struct SchedulerDispatchFillTiming {
+    uint64_t prepare_cycles{0};
+    uint64_t materialize_cycles{0};
+    uint64_t publish_cycles{0};
+};
+
+inline __aicore__ uint64_t scheduler_cycles() {
+#if defined(__CCE_AICORE__)
+    return get_sys_cnt_aicore();
+#else
+    return 0;
+#endif
+}
+
+inline __aicore__ uint32_t scheduler_core_type_index(int32_t core_type) {
+    return core_type == static_cast<int32_t>(CoreType::AIC) ? 0U : 1U;
+}
+
+inline __aicore__ uint32_t scheduler_metadata_core_type_index(uint8_t subtask_slot) {
+    return subtask_slot == 0 ? 0U : 1U;
+}
+
+inline __aicore__ uint8_t scheduler_metadata_single_subtask_slot(uint8_t active_mask) {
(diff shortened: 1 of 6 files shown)


src/a5/runtime/host_build_graph/runtime/scheduler/scheduler_ready.h
```diff
@@ -0,0 +1,1101 @@
...
+        scheduler_dispatch_slot_at(scheduler_state_base, resolver, slot_claim.worker_id, slot_claim.slot_index);
+    uint32_t generation = slot_claim.generation + 1;
+    if (generation == 0) generation = 1;
+    __gm__ uint64_t *callable_addresses =
+        scheduler_state_at<uint64_t>(scheduler_state_base, resolver->callable_addresses_offset);
+    const bool inline_task = scheduler_task_is_inline(metadata.flags);
+    uint64_t callable_address = UINT64_C(1);
+    if (!inline_task) {
+        scheduler_observe_cache_line(&callable_addresses[kernel_id]);
+        callable_address = callable_addresses[kernel_id];
```
ChaoZheng109: **`kernel_id` is used as a table index without a bounds check — up to a 512 KB out-of-bounds GM read.**

`callable_addresses` holds `SCHEDULER_CALLABLE_CAPACITY` (1024) entries = 8 KB. `kernel_id` is a `uint16_t` loaded from GM, so its range is 0..65535; at the top of that range this reads ~512 KB past the end of the table.

This is not just theoretical. The test fixture in this PR writes `UINT16_MAX` into the unused kernel slots:

```cpp
metadata[task].kernel_ids[1] = UINT16_MAX;
metadata[task].kernel_ids[2] = UINT16_MAX;
```

which is the natural result of the host projecting `INVALID_KERNEL_ID` (-1) into a `uint16_t` field. Any inconsistency between `active_mask` and `kernel_ids` — a mask bit pointing at a slot with no kernel — lands here.

Everything else reachable from GM in this function is validated: task id range, fanin count, fanin topological order, tensor/scalar counts, `block_idx`, core-type match, executable, gang. `kernel_id` is the only externally-sourced value used directly as an array subscript, and it is the one that is unchecked. One line closes it, in the same style as the neighbouring guards:

```cpp
if (kernel_id >= SCHEDULER_CALLABLE_CAPACITY) {
    scheduler_record_error(run_control, ready_claim.task_id, SchedulerGraphResult::INVALID_CALLABLE,
                           &graph, resolver, <site>);
    return false;
}
```

src/a5/runtime/host_build_graph/runtime/scheduler/scheduler_ready.h
```diff
@@ -0,0 +1,1101 @@
...
+    case 2:
+        value = *reinterpret_cast<__gm__ int16_t *>(operand);
+        break;
+    case 4:
+        value = *reinterpret_cast<__gm__ int32_t *>(operand);
+        break;
+    case 8:
+        value = *reinterpret_cast<__gm__ int64_t *>(operand);
+        break;
+    default:
```
ChaoZheng109, line 177: **Malformed predicate data silently skips the task instead of raising an error.**

`return false` here means "predicate did not pass", and the caller turns that into `payload->function_bin_addr = 0`, i.e. the task is not executed. But this `default:` branch is reached when `elem_size` is not 1/2/4/8 — that is corrupt data, not a failed comparison. The same applies to the `default:` in the `op` switch below and to the `predicate->addr == 0` early return.

So a malformed predicate produces **a task that silently does not run**, with no `scheduler_error` and no diagnostic. That is the hardest scheduler failure mode to chase: wrong results, everything green. It is also inconsistent with the rest of this file, where every malformed GM input goes through `scheduler_record_error`.

Suggest separating the two outcomes — e.g. return a tri-state (`PASS` / `FAIL` / `MALFORMED`), and have the caller route `MALFORMED` to `scheduler_record_error(..., UNSUPPORTED_SHAPE, ...)`.

Related, same function: `predicate->addr` is an arbitrary GM address read straight from the payload and dereferenced with no validation. Same class of unchecked external input as the `kernel_id` comment above.

src/a5/runtime/host_build_graph/runtime/scheduler/scheduler_ready.h
```diff
@@ -0,0 +1,1101 @@
...
+        scheduler_gm_store(run_control->error_graph_task_count, graph->task_count);
+        scheduler_gm_store(run_control->error_descriptors_address, graph->descriptors_address);
+        scheduler_gm_store(run_control->error_payloads_address, graph->payloads_address);
+        scheduler_gm_store(run_control->error_task_window_mask, graph->task_window_mask);
+    }
+    if (context != nullptr) {
+        scheduler_gm_store(run_control->error_core_id, static_cast<uint64_t>(context->physical_core_id));
+        scheduler_gm_store(run_control->error_core_type, static_cast<uint64_t>(context->core_type));
+    }
+    scheduler_gm_store(run_control->error_reserved[0], error_site);
```
ChaoZheng109: **The 13 error-site codes are bare magic numbers with no enum and no table.**

`error_site` takes `UINT64_C(2)`, `40`, `41`, `44`, `45`, `46`, `60`, `61`, `62`, `63`, `64`, `65`, `66` at the various call sites. Once the resident scheduler is live, this word plus `scheduler_error` is very close to the only localisation signal available when something goes wrong on device — there is no host-side stack to fall back on.

Six months from now nobody will be able to get from `error_site=64` back to "`scheduler_resolve_completion` found a waiter whose metadata is not executable". Suggest an `enum class SchedulerErrorSite : uint64_t` next to `SchedulerGraphResult`, or at minimum a table in the header comment block mapping each number to its site.

Separately, this stores the site into `error_reserved[0]`. That is the reserved region declared in #2056 — repurposing `[0]` as a de-facto named field is how reserved space drifts; the next person to extend it will not know the first slot is taken. Consider giving `SchedulerRunControl` a real `error_site` field (the reserved array can shrink by one, so the layout `static_assert`s stay put).

tests/ut/cpp/a5/test_hbg_scheduler_ready.cpp
```diff
@@ -0,0 +1,670 @@
...
+    ASSERT_TRUE(scheduler_claim_ready_for_slot(
+        graph.graph(), storage.scheduler_state->base(), &storage.contexts[7], storage.run_control, 14, 0, &cursor,
+        &stats, &claim
+    ));
+    EXPECT_EQ(claim.task_id, 0);
+    EXPECT_EQ(claim.inbox_index, 8u);
+    EXPECT_EQ(claim.source, SchedulerReadySource::STOLEN);
+}
+
+TEST(SchedulerReadyInbox, ConcurrentConsumersNeverDuplicateTask) {
```
ChaoZheng109, line 676: **This is the only multi-threaded test, and it exercises pop-vs-pop only — none of the three races the PR's own reviewer guide asks to verify are covered under contention.**

All 8 threads pop from the same inbox; there is no concurrent push at any point. So the following remain tested only in serialized orderings:

1. The owner's non-CAS `scheduler_gm_store(inbox->head, batch->head)` in `scheduler_ready_batch_push` racing a thief's pop.
2. The pending -> public promotion in `scheduler_ready_owner_maintain_type` (also a blind store) racing a thief's pop.
3. **`scheduler_route_task` registering on a wake list while `scheduler_resolve_completion` closes that same list.** This is the single most delicate part of the design — the `CAS failed -> observed CLOSED -> advance the fanin cursor` branch — and `WakeResolvePublishesConsumerToResolverLocalInbox` walks it strictly single-threaded.

I convinced myself (1) and (2) are safe: a thief's CAS can only succeed after reading a non-empty head, and only the owner can make the head non-empty, so the blind store cannot be lost. But that argument rests entirely on "exactly one worker writes a given `inbox_index`", which this increment neither asserts nor establishes — `inbox_index` assignment lands in a later PR.

(3) is the one I would most like to see covered: one producer thread repeatedly doing `exchange -> CLOSED`, N consumer threads repeatedly registering, asserting every consumer is woken exactly once and every fanin cursor advances monotonically. Serial tests show the algorithm is correct under *one* interleaving; the next increment wires these functions into the real loop, and after that a defect here surfaces as a 507018 on device rather than a red UT.

Minor robustness point on this test: the lambda `return`s if `scheduler_ready_pop_from_inbox` fails, but the other threads keep spinning on `while (claimed < kTasks)`. A failure would hang the suite instead of failing it.

ChaoZheng109 (commented):
Reviewed the Ready-routing increment. The algorithm itself holds up — I traced the four invariants the PR body claims (single claim under stealing, thief cannot see the owner's pending bank, older pending work stays ahead, per-owner/per-type FIFO) and they are all sound. The `exchange -> CLOSED` wake-list close paired with the "CAS failed, observed CLOSED, advance the fanin cursor" branch is a neat way to make registration and completion race-free without a lock, and reading `next` before calling `scheduler_route_task` on a waiter is exactly the ordering that makes migration safe.

Four things I would like addressed before this lands. One is a concrete out-of-bounds read; the rest are about this increment being the definition site for contracts that the next PR in the stack will consume.

ChaoZheng109 (dismissed):

src/a5/runtime/host_build_graph/runtime/scheduler/scheduler_ready.h
zhusy54: Fixed in 92718cd8. Callable lookup now validates kernel_id < SCHEDULER_CALLABLE_CAPACITY before cache observation or table access. Ready UTs cover the last valid ID, the capacity boundary, and the inline sentinel; the downstream Gang path reuses the same helper.

src/a5/runtime/host_build_graph/runtime/scheduler/scheduler_ready.h
zhusy54: Fixed in 92718cd8. Predicate evaluation is now PASS/FAIL/MALFORMED; malformed op/address/size/alignment records INVALID_ARGUMENTS at DISPATCH_INVALID_PREDICATE and does not publish a slot. Direct Host submit also validates the resolved operand before task allocation, with negative UT coverage.

src/a5/runtime/host_build_graph/runtime/scheduler/scheduler_ready.h
zhusy54: Fixed in 92718cd8 and propagated through the feature stack. SchedulerErrorSite provides stable named diagnostic values, SchedulerRunControl now has a real error_site field while preserving its ABI size/offsets, and AICPU/Host diagnostics read the named field.

tests/ut/cpp/a5/test_hbg_scheduler_ready.cpp
zhusy54: Fixed in 92718cd8. Added owner-push vs thief, pending-promotion vs thief, 64-consumer register-vs-close exactly-once, and concurrent no-duplicate-claim tests. The original pop contention test now publishes failure so peers cannot spin forever. The four contention tests also passed 100 repeated runs locally.

ChaoZheng109 (approved):

PR #2072, opened 2026-08-31
Add: complete A5 HBG single-lane scheduling

zhusy54:
## Why

This is PR3 of the A5 HBG resident AICore scheduler stack. It builds on the
scheduler contracts and Ready routing from #2063, including the Ready
correctness fixes merged through #2077, and completes the ordinary single-lane
scheduling library before the cross-layer production cutover.

Keeping dispatch/completion protocol work separate from the Host/AICPU/AICore
cutover lets reviewers validate slot ownership, generation handling, and
publication ordering without mixing in entry-point and lifecycle wiring.

### Completion service

- Consume generation-tagged completion inboxes and reject stale generations,
  invalid workers, inactive lanes, and unexpected Gang slots before
  scheduler-state access.
- Retire completed tasks, resolve and publish newly ready waiters, and either
  refill the released executor slot directly or return it to `FREE`.
- Propagate `trace_enabled` through completion resolution so resolver timing,
  worker attribution, and waiter ready-transition tracing are preserved.

### Single-lane dispatch

- Dispatch ordinary AIC and AIV work through the two-slot
  `FREE(g) -> FILLING(g) -> READY/GATED(g + 1) -> FREE(g + 1)` protocol.
- Treat completion direct refill as a distinct
  `READY(g, completed) -> READY(g + 1)` fast path. Executors recognize work by
  the full publication value (state + generation), not a `FREE -> READY` edge.
- Add deferred AIV reservations so normal AIV work cannot consume the Resolver
  lane required to keep scheduling forward progress.
- Drain peer reservations before local publication, and validate reservation
  owner, slot bounds, generation, state, and unmaterialized `task_id` before
  computing or reading the reserved GM slot.
- Report pass progress separately from fatal fill failure. A failed fill may
  leave the claimed task and `FILLING` slot as terminal fatal residue; the
  production caller added by PR4 must stop whenever `failed` is set.
- Add stable named error sites for unexpected completion slots, generation
  mismatches, and invalid deferred-reservation states.

### Regression coverage

- Add portable C++ coverage for normal AIC/AIV dispatch, completion catch-up,
  direct refill, waiter tracing, deferred reservation/publication, peer
  priority, and all five new error sites.
- Verify partial progress plus fatal fill failure, including the failure flag,
  scheduler error, named error site, and failed slot terminal state.
- Keep task payload and fanin data in one contiguous aligned graph image so
  self-relative pointers follow the production memory contract.

## Correctness and scope

- Payload and metadata are written before slot publication; stale completion
  generations cannot affect a reused slot.
- Worker IDs, slot indices, and deferred reservation ownership are validated
  before address calculation or GM dereference.
- The implementation remains A5-owned and does not add A5 scheduler state to
  the common A2/A3 runtime.
- This PR intentionally does **not** change the production entry path. PR4 owns
  Host state planning/upload, AICPU bootstrap/lifecycle wiring, and AICore
  Resolver/Executor entry-point cutover, including both callers' fatal failure
  handling.
- Gang scheduling and teardown convergence remain later stack increments;
  Graph replay continues to use the legacy isolated path.

## Reviewer guide

- Check ordinary and direct-refill generation/state transitions and the
  executor's full-publication observation contract.
- Check that deferred AIV reservation always preserves Resolver progress and
  that every invalid reservation is rejected before GM access.
- Check publication ordering, trace propagation, and first-error localization
  on failure paths.
- Check that a return value reporting prior progress cannot hide `failed`.

## Testing

- [x] Targeted scheduler dispatch CTest: `test_a5_hbg_scheduler_dispatch`.
- [x] Full non-hardware C++ CTest: 128/128 passed.
- [x] Pre-commit on all five PR files.
- [x] Full GitHub CI after the latest rebase and review fixes.


src/a5/runtime/host_build_graph/runtime/scheduler/scheduler_completion.h
```diff
@@ -0,0 +1,180 @@
...
+    uint64_t *direct_refilled_slot_mask = nullptr, SchedulerCompletionServiceTiming *timing = nullptr,
+    __gm__ SchedulerReadyOwnerState *owner_state = nullptr
+) {
+    if (resolver->is_resolver == 0) return false;
+    if (direct_refilled_slot_mask != nullptr) *direct_refilled_slot_mask = 0;
+    bool progress = false;
+    for (uint32_t cluster_lane = 0; cluster_lane < 3; ++cluster_lane) {
+        const uint64_t worker_id = resolver->cluster_worker_ids[cluster_lane];
+        __gm__ SchedulerCompletionInbox *completion_line =
+            scheduler_completion_inbox_at(scheduler_state_base, resolver, worker_id);
```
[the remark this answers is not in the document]
zhusy54: Fixed: cluster worker IDs are now bounds-checked and inactive lanes are skipped before completion-inbox access. Coverage includes the sentinel worker-ID case.

src/a5/runtime/host_build_graph/runtime/scheduler/scheduler_dispatch.h
```diff
@@ -0,0 +1,400 @@
...
+    if (queue == nullptr || queue->count == 0) return true;
+    const SchedulerDeferredAivDispatch &entry = queue->entries[0];
+    __gm__ SchedulerDispatchSlot *slot = scheduler_dispatch_slot_at(
+        scheduler_state_base, resolver, entry.reserved_slot.worker_id, entry.reserved_slot.slot_index
+    );
+    const uint64_t publication = scheduler_gm_query(slot->publication);
+    if (entry.reserved_slot.worker_id != resolver->worker_index ||
+        entry.reserved_slot.slot_index >= SCHEDULER_PENDING_SLOT_COUNT ||
+        scheduler_dispatch_state(publication) != SchedulerDispatchSlotState::FILLING ||
+        scheduler_dispatch_generation(publication) != entry.reserved_slot.generation) {
```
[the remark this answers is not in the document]
zhusy54: Fixed: deferred reservation ownership and slot bounds are now validated before computing or reading the slot address. Added invalid-reservation coverage.

ChaoZheng109 (commented):
Four findings on the new AICore scheduler library. The first is a correctness issue I would treat as blocking; the other three are consistency problems that are cheap to fix now and expensive to rediscover from a device log later.

The design itself reads well — the deferred AIV reservation (every queued entry owning a `FILLING` Resolver slot so a peer miss can always fall back locally) is a solid answer to keeping Resolver progress alive, and the error-site coverage in the new tests is thorough.

src/a5/runtime/host_build_graph/runtime/scheduler/scheduler_dispatch.h
```diff
@@ -0,0 +1,408 @@
...
+    const uint32_t aic_core_type = static_cast<uint32_t>(CoreType::AIC);
+    uint64_t stage_start = timing == nullptr ? 0 : scheduler_cycles();
+    uint64_t detail_start = timing == nullptr ? 0 : scheduler_normal_dispatch_detail_cycles(*timing, aic_core_type);
+    bool progress = false;
+
+    // AIC has no peer lane in its Cluster, so preserve the existing slot order.
+    if (scheduler_ready_directory_nonempty(scheduler_state_base, resolver, resolver->resolver_count, aic_core_type)) {
+        bool aic_ready_available = true;
+        for (uint32_t cluster_lane = 0; cluster_lane < 3 && aic_ready_available; ++cluster_lane) {
+            const uint64_t worker_id = resolver->cluster_worker_ids[cluster_lane];
```
ChaoZheng109, line 89: **Missing `worker_id` bound check before the GM dereference — out-of-bounds read on an unpopulated cluster lane.**

`scheduler_worker_context_at()` (`scheduler_ready.h:275-281`) is raw pointer arithmetic with no bound check:

```cpp
return scheduler_state_at<SchedulerWorkerContext>(
    base, context->worker_contexts_offset + worker_id * sizeof(SchedulerWorkerContext));
```

and `scheduler_initialize_state()` (`scheduler_types.h:1390-1392`) initialises unpopulated cluster lanes to `UINT64_MAX`:

```cpp
contexts[worker].cluster_worker_ids[0] = UINT64_MAX;
contexts[worker].cluster_worker_ids[1] = UINT64_MAX;
contexts[worker].cluster_worker_ids[2] = UINT64_MAX;
```

The completion path guards against exactly this — `scheduler_completion.h:27` and `:148` both do `if (worker_id >= resolver->runtime_worker_count) ...`. The dispatch path does not, in three places:

- here (AIC loop, L83 → dereferenced at L85)
- L151 → dereferenced at L153 (AIV loop)
- L280 → dereferenced at L282 (`scheduler_deferred_aiv_peer_lane`)

All three read `worker_id` out of `cluster_worker_ids[]` and immediately dereference `target->active`.

**Failure scenario:** a Cluster with only two populated lanes — the case this file's own comment at L299 calls out ("A Resolver may be the only active AIV in its Cluster (for example, a single-root AIV graph)") — leaves the third lane at `UINT64_MAX`. The offset `UINT64_MAX * sizeof(SchedulerWorkerContext)` wraps, so reading `target->active` is a wild GM access on AICore. On device that surfaces as a generic `507018` with no allocator fatal and no producer/consumer timeout in the device log, which per `.claude/rules/running-onboard.md` is the hardest class to triage.

**Why the tests do not catch it:** `configure_normal_aiv_cluster` (test L169-171) populates all three lanes with valid ids `0/1/2`, and `contexts[2]` happens to land inside the 108-entry buffer with `active == 0`, so the loop simply `continue`s. `KeepsReservationForResolverWhenNoPeerIsActive` (test L618) also models "no peer" as `contexts[2].active = 0`, never as an unpopulated lane. Meanwhile the completion test at L238 *does* use `cluster_worker_ids[2] = UINT64_MAX`. The two paths are modelled differently in the same PR, which is what makes this look like an oversight rather than a deliberate asymmetry.

**Suggested fix:** apply the same guard the completion path already uses at all three sites, and add a dispatch test that leaves a lane at `UINT64_MAX`.

src/a5/runtime/host_build_graph/runtime/scheduler/scheduler_dispatch.h
```diff
@@ -0,0 +1,408 @@
...
+                __gm__ SchedulerDispatchSlot *slot =
+                    scheduler_dispatch_slot_at(scheduler_state_base, resolver, worker_id, pending_slot);
+                const uint64_t publication = scheduler_gm_query(slot->publication);
+                if (scheduler_dispatch_state(publication) != SchedulerDispatchSlotState::FREE) continue;
+                SchedulerReadyClaim ready{};
+                if (!scheduler_claim_ready_for_slot(
+                        graph, scheduler_state_base, resolver, run_control, resolver->resolver_count, aic_core_type,
+                        &ready_victim_cursors[aic_core_type], ready_stats, &ready, trace_enabled, owner_state
+                    ))
+                    return progress;
```
ChaoZheng109, line 108: **The return value conflates "error" with "no progress", and the three error paths disagree with each other.**

`scheduler_fill_cluster_normal_slots` returns a bool meaning "made progress", but errors inside it are handled three different ways:

- `scheduler_claim_ready_for_slot` fails (error already latched) → `return progress;` — here and at L212, where `progress` may be **`true`**
- `scheduler_fill_dispatch_slot` fails (error already latched) → `return false;` — L118 and L222
- unexpected cluster topology → `return false;` with **no error recorded at all** — L155

So a caller receiving `false` cannot distinguish "nothing to dispatch" from "something failed", and a caller receiving `true` cannot conclude that no error was latched. Since PR4 owns the entry-point cutover, whichever contract is intended should be pinned down before that lands.

Either give the error signal its own out-parameter, or commit to "any error returns false and has always latched into `run_control`" and state it in a comment on the function — plus a test that fixes it. Right now the contract is neither documented nor constrained by any test.

Related, on L155: that branch also deserves a named error site. It guards the `aiv_workers[2]` bound so it is doing necessary work, but `PLATFORM_AIV_CORES_PER_BLOCKDIM == 2` makes it unreachable under a sane topology — so if it ever does fire, scheduler state is already corrupt, and silently spinning until the op-execute timeout is the worst possible way to find that out. This PR already adds five named sites; a sixth here is consistent with that.

src/a5/runtime/host_build_graph/runtime/scheduler/scheduler_dispatch.h
```diff
@@ -0,0 +1,408 @@
...
+        uint64_t publications[SCHEDULER_PENDING_SLOT_COUNT]{};
+        uint32_t free_mask{0};
+        uint32_t occupied_slots{0};
+        bool is_resolver{false};
+    };
+    const uint32_t aiv_core_type = static_cast<uint32_t>(CoreType::AIV);
+    stage_start = timing == nullptr ? 0 : scheduler_cycles();
+    detail_start = timing == nullptr ? 0 : scheduler_normal_dispatch_detail_cycles(*timing, aiv_core_type);
+    if (scheduler_ready_directory_nonempty(scheduler_state_base, resolver, resolver->resolver_count, aiv_core_type)) {
+        AivWorkerSlots aiv_workers[2]{};
```
ChaoZheng109: **Hardcoded cluster geometry, where the platform constants are already in scope.**

`src/a5/platform/include/common/platform_config.h:43-45` already defines:

```cpp
constexpr int PLATFORM_CORES_PER_BLOCKDIM = 3;
constexpr int PLATFORM_AIC_CORES_PER_BLOCKDIM = 1;
constexpr int PLATFORM_AIV_CORES_PER_BLOCKDIM = 2;
```

and `scheduler_types.h:18` already does `#include "common/platform_config.h"`, so all three are in scope here at zero cost. This PR adds six new literals instead:

| Location | Current | Should be |
| --- | --- | --- |
| `scheduler_dispatch.h:82, 150, 279` | `cluster_lane < 3` | `PLATFORM_CORES_PER_BLOCKDIM` |
| `scheduler_dispatch.h:148` (this line) | `AivWorkerSlots aiv_workers[2]` | `PLATFORM_AIV_CORES_PER_BLOCKDIM` |
| `scheduler_dispatch.h:155` | `aiv_worker_count >= 2` | `PLATFORM_AIV_CORES_PER_BLOCKDIM` |
| `scheduler_completion.h:25, 146` | `cluster_lane >= 3` / `< 3` | `PLATFORM_CORES_PER_BLOCKDIM` |

Pre-existing code has the same problem (`scheduler_types.h:634`'s `SCHEDULER_WORKER_CAPACITY = 108` is `PLATFORM_MAX_CORES`; `:638`'s `/ 3` and `:1053`'s `cluster_worker_ids[3]` are both `PLATFORM_CORES_PER_BLOCKDIM`) and I am not asking this PR to clean that up. But adding six more moves the count the wrong way, and the replacement is mechanical given the include is already there.

src/a5/runtime/host_build_graph/runtime/scheduler/scheduler_dispatch.h
```diff
@@ -0,0 +1,408 @@
...
+ * Please refer to the License for details. You may not use this file except in compliance with the License.
+ * THIS SOFTWARE IS PROVIDED ON AN "AS IS" BASIS, WITHOUT WARRANTIES OF ANY KIND, EITHER EXPRESS OR IMPLIED,
+ * INCLUDING BUT NOT LIMITED TO NON-INFRINGEMENT, MERCHANTABILITY, OR FITNESS FOR A PARTICULAR PURPOSE.
+ * See LICENSE in the root of the software repository for the full text of the License.
+ * -----------------------------------------------------------------------------------------------------------
+ */
+
+#pragma once
+
+#include "scheduler_completion.h"
```
ChaoZheng109, line 16: **File naming: `scheduler_dispatch.h` and `scheduler_dispatch.cpp` now sit in the same directory and are unrelated.**

After this PR, `src/a5/runtime/host_build_graph/runtime/scheduler/` contains:

- `scheduler_dispatch.cpp` — 1672 lines, the **AICPU-side** legacy scheduler, includes `scheduler_context.h`
- `scheduler_dispatch.h` — this PR, the **AICore-resident** scheduler, a header-only library that includes `scheduler_completion.h`

and the same collision for `scheduler_completion.{h,cpp}` (772 lines on the `.cpp` side).

Neither `.cpp` includes its same-named `.h`, and the pairs share no declaration at all. Every reader — and every editor's "switch between header and source" binding — will assume they belong together. They do not.

This is cheap to fix now and gets considerably worse in PR4, when both schedulers are on screen simultaneously. Either give the new headers a distinguishing prefix (`aicore_scheduler_dispatch.h` / `aicore_scheduler_completion.h`, matching the existing `aicore_completion_mailbox.h` one directory up in `runtime/`), or add a one-line header comment to each stating which tier it belongs to and that the same-named `.cpp` is unrelated.

src/a5/runtime/host_build_graph/runtime/scheduler/scheduler_dispatch.h
zhusy54: Fixed in a4e567c7: scheduler_fill_cluster_normal_slots now returns progress and reports an aborted pass independently through the failed output parameter. Claim and fill failures set failed while preserving prior progress; excessive AIV topology latches NORMAL_DISPATCH_INVALID_TOPOLOGY. Tests cover no-progress/no-error, partial progress followed by failure, and the named topology error.

src/a5/runtime/host_build_graph/runtime/scheduler/scheduler_dispatch.h
zhusy54: Fixed in a4e567c7: both headers now state that they are AICore-resident libraries and that the same-named cpp files are the independent AICPU scheduler implementations.

src/a5/runtime/host_build_graph/runtime/scheduler/scheduler_dispatch.h
zhusy54: Fixed in a4e567c7: cluster scans and completion validation now use PLATFORM_CORES_PER_BLOCKDIM, while AIV worker storage and limits use PLATFORM_AIV_CORES_PER_BLOCKDIM.

src/a5/runtime/host_build_graph/runtime/scheduler/scheduler_dispatch.h
zhusy54: Fixed in a4e567c7: all three cluster-worker dereferences now skip IDs outside runtime_worker_count. The dispatch tests leave cluster_worker_ids entries at UINT64_MAX and exercise both AIC scanning and deferred AIV peer lookup.

poursoul (approved):

Issue #2071, opened 2026-08-29
[Performance][A5 TMR] Unexpected cross-commit performance shifts disproportionate to code changes


yanghaoran29:
## Platform

a5 (Ascend 950 hardware), card 1

## Runtime Variant

`tensormap_and_ringbuffer`

## Summary

On a local A5 TMR benchmark environment, several commits that were not intended as performance changes produced 5%–8% shifts in Device or Orchestrator time even though they did not add corresponding work to the measured hot path. The scale of the performance change is disproportionate to the scale and execution stage of the code change.

Two cases have been confirmed:

| Commit | Functional change | Extra work in the measured hot path | Original anomaly |
|---|---|---|---|
| `de32536a` | Fix the callable dispatch-table lifetime by clearing an 8 KiB table during Host bind | No; the clearing happens before the Device Orchestrator window | PA Orchestrator became 7.37%–8.47% slower |
| `35f195bd` | Preserve dependency flags for dep-gen replay | No; dep-gen was disabled in these benchmarks | BGEMM/PA Device or Orchestrator became 5.23%–8.13% faster |

The executed algorithmic steps do not explain these shifts. A hypothesis more consistent with the observations is that A5 TMR performance is sensitive to the binary instruction layout of hot functions, and that unrelated source changes can indirectly alter that layout.

To test this hypothesis without mixing in a one-sided alignment change, I applied the same 64-byte alignment to three hot entry points in both each commit and its direct parent, then compared aligned Commit against aligned Parent. All eight original `|delta| >= 5%` anomalies moved below 5%:

- `de32536a`: PA Case1/Case2 Orchestrator changed from **7.37%/8.47% slower** to **4.60%/4.18% slower**.
- `35f195bd`: BGEMM Device/Orchestrator changed from **5.59%/8.13% faster** to **2.41%/3.58% faster**.
- `35f195bd`: PA Case1/Case2 Orchestrator changed from **5.23%/5.91% faster** to **1.53%/1.95% slower**.
- `35f195bd`: manual PA Case1/Case2 Orchestrator changed from **5.23%/6.32% faster** to **0.86%/1.23% slower**.

This does not prove that 64-byte alignment is the final fix. It does indicate that function layout is likely an important contributor to the unexpected variance and that explicit layout control is worth investigating as a mitigation.

## Why this is a performance-debugging problem

This behavior makes performance results difficult to trust and interpret:

- A correctness or DFX commit can appear to provide a 5%–8% optimization or regression.
[...]

PR #2076, opened 2026-08-31
Perf: align A5 TMR hot function entries

yanghaoran29:
## Summary

- add an internal 64-byte function-entry alignment attribute for Linux AArch64 builds
- align the A5 tensormap-and-ringbuffer submit, scheduler dispatch, and completion hot entries
- leave non-Linux and non-AArch64 targets unchanged

This applies the mitigation discussed in #2071. Without an explicit compile-time alignment requirement, unrelated code changes can move hot function entries and make performance-debugging results unpredictable. This change stabilizes the selected entry boundaries; it does not claim that alignment always improves performance.

## Binary verification

The locally built A5 onboard AICPU shared object is AArch64 and places all three selected entries on 64-byte boundaries:

- submit_task_common: 0x1c6c0
- SchedulerContext::complete_slot_task: 0x25180
- SchedulerContext::resolve_and_dispatch: 0x2edc0

Artifact SHA-256: 9a80175755e086400bc96ab480209965ad41ba77b4db3be7e706d3d978151d8a

## Testing

- editable package/runtime build: passed
- targeted pre-commit hooks for all changed files: passed
- C++ unit tests: 126/126 passed
- Python no-hardware unit tests: 2078 passed; 2 existing workflow compiler-symlink contract tests failed locally without touching the tested workflow files

No performance benchmarks were run, and this PR makes no performance-benefit claim.

### DIFF
--- a/src/a5/runtime/tensormap_and_ringbuffer/runtime/hot_entry_alignment.h
+++ b/src/a5/runtime/tensormap_and_ringbuffer/runtime/hot_entry_alignment.h
@@ -1,0 +1,18 @@
+/*
+ * Copyright (c) PyPTO Contributors.
+ * This program is free software, you can redistribute it and/or modify it under the terms and conditions of
+ * CANN Open Software License Agreement Version 2.0 (the "License").
+ * Please refer to the License for details. You may not use this file except in compliance with the License.
+ * THIS SOFTWARE IS PROVIDED ON AN "AS IS" BASIS, WITHOUT WARRANTIES OF ANY KIND, EITHER EXPRESS OR IMPLIED,
+ * INCLUDING BUT NOT LIMITED TO NON-INFRINGEMENT, MERCHANTABILITY, OR FITNESS FOR A PARTICULAR PURPOSE.
+ * See LICENSE in the root of the software repository for the full text of the License.
+ * -----------------------------------------------------------------------------------------------------------
+ */
+
+#pragma once
+
+#if defined(__aarch64__) && defined(__linux__) && (defined(__GNUC__) || defined(__clang__))
+#define SIMPLER_A5_TMR_HOT_ENTRY_ALIGN __attribute__((aligned(64)))
+#else
+#define SIMPLER_A5_TMR_HOT_ENTRY_ALIGN
+#endif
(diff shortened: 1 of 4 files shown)


ChaoZheng109 (commented):
Thanks for the careful scoping and for verifying the result in the binary — the three symbol addresses check out (all divisible by 64).

Three questions before I can tell whether this change is necessary and effective.

**1. Why does 64-byte alignment fix the variance?**

#2071 establishes that unrelated changes move hot function entry addresses, but not *why* a moved address produces a 5–8% delta — there is no I-cache / branch-predictor / PMU evidence, and the issue itself labels the layout sensitivity a hypothesis. Without a mechanism it is hard to judge whether 64 is the right constraint (#2071's own follow-up notes 16 and 32 were never compared) or whether these three entries are the right ones (also an unchecked follow-up: "Align each of the three functions independently to identify which entries matter").

**2. Why is the same alignment not applied to a2a3?**

All three functions exist under `src/a2a3/runtime/tensormap_and_ringbuffer/` under the same names (`orchestrator.cpp:884`, `scheduler_dispatch.cpp:892`, `scheduler_completion.cpp:86`), the two arches' `target_compile_options(aicpu_kernel ...)` are identical, and the TMR runtime sources differ by about 1.6% in size. Has the paired commit/parent comparison been run on a2a3?

If it has not, the a5-only scope rests on "no data elsewhere" rather than "a2a3 is unaffected" — but the name `SIMPLER_A5_TMR_HOT_ENTRY_ALIGN` reads as the latter to anyone picking this up later.

**3. What is the measured result?**

The PR verifies the binary layout and functional regression, but carries no performance data. Two numbers would settle it:

- cross-commit variance with this change landed, in the form it lands (#2071's data is `aligned Commit` vs `aligned Parent`, i.e. applied to both sides, which is not what merging this does);
- full a5 benchmark suite, `main` vs `main` + this PR. #2071 records that aligning these same three entries on main cost **7.63% on BGEMM's full AICore window**. Since that is exactly what this PR does, it would be good to address that number directly rather than leave it at "no performance benchmarks were run".

CI will not catch a regression here: `_st-npu-a5.yml` runs the correctness pytest with no `--rounds`, and the repo has no performance gate — so a 7.63% slowdown passes green.

This is also #2071's own acceptance criterion: "Use reduced cross-commit variance with no new regressions as the acceptance criterion."

PR #2080, opened 2026-08-31
Fix: reject hbg scalar access to a tensor with a producer

poursoul:
## Summary

`host_build_graph` completes orchestration before the device scheduler starts, so `wait_for_tensor_ready()` could observe neither a producer completing nor a consumer retiring. Its consumer half compared the **host mirror's** `completed_watermark` — which only the device advances, in its own copy — against a `last_consumer_local_id` seeded to the task's own id, so the comparison was unconditionally true and any caller that reached it stalled the full 15 s `TENSOR_DATA_TIMEOUT_MS`. This closes [`docs/investigations/2026-08-hbg-consumer-wait-cannot-observe-device.md`](docs/investigations/2026-08-hbg-consumer-wait-cannot-observe-device.md) by its reading 1.

- `get_tensor_data` / `set_tensor_data` now fail immediately with `INVALID_ARGS`, naming the producer found either through the tensor's `owner_task_id` or through an overlapping TensorMap entry. A runtime allocation is rejected on the same rule: its creator completes on the host, but the buffer is uninitialized and has no host view, so there is no value to read and no write that survives.
- `last_consumer_local_id` and `completed_watermark` existed only for that wait and are gone with it. Removing the watermark also takes `update_completed_watermark()`'s per-completion CAS prefix walk out of the scheduler completion path on both arches; device fanin readiness reads `completion_flags[]`, which is untouched.
- `ChipTaskSlotState` is reordered widest-first and `ChipTaskState` gets a `uint8_t` base, so the record stays one cache line with no interior padding (`reserved` 16 → 23 bytes, pinned by a new `offsetof` assert). `SharedMemoryTaskHeader` is now 64 bytes and `SharedMemoryHeader` 128.
- `TENSOR_WAIT_TIMEOUT` (8) is raised only by `tensormap_and_ringbuffer` now; the number stays reserved across runtimes.
- Docs updated in the same change: both arches' `SCALAR_DATA_ACCESS.md` / `RUNTIME_LOGIC.md`, the dsv4 example README's "runtime gap" note, `war-anti-dependency.md`, and the error-code troubleshooting pages.

## Testing

- [x] `host_build_graph_validation` gains three negative cases — reading a submitted task's output, writing an external tensor a task produces, reading an `alloc_tensors` output — all asserting status `-5` / `orch_error_code=5`; 8/8 pass on a2a3sim
- [x] Simulation tests pass: `pytest examples tests/st --platform a2a3sim --runtime host_build_graph` (18 passed / 7 skipped) and `--platform a5sim` (15 passed)
- [x] `ctest -LE requires_hardware`: 126/126, including the `test_hbg_sm_compaction` layout case
- [x] All four hbg runtime variants build (a2a3sim / a5sim / a2a3 / a5)
- [ ] Hardware tests — the onboard-only hbg cases (`available_aicore_counts`, `paged_attention_unroll`, `batch_paged_attention`, the examples) are the ones that exercise `set_tensor_data` and the modified completion path, and have not been run yet


zhusy54 (dismissed):
Verdict: Request changes. Producer detection currently relies on TensorMap, but manual dependency policies intentionally bypass TensorMap, so the implementation does not yet reject every producer-backed tensor. See the inline must-fix comment.

src/common/host_build_graph/shared/runtime_core.cpp
```diff
@@ -169,164 +159,40 @@ void rt_report_fatal(RuntimeContext *rt, int32_t error_code, const char *func, c
...
-        if (owner.is_valid()) {
-            const auto *slot = resolve_producer(owner);
-            if (slot == nullptr) return;
-            try_push(*slot);
-        }
-
-        // Step B: modifier writer lookup (OverlapMap), direct callback
+    TaskId producer = tensor.owner_task_id;
+    if (!producer.is_valid()) {
         orch.tensor_map.lookup(tensor, [&](ChipTensorMapEntry &entry, OverlapStatus) -> bool {
```
zhusy54, line 175: **Must fix: manual dependency paths can hide a real producer.**

This code treats `owner_task_id` and TensorMap entries as exhaustive sources of producer information. However, `register_task_outputs()` returns immediately in a manual scope and skips tensors with `tensor.manual_dep`. As a result, an external host-staged tensor (with an invalid `owner_task_id`) can be submitted as `INOUT` or `OUTPUT_EXISTING` through either path and then passed to `get_tensor_data()` or `set_tensor_data()` during orchestration. With no TensorMap entry, `require_no_producer()` returns true: get reads stale pre-kernel host bytes, while set succeeds even though the kernel will later overwrite the value. This violates the PR's contract that scalar access must reject every producer-backed tensor.

Please decouple writer/producer tracking from automatic dependency construction (for example, maintain a separate produced-region registry), and add regression tests for both manual-scope and `manual_dep` producers.

zhusy54 (approved):

src/common/host_build_graph/shared/runtime_core.cpp
poursoul: https://github.com/hw-native-sys/simpler/issues/2084

PR #2144, opened 2026-09-07
Add: hbg in-graph early-dispatch prerequisites — recorded verdicts and a sorted CSR

ChaoZheng109:
## Summary

Early dispatch inside a Graph body needs two things the Definition did not carry, and one of them pays off on its own. Nothing here turns in-graph early dispatch on — that is the follow-up that wires the publish chain to in-graph tasks.

**Qualification moves to recording time.** `graph_fill_definition` decides `ED_FLAG_CANDIDATE` and `ED_FLAG_TRACKED` against the Definition's own fanin CSR, applying the same conjunction the top-level submit path applies to a payload's inline fanin: every producer carries `allow_early_resolve`, the task has no dispatch predicate, its shape is dispatchable, and its internal fanin is at least one. That last term excludes a body root, whose real gate is the outer shell's activation rather than this CSR. A body's structure is fixed by its Definition, so the verdict is decided once per recorded shape and every execution replays it — per-invocation cost is zero.

The caller's early-resolve intent now survives into the recording, since that is the input being qualified, and is cleared on the way into the Definition instead. No in-graph task reaches the device carrying the bit, as before.

The device validates the flags but nothing propagates them to a slot, so **this half is inert on device**.

**A candidate's CSR row is sorted by ascending producer index.** The builder emits a producer before its consumers, so a sorted row's tail names its deepest producer — the entry a reverse scan should bet on. Non-candidate rows keep their record order, so a body with no early-resolve flags schedules exactly as it did before. That scoping is deliberate: an unconditional sort drifted flag-free workloads by ~0.5% in the qwen A/B done for #2095.

**`graph_first_unmet_producer` scans backward** to make that bet, which finishes #1924 — the top-level classifier was flipped then and its in-graph twin in the same file was not. It now resumes at a wake-scan cursor, and the drain routes a single-producer waiter straight to ready rather than rescanning a row whose only entry is the producer that just completed. Both reduce wake-list transfers and the CAS traffic they put on the lists.

Two mechanical consequences worth calling out in review:

- The classifier returns the consumer's own **CSR row index** rather than a producer index, since a row index is what a cursor can mean; `graph_producer_at` maps it back. All call sites move with it.
- `wake_scan_cursor` widens to `uint16_t`. An in-graph row is bounded by the in-graph task cap, not by `CHIP_MAX_FANIN`, so a byte would silently truncate on a large body. Its early-dispatch twin still indexes a payload's inline fanin alone and stays a byte, with a `static_assert` on each. `ChipTaskSlotState` remains one cache line.

## Testing

Re-verified after rebasing onto main now that #2130 has landed:

- [x] Simulation tests pass — a2a3 `host_build_graph` 12 passed / 7 skipped, a5 `host_build_graph` 13 passed
- [x] C++ unit tests — 137/137, including a new `test_hbg_graph_ed_qualification` (both arches) that drives the real recording path to pin the Definition's verdicts and a candidate's sorted row, and asserts a non-candidate row keeps its record order
- [x] Device A/B against `upstream/main`, 100 rounds, 8 cases, both arms sequential on pinned die 12, baseline built in its own worktree venv. Every device delta lands in -0.13% .. +0.44%, nothing near the 2% threshold, `qwen3_14b_decode` +0.01%. Full table in [this comment](https://github.com/hw-native-sys/simpler/pull/2144#issuecomment-5580444937).

  The direction flip is therefore device-neutral on this corpus, which is the honest claim — not that it helps. Worth reading alongside the coverage note above: until this PR no scene test produced an in-graph candidate, so every row the flip lands on today is one whose backward-scan rationale is heuristic rather than exact.

Checks made while reviewing the change, recorded so they need not be re-derived: the fanout CSR build and the device-side topology validation read a fanin row as an unordered set, so sorting has no positional reader; the cursor is reset per materialization by `reset_for_reuse`, so no stale cursor can survive into a later execution and shorten a scan.

### DIFF
--- a/src/common/host_build_graph/host/orchestrator.cpp
+++ b/src/common/host_build_graph/host/orchestrator.cpp
@@ -1086,7 +1086,10 @@ std::optional<GraphDefinition> graph_layout_definition(const GraphRecording &rec
 // Every section is written in full here, so the destination's prior content does not
 // reach the device — with one exception, `fanout_offsets`, which is accumulated
 // rather than assigned and is therefore zeroed below before its first increment.
-// The alignment slack between sections is written by nobody and read by nobody.
+// Two kinds of byte are written by nobody and read by nobody: the alignment slack
+// between sections, and `InGraphTaskDefinition`'s interior padding, which the
+// per-field assignment below cannot reach and a static_assert on that struct's
+// size pins against further growth.
 bool graph_fill_definition(const GraphRecording &recording, GraphDefinition definition, std::byte *image) {
     if (image == nullptr) return false;
     always_assert(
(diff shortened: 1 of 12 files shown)


src/common/host_build_graph/host/orchestrator.cpp
```diff
@@ -1133,18 +1133,51 @@ bool graph_fill_definition(const GraphRecording &recording, GraphDefinition defi
...
+        // execution replays it, so the verdict costs nothing per invocation.
+        bool ed_candidate = source.fanin_count > 0 && !source.task_attrs.has_predicate() &&
+                            source.active_mask.to_shape() != ResourceShape::DUMMY;
+        const size_t row_begin = fanin_cursor;
         for (int32_t f = 0; f < source.fanin_count; ++f) {
             const int32_t producer = recording.internal_fanins[source.fanin_offset + f];
             if (producer >= i) return false;
             fanin_indices[fanin_cursor++] = static_cast<uint16_t>(producer);
             fanout_offsets[producer + 1]++;
+            if (!recording.tasks[producer].task_attrs.allow_early_resolve()) ed_candidate = false;
```
zhusy54, line 1156: [must-fix] Graph-recorded hidden allocation tasks are currently created with `TaskAttrs{}` in `alloc_tensors()`, while the ordinary path explicitly sets `allow_early_resolve` because an allocation is a transparent producer whose output is ready at creation. As a result, this check disqualifies a consumer whenever one of its direct producers is a hidden allocation, even if its other producer(s) opted into early resolve.

I reproduced this with a graph consumer depending on a flagged normal producer plus a hidden allocation producer: the expected candidate flag was nonzero, but the recorded flag was 0.

Please preserve the same transparency semantics when recording graph allocations (for example, record the allocation task with `allow_early_resolve` set) and add a regression test covering this case.

ChaoZheng109: fixed

poursoul (approved):

PR #2157, opened 2026-09-08
feat(a5): add RDMA deferred completion backend for tensormap_and_ringbuffer

wxwnnzdyd:
## Summary

  This PR introduces the RDMA (HNS1825 RoCE) asynchronous transfer and deferred completion backend for the A5 `tensormap_and_ringbuffer` runtime.

  With pto-isa adding `TGET_ASYNC` and `TPUT_ASYNC` support under `DmaEngine::RDMA`, this change completes the runtime glue in `simpler`, enabling AICore
  kernels to issue asynchronous RDMA READ/WRITE operations and defer task retirement to the AICPU CQ/SQ poller.

  The feature is gated via the opt-in CMake / environment flag `SIMPLER_ENABLE_PTO_RDMA_WORKSPACE=ON`, ensuring zero regression on the default SDMA and URMA
  backends.

  ---

  ## Key Changes

  ### 1. Backend Kernel Adapter (`src/a5/runtime/tensormap_and_ringbuffer/runtime/backend/rdma/`)
  - **`rdma_completion_kernel.h`**:
    - Implements `RdmaTget` and `RdmaTput` descriptors mapping to `pto::comm::TGET_ASYNC` / `TPUT_ASYNC<DmaEngine::RDMA>`.
    - Calculates remote WQE target addresses using the peer's MR base address (`PeerMrBaseAddr`) plus local window offset, bypassing the need for peer VA
  mapping in device context.
    - Registers completion events with engine `COMPLETION_ENGINE_ROCE` and type `COMPLETION_TYPE_RDMA_EVENT_HANDLE` (3).

  ### 2. AICPU Scheduler Poller (`src/a5/runtime/tensormap_and_ringbuffer/runtime/`)
  - **`rdma_completion_scheduler.h`**:
    - Implements lock-free polling of HNS1825 Send/Receive Completion Queues (`RdmaCqCtx`) and updates Send Queue tail (`RdmaWqCtx`).
    - Includes detailed queue state dump diagnostics (`log_rdma_event_handle_snapshot`) for stall triage.
  - **`async_wait.h` & `aicore_completion_mailbox_types.h`**:
    - Registers `COMPLETION_TYPE_RDMA_EVENT_HANDLE` into the global completion backend table (`completion_backend_ops_for`).
    - Implements `AsyncWaitList::log_diagnostics` hook wired into shutdown stall snapshots.
  - **`scheduler_completion.cpp`**:
    - Invalidates the `DeferredCompletionSlab` cache line prior to reading error codes and condition counts.

  ### 3. Host Communication & Workspace Setup (`src/a5/platform/onboard/host/comm_hccl.cpp`)
  - **Low-segment Memory Allocation**:
    - Implements `aclrt_malloc_low_segment` to ensure symmetric windows fall below the VMM high region (`0x124000000000`), allowing HNS1825 NIC DMA
  reachability.
  - **Bootstrap Resolution**:
    - Automatically resolves RoCE physical IDs and local IPs from `hccl_rootinfo.json`, `virtualTopology.xml`, environment variables
  (`PTO_ROCE_LOCAL_IP`/`PTO_ROCE_IPS`), and `hccn_tool`.
  - **SQ Doorbell Class-of-Service Patch**:
    - Adds `patch_rdma_workspace_db_cos` to automatically patch SQ context `dbCos` in the device workspace from 0 to the NIC-configured value (4).

  ### 4. Build System & Tooling
  - **`src/a5/platform/onboard/host/CMakeLists.txt`**:
    - Adds `SIMPLER_ENABLE_PTO_RDMA_WORKSPACE` overlay option with strict mutual exclusion against `SDMA` and `URMA`.
    - Adds pto-isa `pkg_inc` to host include search paths.
  - **`simpler_setup/runtime_builder.py`**:
    - Forwards `SIMPLER_ENABLE_PTO_RDMA_WORKSPACE` to host CMake builds and automatically disables the default URMA overlay when RDMA is explicit.
  - **`simpler_setup/kernel_compiler.py`**:
    - Forwards `-DPTO_RDMA_SUPPORTED` and `-DPTO_RDMA_BACKEND_HNS_1825_SUPPORTED` to incore compiler invocations.
    - Automatically discovers CANN ascend device include directories needed for backend headers.

  ### 5. Verification & Tests
  - **Hardware Demo**:
    - `examples/a5/tensormap_and_ringbuffer/rdma_deferred_completion_demo/`: end-to-end AIV TGET / TPUT / Consumer verification on 2 ranks.
  - **Unit Tests**:
    - `tests/ut/py/test_rdma_backend_source.py`: source contract and interface consistency tests.
    - `tests/ut/cpp/a5/test_rdma_completion_scheduler.cpp`: unit tests for synthetic CQ/SQ polling logic, registered in `tests/ut/cpp/CMakeLists.txt`.
  - **Documentation**:
    - Updated `docs/comm-domain.md` and `docs/developer-guide.md`.

  ---

  ## Hardware Validation

  Tested on real A5 NPU hardware (Device pair 1-2):
  - **Command**:
    ```bash
    SIMPLER_ENABLE_PTO_RDMA_WORKSPACE=ON pytest examples/a5/tensormap_and_ringbuffer/rdma_deferred_completion_demo/ -p a5 -d 1-2
  - Results:
    - 64 elements (256 B): PASSED (Multi-chunk WQE issue & deferred retirement verified)
    - 16384 elements (64 KiB): PASSED (Data integrity and full consumer pipeline verified)
    - All status words return kOk (0).

### DIFF
--- a/docs/comm-domain.md
+++ b/docs/comm-domain.md
@@ -216,7 +216,7 @@ symmetric window is realized:
 | Window memory | POSIX shm + `ftruncate`, mmap'd per rank | a2a3: Fabric V2 handle exchange (`ACL_MEM_SHARE_HANDLE_TYPE_FABRIC`), falling back to VMM + shareable-handle IPC where Fabric is unsupported. a5: VMM shareable handles only. Cross-card P2P via `aclrtDeviceEnablePeerAccess` on both |
 | Subset barrier | shm-header atomic, `allocation_id`-scoped | file barriers, `allocation_id`-scoped |
 | Window init | window zeroed before the subset barrier (`memset`) | window zeroed before the handle is announced (`aclrtMemset`) |
-| Async-DMA workspace | opt-in per Worker (`enable_sdma`, provisions inert 16 KiB scratch without STARS streams) | a2a3: opt-in per Worker (`enable_sdma`); a5: SDMA by default, URMA as an opt-in alternative |
+| Async-DMA workspace | opt-in per Worker (`enable_sdma`, provisions inert 16 KiB scratch without STARS streams) | a2a3: opt-in per Worker (`enable_sdma`); a5: SDMA by default, URMA or RDMA as an opt-in alternative |
 
 The window is zero-initialized on both backends so scratch/signal protocols see
 a known starting state (matching the historical static-path contract).
@@ -277,6 +277,17 @@ kernel's `get_dma_workspace` is unverified rather than closed. Simulation, a5,
 and provider-disabled builds fail Worker init fast when it is set. A5 provisions
 its communication-context SDMA workspace by default; this is separate from
 the Worker-level workspace mechanism controlled by `enable_sdma`.
+
+On a5, `SIMPLER_ENABLE_PTO_RDMA_WORKSPACE=ON` selects the HNS1825 RDMA
+workspace overlay instead. RDMA is mutually exclusive with the SDMA and URMA
+overlays because `CommContext` exposes a single `workSpace`/`workSpaceSize`
+pair. The first RDMA integration only supports `orch.allocate_domain()`
+dynamic domains; it does not support `comm_derive_context()`, sim platforms,
+a2a3, or `host_build_graph`. RDMA kernels derive remote WQE addresses from the
+RDMA workspace peer MR base plus a local-window offset, not from
+`CommContext::windowsIn[peer]` — the symmetric window is a low-GM-segment
+`aclrtMalloc` buffer the HNS1825 NIC can DMA to, and peers are reached over
+RoCE by rkey + VA.
 
 ---
 
--- a/docs/developer-guide.md
+++ b/docs/developer-guide.md
@@ -214,6 +214,11 @@ A `pto_isa.pin` bump changes the SDMA headers embedded by
 `host_runtime.so`. Install-time runtime builds and run-time kernel compilation
 both read `pto_isa.pin`; use a different ISA revision by updating that file.
 
+The a5 RDMA workspace is an explicit opt-in overlay enabled with
+`SIMPLER_ENABLE_PTO_RDMA_WORKSPACE=ON`. It requires a `pto_isa.pin` revision
+that provides the PTO-ISA RDMA/HNS1825 headers and is mutually exclusive with
+the SDMA and URMA workspace overlays.
+
 ### Runtime binary lookup
 
 Scene tests load pre-built runtime binaries from `build/lib/`. These are produced
--- a/examples/a5/tensormap_and_ringbuffer/rdma_deferred_completion_demo/kernels/aiv/kernel_rdma_deferred_completion_consumer.cpp
+++ b/examples/a5/tensormap_and_ringbuffer/rdma_deferred_completion_demo/kernels/aiv/kernel_rdma_deferred_completion_consumer.cpp
@@ -1,0 +1,120 @@
+/*
+ * Copyright (c) PyPTO Contributors.
+ * This program is free software, you can redistribute it and/or modify it under the terms and conditions of
+ * CANN Open Software License Agreement Version 2.0 (the "License").
+ * Please refer to the License for details. You may not use this file except in compliance with the License.
+ * THIS SOFTWARE IS PROVIDED ON AN "AS IS" BASIS, WITHOUT WARRANTIES OF ANY KIND, EITHER EXPRESS OR IMPLIED,
+ * INCLUDING BUT NOT LIMITED TO NON-INFRINGEMENT, MERCHANTABILITY, OR FITNESS FOR A PARTICULAR PURPOSE.
+ * See LICENSE in the root of the software repository for the full text of the License.
+ * -----------------------------------------------------------------------------------------------------------
+ */
+#include "rdma_deferred_completion_common.h"
+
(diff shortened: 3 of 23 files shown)


wxwnnzdyd:
Addressed in ea8fd954. Summary and responses below.

## Required changes

**1. [P2] Advance the SQ reclamation state when consuming completions**

Fixed. `update_tail_info()` now writes the SQ tail mirror alongside the CQ tail
and CQ doorbell, using the same consumer index — matching the pinned PTO
`RingCqDoorbell`, which writes both mirrors from the same `curTail`.

On coordinating the two completion-consumption paths: the AICPU scheduler poller
is the sole CQ consumer for deferred completions, and the AICore post path's
drain is a queue-full backstop. Both advance the same monotonic index, so an
AICore-side drain consumes only the CQEs still outstanding at that moment and
the poller never re-consumes them. This is documented in `docs/comm-domain.md`
under "Completion consumption".

The `...Advances...SqTail` tests were asserting that the tail did *not* advance,
as you noted; they now assert the advance, and a new case reuses one workspace
across the ring-depth threshold.

**2. [P1] Explicitly reset backend options when reconfiguring a cached build**

Fixed. `SIMPLER_ENABLE_PTO_URMA_WORKSPACE` and `SIMPLER_ENABLE_PTO_RDMA_WORKSPACE`
are the host CMakeLists' only cached options, so both are now forwarded with an
explicit ON/OFF on every host configure — a value left in the CMake cache by an
earlier build can no longer survive a reconfigure. `SIMPLER_ENABLE_PTO_SDMA_WORKSPACE`
is not forwarded: it is an internal marker the CMakeLists derives from those two,
and a `-D` value for it is overwritten unconditionally, so it carries no cache
state to reset. Tests cover ON->OFF, ON->unset, and RDMA->SDMA against the same
build directory.

## Contract and compatibility

**3. Same-peer multi-core submission**

Confirmed as an intentional single-producer constraint, now documented in
`docs/comm-domain.md`: callers must serialize submissions to a given peer/QP.
The demo's inter-kernel marker dependencies are what order its submissions to
each peer. We are not proposing multi-producer support in PTO-ISA here.

**4. Compatibility with kernels using mapped peer addresses**

Confirmed: the RDMA domain is intentionally RDMA-only. Under the overlay,
`domain_alloc_via_ipc()` skips peer import/mapping, so `windowsIn[peer]` stays
zero and there is no locally addressable mapped VA for a peer window. A peer MR
address (rkey + VA) is not a substitute — only the NIC can dereference it. This
caller precondition is now stated explicitly in `docs/comm-domain.md`; switching
a workload from SDMA/URMA to RDMA is a kernel contract change, not a
configuration flip.

## Test coverage

**5. [P2] Add an RDMA-enabled A5 end-to-end CI configuration**

We tried this and removed it. The CI A5 runners have no usable RoCE endpoint, so
the lane cannot exercise NIC submission at all:

    [RoCE] HcommEndpointCreate failed: 19
    [comm rank 1] RDMA workspace init failed (result=2 rank_id=1 rank_count=2 ...)

Both RDMA workspace inits fail at `HcommEndpointCreate`, before any transfer is
posted. Adding the lane therefore only produces a permanently red check without
validating anything the source-contract and synthetic CQ tests do not already
cover.

End-to-end RDMA validation is currently manual, on a RoCE-capable A5 pair, with
`SIMPLER_ENABLE_PTO_RDMA_WORKSPACE=ON`; the demo (64 and 16384 elements, 2 ranks)
passes with `status = kOk`. If a RoCE-capable A5 runner becomes available, the
lane can be re-added.

## Non-blocking

**6. [P3] Bound diagnostic doorbell re-ringing and logging**

Removed. The probe's rate limit tracked only the previous handle, so alternating
pending handles re-rung the SQ doorbell and logged on every poll. The condition
it was probing (a `dbCos` mismatch causing the NIC to ignore the SQ doorbell) has
been fixed, so the probe was no longer needed.

**7. `patch_rdma_workspace_db_cos` and its failure handling**

Failure now propagates: the function returns `bool` and `init_rdma_workspace()`
fails when the patch cannot be applied, rather than logging and continuing. The
patch is required for correct operation — with a mismatched `dbCos` the NIC
silently ignores SQ doorbells — so a failed patch leaves the workspace unusable.

On the hardcoded value: 4 is the class of service the NIC is configured with,
observed from native pto-isa. Deriving it from an authoritative source is the
right long-term fix, but that requires an interface PTO-ISA exposes for it; we
keep the constant with a comment recording where the value comes from.

---

## CI status

The code-related checks pass, including `st-onboard-a5` (which exercises the
allreduce non-Sum cases), both `st-sim-*` matrices, `ut` / `ut-a2a3` / `ut-a5`,
`packaging-matrix`, and `pre-commit`.

Two checks are red from self-hosted runner failures rather than this change:

- `ut-a5` failed at the **"Set up environment"** step (dependency install) and
  never reached the tests. It passed on healthy runners in adjacent runs.
- `st-onboard-a2a3` reported 108 x `simpler_init failed with code 507018`
  (device init) on one runner, while `st-deepseek-onboard-a2a3` and
  `st-network1-onboard-a2a3` passed on other a2a3 runners in the same run.

Both are being retried; the failures do not touch the paths this change affects.

YunjiQin (approved):

PR #2168, opened 2026-09-09
Refactor: stage hbg device args through the retained temporary buffer

### LINKED ISSUE #2151: [Code Health] hbg tensor staging is the last bind-path storage still malloc'd per run
### Category

### Component

Host Runtime

### Description

`host_build_graph`'s bind path still does a per-run `device_malloc` + H2D + `device_free` for **every** host-side tensor argument. `tensormap_and_ringbuffer` stopped doing that in #1198, which added a runner-scoped retained staging buffer (`RetainedTempBump`) that grows to the high-water packed size and is bump-sliced each run, so steady-state converges to **zero** temporary device allocations.

This is not "port a foreign mechanism" — it is an inconsistency inside hbg itself, and hbg is now the only holdout on its own bind path. `DeviceRunnerBase` carries four per-pipeline-slot retained storages, and hbg has been converted to three of them:

| Retained storage | Declared at | hbg uses it? |
| --- | --- | --- |
| Graph Definition blocks | `device_runner_base.h:1253` | yes — #1988 |
| Host SM mirror | `device_runner_base.h:1273` | yes — #2013 |
| ChipCallable / orch SO buffer | `device_runner_base.h:1123` | yes |
| **Temp arg-staging buffer** | `device_runner_base.h:1237` | **no — TRB only** |

The last row's own comment scopes itself to one runtime — *"Retained temporary buffer for **TRB** device-arg staging, one per pipeline slot"* — even though nothing about the slot is TRB-specific: it is just a remembered `{addr, size}` per `pipeline_slot`, and `HostApi` has exposed it generically since #1198 (`host_api.h:54-55`). **No platform change is needed to adopt it; TRB is simply the only consumer so far.**

Why it matters beyond tidiness: #1841 measured an a5 hbg qwen3-14b decode where the bind H2D phase was dominated by device-side *allocation* rather than by PCIe/DMA transfer (5.5 MB copied, 107 ms wall). That issue's specific object (`acquire_graph_execution_buffer`) no longer exists in the tree and the graph-side allocations have since been retained by #1947 / #1988 / #2013, so per-tensor arg staging is the remaining un-retained allocation on that path.

One note on where future evidence will come from. #2041 made the qwen decode driver keep every parameter device-resident, and a device-memory tensor takes the `is_device_memory()` pass-through at `runtime_maker.cpp:1184` — skipping the whole staging block, allocation included. As more callers become device-resident they stop exercising this path, so the case for fixing it should be measured against callers that still hand over host tensors, which is what the scene-test corpus and the external repos still do.

### Location

hbg — allocate and free per run, per tensor:

- `src/a2a3/runtime/host_build_graph/host/runtime_maker.cpp:1194` — unconditional `api->device_malloc(size)` for each host-side tensor
- `src/a2a3/runtime/host_build_graph/host/runtime_maker.cpp:1449` — `api->device_free(tensor_pairs[i].dev_ptr)` in the end-of-run cleanup loop (`:1461` clears the table)
- `src/a5/runtime/host_build_graph/host/runtime_maker.cpp:1548` and `:1803` — same, arch sibling

Neither hbg `runtime_maker.cpp` mentions `retained_temp` or `RetainedTempBump` (`grep -c` = 0 in both).

Reference implementation (TRB):

- `src/a2a3/runtime/tensormap_and_ringbuffer/host/runtime_maker.cpp:256` — `RetainedTempBump`: pack to a 1024-aligned required size, read the slot, grow only when too small, bump-slice
- `src/a2a3/runtime/tensormap_and_ringbuffer/host/runtime_maker.cpp:530` / `:537` — slice when a bump is present, else fall back to `device_malloc`
- `src/a2a3/runtime/tensormap_and_ringbuffer/host/runtime_maker.cpp:224` — `release_tensor_leases` honours the release kind

Platform side (already generic, no change expected):

- `src/common/platform/include/common/host_api.h:54-55` — `get/set_retained_temp_buffer`
- `src/common/platform/onboard/host/device_runner_base.cpp:162,172` — the slot accessors
- `src/common/platform/onboard/host/device_runner_base.cpp:278` — `clear_temporary_buffer()` frees the slot once at finalize

### Proposed Fix

Give hbg the same mechanism, mirroring TRB rather than inventing a second shape. Two deltas are specific to hbg and are where the work actually is:

1. **hbg's `TensorPair` has no release kind.** `src/common/host_build_graph/runtime.h:105` carries only `{host_ptr, dev_ptr, size, needs_copy_back}`, and the cleanup loop assumes every recorded `dev_ptr` was `device_malloc`'d this run. TRB replaced that assumption with `TensorLease` + `TensorReleaseKind{Free, BufferNoop, ExternalNoop}` in #1198. Without it a bump slice would be passed to `device_free`. Since #2060 this struct is shared by both arches, so it is a single edit rather than one per sibling — but the two `runtime_maker.cpp` call sites are still per-arch and must move together.
2. **Re-derive the safety argument for hbg.** #1198 justified TRB's version with "TRB runs are already serial per runner (enforced by the static arena pool), so there is no correctness reason to gate it." hbg needs its own statement — most likely leaning on the slot already being per-`pipeline_slot` rather than on a global serialization claim.

Also worth checking while in there: hbg registers each staged `dev_ptr` with `HostTensorAccessor` (`runtime_maker.cpp:1232`) so host orchestration can read control tensors. The H2D currently precedes that registration, so a reused slice cannot expose a previous run's bytes through the accessor — a port must preserve that ordering.

Related: #1841 (the perf observation this sits under; its own object is gone from the tree and it likely wants closing or refreshing), #1988, #2013, #2041, #1198 (the TRB original).

Not covered by PR #1854: that change is one layer up — it makes the ST framework pre-upload `IN` tensors so they arrive as device memory and skip staging entirely when `--rounds > 1`. `rounds == 1`, every `OUT`/`INOUT`, and any non-scene-test caller still pay a `device_malloc` + `device_free` round trip per run.

### Priority

ChaoZheng109:
Fixes #2151

## What

`host_build_graph`'s bind path did a `device_malloc` + H2D + `device_free` for **every** host-side tensor argument on **every** run. `tensormap_and_ringbuffer` stopped doing that in #1198, which added a runner-scoped retained staging buffer that grows to the high-water packed size and is bump-sliced each run, so steady state converges to **zero temporary device allocations**.

Nothing about that mechanism was ever TRB-specific: `HostApi` has exposed `get/set_retained_temp_buffer` generically since #1198, and hbg already consumes three of `DeviceRunnerBase`'s four per-pipeline-slot retained storages (Graph Definition blocks #1988, SM mirror #2013, ChipCallable/orch SO). This makes it consume the fourth. No platform change was needed.

## How

**One definition of the mechanism, and of the ledger.** `RetainedTempBump` moves out of the two TRB `runtime_maker.cpp` files into `src/common/utils/retained_temp_bump.h`; the `TensorLease` ledger moves out of the three `runtime.h` files into `src/common/utils/tensor_lease.h`, with its release loop in `utils/tensor_lease_release.h`. The type is split from its release because both `runtime.h` files are included by AICore and AICPU translation units, which cannot see the host API.

`RetainedTempBump::begin()` takes a byte count instead of `ChipStorageTaskArgs`, which keeps that header free of `task_interface` types; the packing that produces the count is shared too, in `utils/temp_buffer_plan.h`. Diagnostics the helpers used to log are now emitted by the callers, which have the tensor index and the logging backend.

**TRB does not come through unchanged, and the difference is a fix.** Kernels require 1024-byte-aligned device pointers, and the old `RetainedTempBump` got them by assuming `device_malloc` had already aligned the base — true onboard, false on sim, where `device_malloc` is `std::malloc` and guarantees only `max_align_t`. So every slice TRB handed out on `a2a3sim` / `a5sim` was as misaligned as its base, and nothing could see it: a sim "device" pointer is host memory and never faults. `begin()` now over-allocates by `kAlignment - 1` and aligns inside the allocation, keeping the raw pointer in the slot for `device_free`. TRB therefore allocates 1023 bytes more and re-aligns, which is why three assertions in `test_trb_runtime_temp_buffer.cpp` move with it.

After this the path carries no duplicated logic at all: the bump, the packing, the lease type and its release are each one definition shared by both runtimes and both architectures.

**hbg's ledger takes TRB's shape.** `TensorPair` becomes `TensorLease` and gains a `TensorReleaseKind`; `validate` releases through `release_tensor_leases` instead of `device_free`ing every recorded pointer, because a bump slice is not a separate allocation and must not be freed. Slices are recorded as `BufferNoop`; the buffer itself is freed once per Worker in `DeviceRunnerBase::clear_temporary_buffer`.

**The sizing and the slicing cannot disagree.** `packed_temp_bytes` is one shared definition, and each runtime's argument loop skips exactly what it skips (`is_device_memory()`, `size == 0`). That pairing is what the single copy protects: a tensor counted by the sizing but not sliced by the loop — or the reverse — shifts every later slice off the offsets the buffer was sized from, and nothing detects it until a kernel reads the wrong bytes.

**bind now clears the ledger on entry.** A run whose `validate` never executes otherwise leaves its entries for the next bind. On the onboard native path a failed *prepare* still reaches `validate_runtime_impl` via `cleanup_failed_prepare`, so that path was already clean; the gap is a *finalize* whose `attach_current_thread` fails, which skips validation entirely. Stale entries were a leak before; with a reused buffer they name offsets the next bind re-slices, so `validate` would copy that run's bytes back to the earlier run's host pointer.

**The grow lands inside the `args` bind-phase span.** The one device allocation this phase can make is now attributed to it, so "steady state allocates nothing here" is a measurement rather than a definition — which matters because `hbg-bind-phases` is the tool this change is meant to be observed through.

## Why it is safe

The retained slot is per `pipeline_slot`. A native run holds its slot from bind through validate, and a concurrent reservation is admitted only when it selects a **distinct** slot (`device_runner_base.h`, `try_reserve_native_run`), so no other run can re-slice a buffer whose slices are live. That is the same property hbg's Graph Definition blocks and SM mirror already rely on — hbg does not need TRB's "serial per runner" argument.

The H2D of a staged tensor still precedes its registration with the run's `HostTensorAccessor`, so a reused slice cannot expose the previous run's bytes to orchestration. (`add()` uses the caller's host buffer as the view when one is supplied, so slices install no platform mapping and cannot produce overlapping `halHostRegister` calls.)

One property worth knowing, now documented in both runtimes' `RUNTIME_LOGIC.md`: a pure `OUT` tensor is staged into a slice but never copied in and never zero-filled, so bytes a kernel does not write are whatever the slice last held. That was already true of a fresh `device_malloc` (the allocator pools device memory), but the residue is now *this workload's own previous round* rather than arbitrary — a kernel writing only part of its output now fails deterministically instead of randomly. `device_memset` exists for zero-filling pure outputs; neither runtime uses it, because it would cost a device operation per output per run.

## Tests

- `test_depth_two_slots_own_separate_resources` asserted that neither hbg pipeline slot held a retained buffer. It now asserts both hold one and that the two are **distinct** — the regression barrier for this change. Verified passing onboard.
- New `tests/ut/cpp/common/test_retained_temp_bump.cpp` covers the mechanism against a fake `HostApi`: first allocation, reuse without allocation, grow (free old + malloc new), slice alignment/contiguity/disjointness, slice miss, a zero-byte run leaving the slot untouched, and a failed grow clearing the slot rather than leaving it naming the buffer it just freed.
- Existing `test_trb_runtime_temp_buffer` (both arches) stays green — that is the evidence the extraction did not change TRB.

### Verification run

| Suite | Result |
| ----- | ------ |
| `ctest -LE requires_hardware` | 141/141 passed |
| `pytest tests/ut/py` | 2242 passed, 24 skipped |
| a5sim / a2a3sim `--runtime host_build_graph` | 21 passed / 20 passed 7 skipped |
| a5sim / a2a3sim `--runtime tensormap_and_ringbuffer` | 50 passed / 55 passed 1 skipped |
| a2a3 **onboard** `--runtime host_build_graph` | 46 passed, 1 skipped |
| a2a3 **onboard** `--runtime tensormap_and_ringbuffer` | 94 passed, 18 failed — see below |

The 18 onboard TRB failures are **pre-existing and unrelated**: all are L3/L4 cases whose forked chip worker exits with `wait status 139` during `init`, before any bind. The identical 18 fail on an unmodified `upstream/main` build of the same tree, each passes when run on its own, and the set is byte-identical before and after this change.

## Note on measurement

The two cases the `hbg-bind-phases` tooling instruments (dsv4 FLASH decode, qwen3-14b decode) both hand over device-resident buffers since #2041, so they take the `is_device_memory()` pass-through and never reach this code. There is therefore **no perf number in this PR** — the benefit lands on callers that still hand over host tensors, which is what the scene-test corpus and the external repos do. #1841 is the standing observation this sits under.


src/common/utils/retained_temp_bump.h
```diff
@@ -0,0 +1,103 @@
...
+ *
+ * Reports failure through return values rather than logging: this header is
+ * included by translation units that do not all share one logging backend, and
+ * the caller has the tensor index the message needs anyway.
+ */
+class RetainedTempBump {
+public:
+    static constexpr size_t kAlignment = 1024;
+
+    static size_t align_up(size_t v) { return (v + (kAlignment - 1)) & ~(kAlignment - 1); }
```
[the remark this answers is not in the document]
ChaoZheng109: Confirmed and fixed in c185229d — this one is real, and I checked it is reachable rather than theoretical: `ChipTensor::nbytes()` is `numel() * get_element_size(dtype)` (`src/common/task_interface/tensor.h:132`), an unchecked `uint64_t` product, so a malformed descriptor can produce a value that wraps `align_up` and the packed total.

`acquire()` now tests `aligned > capacity_ || bytes > capacity_ - aligned`. I verified the old form really did hand back an out-of-bounds pointer rather than merely being untidy — reverting just that line makes the new test fail with:

```text
RetainedTempBumpTest.AnOversizedRequestCannotWrapPastTheBuffer
Expected equality of these values:
    Which is: 0xf8e2c00
    Which is: (nullptr)
```

i.e. `acquire(SIZE_MAX)` on a 2 KiB buffer returned a valid-looking pointer, which `copy_to_device` would then have written past. `AnOversizedRequestCannotWrapPastTheBuffer` covers `SIZE_MAX` and `SIZE_MAX - kAlignment + 1`, and asserts the cursor is untouched by a refusal so a later legitimate slice still fits.

One thing I did **not** do, deliberately: `align_up()` stays unchecked. Making it reject overflow means giving it a failure channel, which threads an error path through both callers' packing loops — for a case whose only effect is an undersized `required`, which `acquire()` now catches as an ordinary slice miss. So the containment is at the single point that bounds the returned pointer, and the header says so. Happy to add the checked form if you would rather have the failure raised earlier.

Note this arithmetic predates the PR — it is TRB's original `RetainedTempBump`, which this change extracts to `src/common/utils/`. Fixing it here rather than leaving it is the right call, but it is not a regression the PR introduced.

src/a2a3/runtime/host_build_graph/docs/RUNTIME_LOGIC.md
```diff
@@ -56,6 +56,34 @@ An orchestration fatal stops this sequence before the upload. The orchestrator
...
+
+A pure `OUT` tensor is staged into a slice but never copied in, and nothing
+zero-fills it, so the bytes a kernel does not write are whatever the slice last
+held. That was already true of a fresh `device_malloc` — the allocator pools and
+reuses device memory — but the residue is now *this workload's own previous
+round* rather than arbitrary. A kernel that writes only part of its output
+therefore fails deterministically rather than randomly, which a golden check is
+less likely to catch on the second and later rounds. `device_memset` exists for
+zero-filling pure outputs; neither runtime uses it, because it would cost a
+device operation per output per run.
```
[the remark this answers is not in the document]
ChaoZheng109: You are right, and the wording was mine to get wrong — fixed in c185229d across all four passages.

`begin()` neither preserves nor initializes a buffer it replaces, so "the residue is now this workload's own previous round" only holds for a run that reuses the retained buffer **unchanged**. A Worker's first run, and any run that grew the buffer, gets uninitialized allocator memory exactly as before.

I also tightened a second condition you did not raise but which follows from the same mechanism: the slices only land on the same offsets when the run's tensors pack the same way, so a shape change moves them even on an unchanged buffer. The passages now read:

> On a run that reuses the retained buffer unchanged, and whose tensors pack to the same offsets, that is the previous round's own bytes; a run that allocated or grew the buffer gets uninitialized allocator residue, since `begin()` neither preserves nor initializes a replacement buffer. A kernel that writes only part of its output therefore fails deterministically across a fixed-shape workload's steady-state rounds and randomly elsewhere.

which keeps the point worth making — a steady-state decode loop is exactly where a partial-output kernel would stop being caught — without claiming it unconditionally. The explanation that pure `OUT` staging skips `copy_to_device`, and that `device_memset` is not used because it would cost a device operation per output per run, is unchanged.

src/common/utils/retained_temp_bump.h
```diff
@@ -0,0 +1,116 @@
...
+ * every host runtime that stages device arguments.
+ *
+ * This is the whole temporary-buffer mechanism: the platform only remembers a
+ * {addr, size} slot across runs (HostApi get/set_retained_temp_buffer); the
+ * grow and slice logic lives here. A bind packs its own staged tensors to a
+ * required size, calls begin() to grow the slot to it, and acquire()s one slice
+ * per tensor. The slot is per pipeline slot, so two runs holding different slot
+ * leases never share a staging buffer.
+ *
+ * Kernels require 1024-byte-aligned device pointers, which device_malloc
```
zhusy54: `RetainedTempBump` treats a 1024-byte-aligned `device_malloc` base as a hard invariant, but the real sim path ultimately calls `std::malloc` in `src/common/platform/sim/host/memory_allocator.cpp`, which does not guarantee 1024-byte alignment. The new unit-test fake explicitly uses `aligned_alloc`, masking this backend/contract mismatch; aligning only the slice offset cannot repair a misaligned base. Please make the sim allocation path satisfy this contract and assert it through the real sim `HostApi`, or change the bump ownership model to retain raw and aligned addresses. If sim intentionally does not need this constraint, narrow and justify the cross-backend contract here.

src/a2a3/runtime/host_build_graph/host/runtime_maker.cpp
```diff
@@ -1171,7 +1198,30 @@ extern "C" int bind_callable_to_runtime_impl(
     // the point at which a task could make it stale.
     HostTensorAccessor tensor_access(api);
 
+    // A lease recorded by an earlier bind names an offset this bind is about to
+    // re-slice, so carrying one over would copy this run's bytes back to that
+    // run's host pointer. Every exit path from here on leaves the ledger owned
+    // by the caller's Runtime, and this is where it starts empty.
+    runtime->tensor_leases_.clear();
```
zhusy54, line 1210: This line fixes a real failure path: if `attach_current_thread` fails during finalize, validation is skipped; without clearing the ledger, the next bind can re-slice the retained buffer and later copy the successor run bytes into the previous run host pointer. The bump unit tests and normal two-slot scene test do not cover “validation skipped → same slot rebound → validate.” Please add either a stale-ledger bind regression test or a finalize-attach-failure injection test that verifies the old host buffer is not modified. The same test intent should cover the a5 sibling.

src/common/utils/tensor_lease.h
```diff
@@ -0,0 +1,48 @@
...
+ * and then release each one the way its provenance requires. The release kind
+ * is what keeps that decision out of the release loop's hands: a bump slice and
+ * an owned allocation are both a `dev_ptr`, and only the staging site knows
+ * which it handed over.
+ *
+ * Declared here rather than beside either `Runtime` because both are included
+ * by AICore and AICPU translation units, which is also why this header pulls in
+ * no host API: the release itself lives in `utils/tensor_lease_release.h`.
+ */
+enum class TensorReleaseKind {
```
zhusy54, line 30: This shared ownership model retains states that no longer have producers. HBG records only `BufferNoop`; the sole TRB call always passes `&bump`, making the `Free` fallback unreachable; and device-memory tensors `continue` before entering the ledger, so `ExternalNoop` is also unreachable. Consequently, the new release helper only counts no-ops and clears the vector. Prefer making the bump parameter a non-null reference and removing the obsolete per-tensor allocation fallback, `Free`/`ExternalNoop`, and their counters, leaving a staged copy-back record. Otherwise, please identify the current producer of each retained state.

examples/workers/l3/child_memory/main.py
```diff
@@ -16,8 +16,8 @@
   * ``orch.copy_to(dev_handle, host_buffer)`` — H2D upload of the weight.
   * ``Tensor.make(dev_ptr, shape, dtype, child_memory=True)`` —
     wrap the worker pointer as a tensor that the runtime treats as
-    *already on device*. ``init_runtime_impl`` skips malloc + H2D copy
-    for these and does not record them in ``tensor_pairs``, so the buffer
+    *already on device*. ``init_runtime_impl`` skips staging + H2D copy
```
zhusy54: The terminology changed here, but both backticked identifiers still disagree with the source. There is no `init_runtime_impl`; the current entry point is `bind_callable_to_runtime_impl`. The member is `Runtime::tensor_leases_`, not `tensor_leases`. Please use the actual symbols, or avoid coupling this example to an internal member name by saying “the host-side staging ledger.”

src/a2a3/runtime/host_build_graph/host/runtime_maker.cpp
```diff
@@ -1409,57 +1468,49 @@ extern "C" int validate_runtime_impl(Runtime *runtime, const HostApi *api, int e
...
 
     // Cleanup device tensors
     LOG_INFO("=== Cleaning Up ===");
-    for (int i = 0; i < tensor_pair_count; i++) {
-        if (tensor_pairs[i].dev_ptr != nullptr) {
-            api->device_free(tensor_pairs[i].dev_ptr);
-        }
-    }
-    LOG_INFO("Freed %d device allocations", tensor_pair_count);
+    release_run_tensor_leases(runtime, api);
```
zhusy54, line 1511: This change turns per-run allocation release into a retained-slice no-op, with the buffer released only at Worker finalization. However, the file banner still says this file provides the nonexistent `init_runtime_impl`, claims that `validate_runtime_impl` “Frees device memory,” and retains stale `rt2` terminology. Please update the four touched HBG/TRB, a2a3/a5 `runtime_maker.cpp` banners so their entry-point names and retained-buffer lifetime match the implementation.

src/a2a3/runtime/tensormap_and_ringbuffer/host/runtime_maker.cpp
```diff
@@ -530,7 +457,10 @@ static bool stage_device_args(
             dev_ptr = bump->acquire(size);
             release_kind = TensorReleaseKind::BufferNoop;
             if (dev_ptr == nullptr) {
-                LOG_ERROR("Retained temp buffer slice failed for tensor %d: tensor bytes=%zu", i, size);
+                LOG_ERROR(
```
zhusy54, line 461: The contract above this function still says a pure `OUT` tensor receives “an on-device zero,” while the implementation deliberately performs neither H2D nor `device_memset`, and the new `RUNTIME_LOGIC.md` text says the slice retains residue. Please correct this comment in both architecture siblings so readers do not debug against a false zero-initialization contract.

zhusy54 (commented):
Review comments are attached to the corresponding source lines. The main blocking issue is the alignment contract mismatch; the other comments cover the missing failure-path regression, obsolete lease states, readability, and documentation consistency.

src/a2a3/runtime/host_build_graph/docs/RUNTIME_LOGIC.md
```diff
@@ -56,6 +56,39 @@ An orchestration fatal stops this sequence before the upload. The orchestrator
...
+The H2D copy of a staged tensor precedes its registration with the run's host
+accessor, so a reused slice can never expose the previous run's bytes to
+orchestration.
+
+A pure `OUT` tensor is staged into a slice but never copied in, and nothing
+zero-fills it, so the bytes a kernel does not write are whatever the slice last
+held. That was already true of a fresh `device_malloc` — the allocator pools and
+reuses device memory — but the residue is no longer always arbitrary. **On a run
+that reuses the retained buffer unchanged, and whose tensors pack to the same
+offsets, it is that tensor's own bytes from the previous round.** A run that
```
zhusy54: Using the same offsets does not imply that the residue belongs to the same tensor. The retained buffer is keyed only by pipeline slot, not by callable or tensor identity, and a slot can alternate between different callables. The precise contract is that the slice contains bytes left by whichever previous tensor occupied that byte range; it is the same tensor only when the same callable and argument layout run consecutively. Please make this distinction in all four new residue discussions.

poursoul (commented):
## 检视意见

**结论:approve。** 未发现正确性缺陷;slot 独占性的安全论证经独立核验成立。下面两条 should-fix 都不阻塞合并。

### Stated goal 与 Real goal

一致。改动确实只做了三件事 —— 把机制上移到 `src/common/utils/`、让 hbg 接入、统一台账(ledger),TRB 一侧是纯搬迁。没有夹带范围外改动。

### 我独立核验过的点

- **slot 独占性是真的,hbg 确实不需要 TRB 那套"每个 runner 串行"的论证。** `try_reserve_native_run` 在 `reservation.pipeline_slot == pipeline_slot` 处直接拒绝(`device_runner_base.cpp:2010`),`allow_prepared_successor` 也绕不过这一条;reservation 在 validate 之后才释放(`onboard/host/c_api_shared.cpp:734`)。所以切片存活期内不可能被另一个 run 重新切分。
- **确实不需要平台改动,也不会漏释放。** 两个 backend 早就各自持有槽位并在 finalize 释放一次 —— `onboard/host/device_runner_base.cpp:279-283`、`sim/host/device_runner_base.cpp:515-519`。
- **arch 对称性:** a2a3 与 a5 的 diff 除上下文行外逐字相同,两个 runtime 都是。
- 抽出的 `acquire()` 相比 TRB 原版有两处严格改进,且都被新单测钉住了:边界检查改写成减法(于是溢出的 `nbytes()` 落成一次 miss,而不是交出越过缓冲末尾的指针);grow 失败时清空槽位,而不是让它继续指着刚被 free 掉的缓冲。
- CI 全绿,含两个 onboard arch —— 这同时也证明了 AICore/AICPU 的 translation unit 能解析新的 `utils/tensor_lease.h`。

### Should fix

**1. 零字节 tensor 这一条对 hbg 是行为变更,不只是 predicate 对齐。**

PR 描述把新增的 `if (size == 0)` 跳过写成 "hbg's staging loop skips empty tensors exactly as TRB's does"。但在这次改动之前,一个零字节的非 child、非纯 `OUT` tensor 会让 hbg 的 bind **直接失败**:`HostTensorAccessor::add` 对 `size == 0` 返回 false(`src/common/host_build_graph/host/host_tensor_access.cpp:72`),于是循环落到 `LOG_ERROR("host-orch: no host view for tensor %d")` 并返回 `PTO_RUNTIME_ERR_INTERNAL`。改动之后这个 tensor 被接受,以 `buffer.addr = 0` 透传。

方向是对的 —— TRB 一直就是这个语义,已经在线上跑 —— 但这是一个藏在重构里的修复:

- 只读 PR 描述的人看不出 hbg 从**拒绝**这类输入变成了**接受**;
- 没有测试。按 `.claude/rules/discipline.md` §3,行为修复应当留下回归屏障。

建议在 PR 描述和 hbg 的 `RUNTIME_LOGIC.md` 里明确写出这一条,并补一个零字节 tensor 的用例。

**2. `packed_temp_bytes` 复制了四份,而它恰好就是 PR 自己声明"绝不能分叉"的那个 predicate。**

四个副本(hbg/TRB × a2a3/a5)函数体逐字节相同,只有文档注释不同。而它们上方的注释已经把失效模式说得很准 —— "a tensor counted here but not sliced there ... shifts every later slice off the offsets this size was computed from" —— 也就是一类静默的偏移错位,唯一的防线是人的自觉。

把它排除在 `retained_temp_bump.h` 之外的理由(保持该 header 不依赖 `task_interface` 类型)是偏好而非约束:`ChipStorageTaskArgs` 就定义在共享的 `src/common/task_interface/task_args.h:197`,两个 runtime、两个 arch 用的是同一个类型。既然这个 PR 本身就是在做收敛,建议把 plan/size 这一步也一起收进来 —— `RetainedTempBump::plan(orch_args)`,或者单独一个 `utils/staged_tensor_plan.h` —— 让这个 predicate 只存在一份。这正是 PR 想守住的那个不变量。

### Consider

**3. ~~`TensorReleaseKind` 三个取值里有两个没有生产者~~ —— 撤回,与 @zhusy54 在 `src/common/utils/tensor_lease.h:30` 的意见重复,以那条为准。**

我提交时没有先读完已有的 inline 评审,这条与 @zhusy54 已经提出的完全撞车:同样指出 `ExternalNoop` 零生产者、`Free` 因唯一调用点永远传 `&bump` 而不可达,连"把 bump 参数改成非空引用"的建议也是同一个。请按那条线程讨论,不要在两处分头回复。

下面第 4 条与那条线程有部分重叠(它提到"the new release helper only counts no-ops"),仅"四份逐字相同的包装函数"这一点是额外的。

**4. `release_run_tensor_leases` 同样是四份逐字相同的 static 包装,唯一的职责是打一行日志。** `release_tensor_leases` 已经返回 counts 了,两个调用点直接打日志即可,包装可以去掉。

**5. hbg 原来的 `LOG_INFO("Freed %d device allocations")` 变成了共享 release 内部的 `LOG_DEBUG`。** 一条一直出现在 onboard 日志里的行在默认 verbosity 下消失了 —— 如果有人靠它排查释放数量,值得在 PR 描述里提一句。

### pto-isa pin

ℹ️ `pto_isa.pin` 固定在 `5a4f74cbf627d4aac2e0ce10d5e0d8b118343265`。本 PR 未改动任何 pto-isa header 引用,所以这里只是例行提醒:确认该 commit 仍然够用即可,不涉及 bump 或重建 onboard `host_runtime.so`。

### 体量

总 churn 1144 行,超过通常的 1000 行拆分阈值 —— 但其中 Core 占 763 行,而这 763 里约 200 行是纯搬迁,整体是同一个关注点镜像了四份。这个 PR 我不建议拆。

src/common/utils/retained_temp_bump.h
ChaoZheng109: Fixed — you are right that aligning only the slice offset cannot repair a misaligned base, and that the test double was hiding it.

`RetainedTempBump` no longer treats an aligned `device_malloc` as an invariant it may rely on. `begin()` over-allocates by `kAlignment - 1` and aligns the base *inside* the allocation, keeping the raw pointer in the slot because that is the one `device_free` must receive. The alignment is now a property of the class rather than a cross-backend contract nobody enforced, so it holds on sim as well without the sim allocator changing.

Both test doubles now allocate with plain `std::malloc` — the TRB one used `posix_memalign` and had the same masking problem, so it changed too. `SlicesAreAlignedEvenWhenTheBackendReturnsUnalignedMemory` asserts every slice is 1024-aligned against that deliberately unaligned backend and that the slot still names the raw allocation. Several existing expectations encoded the old "slot size == packed size" contract and were updated to the new one (raw size, `capacity() >= required`).

src/a2a3/runtime/host_build_graph/host/runtime_maker.cpp
ChaoZheng109: Fixed — the coverage gap was real, and my first read of its cost was wrong. I assumed an hbg bind test needed an orchestration `.so`; it does not. `dlopen` happens in `register_callable_impl`, and `bind_callable_to_runtime_impl` receives the resolved entry points through its `host_orch_func_ptr` parameter, so a test can pass two of its own function pointers. TRB already had the equivalent harness (`test_trb_runtime_temp_buffer.cpp`); hbg simply lacked it.

New `tests/ut/cpp/common/test_hbg_bind_ledger.cpp` drives the real bind/validate against a fake `HostApi`:

- `SecondBindDoesNotInheritTheFirstBindsLeases` — binds one `Runtime` twice with no validate between, the shape a finalize whose `attach_current_thread` failed leaves behind, and asserts the ledger holds only the second bind lease.
- `ValidateAfterARebindLeavesTheEarlierRunsBufferAlone` — writes a sentinel through every recorded slice, validates, and asserts the first run host buffer is untouched.

Verified as a barrier in both directions: with the `clear()` removed the ledger holds 2 leases instead of 1 and the first buffer goes from `0x11` to `0x5a`, i.e. the second run bytes land in the first run caller buffer.

One asymmetry worth recording rather than leaving implicit: TRB defends this twice — the clear at bind entry plus an RAII `bind_cleanup` guard that releases on every early return, dismissed on success. hbg has only the clear. Its 10 early returns after staging begins are covered instead by the callers, since every bind-failure path reaches `validate_runtime_impl` (`cleanup_failed_prepare` onboard, the equivalent in sim) with a non-zero `execution_rc`, which skips copy-back and releases. That is a longer argument than TRB internal one, and I did not add the guard here to keep this change scoped.

src/common/utils/tensor_lease.h
ChaoZheng109: Confirmed, and it is worse than the comment says — filed as #2196 rather than changed here.

`ExternalNoop` has **two** occurrences in the whole tree, the declaration and the switch arm, and zero producers. `git log -S "TensorReleaseKind::ExternalNoop"` returns only #1198 and this PR, and the #1198 diff adds just the `case`: it has had no producer since the day it was written. `Free` is reachable only through the `bump == nullptr` branch of a `stage_device_args` whose single call site always passes `&bump`. So both runtimes record `BufferNoop` exclusively and the release helper counts no-ops.

This PR did not create those states, it moved them from TRB private header into a shared one, which is what made them visible. Removing them is the right call but it touches both runtimes shared ledger and TRB staging function, whereas this change is scoped to hbg staging path — so it is #2196 with your suggested shape (non-null bump reference, drop the fallback, drop `Free`/`ExternalNoop` and their counters), including a note to re-check the single-call-site premise before deleting.

examples/workers/l3/child_memory/main.py
ChaoZheng109: Fixed — and you are right that the example should not name an internal member at all. It now reads "The runtime bind skips staging + H2D copy for these and keeps them out of its host-side staging ledger", so it names neither `init_runtime_impl` nor `tensor_leases_`.

Grepping for the same defect turned up two more: `src/common/{host_build_graph,tensormap_and_ringbuffer}/tensor.h` both said "host skips the H2D copy in init_runtime_impl". Both fixed.

src/a2a3/runtime/host_build_graph/host/runtime_maker.cpp
ChaoZheng109: Fixed — all four banners rewritten. They now name `register_callable_impl` / `bind_callable_to_runtime_impl` / `validate_runtime_impl`, drop the `rt2` spelling, describe staging as slices of the pipeline slot retained buffer, and replace "Frees device memory" with the actual lifetime: validate releases the run leases as no-ops and the buffer is freed once at Worker finalization.

One more of the same class while in there: TRB emitted `LOG_INFO("TIMING: total_init_runtime_impl = ...")`, naming a function that does not exist. Renamed to `total_bind` after confirming no tooling parses that label.

src/a2a3/runtime/tensormap_and_ringbuffer/host/runtime_maker.cpp
ChaoZheng109: Fixed in both siblings — the line now reads "nothing at all for pure-OUTPUT buffers" instead of "an on-device zero", which matches the implementation and the `RUNTIME_LOGIC.md` text.

src/a2a3/runtime/host_build_graph/docs/RUNTIME_LOGIC.md
ChaoZheng109: Fixed in all four — the distinction is exactly right and my wording asserted more than the mechanism gives.

They now say the slice contains whatever the previous tensor occupying that byte range left there, and that it is the same tensor own bytes only when the same callable runs consecutively with the same argument layout, since the slot is keyed by pipeline slot alone and can alternate between callables. The first-run and post-growth cases (uninitialized allocator residue, because `begin()` neither preserves nor initializes a replacement buffer) are kept.

poursoul (dismissed):
## 复审意见（第二轮）

**结论：approve。** 上一轮我提的两条 should-fix、以及 @zhusy54 的 blocking 对齐问题都已落地，9 条线程全部 resolved，CI 20/20 全绿。重新独立核验了对齐修复和容量账，未发现正确性缺陷。下面两条都是文档/描述口径问题，不阻塞合并。

### 独立核验通过的点

**对齐修复是真的，而且顺带补掉了一个潜在缺陷。** 旧 `RetainedTempBump` 假定 `device_malloc` 返回 1024 对齐的 base，只对 offset 做对齐。但 sim 后端的 `device_malloc` 就是 `std::malloc`（`src/common/platform/sim/host/memory_allocator.cpp:24-25`，只保证 `max_align_t`），所以**改动前 TRB 在 a2a3sim / a5sim 上分出来的 slice 绝对地址并非 1024 对齐**——sim 下"设备"是 host 内存不会 fault，所以 ST 一直是绿的，这个不一致没人看得见。新 `begin()` 超分 `kAlignment - 1` 并在分配内部重新对齐 base、slot 仍保存 raw 指针供 `device_free`，这条路径是对的，且被 `SlicesAreAlignedEvenWhenTheBackendReturnsUnalignedMemory` 钉住了。

**容量账算得住。** `capacity_ = size - head >= wanted - 1023 = required`，无论 backend 返回什么地址，`required` 字节一定可切。`acquire()` 写成 `bytes > capacity_ - aligned` 的减法形式，上游 `nbytes()` 溢出会落成一次 miss 而不是越界指针，`AnOversizedRequestCannotWrapPastTheBuffer` 覆盖了这一类。

**两个 predicate 现在不可能漂移**——因为只剩一份 `src/common/utils/temp_buffer_plan.h:44-53`；hbg / TRB 两边循环的跳过条件（`is_device_memory()` / `size == 0`）与之逐字一致。`ChipTensor` / `ChipStorageTaskArgs` 来自 `src/common/task_interface/`，是单一类型，共享 inline 函数不存在 ODR 隐患。a2a3 与 a5 四个文件的改动逐行 diff 完全一致。

**`HostTensorAccessor` 的安全论证成立。** hbg 传的 `fallback_host_view = host_ptr` 非空，走 `needs_push_back` 分支，不会调用 `register_device_memory_to_host`（`src/common/host_build_graph/host/host_tensor_access.cpp:74-84`），所以相邻 slice 不会产生重叠的 `halHostRegister`；`find_region` 按不相交区间线性扫描，切片相邻也不会误命中。

### Should fix 1 — TRB 的 `RUNTIME_LOGIC.md` 还在说"不需要 base fix-up"，与新代码直接矛盾

两份文件（a2a3 / a5，逐字相同）的这一条 bullet 现在是假的：

> packs the run's non-child, non-empty tensors to a required size, aligning each slice up to 1024 bytes (kernels require 1024-aligned device pointers, **which `device_malloc` already guarantees, so the retained base and every 1024-aligned slice stay aligned without a base fix-up**);

`src/common/utils/retained_temp_bump.h` 的头注释写的恰恰是相反的事实（sim 的 `device_malloc` 是 `std::malloc`，只保证 `max_align_t`，所以 `begin()` 超分并自己对齐 base）。这次 PR 改的就是这一行（`TRB kernels` → `kernels`），却把已经过时的后半句留下了。同段下一条 "if `required` exceeds the retained size" 也应改成 `required + 1023`。

- `src/a2a3/runtime/tensormap_and_ringbuffer/docs/RUNTIME_LOGIC.md:118-121`
- `src/a5/runtime/tensormap_and_ringbuffer/docs/RUNTIME_LOGIC.md`（同位置，需同一 commit 一起改）

### Should fix 2 — PR 描述有两处已不成立，而且掩盖了这次改动最值得说的一半

- "the only duplication left on this path is the packing function — duplicated on purpose" / "both `packed_temp_bytes` **copies** skip the same two cases"：代码里只有一份共享实现，TRB 的 `RUNTIME_LOGIC.md` 自己也写了 "Both are one copy shared by `host_build_graph` and both architectures"。描述里那整段"两份 predicate 必须逐字一致"的安全论证已经不描述这份代码了。
- "so **TRB behaves exactly as before**"：对齐修复之后不成立——TRB 现在多分配 1023 字节并重新对齐 base，`test_trb_runtime_temp_buffer.cpp` 有三处断言为此改动。更关键的是真实情况比这句话强：**这次顺带修掉了 TRB 在 sim 上 slice 未对齐的潜在缺陷**。读描述的人看到"行为完全不变"就不会去找它了，建议把它写成一条明确的收益。

squash 合并会把 body 带进 commit message，建议合并前更新。

### Consider（不影响合并）

- `src/common/utils/` 原有 5 个 header 依赖极轻，这次新进的三个把 `common/host_api.h` 和 `task_args.h` 拉进了 `utils/`，目录语义被拓宽了一点。如果后续还有 host-runtime 共享件进来，可以考虑另起目录。当前对"为什么 `tensor_lease.h` 与 `tensor_lease_release.h` 要拆开"的说明是清楚的，不算欠账。
- TRB `copy_in_device_args` 的 `bump != nullptr` 分支与 `TensorReleaseKind::Free` 目前没有生产者（唯一调用点恒传 `&bump`）。保留作为通用契约可以接受，但建议在 `tensor_lease.h` 上标一句"`Free` 当前无生产者，为无切片后端保留"，免得下一个人以为有活路径。

poursoul (approved):

PR #2183, opened 2026-09-10
Update: select A5 cluster ownership from trusted topology

yanghaoran29:
## Summary
- Select balanced contiguous AICore cluster ownership when live CPU_TOPO or a verified JSON fallback fully matches the device OCCUPY pool.
- Preserve round-robin ownership for incomplete, unavailable, or unmatched topology.
- Carry the internal mode through both A5 runtimes and cover topology selection and partition boundaries.

## Testing
- [x] `test_a5_aicpu_topology_fallback`
- [x] Syntax checks for both A5 scheduler cold paths
- [x] `clang-format --dry-run --Werror` and `git diff --check`
- [ ] Onboard hardware validation (not run)


ChaoZheng109:
## Why this branch now reverts its own first commit

I got access to an a5 box (`Ascend950DT_9581`, FG) and measured the thing this PR was trying to improve. The result says contiguous ownership moves in the wrong direction on that machine, so I reverted it and pushed a replacement rather than adjusting the gate. Pushed on top of `c612e7fa`, not over it — the original commit is untouched.

### What the measurement says

The cost this PR is ultimately about is the COND read a scheduler thread issues per owned core per completion sweep (`scheduler_completion.cpp`). Timing it in place, bucketed by the die the target core sits on:

| | same die | cross die |
| --- | ---: | ---: |
| per MMIO read | **90 ns** | **139 ns** (+54%) |

Three runs, under 5% spread. The same-die figure lines up with the ~95 ns per MMIO LDR already recorded in `.claude/rules/ascend.md`, which is the main reason I trust the number. The penalty is not uniform — it ranged 1.25x to 1.99x across the four scheduler threads — so AICPU cluster position matters too, but CPU_TOPO exposes nothing that expresses that distance.

### Why contiguous is the wrong split here

On this box, FG resolves to `ALLOWED_CPUS = [7, 5, 6, 3, 8(orch)]` → phy `[6, 4, 5, 2]` → die **`[1, 1, 1, 0]`**, while the 32 clusters split **16/16** by `physical_core_id`. Three scheduler threads on one die, clusters evenly divided across two: the two distributions do not line up, so an even split of the cluster index space hands the first two threads a full quota from the die neither of them sits on.

| ownership | same-die clusters | total MMIO per sweep |
| --- | ---: | ---: |
| `ci % active_sched_threads_` (today) | 16/32 (50%) | 11008 ns |
| contiguous (this PR's first commit) | 8/32 (25%) | 11448 ns (+4%) |
| die affinity (replacement) | 24/32 (75%) | 10569 ns (−4%) |

Separately, the gate did not guard what it selected: it asked whether the AICPU CPU_TOPO covers the OCCUPY pool — a fact about AICPU cores — and used the answer to choose how *AICore* clusters are divided, without reading the per-CPU die placement the topology actually carries.

### What replaces it

Assign each cluster to a thread on its own die, subject to an exact quota. The quota is load-bearing: a plain "same die first" pass would give the lone die-0 thread all sixteen die-0 clusters and leave its three peers splitting the rest.

The two facts needed are produced on opposite sides and neither side can derive both:

- **which die a thread sits on** — host knowledge, from `ALLOWED_CPUS` + CPU_TOPO. The device only ever sees `cpu_id`, never `phy_cpu_id`.
- **which die a cluster sits on** — device knowledge. `worker_id` is monotone in `physical_core_id`, but *which* cores CANN excludes varies between processes (observed `{14,17,61,67}` in one, `{7,17,68,71}` in another), so the host cannot predict it. Only the handshake report settles it.

So the host publishes a packed per-thread die word through a5's `KernelArgs`, and the device derives the cluster side itself. Going through `KernelArgs` rather than `Runtime` is deliberate: a5 `host_build_graph` uses `src/common/host_build_graph/runtime.h`, which a2a3 shares and whose whole object is the H2D image, so the first commit's field there grew a2a3's device image for something a2a3 never reads. Nothing under `src/common/host_build_graph/` is touched now.

On the barrier-free init path a thread must choose its clusters before handshaking them, so it first reads every core's report. That waits only on `aicore_done` — a GM flag each AICore publishes *before* it waits for its own window — so no thread blocks on another thread's MMIO work and the path keeps the property it was built for. The partition is deterministic, which is what lets each thread reach the same answer without exchanging anything.

Round-robin remains the fallback, taken for: any scenario other than FG (`die_id = phy_cpu_id / 4` only holds for the fault-free numbering, and PG renumbering is unresolved — see #2221), an OCCUPY-only topology whose die fields are `-1`, a scheduler `cpu_id` absent from the schedulable pool, a handshake report outside the physical core range, and a gate that had to fill exec slots by report order. That last one is new: `platform_aicpu_affinity_exact_match()` reports whether every survivor is on the cpu its slot names. The gate's recovery path breaks that correspondence silently, and a wrong die label costs about as much as the affinity gains.

### Verified

- 7 unit tests on the partition itself, including the measured FG shape, uneven die splits, the all-one-die degradation, and determinism.
- Built and run on the a5 box. On-device cold-path log:
  ```
  Cluster ownership (die-affine): tidx=0 die=1 clusters=8 same_die=8
  Cluster ownership (die-affine): tidx=1 die=1 clusters=8 same_die=8
  Cluster ownership (die-affine): tidx=2 die=1 clusters=8 same_die=0
  Cluster ownership (die-affine): tidx=3 die=0 clusters=8 same_die=8
  ```
  Exactly the quota, 24/32 same-die, matching what the unit test asserts.
- Also confirmed along the way, and written into `src/a5/docs/hardware.md`: the pcid space really is die-blocked (AIC only in `[0,17]` and `[54,71]`, AIV only in `[18,51]` and `[72,107]` — the gap rules out a 16-per-die layout), and a cluster's `worker_id` triple is the physical cluster (32/32 matched `die_base + 18 + 2*local`).

### Not in this PR

- **End-to-end A/B.** The −4% above is MMIO polling work, not wall clock; what share of the scheduler loop that is has not been measured. This is the number that should decide whether the change earns its keep.
- **PG.** Gated off until the `phy_cpu_id` renumbering question in #2221 has an answer.
- **host_build_graph**, and picking *which* thread absorbs the cross-die quota. The choice matters (the per-thread penalty spread is 1.25x–1.99x) but needs runtime calibration to do better than index order.

@yanghaoran29 — happy to split the replacement into its own PR if you would rather keep this one for the original approach. The title and description are updated to match what the branch now does.

ChaoZheng109:
2026-09-22: updated the title and description to describe the die-affinity implementation the branch now carries; the previous text described the reverted contiguous-partition approach. No code change with this edit.

PR #2176, opened 2026-09-09
Add: persistent kernel-context execution resources

YunjiQin:
> Rebased onto `4860af980` (current main). Scope is the split agreed in [this comment](https://github.com/hw-native-sys/simpler/pull/2176#issuecomment-5680058714) and frozen in [this one](https://github.com/hw-native-sys/simpler/pull/2176#issuecomment-5695273000): execution-resource ownership, initialization, callable registration, explicit close, and failure/quarantine state. `simpler_kernel_mode_supported()` stays `0` and launch stays a rejecting stub.

## Summary

- Own a dedicated AICPU stream, a hidden AICore stream, and the context event set, created at init and released at explicit close.
- Establish the device Runtime image, architecture register table, device KernelArgs, and callable uploads once, and reuse them across preparations.
- Implement kernel init and prepare on K1's write-once mode latch. Preparation neither accepts nor retains a caller stream.
- Validate the entire flexible-length callable image before parsing or uploading any variable tail data.
- Give every resource this PR acquires a coherent success, rollback, retry, or quarantine lifetime — and bound each one by the thing that actually invalidates it.

## The capability gate, and why it precedes the latch

The contract build and validation run ahead of `latch(SIMPLER_MODE_KERNEL)`, and a comment there says they must stay ahead of it. The latch never changes — not on finalize, not on error — so a config the builder rejects *after* it is taken leaves a context permanently bound to kernel mode with no resources. Validating first returns the same code on an untouched, still-reusable context, and the builder allocates nothing and touches no device, so nothing is lost by running it early.

A runtime without kernel-mode sizing reports `UNSUPPORTED` there, and that refusal **ends the call**. Everything after it — the latch, the streams and events, the runtime-image upload — is runtime-agnostic, so a runtime that declared it cannot size kernel resources would otherwise reach a permanent kernel latch with device work already issued.

## One ownership rule, three different bounds

The rule is the same everywhere: **a release that fails keeps an owner able to retry it**, because `MemoryAllocator::finalize()` clears its tracking map even when `rtFree` fails and is therefore not a reportable owner.

What differs is what bounds the retention, and getting that wrong is what three rounds of review caught. Retention that outlives the thing which invalidates the resource is not safety — it converts a recoverable leak into a permanent refusal, or hands out an address from a dead device generation.

| Owner | Retention bounded by | Program-mode behaviour |
| --- | --- | --- |
[...]

PR #2064, opened 2026-08-29
Add: kernel-mode C ABI skeleton and wire headers (K1)

sunkaixuan2018:
## Summary

Add the shared kernel-mode C ABI and contracts for contexts that borrow the caller's device and stream. The program entry latches PROGRAM before changing runner or process state. `ExecutionModeLatch` is write-once and idempotent for the same mode; failures and finalization never roll it back or permit a switch to the other mode.

All current backends report kernel mode unsupported. This PR defines and tests the shared contracts; it does not enable kernel launches or allocate kernel execution resources in production.

**Scope.** This gate carries the surface the downstream PRs must agree on, plus the guards that make *program* mode safe today. The persistent execution resources — hidden stream pair, event set, and phase machine — are **not** here: they have no consumer in this change, ship in no library, and the K2 branch already stacked on this one revises the event set from four entries to five with an ordering constraint this change has no way to see. They belong to the change that implements them.

## Contracts and implementation

- Four C entries: `simpler_kernel_mode_supported`, `simpler_kernel_mode_init`, `simpler_kernel_mode_prepare_callable`, and `simpler_kernel_mode_launch`. Kernel close reuses `finalize_device`. Every built component exports all four, and `ChipWorker` resolves them unconditionally like the rest of the uniform ABI — shipping not-supported stubs is already this repo's answer for an unsupported capability, so the probe answers the capability question at call time rather than deciding which symbols get resolved.
- `simpler_kernel_mode_supported` runs before any init, and `create_device_context()` takes no device, so an implementation answers from the runtime build alone; `ctx` is accepted for signature symmetry and must not be dereferenced for a device.
- The invocation envelope is exactly 40 bytes, with assertions for every field offset and explicit zero-only `host_copy_tensor_count` and `reserved_` fields. Runtime-specific payloads follow it. Both wire endpoints are built together, without version negotiation.
- Kernel-entry structural argument errors return `INVALID_ARGUMENT` (-1004), explicit mode/lifecycle guards return `INVALID_STATE` (-1003), and an unavailable capability returns `UNSUPPORTED` (-1001). `INTERNAL` (-1000) remains reserved for internal invariants, including forbidden changes to committed kernel arena capacity.
- `ChipCallable::scalar_count()` and the Python read-only property derive the count from the signature through `count_scalar_args`. No scalar-count field, sentinel, or extra factory argument is added. The original C++/Python construction interfaces, callable sizes, and field offsets remain intact and are protected by layout assertions.
[...]

PR #2190, opened 2026-09-10
Add: bounded kernel callable registration

YunjiQin:
Each successful kernel `prepare_callable` call uploads and registers a new callable and returns a fresh context-local `int32_t` ID, even for identical content. The caller owns ID reuse and content deduplication. Registrations retain independent device addresses until context close.

Kernel residency supports 8192 registrations. Small images share 2 MiB device blocks, while images larger than a block receive an independent allocation rounded to 64 bytes. Blocks grow on demand only during prepare; published device addresses never move. The 2 GiB cap applies to all allocated code blocks, including alignment and unused block tails. Entry count and device capacity are independent limits; neither preallocates execution resources.

The shared callable-ID protocol bound covers Host runner checks and the A2A3/A5 AICPU registration tables and dispatch guards. Host kernel launch resolves IDs directly without allocation. Admission failures preserve existing residents; failed unpublished uploads retain reusable block storage within the same budget.

Kernel preparation is supported inside GLOBAL and THREAD_LOCAL capture after initialization outside capture. Callable images, persistent Runtime/KernelArgs, and register-address tables use synchronous `aclrtMemcpy` under a temporary `ACL_MODEL_RI_CAPTURE_MODE_RELAXED` thread mode, restored after the copy even on failure. These uploads execute immediately and their device allocations remain valid through graph replay. Prepare performs no stream/device synchronization: success guarantees upload and registration enqueue, while AICPU execution errors surface through subsequent warmup and synchronization. The kernel Host launch binder is still unconnected, so this PR does not validate full kernel graph execution.

### Interface and lifetime

```cpp
int simpler_kernel_mode_prepare_callable(
    DeviceContextHandle ctx, const void *callable,
    size_t callable_size, int32_t *out_callable_id);
int simpler_kernel_mode_launch(
    DeviceContextHandle ctx, int32_t callable_id,
    const void *args, void *caller_stream);
```

Failed preparation writes `-1`. Successful IDs are never reused within a context and become invalid at close. Callers must pair IDs with their issuing context and drain executions and destroy referencing graphs before close. There is no callable generation field or device residency descriptor. K9 carries the callable ID and payload metadata without a callable generation.

The separate device entry `simpler_aicpu_kernel_exec` accepts a `SimplerKernelDispatchArgs` prefix followed by an immutable runtime payload. It validates packet size, address arithmetic, kernel mode, ID range, argument counts and reserved fields before calling `consume_kernel_invocation`. The packet also carries `chip_callable_address` and `chip_callable_bytes`, filled from the issuing context's committed residency. After null/alignment/size/overflow checks, the entry passes a `const ChipCallable &` and image length to the consumer. This gives the AICPU access to child function IDs and patched device code addresses for per-invocation function-table reconstruction. The consumer must establish cache visibility and validate child offsets/signatures before consuming the image. The producer must supply an actual allocation matching `packet_bytes`; the entry ABI has no independent allocation length. This protocol does not share the program-mode `simpler_aicpu_exec` / `KernelArgs` entry. The default consumer is weak and returns `UnsupportedPayload`; a runtime-specific strong consumer overrides it in the same link target without duplicate symbols. In this branch, runtime-specific function-table reconstruction and payload execution remain unconnected, as does the Host launch binder.

### Dependencies and scope

- Depends on #2176 (K2), which depends on #2064 (K1). This draft is rebased onto the latest K2 head `abbe07f5`; dependency commits remain separate from K10a.
- K10a-only diff: [review the independent commit](https://github.com/YunjiQin/simpler/commit/a8b19e8a0a707797ab24d2832c48d078b971686b).
- K2 context lifecycle parameters, canonical image validation, and close retry behavior are retained. Kernel registration enqueues without stream synchronization; program registration retains its synchronous completion behavior. Callable errors use -1005/-1006/-1007, preserving K2 INVALID_ARGUMENT (-1004).
- Program-mode registration retains its existing behavior.
- Host kernel launch and ID residency checks remain; binder and actual kernel execution are not connected yet, and kernel support remains disabled.
- Chinese call chain: `docs/zh-cn/kernel-callable-residency.md`.

### Validation

- Capture regressions reproduce a failure before this fix and pass after it: first prepare plus 64 additional registrations inside GLOBAL and THREAD_LOCAL capture, both runtimes, A3 hardware. The full kernel C ABI suite passes 34 tests (22 platform cases deselected). A preload guard rejects and counts synchronization calls during preparation; all prepare paths record zero calls.
- Capture-copy unit tests cover all thread modes, exchange failure, copy failure with restoration, and restore failure propagation; all five relevant C++ targets pass. Program vector execution passes on both a2a3sim and A3 hardware for both runtimes. CPU-only C ABI checks pass (12 passed, 44 skipped).

- Canonical image validation rejects child function IDs outside `[0, 1024)` and duplicates within an image before allocation/upload. The bound is checked against every runtime at compile time. All 33 relevant C++ tests pass, including the new regression that failed before this fix.

- Integration link regression: the test links both the default consumer and a strong runtime substitute. It reproduces duplicate symbols before the weak attribute and passes after the fix; the default-only consumer test also passes. All eight AICPU builds expose the fallback as a weak symbol.
- Removed the always-false uint64 device-address comparison with uintptr maximum; retained the address-plus-length overflow check.

- Editable runtime rebuild succeeds; all pre-commit checks pass.
- All eight AICPU runtime libraries export `simpler_aicpu_kernel_exec`. The two device-entry test targets pass; six entry-validation tests also pass ASan/UBSan, including A/B/A callable switching and invalid device-image spans. The updated kernel C API/Host ABI regression passes (16 passed, 40 skipped).
- Five relevant C++ targets pass. Four new capacity/pool tests failed against the previous implementation and pass with this change.
- Fourteen registration tests pass ASan/UBSan: 8192 IDs, small-image sharing, large-image allocation, stable addresses, block-slack accounting, allocation/upload failures and rollback.
- Python/C ABI/callable-identity/task-interface/worker ABI regression: 296 passed, 40 skipped.
- A3 onboard kernel lifecycle: 18 passed. Both runtimes register 65 identical images, check bounded first allocation and incremental block accounting, and verify persistent-free retry and zero memory after close.
- Program-mode vector execution passes on a2a3sim for both runtimes. Additional program-mode hardware smoke was cancelled while all devices were occupied.
- The Chinese call-chain document is included in site navigation; the new strict docs CI build passes (run 34817495346).
- Full PR CI reruns after push. Prior-head macOS remote transport and A3 scene failures were observed; this change does not claim those checks green. Actual kernel launch/binder and graph replay remain outside the validation scope.


src/common/platform/include/host/kernel_callable_cache.h
```diff
@@ -0,0 +1,225 @@
...
+    size_t bytes{0};
+    uint64_t descriptor_address{0};
+};
+
+// The caller serializes prepare/resolve/close. Entries and arena addresses
+// remain immutable until external quiescence permits context close.
+class KernelCallableCache {
+public:
+    static constexpr size_t kDescriptorBytes = MAX_REGISTERED_CALLABLE_IDS * sizeof(KernelCallableDeviceResidency);
+    static constexpr size_t kByteLimit = 512ULL * 1024 * 1024;
```
nalinaly: 建议将 kernel 模式的驻留 callable 容量规格调整为 **8192**，并将条目规格与 device 内存预分配解耦。

当前实现存在两个问题：第 65 个不同 callable 会在仍有可用 device 内存时被拒绝；第一个很小的 callable 也会触发约 512 MiB 的 arena 分配。仅把 64 调大、继续保留这个 arena 模型，不能满足大量 kernel 按需注册的场景。

建议本 PR 按以下规格修改：

1. **最多支持 8192 个不同的驻留 callable。** 需要同步处理 Host cache/resolve、runner、A2A3/A5 AICPU 注册表和 Device dispatch 的相关边界，不能只修改 cache 一处；8192 是条目容量规格，不代表预分配 8192 套 device 执行资源。
2. **Device 内存按实际驻留需求申请。** 可参考 CANN runtime 的 2 MiB 分块池：小镜像在已有块内子分配，空间不足才追加新块；超过单块大小的镜像按实际大小加必要对齐独立分配。2 MiB 是共享分配块的粒度，不是每个 callable 至少占 2 MiB。取消固定 512 MiB 首次预留，按需增长。
3. **驻留代码缓存的 device 总容量上限暂定为 2 GiB。** 这是容量上限，不是首次预分配量；仍按上述方式按需申请。达到 8192 项或 2 GiB 容量上限时返回明确的容量错误，不影响已经成功驻留的条目。
4. **扩容只追加，不迁移已发布的代码和 descriptor。** 保持已有地址及 handle 有效；新增分配、上传和注册仍放在 capture 外的 prepare，launch/replay 不承担扩容工作。

Host 内存目前不是本条 review 的关注点，现有 Host 镜像保留方式可以不改。

参考 CANN 官方源码：[BinaryLoad 按实际加载大小分配](https://gitcode.com/cann/runtime/blob/master/src/runtime/core/src/runtime.cc#L4937)、[2 MiB 单池定义](https://gitcode.com/cann/runtime/blob/master/src/runtime/core/src/pool/memory_pool.hpp#L22)、[池不足时追加新池](https://gitcode.com/cann/runtime/blob/master/src/runtime/core/src/pool/memory_pool_manager.cc#L69)。

nalinaly (commented):
请调整 callable 驻留容量及 device 内存分配方式：8192 项规格，参考 CANN 的 2 MiB 分块按需增长。具体意见见行内评论。

src/common/platform/include/host/kernel_callable_cache.h
YunjiQin: 已按建议调整：

- 条目上限改为 8192，统一使用 `MAX_REGISTERED_CALLABLE_IDS`，覆盖 Host cache/resolve、runner 和 A2A3/A5 AICPU 注册表及 dispatch 边界；不预分配 8192 套执行资源。
- 取消首次固定预留 512 MiB。完整 ChipCallable 镜像不超过 2 MiB 时，按 64 字节对齐在已有块内子分配；已有块都放不下时才追加一个 2 MiB 块。各块独立分配，不要求地址连续，单个镜像不跨块。
- 大于 2 MiB 的镜像独立申请 `align_up(size, 64)` 空间；已有小块的剩余空间仍可供后续小镜像使用。
- 2 GiB 上限按所有已分配代码块的容量统计，包含对齐及空闲尾部，与 8192 项上限独立检查。超限返回容量错误，已有驻留条目不受影响。
- 扩容只在 capture 外的 prepare 追加分配，不搬迁已发布镜像，已有地址和 handle 保持有效。每次成功 prepare 仍注册新 callable，上层负责 handle 复用。

验证：5 个相关 C++ 测试目标、14 项缓存 ASan/UBSan 测试、296 项 Python/ABI 回归及 18 项 A3 设备生命周期测试通过；两种 runtime 的 program 模式模拟执行通过。设备测试覆盖 65 次连续注册、首次小块分配、按块增长及 close 重试；8192 项边界和容量超限不变性由 C++ 单测覆盖。中文调用链文档已同步更新。

ChaoWao:
## 关闭：交付路线已替代（superseded）

根据维护者确认，关闭这份旧 kernel-mode PR。

**原因：**有界、稳定地址、不自动淘汰的注册方向仍符合目标。但当前整包同时引入 capture 内 prepare、RELAXED 同步上传及独立 dispatch 接线；集成线 #2290 / INTEGRATION-LOG D22 已吸收并继续修改。这不是当前 main 路线可直接合入的单元。

**仍需保留／迁移：**后续拆出面向当前 main 的有界代码池、稳定 callable handle、容量失败保护和保活测试；注册等冷路径在 capture 前完成，错误与退休接共同执行层。关闭不取消常驻注册能力，也不否认现有硬件证据。

当前确定路线是 **Pipeline A → kernel eager → capture/replay**：两种资源模式接共同执行与退休层，每次 eager 重新准备并提交一次。集成分支中的采用不等于已合入 main，也不代表本 PR 的全部目标已交付。

本次只关闭旧 PR，保留分支、提交、作者贡献、测试及实验记录，供后续按能力迁移。

本次核查 head：`a8b19e8a0a707797ab24d2832c48d078b971686b`；main 基线：`2d3661517`。

PR #2197, opened 2026-09-11
Refactor: unify scalar reads across host_build_graph and TaskArgsTpl

### LINKED ISSUE #2170: Migrate boundary-scalar value reads off the deprecated InheritableScalar conversion
## Background

#2160 makes a Graph boundary scalar a formal parameter. `args.scalar(i)` now
returns an `InheritableScalar` carrying the value together with the parameter it
came from. Forwarding it — `task_args.add_scalar(args.scalar(i))` — keeps the
parameter, so the destination slot is refreshed on every replay. Reading it as a
number goes through a **deprecated** conversion, which turns what used to be a
silent change of behaviour into a compiler diagnostic.

That deprecation currently has no consumer: every existing value read still
compiles, and the warning is an inventory rather than a migration. This issue
tracks closing it.

## Why this is not only about noise

`args.scalar(i).to<T>()` applies `to_u64`'s actual inverse. `static_cast` does
not: a float slot holds a bit pattern, so `static_cast<float>(args.scalar(i))`
of `1.0f`'s pattern yields `1065353216.0`. `to<T>()` is also the only spelling
that reaches an enum, since a conversion to an enumeration does not accept a
user-defined one on the way.

Each site therefore needs a decision, not a mechanical rewrite:

- the body forwards the parameter → `add_scalar(args.scalar(i))`
- the body genuinely wants the number → `args.scalar(i).to<T>()`
- the body wants the number *and* the value should gate cache reuse →
  `add_static_scalar(args.scalar(i))`

## Scope

| Repo | Sites | Where |
| ---- | ----- | ----- |
| `simpler` | 98 | `host_build_graph` orchestration under `examples/` and `tests/st/` |
| `pypto` | 2 | `src/codegen/orchestration/orchestration_codegen.cpp` |
| `pypto-lib` | 0 | no direct call site; covered by the `pypto` generator |

The two `pypto` sites are **code-generator templates**, not ordinary call sites:

- `orchestration_codegen.cpp:218` emits `<T> var = from_u64<T>(orch_args.scalar(i));`
- `orchestration_codegen.cpp:3449` emits `uint64_t ext_<name>_ctx = orch_args.scalar(i);`

One template each, so every generated orchestration inherits the value read.
This is why the work is not a `simpler`-only cleanup: without the `pypto`
change, generated orchestration keeps re-introducing the reads, and users never
get the chance to declare which scalars are static.

## Ordering constraint

The two declarations have asymmetric failure costs:

- a forward-only scalar wrongly declared **static** → stricter match, fewer
  cache hits, still correct
- a scalar the body freezes into the Definition wrongly declared **dynamic** →
  matching skips its value → a hit on a Definition baked with a stale number,
  **silently wrong**

Callers today declare everything dynamic (`add_scalar`); `add_static_scalar` has
no caller yet. Nothing is wrong at present because Graph matching does not
compare scalar values at all — `graph_full_key` is `callable_hash + graph_key`.

**This migration must land before scalar values enter the cache-reuse
condition.** Doing that work first would immediately start mis-matching every
frozen-but-dynamic parameter, with no diagnostic.

## Known blind spot

GCC suppresses a deprecation diagnostic instantiated inside a system header, so
a value read whose conversion happens in third-party template code stays silent.
`EXPECT_EQ(args.scalar(i), v)` in the unit tests is the case found so far. The
warning is an inventory of the value reads *written in this repo*, not a proof
that none exists — "no warnings" is not a valid exit criterion for this
migration.

poursoul:
## Summary

- `Arg::scalar(i)` on host_build_graph and `TaskArgsTpl::scalar(i)` (TMR and every other `Arg` built on it) disagreed on what a caller writes for a static read, even though both slots are the same `uint64_t`. `scalar(i)` is now a template on both sides, with the parameter spelled `ScalarT` in both (`TaskArgsTpl`'s `T` is its tensor type): host_build_graph defaults it to `InheritableScalar` (a bare `scalar(i)` still forwards with its origin), `TaskArgsTpl` defaults it to `S`/`uint64_t` (unchanged), and either side accepts an explicit type for a static value read (`args.scalar<T>(i)`). `TaskArgsTpl` bounds that read by `sizeof(S)` — the slot it has to fit in — rather than by a hardcoded 8.
- Removes the deprecated `InheritableScalar::operator uint64_t()`. This is strictly stronger than the deprecation it replaces: with no conversion left to suppress, a value read is a **compile error** rather than a warning, which closes #2170's "known blind spot" — a read inside third-party template code (`EXPECT_EQ(args.scalar(i), v)`) no longer slips past GCC's system-header suppression.
- Migrates all 98 call sites (20 orchestration files) that were hitting the deprecation warning to the explicit-`T` spelling. Every one of them was already a value read (`uint64_t v = args.scalar(i)`, `static_cast<T>(...)`, `from_u64<T>(...)`, `.to<T>()`); no forwarding site is touched, so no call site changes meaning. Deciding which of these *should* become forwards is a separate change against the individual examples — see #2170.
- `InheritableScalar::to<T>()` stays. Removing it is what forced every value read to name its type while the call sites were migrated; now that they have been, it is the only way to read a handle that was passed on as a function argument, where `Arg::scalar<T>(i)` is unavailable because the `Arg` is not in hand.
- Python's `add_scalar` took a pre-encoded `uint64_t`; it now takes the value directly (int, float, bool, or a ctypes scalar) and encodes it natively in the bindings (`encode_scalar`, exposed to Python as `scalar_to_uint64`). `scene_test.py`'s three call sites drop their `scalar_to_uint64` wrapping accordingly. The encoder matches C++ `to_u64()` bit for bit (see Behavior changes) and reads a ctypes scalar through the buffer protocol, so it needs no per-type dispatch and handles subclasses correctly.

- The `tensormap_and_ringbuffer` orchestrations that read a slot as some type *other than* the slot's own move with them (15 files, 28 sites). Two reasons beyond consistency: it puts the same spelling in front of both runtimes in-tree, which is the claim this PR is named after, and it gives `TaskArgsTpl::scalar`'s non-`S` branch its first instantiations — a template body is only checked when instantiated, and until now every `scalar<T>` call site was `host_build_graph`, whose `Arg` hides the base's `scalar` with its own. A bare `uint64_t` read is deliberately left as `scalar(i)`: that is already what it answers, so naming the type would add nothing.

Progress on #2170 — this PR migrates the `simpler`-repo side only (the 98 sites above). #2170 also scopes a `pypto`-side change (the codegen templates that emit these value reads) that is not part of this PR; see the issue for the remaining work. Not closing the issue yet.

## Behavior changes

Five, all on the Python scalar-encoding boundary (`scalar_to_uint64` and `add_scalar`). The first three make the encoding agree with C++ `to_u64()`; the last two replace a silent wrong answer with an error:

1. **A ctypes scalar is zero-extended from its own width, not sign-extended.** `c_int8(-1)` now encodes to `0xFF`, matching `to_u64(int8_t{-1})`; it was `0xFFFF_FFFF_FFFF_FFFF`. Same for `c_int16` / `c_int32`. Reading the slot back at the matching width (`scalar<int8_t>`) yields `-1` either way, so this is only observable by reading a narrow-signed slot as `scalar<int64_t>`, which is already a width mismatch. This makes a slot written from Python and one written from orchestration identical for the same value.
2. **`scalar_to_uint64` no longer accepts an object that only implements `__int__`** (e.g. `numpy.float32`, `Decimal`). The old fallback was `int(value) & 0xFFFF…`, which silently truncated. `bool`, `IntEnum` members and numpy integer scalars are still accepted, via `__index__`.
3. **An integer outside the 64-bit two's-complement range raises** instead of being truncated to its low 64 bits.
4. **A finite `float` outside IEEE-754 single-precision range raises** (`1e100`) instead of being stored as an infinity. This matches the old `struct.pack("<f", ...)`, which raised `OverflowError` for the same set; the tests assert against `struct.pack`'s own verdict rather than hardcoding a range. `inf` and `NaN` pass through as themselves.
5. **A ctypes scalar not in host byte order is refused** — `c_uint32.__ctype_be__` on a little-endian host, and `__ctype_le__` on a big-endian one. The old `.value` path read the number and so was order-blind; reading raw bytes is not. A refusal, rather than a byte swap, because `to_u64` has no reversed-order form for such a value to agree with. Pointer and character types (`c_void_p`, `c_char_p`, `c_char`, `c_wchar`) and `c_longdouble` are refused for the same "no number a slot can carry" reason; `c_char_p` already raised.

A native Python `float` still narrows to IEEE-754 single precision. It is the one encoding that cannot align with its C++ counterpart — a Python float carries no width, where `to_u64(1.5)` in orchestration is a double. Pass `ctypes.c_double` for full precision.

## Test plan

- [x] `pip install --no-build-isolation -e .` — full rebuild across all 4 `host_build_graph` variants (a2a3/a5 × onboard/sim), no compile errors or warnings.
- [x] ut-py: 2350 passed, 11 skipped. Includes 96 new cases in `test_task_interface.py` pinning every encoding across `scalar_to_uint64`, `TaskArgs.add_scalar` and `ChipStorageTaskArgs.add_scalar`. Regression anchors, one per behavior above: `c_int8(-1)` → `0xFF`, a `c_double` subclass, both byte-order qualifiers (the foreign one refused, the host one accepted), the single-precision boundary checked against `struct.pack`'s own verdict, out-of-range integers, and an `__index__` that raises.
- [x] ut-cpp: `ctest -LE requires_hardware` — 140/140 passed.
- [x] a2a3sim: `predicated_dispatch`, `paged_attention`, `batch_paged_attention`, `host_build_graph_validation` — 0 failed.
- [x] a2a3 **onboard** (`task-submit`, device 7): the TMR files no sim platform reaches — `alternating_matmul_add`, `fanin_lookup_perf`, `paged_attention_unroll`, `dfx/chip_swimlane` — 6 passed, 0 failed.
- [x] TMR sim: `sliding_window_deps` (a2a3+a5), `benchmark_bgemm` (a5), `dfx` — 0 failed.
- [x] The 7 changed files no test on this host reaches (a5 `alternating_matmul_add` / `fanin_lookup_perf` / `chip_swimlane` / `sdma` / `urma`, a2a3 `sdma` / `decode_fwd`) syntax-checked against each arch's own `compile_commands.json` include set — all clean. a5 onboard is not runnable here (host is a2a3 silicon); CI's `st-onboard-a5` covers them.
- [x] a5sim: `predicated_dispatch`, `single_core_dag`, `multi_core_dag`, `paged_attention`, `alternating_matmul_add`, `benchmark_bgemm`, `host_build_graph_validation` — 0 failed.
- [x] `pre-commit run` on all changed files — all hooks pass.
- [ ] Remaining hardware-only files (no sim platform variant) are not exercised by this test plan: `deepseek_v4_flash_decode`, `spmd_paged_attention` (a2a3+a5), `worker_async_fifo`, `alternating_matmul_add` (a2a3), `paged_attention_unroll` (a2a3+a5), `paged_attention_unroll_manual_scope` (a2a3+a5) — covered by the onboard CI jobs on this PR.

### DIFF
--- a/python/bindings/task_interface.cpp
+++ b/python/bindings/task_interface.cpp
@@ -1666,6 +1666,67 @@ nb::tuple dims_tuple(const uint32_t *dims, uint32_t ndims) {
     return nb::tuple(out);
 }
 
+// `ctypes._SimpleCData` is the common base of every ctypes scalar. Cached after the first
+// lookup: importing `ctypes` and walking its attributes on every add_scalar call would cost far
+// more than the check it guards.
+PyObject *ctypes_simple_cdata_type() {
+    static PyObject *cached = nb::module_::import_("ctypes").attr("_SimpleCData").inc_ref().ptr();
+    return cached;
+}
+
+// A ctypes scalar's buffer format is a byte-order prefix plus one `struct` format character
+// ("<I", ">d"). Both halves decide whether it can encode, which is why the buffer's format
+// answers this rather than the type's `_type_`: `_type_` carries the character alone.
+//
+// **The prefix is the scalar's actual byte order**, and ctypes spells it out even for a native
+// type, so admission is a match against the host's. `ctypes.c_uint32.__ctype_be__` is an
+// ordinary `_SimpleCData` subclass whose `_type_` is `'I'`, indistinguishable from `c_uint32`,
+// but whose format is `">I"` and whose bytes for the value 1 are `00 00 00 01`. Copying those
+// raw would store `0x01000000` where `to_u64(uint32_t{1})` is `0x1`. A reversed-order scalar has
+// no native C++ counterpart to agree with, so it is refused rather than byte-swapped into one.
+//
+// **The character decides whether the value is a number at all**: the integer widths, `f`, `d`
+// and `?` encode. The pointer and character types (`P`, `z`, `Z`, `c`, `u`) and long double
+// (`g`) do not — they name no number a slot can carry, and encoding a host pointer into a
+// device-bound slot is a defect wherever it happens. A subclass inherits its base's format, so
+// it is admitted with the base, which dispatching on the type's `__name__` would not do.
+enum class CtypesFormat { Encodable, ForeignByteOrder, NotNumeric };
+
+#if defined(__BYTE_ORDER__) && __BYTE_ORDER__ == __ORDER_BIG_ENDIAN__
+constexpr char NATIVE_BYTE_ORDER = '>';
+#else
+constexpr char NATIVE_BYTE_ORDER = '<';
+#endif
+
+CtypesFormat classify_ctypes_format(const char *format) {
+    // ctypes always emits the prefix, so a format without one did not come from one of its
+    // scalars and is refused rather than guessed at.
+    if (format == nullptr || (format[0] != '<' && format[0] != '>')) {
+        return CtypesFormat::NotNumeric;
+    }
+    if (format[1] == '\0' || format[2] != '\0') {
+        return CtypesFormat::NotNumeric;
+    }
+    switch (format[1]) {
+    case 'b':
+    case 'B':
+    case 'h':
+    case 'H':
+    case 'i':
+    case 'I':
+    case 'l':
+    case 'L':
+    case 'q':
+    case 'Q':
+    case 'f':
+    case 'd':
+    case '?':
+        return format[0] == NATIVE_BYTE_ORDER ? CtypesFormat::Encodable : CtypesFormat::ForeignByteOrder;
+    default:
+        return CtypesFormat::NotNumeric;
+    }
+}
+
 // Resolve one wire tensor onto a local base and build the address-bearing device POD.
 // `resolved` maps CanonicalIdentity -> (local_base, address_space); the caller populates it by
 // materializing each embedded descriptor.
@@ -1690,6 +1751,129 @@ ChipTensor materialize_one(const Tensor &r, nb::dict resolved) {
         reinterpret_cast<void *>(static_cast<uintptr_t>(base + r.byte_offset)), r.shapes, r.strides, r.ndims, r.dtype,
         static_cast<AddressSpace>(addr_space)
     );
+}
+
+// Read an exact PyLong as the 64-bit pattern a slot stores, accepting the whole two's-complement
+// range: signed first, then unsigned for anything above INT64_MAX. Out of that range is an error
+// rather than a silent truncation to the low 64 bits.
+uint64_t encode_python_int(PyObject *long_ptr) {
+    int64_t signed_value = PyLong_AsLongLong(long_ptr);
+    if (signed_value == -1 && PyErr_Occurred()) {
+        PyErr_Clear();
+        unsigned long long unsigned_value = PyLong_AsUnsignedLongLong(long_ptr);
+        if (PyErr_Occurred()) {
+            PyErr_Clear();
+            throw std::overflow_error("add_scalar: integer value out of 64-bit range");
+        }
+        return static_cast<uint64_t>(unsigned_value);
+    }
+    return static_cast<uint64_t>(signed_value);
+}
+
+// The one place a Python scalar becomes the uint64_t a wire slot stores. Its output must match
+// C++ `to_u64` bit for bit for the same value, so that a slot written from Python and one written
+// from orchestration read back identically.
+//
+// Three encodings, in the order checked:
+//
+// - A Python int is read as a 64-bit quantity, which is what `to_u64(int64_t)` / `to_u64(uint64_t)`
+//   also produce. PyLong_CheckExact short-circuits the common case; the general branch further
+//   down goes through __index__ (PyIndex_Check), not PyLong_Check, so bool, IntEnum members, and
+//   numpy integer scalars are admitted -- none of which are actual `int` instances -- while every
+//   float type, numpy's included, is still rejected, since none of them define __index__.
+// - A ctypes scalar is read through the buffer protocol and **zero-extended** into the slot, which
+//   is exactly what to_u64's union does. Reading `.value` instead would sign-extend a narrow
+//   signed type (c_int8(-1) -> 0xFFFF'FFFF'FFFF'FFFF where to_u64(int8_t{-1}) is 0xFF), and
+//   dispatching on the type's __name__ to find the float types would miss their subclasses and
+//   silently narrow a c_double subclass to single precision. One buffer read has neither problem,
+//   and its format answers admission and byte order in the same step -- see classify_ctypes_format.
+// - A native Python float narrows to IEEE-754 single precision, matching to_u64(float). Out of
+//   single-precision range is an error, not an infinity: a finite input that narrowed to inf would
+//   be stored as a different number. This is the one encoding that cannot align with its C++
+//   counterpart by construction, because a Python float carries no width: to_u64(1.5) in
+//   orchestration is a double. ctypes.c_double is the spelling for full precision, and the
+//   docstrings say so.
+uint64_t encode_scalar(nb::handle value) {
+    PyObject *ptr = value.ptr();
+
+    if (PyLong_CheckExact(ptr)) {
+        return encode_python_int(ptr);
+    }
+
+    // PyObject_IsInstance answers -1 on error, which is truthy -- test for it explicitly rather
+    // than walking into the ctypes branch with a Python exception already set.
+    int is_ctypes = PyObject_IsInstance(ptr, ctypes_simple_cdata_type());
+    if (is_ctypes < 0) {
+        throw nb::python_error();
+    }
+    if (is_ctypes) {
+        // One buffer request answers both questions: its format decides admission and byte
+        // order, its length decides the width. Taking the width from the buffer rather than
+        // inferring it from the format character is what keeps a subclass whose format
+        // disagrees with its layout from reading past its own storage.
+        Py_buffer view;
+        if (PyObject_GetBuffer(ptr, &view, PyBUF_FORMAT) != 0) {
+            throw std::invalid_argument("add_scalar: ctypes scalar does not support the buffer protocol");
+        }
+        CtypesFormat verdict = classify_ctypes_format(view.format);
+        size_t width = static_cast<size_t>(view.len);
+        uint64_t bits = 0;
+        if (verdict == CtypesFormat::Encodable && width <= sizeof(uint64_t)) {
+            std::memcpy(&bits, view.buf, width);
+        }
+        PyBuffer_Release(&view);
+
+        if (verdict == CtypesFormat::ForeignByteOrder) {
+            throw std::invalid_argument(
+                "add_scalar: ctypes scalar is not in host byte order (e.g. c_uint32.__ctype_be__ on a "
+                "little-endian host); its bytes would encode to a different number than the same value "
+                "written from orchestration"
+            );
+        }
+        if (verdict == CtypesFormat::NotNumeric) {
+            throw std::invalid_argument(
+                "add_scalar: ctypes scalar must be an integer width (c_int8..c_uint64), c_float, "
+                "c_double or c_bool -- a pointer or character type carries no value a slot can hold"
+            );
+        }
+        if (width > sizeof(uint64_t)) {
+            throw std::invalid_argument("add_scalar: ctypes scalar is wider than the 8-byte slot");
+        }
+        return bits;
+    }
+
+    if (PyFloat_Check(ptr)) {
+        double d = PyFloat_AsDouble(ptr);
+        if (d == -1.0 && PyErr_Occurred()) {
+            throw nb::python_error();
+        }
+        float f = static_cast<float>(d);
+        // A finite value that narrows to an infinity is out of single-precision range, and
+        // storing that infinity would silently be a different number. An input that is already
+        // infinite or NaN passes through as itself.
+        if (std::isfinite(d) && !std::isfinite(f)) {
+            throw std::overflow_error(
+                "add_scalar: float value out of IEEE-754 single-precision range; pass ctypes.c_double "
+                "for full precision"
+            );
+        }
+        uint32_t bits;
+        std::memcpy(&bits, &f, sizeof(bits));
+        return static_cast<uint64_t>(bits);
+    }
+
+    if (PyIndex_Check(ptr)) {
+        // Normalize through __index__ into an exact PyLong first: PyLong_As* is only
+        // guaranteed against one, and ptr may be a numpy integer scalar or an IntEnum member,
+        // neither of which is a PyLong itself.
+        nb::object as_long = nb::steal<nb::object>(PyNumber_Index(ptr));
+        if (!as_long.is_valid()) {
+            throw nb::python_error();
+        }
+        return encode_python_int(as_long.ptr());
+    }
+
+    throw std::invalid_argument("add_scalar: value must be int, float, bool, or a ctypes scalar");
 }
 
 // Which device allocations a dispatch's operands may name: the private snapshot of every live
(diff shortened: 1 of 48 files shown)


python/bindings/task_interface.cpp
```diff
@@ -1692,6 +1700,64 @@ ChipTensor materialize_one(const Tensor &r, nb::dict resolved) {
...
+        uint32_t bits;
+        std::memcpy(&bits, &f, sizeof(bits));
+        return static_cast<uint64_t>(bits);
+    }
+
+    if (PyIndex_Check(ptr)) {
+        // Normalize through __index__ into an exact PyLong first: PyLong_As* is only
+        // guaranteed against one, and ptr may be a numpy integer scalar or an IntEnum member,
+        // neither of which is a PyLong itself.
+        nb::object as_long = nb::steal<nb::object>(PyNumber_Index(ptr));
```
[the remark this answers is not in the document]
poursoul: Valid, and fixed in `7ea5551e` (the branch is now a single squashed commit; this thread was against the earlier `75673550`).

`PyNumber_Index` returning `nullptr` is now checked before the value is used:

```cpp
nb::object as_long = nb::steal<nb::object>(PyNumber_Index(ptr));
if (!as_long.is_valid()) {
    throw nb::python_error();
}
return encode_python_int(as_long.ptr());
```

Checking after the `steal` rather than before it is equivalent here — `nb::steal` only takes ownership, it does not dereference, and `nb::object`'s destructor is `Py_XDECREF`, which is a no-op on null. `is_valid()` is exactly the `m_ptr != nullptr` test.

The regression test you asked for is in `tests/ut/py/test_task_interface.py::TestScalarToUint64::test_propagates_an_exception_raised_by_index`, covering both entry points:

```python
class BrokenIndex:
    def __index__(self):
        raise RuntimeError("boom")

with pytest.raises(RuntimeError, match="boom"):
    scalar_to_uint64(BrokenIndex())

args = TaskArgs()
with pytest.raises(RuntimeError, match="boom"):
    args.add_scalar(BrokenIndex())
assert args.scalar_count() == 0
```

It confirms the original `RuntimeError` propagates rather than being replaced, and that the failed `add_scalar` leaves no partial slot behind.

The same class of bug was present one branch up and is fixed in the same commit: `PyObject_IsInstance` answers `-1` on error, which is truthy, so the ctypes branch was reachable with an exception already pending. That return is now tested for `< 0` explicitly.

src/common/host_build_graph/types.h
```diff
@@ -208,32 +208,16 @@ class InheritableScalar {
...
-    // Read the parameter as a value, with to_u64's actual inverse applied. static_cast on
-    // the pattern is not that inverse: a float slot holds a bit pattern, so
-    // static_cast<float> of 1.0f's pattern yields 1065353216.0. It is also the only
-    // spelling that reaches an enum, since a conversion to an enumeration does not accept
-    // a user-defined one on the way.
-    template <typename T>
-    T to() const {
-        return from_u64<T>(bits_);
-    }
+private:
```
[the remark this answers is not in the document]
poursoul: Valid, and fixed in `7ea5551e` (single squashed commit now; this thread was against `75673550`).

`GRAPH_EXECUTION.md` had three stale places, not just the one you flagged — the original change updated the two `.to<T>()` spellings inside that section and left the surrounding prose describing a mechanism that no longer exists:

- **L102-107** (your finding) — rewritten. It now says an `InheritableScalar` has no conversion to a number, so `uint64_t v = args.scalar(i)` and `static_cast<int32_t>(args.scalar(i))` do not compile, and a value read has to name its type.
- **L109-113** — the paragraph documenting the GCC system-header blind spot, which only existed because the diagnostic was a `[[deprecated]]` attribute. Replaced with the opposite statement: because the diagnostic is now the *absence of a conversion* rather than a deprecation, it has no blind spot, and `EXPECT_EQ(args.scalar(i), v)` fails inside third-party template code too.
- **L132** — `A derived value (args.scalar(i) + 1) is a value read and freezes the same way` was still describing a warning-based path. Now: `args.scalar(i) + 1` does not compile, `args.scalar<uint64_t>(i) + 1` does and is frozen.

The section also gained a note on `InheritableScalar::to<T>()`, which this PR keeps — it is the read for a handle already passed on as a function argument, where `Arg::scalar<T>(i)` is unavailable because the `Arg` is not in hand.

One doc outside that file was stale for the Python half and is updated in the same commit: `examples/workers/l3/worker_chip_orch_comm_stream/README.md` still taught `scalar_to_uint64(ctypes.c_float(7.0))` as the way to pass a typed scalar, which `add_scalar` now does directly.

zhusy54 (commented):
Two scalar-encoding corner cases need handling before merge.

python/bindings/task_interface.cpp
```diff
@@ -1692,6 +1700,94 @@ ChipTensor materialize_one(const Tensor &r, nb::dict resolved) {
...
+            throw std::invalid_argument("add_scalar: ctypes scalar is wider than the 8-byte slot");
+        }
+        uint64_t bits = 0;
+        std::memcpy(&bits, view.buf, width);
+        PyBuffer_Release(&view);
+        return bits;
+    }
+
+    if (PyFloat_Check(ptr)) {
+        float f = static_cast<float>(PyFloat_AsDouble(ptr));
```
zhusy54: [P1] Preserve overflow errors when narrowing Python floats\n\nThe previous Python implementation used `struct.pack("<f", value)`, which raises `OverflowError` for a finite value outside the float32 range (for example, `1e100`). This `static_cast<float>` instead produces `+/-inf`, so the PR silently stores a different scalar even though its description claims only three encoding behavior changes. Please preserve the old rejection (for example, reject when a finite input becomes non-finite after narrowing) or explicitly document and test this additional behavior change.

python/bindings/task_interface.cpp
```diff
@@ -1692,6 +1700,94 @@ ChipTensor materialize_one(const Tensor &r, nb::dict resolved) {
...
+        if (PyObject_GetBuffer(ptr, &view, PyBUF_SIMPLE) != 0) {
+            throw std::invalid_argument("add_scalar: ctypes scalar does not support the buffer protocol");
+        }
+        size_t width = static_cast<size_t>(view.len);
+        if (width > sizeof(uint64_t)) {
+            PyBuffer_Release(&view);
+            throw std::invalid_argument("add_scalar: ctypes scalar is wider than the 8-byte slot");
+        }
+        uint64_t bits = 0;
+        std::memcpy(&bits, view.buf, width);
```
zhusy54: [P1] Honor ctypes byte order before copying raw bytes\n\n`ctypes._SimpleCData` also includes endian-qualified numeric scalars. On a little-endian host, `ctypes.c_uint32.__ctype_be__(1)` exposes bytes `00 00 00 01`, so this `memcpy` returns `0x01000000`; the old `.value` path and C++ `to_u64(uint32_t{1})` both produce `1`. That violates the advertised bit-for-bit contract and silently corrupts a supported ctypes scalar. Please request/inspect the buffer format and normalize (or explicitly reject non-native-endian formats), with a regression test.

poursoul: Good catch — confirmed and fixed in `198c3f9f`.

Reproduced on the branch before the fix, exactly as you described:

```
scalar_to_uint64(ctypes.c_uint32.__ctype_be__(1)) = 0x1000000
C++ to_u64(uint32_t{1})                           = 0x1
```

Worth spelling out why the commit that landed between your review and this reply did not catch it. That commit restricted admission by reading the `struct` format character out of `_type_` — and the byte-order variants carry the *same* `_type_` as their native counterpart, so that judge is blind to exactly this case:

```
type                     _type_    memoryview.format
c_uint32                 'I'       '<I'
c_uint32.__ctype_be__    'I'       '>I'      <- identical _type_
```

The order only exists in the buffer format, and `PyBUF_SIMPLE` does not populate `view.format` at all. So the fix moves the whole judgement onto one `PyBUF_FORMAT` request, which answers admission, byte order and width together instead of splitting them across `_type_` and the buffer:

```cpp
enum class CtypesFormat { Encodable, ForeignByteOrder, NotNumeric };

CtypesFormat classify_ctypes_format(const char *format) {
    if (format == nullptr || (format[0] != '<' && format[0] != '>')) return CtypesFormat::NotNumeric;
    if (format[1] == '\0' || format[2] != '\0') return CtypesFormat::NotNumeric;
    switch (format[1]) {
    case 'b': ... case 'Q': case 'f': case 'd': case '?':
        return format[0] == NATIVE_BYTE_ORDER ? CtypesFormat::Encodable : CtypesFormat::ForeignByteOrder;
    default:
        return CtypesFormat::NotNumeric;
    }
}
```

I took the **reject** branch of your suggestion rather than normalizing. `to_u64` is a native-order union with no reversed-order form, so byte-swapping would invent a C++ semantic that does not exist; and a big-endian scalar bound for a device slot is not a thing a caller means on purpose. `NATIVE_BYTE_ORDER` is derived from `__BYTE_ORDER__`, so on a big-endian host it is `__ctype_le__` that gets refused and the unqualified type that stays native — ctypes states the real order in the format either way.

Refusal and acceptance are separately tested, parameterized over the host's own order so the pair does not silently invert on a big-endian build:

```python
def _HOST_ORDER(ctype):     return ctype.__ctype_le__ if sys.byteorder == "little" else ctype.__ctype_be__
def _FOREIGN_ORDER(ctype):  return ctype.__ctype_be__ if sys.byteorder == "little" else ctype.__ctype_le__

# refused, for c_uint32 and c_double, after asserting the trap: same _type_, reversed bytes
assert type(foreign)._type_ == type(native)._type_
assert bytes(memoryview(foreign)) == bytes(reversed(bytes(memoryview(native))))
with pytest.raises((TypeError, ValueError)):
    scalar_to_uint64(foreign)

# and the host-order qualifier is still accepted
assert scalar_to_uint64(_HOST_ORDER(ctypes.c_uint32)(1)) == 1
```

PR description Behavior changes now lists this as item 5, noting it is a tightening relative to the old `.value` path, which was order-blind and therefore happened to give the right answer here.

python/bindings/task_interface.cpp
poursoul: Agreed on both counts — it is a real behavior change and it was undeclared. Fixed in `198c3f9f`.

Reproduced before the fix:

```
1e100             old struct.pack: OverflowError   was: 0x7f800000 (inf)
3.4028235678e38   old struct.pack: OverflowError   was: 0x7f800000 (inf)
-1e100            old struct.pack: OverflowError   was: 0xff800000 (-inf)
```

Taking your first option — preserve the rejection — since the second (document the silent `inf`) keeps a wrong number reachable, and this one also removes the UB in `double`→`float` overflow ([conv.double]):

```cpp
double d = PyFloat_AsDouble(ptr);
if (d == -1.0 && PyErr_Occurred()) throw nb::python_error();
float f = static_cast<float>(d);
// A finite value that narrows to an infinity is out of single-precision range, and
// storing that infinity would silently be a different number. An input that is already
// infinite or NaN passes through as itself.
if (std::isfinite(d) && !std::isfinite(f)) {
    throw std::overflow_error(
        "add_scalar: float value out of IEEE-754 single-precision range; pass ctypes.c_double "
        "for full precision");
}
```

`isfinite(d) && !isfinite(f)` is your "finite input becomes non-finite after narrowing" phrasing verbatim, which also keeps `inf`/`-inf`/`NaN` passing through — `struct.pack("<f", ...)` accepts those too, so rejecting them would have been a second undeclared change in the other direction.

Rather than hardcode a range next to the encoder, the tests assert against `struct.pack`'s own verdict, so the two cannot drift:

```python
def _struct_pack_f_raises(value):
    try:
        struct.pack("<f", value)
    except OverflowError:
        return True
    return False

# for 1e100, -1e100, 3.402823806667635e38, 1e308
assert _struct_pack_f_raises(value)
with pytest.raises((OverflowError, ValueError)):
    scalar_to_uint64(value)
```

plus the boundary either side (`FLT_MAX` = `3.4028234663852886e38` encodes; the next double up that does not round back down to it raises) and pass-through for `inf`/`-inf`/`NaN`. Verified matching `struct.pack` on every case.

PR description Behavior changes now lists this as item 4. Your second comment on byte order is answered in its own thread — it was also still live, and for a related reason: reading raw bytes made the buffer, not `.value`, the source of truth.

zhusy54 (dismissed):
Re-reviewed at `198c3f9f`. Both scalar-encoding findings are addressed: finite Python floats that overflow float32 are rejected while existing infinities/NaNs retain their prior behavior, and ctypes admission now validates numeric format plus native byte order before copying. The added regression tests cover overflow boundaries, non-finite values, foreign/native byte-order variants, non-numeric ctypes values, and admitted subclasses. I found no additional blocking issues in the updated diff. A few CI jobs are still running at review time.

ChaoZheng109 (approved):

Issue #2134, opened 2026-09-05
[Feature] Place device and remote work by span containment, and publish the slack as its error bound


ChaoWao:
### Summary

Derive cross-level timeline placement from **span containment** instead of clock
calibration, and make the resulting error bound explicit per span.

Every level already wraps the level below it in a host span, and the same eight
suffixes are reused at every level through the bound level word
(`src/common/log/include/common/host_span_names.h:56`, `level_word()` at `:65`):

```text
network1.dispatch ──────────────────────────────── network1.complete   host A's clock
    └─ node.dispatch ───────────────── node.complete                   host B's clock
           └─ chip.run.runner_run ─────────                            B's host↔device
                  └─ device_wall (ts=0, measured duration)             device counter
```

Each nesting gives one provable statement: *the inner work happened inside the
outer window*. So for any inner event,

```text
slack        = outer_window_duration − inner_measured_duration
position     ∈ [outer_start + δ, outer_start + δ + slack]
```

where `δ` is exact (same counter, same rate inside one level). **The placement
error of any event is bounded by the sum of the slacks along its chain, and every
term is measured, not estimated.**

This is already how the device level works: `device_wall` is emitted with
`ts = 0` and a measured duration, nested under `chip.run.runner_run`
(`src/common/platform/onboard/host/c_api_shared.cpp:565`). The proposal is to
recognise that as the general mechanism, apply it at every level, and publish
`slack` as the error bar.

### Motivation / Use Case

`#2058` asked for one cross-rank timeline; `#2099` delivered it for the same-host
case by mapping each rank's cycles to host `CLOCK_MONOTONIC` through that rank's
own clock anchors, and **refuses to merge when a rank is uncalibrated or comes
from a different host clock domain**.

Containment answers a large part of the same question without any calibration:

- **Same-host cross-rank.** Host span endpoints are `CLOCK_MONOTONIC` and
  `docs/dfx/host-trace.md:141` states they are *"same-host cross-process
  comparable"*. Two ranks' windows are therefore already on one axis, and
  "did A's event precede B's" is decidable whenever the two bounded intervals are
  disjoint — i.e. whenever the gap exceeds `slack_A + slack_B`.
- **Cross-host.** No comparability between the two machines' clocks is needed:
  the L4 process that dispatched to both is a common ancestor, and its
  `network1.dispatch`/`network1.complete` window contains everything the remote
  host did for that dispatch. **The cross-host case is the same operation one
[...]

PR #2217, opened 2026-09-14
Place a peer host's work by the frame that dispatched to it

doraemonmj:
## Why

Two hosts that exchange a task share no clock, so the cross-Rank merge refused
anything from a second machine: comparing the two axes would be wrong rather
than loose. The refusal was wider than the problem. What two axes cannot answer
is where an instant on one sits on the other, and placing a peer needs no such
answer — a remote dispatch blocks from publishing its frame until the completion
arrives, so everything the peer did happened inside that window, and the spare
width between the two durations is the bound. Each duration is timed where it
happened, so the peer's boot time cancels out of the result.

What was missing was a name both sides write down. The wire already carries one:
every `remote_l3` frame header holds `(session_id, worker_id, sequence)`, the
sender allocates the sequence per endpoint, and the peer decodes all three
immediately before it serves the frame. **The join therefore needs no protocol
field and no version bump.**

## What this does

- **`WorkerEndpoint::progress_frame_attrs()`** reports the header of the last
  frame an endpoint published, and the scheduler appends it to
  `<level>.dispatch`. An endpoint that publishes no frame — every local mailbox
  — reports nothing. The header is kept as its fields rather than as text,
  because `submit_progress` is on the dispatch path and a run may never read the
  trace. It is dropped before anything in a submission can throw: a failed
  submission reporting the previous frame's header would name it twice, and the
  later span's window would silently replace the real one.
- **The peer** brackets its service of one frame in `<level>.remote_task`,
  carrying the same three.
- **`containment.remote_windows`** pairs the two sides on that header and
  reports the caller window, the peer window, and the slack between them. A peer
  window wider than its caller's describes no containment and is dropped — the
  usual cause is two logs from different runs whose sequence numbers collide.
  One frame served by two windows refuses outright: that is the whole pile being
  wrong rather than one pair.
- **Chains compose.** A peer that dispatched on to a further peer is reached in
  as many hops as the pile holds, and the slacks add, because the freedom one
  window grants is freedom the windows above it cannot take back. Windows naming
  each other's process reach no axis at all and refuse rather than loop.
- **A peer already on this axis is left where it was recorded.** Its raw window
  sitting inside the caller's means either one machine or two that booted within
  that window's own spare width, and under both readings the recorded position
  is right to within the bound placing it would publish. Placing it would move
  it to the window's start and so collapse the real gap between the dispatch and
  the peer picking the frame up — which is what a loopback run exists to show.
- **Which processes are on the axis is decided, not assumed.** A span names its
  process by pid alone and two machines number theirs independently. A process
  is taken off the axis on positive evidence — its interval overlapping a
  peer's — never on the absence of it, since a local process that ran before the
  dispatch shares no instant with it either. One pid writing two Host logs makes
  every pid ambiguous and refuses the merge.
- **Ranks still have to share one Host clock.** A frame window places the
  process it was served by, never the chip children under it, so nothing puts a
  second machine's Rank on this axis.
- A host-family process is labelled by its own level word, because a spliced
  trace holds `node`, `network1` and the levels above it at once.
- `endpoint_kind_name` gains its MPI group case, which fell through to `unknown`
  and so hid the transport on exactly the dispatches this key exists to follow.

## Cost

Both paths a run pays for are kept off the work they measure.

| | before | after |
| --- | ---: | ---: |
| remote dispatch, trace attributes | 0.652 us | 0.960 us |
| local dispatch, trace attributes | 0.652 us | 0.655 us |
| peer serving a gated-off frame | 1.909 us | 0.452 us |
| placing a 500-frame log | 3964 ms | 117 ms |

The remote dispatch span itself is 20.8 us P50 (`#2134`), so the attribute cost
is ~1.5% of it and the local path is unchanged. The peer's own saving comes from
serving a gated-off frame through one reused do-nothing context manager rather
than building a generator per frame. A chain finds an instant's window by
bisecting that peer's windows rather than scanning every frame the run
dispatched, which leaves the per-span cost flat at 2.3 us where it had been
growing with the frame count.

## Verification

- `tests/ut/py`: 2318 passed, 0 failed. 30 new cases, including that moving one
  machine's clock 900 days changes nothing the placement reports, that a
  three-level chain sums its two slacks, that a failed submission publishes no
  header, and that a peer whose timestamps already land in the window keeps
  them while still reporting its bound.
- `ctest`: the endpoint, scheduler and worker-manager suites pass, with a new
  case reading the reported header back out of the frame that actually left the
  process.
- Onboard a2a3 through `task-submit`: `TestL3Group` (2 cards) and a single-card
  `host_build_graph` case pass. The L3 dispatches carry
  `endpoint_kind=local_mailbox` and no frame header, which is what an endpoint
  that publishes no frame should report.
- A single-file chip trace is unchanged down to the bytes of `traceEvents`,
  against a capture taken today on this branch.

## Not covered

- **The L3 merge is not verified end to end on this branch.** A fresh L3 run
  writes no `chip_swimlane_records.json` on `main` today — bisected to
  `0cc51adf` (#2204), reproduced on a clean `748c39fb` with no local changes —
  so there is no current capture for the cross-Rank path to convert. What is
  verified against fresh data is the single-card chip trace; the L3 comparisons
  this branch was developed against used a capture from before that regression,
  which shows the converter unchanged but not the path end to end.
- **No real cross-host run.** The pairing and the placement are covered by unit
  tests over synthetic logs from three machines, and the single-host half of the
  key is exercised onboard, but nothing here has been through two machines. The
  environment that would do it is the one `#2134` measured on, where a ~21.5 s
  timeout fires in batches.
- **L5 and L6 inherit this for free once they have a remote path.** Nothing in
  the production code names a level: the pairing reads the leaf after any level
  word and the peer builds its span name from its own level. Their remote
  composition is still proposed rather than implemented
  (`docs/hierarchical-level-runtime.md`).

Refs #2134


simpler_setup/tools/containment.py
```diff
@@ -726,3 +730,339 @@ def sum_two_widest_slacks(slacks):
...
+            continue
+        handoff = _attr_ints(span, _DISPATCH_FIELDS)
+        if handoff is None:
+            continue
+        if leaf == COMPLETE_LEAF:
+            completions.setdefault((span.pid, handoff), []).append(span)
+            continue
+        frame = _attr_ints(span, _FRAME_FIELDS)
+        if frame is not None:
+            dispatches[frame] = (span, handoff)
```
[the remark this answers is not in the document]
doraemonmj: Valid, fixed in 0686386.

The cause on the caller side was the endpoint keeping its published header after a failed submission, which this branch already dropped — but the asymmetry you name is the real point: the peer side refuses a repeated frame while the caller side silently kept the later one. `remote_windows` now raises on both, so a pile that names one frame twice is rejected rather than quietly re-pointed at a different, later window whose completion search then closes the wrong end.

Covered by `test_one_frame_dispatched_twice_refuses_rather_than_keeping_the_later`.

simpler_setup/tools/swimlane_converter.py
```diff
@@ -3746,12 +3760,32 @@ def _dispatcher_spans(spans, chip_pids, window_ns):
...
         span for span in spans if span.pid not in chip_pids and not span.is_device and span_family(span.name) != "chip"
     ]
 
+    # A peer process ran against its own machine's clock, so the window test
+    # below would reject every one of its spans by many days. What selects
+    # those is the peer interval each frame's own window names, and a span
+    # outside all of them belongs to a frame this merge did not ask for.
+    peer_pids = chain.peer_pids
+    peer_spans = [span for span in host_spans if span.pid in peer_pids and chain.place(span.pid, span.ts) is not None]
+    host_spans = [span for span in host_spans if span.pid not in peer_pids]
```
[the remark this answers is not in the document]
doraemonmj: The crash path is real and the filter is now `span.pid in chain.axis_pids`, as suggested.

But applying only that would have traded a loud failure for a silent one, because the classification behind it was wrong in a regime I had not tested. `beside_a_peer` reads "interval overlaps a peer's" as "runs on the peer's machine", and two Host clocks that read close enough put *local* processes inside the peer's interval too. Reproduced: with the peer's raw window overlapping the local ones, `axis_pids` came out empty — the dispatching process included — so every window was dropped and the merge failed with a misleading "no runner_run pair". With your filter and nothing else, that becomes a silent empty trace instead.

So the classification is now three-way. A process beside a peer *and* beside the processes that dispatched to it belongs to whichever machine's clock reads that way, and nothing in the logs says which — those go to `ambiguous_pids` and the merge refuses, naming them. Only a process beside a peer alone is taken off the axis.

Measured on both regimes: clocks 17 days apart convert as before (peer placed, its children left out), clocks reading close refuse with `pids 1000, 1001, 1500, 2600 sit beside both a peer and the processes that dispatched to it`.

Covered by `test_clocks_that_read_close_leave_a_process_undecided`, `test_a_peer_days_away_leaves_nothing_undecided`, `test_clocks_too_close_to_tell_the_machines_apart_refuse_the_merge`, and `test_a_peer_side_process_is_not_drawn_on_the_local_path`.

simpler_setup/tools/containment.py
```diff
@@ -726,3 +730,359 @@ def sum_two_widest_slacks(slacks):
...
+        either the same machine or one that booted within this window's own
+        spare width, and under either reading drawing it where it was recorded
+        is wrong by no more than the bound that placing it would publish. So it
+        is left alone: the observed gap between the dispatch and the peer
+        picking the frame up is a real latency, and placing would collapse it
+        to zero.
+        """
+        return (
+            self.caller_start_ns <= self.peer_start_ns
+            and self.peer_start_ns + self.peer_duration_ns <= self.caller_start_ns + self.caller_duration_ns
```
YunjiQin, line 814: **Must fix:** 这里不能仅凭 peer 的原始时间戳恰好落入 caller window，就断定二者 `same_axis`。两台不同主机的 `CLOCK_MONOTONIC` 没有共同零点，但 uptime 恰好接近时同样会满足这个条件；随后 `remote_chain()` 会删掉该 remote window，把 peer 的原始时间戳直接当成本地时间戳，并报告 0 slack。

例如 caller window 是 `[1000, 2000] us`，另一台主机记录的 peer window 恰好是 `[1200, 1800] us`。真实结论只能是 peer 起点位于 `[1000, 1400] us`、误差为 400 us；当前结果却把 1200 us 当成精确位置。必须通过明确的 host/clock-domain identity 证明同轴；证明不了时应保留 containment window 和 slack。

simpler_setup/tools/containment.py
```diff
@@ -726,3 +730,359 @@ def sum_two_widest_slacks(slacks):
...
+    caller_start_ns: int
+    caller_duration_ns: int
+    peer_pid: int
+    peer_start_ns: int
+    peer_duration_ns: int
+
+    @property
+    def slack_ns(self):
+        """Spare width of the caller's window — the peer block's freedom to slide."""
+        return self.caller_duration_ns - self.peer_duration_ns
```
YunjiQin: **Must fix:** 两台机器测得的 duration 不能直接相减后宣称结果是不会低估的误差上界，因为两边时钟 rate 可能相差 10–100 ppm。

反例：caller 真实/记录 40.000 ms，peer 真实 39.990 ms，真实 slack 为 10 us。若 peer clock 快 100 ppm，它会记录约 39.993999 ms，这里只得到约 6.001 us；8 us 的事件间隔会被工具误判为可排序，实际 10 us 的不确定性仍无法排序。真实 slack 更小时还可能算成负数，并在后面被丢弃。

#2134 已明确要求把 clock-rate 项计入误差。请加入有依据的 relative-rate bound（并说明 caller/peer 两侧如何合成），或者不要把该值描述为保证成立的 bound。

simpler_setup/tools/containment.py
```diff
@@ -726,3 +730,359 @@ def sum_two_widest_slacks(slacks):
...
+        low, high = extent
+        return any(low < other_high and high > other_low for other_low, other_high in others)
+
+    peer_extents = [extents[pid] for pid in peers if pid in extents]
+    root_extents = [extents[pid] for pid in roots if pid in extents]
+    beside_peer = {pid for pid, extent in extents.items() if pid not in peers and touches(extent, peer_extents)}
+    beside_axis = {pid for pid, extent in extents.items() if pid in roots or touches(extent, root_extents)} - peers
+    return RemoteChain(
+        windows,
+        frozenset(log_pids) - peers - beside_peer,
```
YunjiQin: **Must fix:** `not in beside_peer` 不是“属于 caller host”的正面证据，因此这里仍会把远程主机上的未知 PID 猜成本地轴。

例如 caller PID 100 的当前 story 是本机 `[1.0, 2.0] s`；已知 peer PID 200 的 `remote_task` 在远程 clock 的 `[10.0, 10.1] s`；同一远程主机上的另一个 PID 201 此前在它自己的 clock `[1.3, 1.4] s` 运行。PID 201 不与 peer extent 重叠，所以会进入 `axis_pids`；其原始数值又与 caller story 重叠，converter 会把它画成本机 `[1.3, 1.4] s`。

需要用稳定的 host/clock-domain identity 分组，或只接受明确归属于 root/peer 的 PID；其余 PID 应拒绝或丢弃，而不是按 extent 猜测。

simpler_setup/tools/containment.py
```diff
@@ -726,3 +730,359 @@ def sum_two_widest_slacks(slacks):
...
+    def _window_holding(self, pid, ns):
+        found = self._by_peer.get(pid)
+        if found is None:
+            return None
+        starts, group, widest = found
+        for index in range(bisect_right(starts, ns) - 1, -1, -1):
+            window = group[index]
+            if ns - window.peer_start_ns > widest:
+                return None
+            if ns <= window.peer_start_ns + window.peer_duration_ns:
```
YunjiQin: **Should fix:** 上面的契约把窗口写成半开区间 `[start, start + duration)`，这里却用 `<=` 把结束点也算进窗口。恰好位于 `peer_start_ns + peer_duration_ns` 的事件会被归到已经结束的 frame。建议改为 `<`，并补一个 end-boundary regression test。

YunjiQin (changes requested):
本次 review 有 3 个 must-fix correctness 问题和 1 个 should-fix 边界问题。核心问题是当前实现仍可能把没有共同 clock domain 的时间戳当成同轴数据，并且发布偏小或为零的误差边界；这与 #2134 的“只能宽松、不能错误”目标不一致。具体见 inline comments。

doraemonmj (commented):
All four addressed in 337e50f5.

**Rate (containment.py:778).** The arithmetic is as you set it out — at 100 ppm a true 10 us slack measures 6 us, and the claim in the description that the rate was "already inside the measured slack" was wrong. `slack_ns` now carries it: the derivation is `S - M = T_p(e_p - e_c) - S·e_c`, bounded by `T_c × MAX_RELATIVE_RATE`, so only the *difference* between the two rates carries — two clocks off by the same amount leave an error proportional to the slack itself. `rate_bound_ns` is published beside the total, since only the measured half shrinks when a window tightens. 1.3 us on the 13.3 ms window `#2134` measured across two hosts.

**same_axis (containment.py:794).** You are right about the consequence and I want to be precise about which half. The *position* was within bound — if the peer's raw window sits inside the caller's, the true offset is at most the slack, so the recorded instant is within `slack_ns` of the truth either way. What was wrong was reporting **0 slack**, which claims exactness. The window now stays in the chain: the peer keeps its recorded instant, the bound is published all the same, and because the truth can lie either side of an observed instant the lane reads `observed, ±N us` rather than `placed, +N us`. `metadata` gained `drawn_at_ns` and `observed` so the bound and the drawn position are two separate facts. Keeping the position matters on a loopback session, where placing would collapse the real dispatch-to-serve latency to zero.

**Axis inference (containment.py:958).** Dropped entirely rather than patched — your counterexample is the general case, and I found a second one while testing: two clocks reading close enough put *local* processes inside the peer's extent too, which made `axis_pids` come out empty and killed the merge with a misleading "no runner_run pair". No rule over the timestamps can separate the two machines. So the merge now vouches rather than infers: a process at the top of a chain, or one whose own device window the merge is of; everything else is left out. Peers are excluded from that set as well, or a peer that also has a device window would be drawn twice.

**Half-open window (containment.py:899).** `<` now, with an end-boundary regression test.

Verification: `tests/ut/py` 2318 passed, onboard `TestL3Group` on a2a3 passes, and a same-host merge is unchanged down to the bytes of `traceEvents`.

simpler_setup/tools/containment.py
doraemonmj: Fixed in 337e50f5, and I want to be precise about which half of this was wrong, because the two halves have different answers.

**The position was within bound.** If the peer's raw window sits inside the caller's, then both the recorded instant and the true one lie in an interval of width `slack_ns` — containment puts the true position in `[caller_start, caller_end - peer_dur]`, and the recorded one is in there too by the test itself — so `|true - recorded| <= slack_ns`. In your example that is the 400 us you name, and drawing at 1200 us is off by at most that. Placing it at 1000 us instead is not more accurate, only differently placed.

**Reporting 0 slack was wrong.** That is what `remote_chain()` did by dropping the window, and it claims exactness where there is none. That is the defect, and it is a straight violation of `#2134`'s "cannot be wrong, only loose".

So the window now stays in the chain and only the *move* is skipped: the peer keeps its recorded instant, and the bound is published exactly as it is for a placed block. Because the truth can lie either side of an observed instant — where a placed block is drawn at the earliest position its window allows and can only be later — the lane reads `observed, ±N us` rather than `placed, +N us`, and `metadata` gained `drawn_at_ns` and `observed` so that where the bound is and where the block sits are two separate facts rather than one conflated field.

On proving same-axis by identity instead: I looked, and there is nothing to prove it with. `host_clock_domain_id` is written only into a capture (`chip_swimlane_collector.cpp:1144`); a Host log carries no clock-domain field at all, so the converter has no identity to read. Adding one is a runtime change worth doing — it would also let `remote_chain` drop its vouching rule entirely — but it is not something this PR can lean on today.

Keeping the recorded position matters for the one configuration that can be tested without two machines: `remote_l3_worker.py:307` defaults to `127.0.0.1`, and on such a run placing the peer would move it to the window's start and collapse the real dispatch-to-serve latency to zero.

Covered by `test_a_peer_already_on_this_axis_is_left_where_it_was_recorded`, `test_metadata_separates_where_the_bound_is_from_where_the_block_sits`, and `test_a_peer_whose_timestamps_already_land_in_the_window_keeps_them`.

PR #2251, opened 2026-09-16
Refactor: resolve Graph recording by provenance and shrink its bind path

poursoul:
## Summary

A recorded Graph body had to guess where each tensor came from — boundary tensors by binary-searching a sorted buffer-address array and comparing every field, internal outputs by whether the address sat above `1<<63`. Both are artefacts of a boundary carrying the caller's own addresses. The same paths also copied and compared more bytes per tensor than their readers need.

The boundary now lives in the recording's own address space and every tensor in it is stamped with `TaskId::Space::PARAM` plus its parameter index; a body's own outputs keep `Space::IN_GRAPH`. Classification reads the stamp — no search, no address-range assumption, and a tensor that never passed through the boundary is refused by name instead of being attributed to whichever parameter its address happened to land on.

- **Provenance** — recorded tensors become relocation records (parameter-relative or graph-heap-relative); `GraphTensorSourceRef`, `GraphBoundarySignature`, `GRAPH_RECORD_VIRTUAL_BASE` and the address-range scheme are gone. Alias partitioning is settled once by a routine both recording and matching call.
- **Alias partition** — grouping parameters by buffer address needs no ordering, so the routine hashes the addresses into a 256-slot byte table instead of sorting them; each parameter's representative is the first parameter that reached its address, and argument order is the scan order. The hash is `tensormap.h`'s own `addr_to_slot`, now shared with the tensor map's buckets so one spelling covers every table this runtime keys by a buffer address. Linear probing is quadratic when addresses cluster onto one slot — which the table's half-load bound does not cap — so the hash concedes to a sort after four probes per parameter, keeping an `O(n log n)` ceiling under the routine.
- **Bytes** — boundary comparison moves to `GraphBoundaryParamMatch` (64 B, sized to the boundary rather than a 16 KB fixed array); the image element becomes `Tensor`'s own `TensorData` base instead of a separate `GraphTensor` wire struct; `init_geometry_from` copies `ndims` slots instead of `MAX_TENSOR_DIMS`.
- **Device replay** — `graph_rebind_tensor` is now one copy plus one addition. It was two struct copies, two wire validations, and three dependent loads into other tasks' descriptors. Validation moves to record time, where it runs once per Definition instead of once per tensor per replay.
- **Contract** — three changes a caller can see, all in [`GRAPH_EXECUTION.md`](src/common/host_build_graph/docs/GRAPH_EXECUTION.md).
  - *Tightened*: a recorded task may no longer name a pre-Graph task in an explicit dependency. The old form was admitted when that task created a boundary tensor; a captured parameter now carries `Space::PARAM` rather than its creator's GLOBAL id, so the supported spelling is `set_dependencies({param.owner_task_id})` and a GLOBAL id reaching a body is refused by name.
  - *Relaxed*: a parameter's `start_offset` may now move between invocations, provided its whole alias partition moves with it — what the reuse condition pins is each parameter's origin relative to the lowest-numbered parameter sharing its buffer, not its absolute origin, so a sliding-window caller keeps reusing its Definition instead of recording per slice.
  - *Moved from a check to a precondition*: that two parameters over different addresses name non-overlapping memory is no longer verified. A runtime-allocated buffer satisfies it for free, since one allocator hands out disjoint blocks; boundary storage is caller-owned, though, so a caller-provided address (`ChipTensor.make(..., child_memory=True)`) is the caller's to supply. Proving it per invocation costs exactly the ordering the hash exists to avoid. What the routine still refuses is answerable from one address's own group — an empty buffer, and one address presented at two sizes, which contradicts the single recording-space window reserved per address.

## Measured on a2a3 silicon

Both sides clean-rebuilt (`build/cache` + `build/lib` cleared) and pinned to the same dies, 8 processes × 9 warm rounds per side, `--skip-golden`. `device_wall` is the device-clock execution span. `P` is P(branch > main) over the pooled samples — a rank-sum estimate, so 0.5 means no difference and 0.0 means fully separated.

### qwen3-14B decode — 72 samples per side, die 1

| metric | main `22385d2b` | this branch | delta | P |
| ------ | --------------- | ----------- | ----- | - |
| **`device_wall`** | 37.042 ms | **36.778 ms** | **−0.264 ms (−0.7%)** | **0.000** |
| `chip.run` | 38.475 ms | 38.073 ms | −0.402 ms (−1.0%) | 0.131 |
| `bind` | 1.133 ms | 1.023 ms | −0.111 ms (−9.8%) | 0.376 |
| `bind.host_orch` | 0.333 ms | 0.265 ms | −0.068 ms (−20.5%) | 0.299 |

Per-process medians do not overlap — main's fastest process is slower than the branch's slowest:

```
main:   36.987 36.996 37.002 37.039 37.062 37.081 37.164 37.227
branch: 36.590 36.642 36.697 36.776 36.815 36.824 36.850 36.858
```

The other nine `bind` segments (`args`, the six arena segments, both H2D segments) move by at most ±0.002 ms with P between 0.43 and 0.55 — the change stays inside orchestration and does not leak into the rest of the bind path. **Superseded:** the 10 × 10 measurement below finds `arena_h2d` at P 0.726 on this case, so it does leak; see "Re-measuring the host side".

### DeepSeek-V4 FLASH decode — 144 samples per side (2 ranks), dies 1 & 3

| metric | main `22385d2b` | this branch | delta | P |
| ------ | --------------- | ----------- | ----- | - |
| **`device_wall`** | 35.468 ms | **33.924 ms** | **−1.544 ms (−4.4%)** | **0.007** |
| `chip.run` | 36.788 ms | 35.606 ms | −1.182 ms (−3.2%) | 0.059 |
| `bind` | 0.890 ms | 1.154 ms | **+0.264 ms (+29.7%)** | 0.772 |
| `bind.host_orch` | 0.409 ms | 0.548 ms | +0.138 ms (+33.8%) | 0.759 |
| `bind.arena_h2d` | 0.051 ms | 0.082 ms | +0.031 ms (+61.9%) | 0.879 |
| `bind.graph_upload` | 0.157 ms | 0.198 ms | +0.041 ms (+26.0%) | 0.697 |

All 16 rank-level medians separate: main 35.205–35.848, branch 33.699–34.556.

**The four `bind*` rows of this table are superseded and not reproduced.** They carry no untouched-segment control, and the 10 × 10 measurement below finds this branch *faster* on dsv4 `host_orch` and `arena_h2d` — while also showing the untouched segments moving by as much, which is why neither direction can be claimed. The `device_wall` and `chip.run` rows are unaffected.

### Reading the two together

The device gain scales with tensors per replay, which is where `graph_rebind_tensor`'s removed work lands:

| | tensor args | `device_wall` gain |
| - | ----------- | ------------------ |
| qwen | 1,059 | −0.264 ms (−0.7%) |
| dsv4 | 3,113 (2.9×) | −1.544 ms (5.8×) |

**The device gain is not free.** The shared-memory image grows 25% (360 KB → 453 KB) because the outer Graph task's payload is now `Tensor`-typed, while the Definition image shrinks 12% (1,020,624 → 895,248 bytes). Those are sizes, and they are exact.

The host *timings* below them turned out to need a better measurement than the one in the two tables above, which sampled 8 processes × 9 rounds and carried no drift control on the dsv4 side. See the next section: it supersedes the four `bind*` rows of those tables, and contradicts three of them.

Both cases net out ahead end to end on the device: `chip.run` −1.0% on qwen, −3.2% on dsv4. The `device_wall` rows are unaffected by anything below — that measurement skips the device run entirely.

### Re-measuring the host side: 10 × 10, with a drift control

10 alternating repetitions (`base, branch, base, branch`) × 10 rounds per arm per case against `22385d2b`, `SIMPLER_SKIP_DEVICE_RUN=1`, both arms' `[stamp]` and `torch_backend_autoload` records matching, each arm in its own worktree so neither rebuilds between repetitions. Every warm bind is a sample: **n=90 per arm on qwen, n=180 on dsv4** (two ranks). `P` is P(branch > base) over the pooled samples, as in the tables above.

The reading rests on the segments this branch **does not touch** — `args`, `arena_build`, `runtime_init`, `static_arena`, `shared_mem`, `gm_heap`, `host_view_close`. They are the ruler: if they move, the arms did not run on the same machine state.

**qwen — the ruler is clean, and two segments move.**

| segment | base median | branch median | delta | P |
| ------- | ----------- | ------------- | ----- | - |
| the 7 untouched segments | — | — | +0.000…+0.001 | **0.511–0.544** |
| `graph_upload` | 0.059 | 0.060 | +0.001 | 0.491 |
| **`host_orch`** | 0.292 | **0.257** | **−0.036 (−12%)** | **0.360** |
| **`arena_h2d`** | 0.044 | **0.049** | **+0.005 (+11%)** | **0.726** |

Untouched segments sit at P ≈ 0.5, so `host_orch` faster and `arena_h2d` slower are both real. `arena_h2d` slowing is the 25% larger shared-memory image arriving as a cost — the table above claimed qwen's non-`host_orch` segments were all neutral at P 0.43–0.55, and that is **wrong**: this is the H2D cost of the `Tensor`-typed payload, visible on the case with the quieter bind rather than the louder one.

**dsv4 — the ruler moves as much as the segments under test, so nothing is attributable.**

| segment | delta median | P |
| ------- | ------------ | - |
| `args` (untouched) | −0.006 | 0.303 |
| `static_arena` (untouched) | −0.000 | 0.293 |
| `shared_mem` (untouched) | −0.000 | 0.323 |
| `arena_build` (untouched) | −0.000 | 0.321 |
| `host_orch` | −0.134 | 0.266 |
| `graph_upload` | −0.069 | 0.220 |
| `arena_h2d` | −0.003 | 0.417 |

The untouched segments land at P 0.29–0.40 against the control plane's 0.22–0.27 — the same band. The branch arm simply ran on a faster machine state, so dsv4's apparent `host_orch` and `graph_upload` wins cannot be separated from it.

That also means the dsv4 table above is **not reproduced**: it reported `host_orch` +33.8% and `arena_h2d` +61.9% on this branch, and this measurement finds the branch *faster* on both (P 0.266 and 0.417). Neither figure survives, and neither direction can be claimed — the earlier dsv4 rows had no untouched-segment control to rule out exactly this.

### Alias partition: why a hash, and why it concedes

The partition runs once per same-key submission, so its shape was chosen by measurement over the boundary sizes that occur — dsv4's eight boundaries are 23/30/23/30/48/30/33/71 parameters, qwen's is 26. Arena-allocated addresses arriving in argument order, ns per call:

| | n=23 | n=26 | n=48 | n=71 | n=128 |
| - | ---- | ---- | ---- | ---- | ----- |
| sort | 159.7 | 181.0 | 383.6 | 606.7 | 1196.6 |
| **hash** | **27.3** | **30.3** | **57.5** | **106.5** | **187.5** |

Three alternatives measured worse and are not taken:

- **Staging the addresses into side arrays before probing** costs 12–27% more at every size. Each key would be written once to be read about once, and the parameter it came from is already in cache from the `buffer.size` test.
- **`std::unordered_map`** costs 10–14× the hash and 3–4× the sort, on node allocation and pointer chasing.
- **Widening the table to 512 or 1024 slots** loses at every size up to 48 — the table is cleared per call, so a lower load factor buys fewer probes than the larger `memset` costs — and does nothing for the clustered case, which is quadratic at any width.

On 128 addresses hashing to a single slot the probe budget fires and the routine costs 1.55 µs against the sort's 1.07 µs, where unbudgeted probing cost 6.94 µs. That 1.5× is the budget's standing price and the reason it is four probes per parameter rather than more. Such a set was found by scanning a 1 TB range keeping 1 address in 256; an allocator handing out consecutive or power-of-two-strided blocks cannot produce one, since that is precisely the input a multiplicative hash spreads.

## Testing

- [x] `qwen3_14b_decode` golden passes on both main and this branch — `out`, `k_cache`, `v_cache` within `rtol=5e-2 / atol=1e-1`
- [x] `deepseek_v4_flash_decode` completes with the real 42.6 GiB/rank fixture uploaded, both ranks `outcome=0` (this case has no golden)
- [x] 8 processes × 10 rounds per case per side, all `PASSED`, no `507*` and no scheduler diagnostics
- [x] 147/147 cpput, including three new cases on the boundary contract: a parameter origin that slides between invocations keeps reusing its Definition and the outer payload carries the slid origin; partition members that slide against each other do not, while a uniform slide of the same pair does; and one address presented at two sizes takes the ordinary path
- [x] pre-commit clean (clang-format, clang-tidy, cpplint, markdownlint, header/name checks)
- [x] `bind` phase A/B against `22385d2b`: 10 alternating repetitions × 10 rounds per arm per case, pooled per-bind samples (n=90 qwen / n=180 dsv4). qwen resolves two segments against a neutral untouched-segment ruler; dsv4 does not resolve any, and does not reproduce the earlier tables' dsv4 host-side rows. Method and numbers above.

### Caveat on the measured absolute values

The A/B numbers above are collected with `--skip-golden`, which leaves child memory uninitialized. dsv4's MoE tile loops dispatch on `recv_count_out[expert][0] > t0` evaluated on device, so uninitialized counts skip tasks that real routing would run — with the fixture uploaded `device_wall` is 49–56 ms rather than 33.9 ms. Both sides use the same seed and the same uninitialized state, so the **relative** comparison holds; the absolute figures should not be quoted as dsv4 decode latency.


src/common/host_build_graph/host/orchestrator.cpp
```diff
@@ -341,24 +321,62 @@ static_assert(
...
+static_assert(sizeof(GraphBoundaryParamMatch) == 64, "one parameter's comparison must stay one cache line");
+
+GraphBoundaryParamMatch graph_boundary_param_match_of(const simpler::hbg::Tensor &tensor, TensorArgType tag) {
+    GraphBoundaryParamMatch match{};
+    match.buffer_size = tensor.buffer.size;
+    match.start_offset = tensor.start_offset;
+    // Only the dimensions in use, which is also the range graph_boundary_param_matches
+    // compares: a value-initialized record then holds zero past ndims rather than
+    // whatever the argument's unused slots happened to carry.
+    std::copy(std::begin(tensor.shapes), std::begin(tensor.shapes) + tensor.ndims, std::begin(match.shapes));
```
zhusy54, line 361: `tensor.ndims` is used to form both source ranges before it is validated. On a first miss, `rt_graph_args_cacheable()` only checks argument counts, so a public/default-constructed `Tensor` with `ndims > MAX_TENSOR_DIMS` reads past both fixed arrays here instead of taking the ordinary path. Please validate the complete boundary before constructing `param_match` and add a first-miss invalid-rank regression test.

src/common/host_build_graph/host/orchestrator.cpp
```diff
@@ -844,132 +1012,93 @@ bool graph_recording_reset(GraphRecording &recording, const GraphInflightRecordi
...
-    const TaskId owner = tensor.owner_task_id;
-    if (owner.is_valid() && !owner.is_global()) {
-        const int32_t owner_index = owner.local_id();
-        if (owner_index >= 0 && owner_index < task_index && window_offset(owner_index, tensor_addr, &offset)) {
-            source->source_kind = GraphRecordedTensorSourceKind::INTERNAL;
-            source->source_index = static_cast<size_t>(owner_index);
-            source->packed_offset = offset;
-            return true;
-        }
+    if (owner.space() == TaskId::Space::PARAM) {
```
zhusy54, line 1047: Please explicitly require `owner.space() == TaskId::Space::IN_GRAPH` before treating the owner as a producer. Today every non-PARAM, non-GLOBAL space falls through to the producer path, and device replay mirrors the same PARAM/else split, so invalid provenance with a plausible local id is not rejected. Please reject unsupported spaces here and add a regression test.

src/common/host_build_graph/host/orchestrator.cpp
```diff
@@ -844,132 +1012,93 @@ bool graph_recording_reset(GraphRecording &recording, const GraphInflightRecordi
...
+            "on: a Definition's edges run from tasks already recorded",
+            task_index, static_cast<unsigned long long>(tensor_addr), producer_index
+        );
+        return false;
     }
-    return false;
+    const RecordedInGraphTask &producer = recording.tasks[producer_index];
+    // Compared as a subtraction on unsigned values so a block ending past the address space
+    // cannot wrap the bound.
+    if (producer.total_output_size == 0 || tensor_addr < producer.record_packed_base ||
```
zhusy54, line 1066: The previous replay path validated both the full tensor geometry and `packed_offset + buffer_size <= producer_bytes`. This replacement only proves that the buffer start is inside the producer window; an oversized buffer or inconsistent extent is accepted, serialized, and rebound without any later check. Please move the complete former `graph_tensor_wire_valid` and producer-span validation to record time and add a regression test, or retain those checks on replay.

src/common/host_build_graph/host/orchestrator.cpp
```diff
@@ -844,132 +1012,93 @@ bool graph_recording_reset(GraphRecording &recording, const GraphInflightRecordi
...
+// recording's output region -- an affine image of the heap a replay commits -- and its
+// origin is already its own.
+//
+// Between them the image holds no address from the recording's own space, and none from
+// the caller's either: it is readable knowing only the heap it will be bound to and the
+// arguments it will be bound against.
+void graph_tensor_relocation_record(
+    const GraphRecording &recording, const simpler::hbg::Tensor &tensor, simpler::hbg::TensorData *record
+) {
+    record->init_from(tensor);
```
zhusy54, line 1094: `TensorData` is now a serialized wire record, but `init_from()` writes only the active shape/stride slots. The retained arena is not cleared before this fill, so unused slots retain bytes from a previous Definition and are uploaded, contrary to the later "every section is written in full" contract. Please canonicalize each wire record before filling it and add a bytewise determinism test; keep the partial-copy optimization only for working tensors.

src/common/task_interface/tensor.h
```diff
@@ -39,7 +39,7 @@ struct PTOBufferHandle {
  * A per-tensor tag carried by TaskArgs (drives dependency inference at submit
  * time; stripped before the args cross the dispatch boundary).
  */
-enum class TensorArgType : int32_t {
+enum class TensorArgType : uint8_t {
```
zhusy54, line 42: Changing the underlying type here changes the layout of every `TaskArgsTpl<..., TensorArgType>` consumer, including TMR structures, although this Graph optimization only needs a one-byte field in `GraphBoundaryParamMatch`. Please keep the shared enum layout stable and store a Graph-local `uint8_t` with explicit conversions instead. If a project-wide byte-sized ABI is intended, please separate that change and validate every cross-component consumer.

src/common/host_build_graph/docs/GRAPH_EXECUTION.md
```diff
@@ -166,23 +166,59 @@ why it cannot be diagnosed instead:
...
+  tensor is rebound; a shift of one member against another is not, because the
+  overlap geometry recording inferred from those origins is baked into the
+  Definition. A sliding-window caller that hands the same cache a different
+  slice each invocation therefore keeps reusing its Definition.
+- The buffers a boundary names must be pairwise identical or disjoint. Two
+  parameters either share a buffer exactly — same address and same size, as two
+  views of one tensor do — or share nothing. This one is a **precondition, not a
+  checked property**: hazard tracking groups by buffer address, so two buffers
+  that partially overlap at different addresses would be called unrelated memory
+  and the edges between them would be lost. Every buffer a boundary can name
```
zhusy54: This invariant is not guaranteed for child-memory arguments: `ChipTensor.make(..., child_memory=True)` accepts a caller-provided device address, HBG passes it through unchanged, and `add_child_memory()` does not reject overlaps. Two distinct bases can therefore partially overlap and this hash will classify them as different partitions, dropping WAR/WAW edges. Please either validate child-memory boundary overlaps or document and enforce the caller contract; the claim that the allocator guarantees this is not accurate.

src/common/host_build_graph/docs/GRAPH_EXECUTION.md
```diff
@@ -390,25 +464,33 @@ Definition. It contains:
...
+`TensorData`, the 96-byte base the 64-byte-aligned runtime `Tensor` derives from:
+the image is built in a byte vector, which has nowhere to put an over-aligned
+element. The upload is therefore one contiguous copy with no raw Host pointers
+and no relocation pass.
 
-Before materialization, the Scheduler recomputes the Definition content hash
-and validates section ranges, topology indices, in-graph task heap offsets, the outer
-heap extent, ChipTensor metadata, and ChipTensor-source bounds. Invalid wire data is
-rejected before an offset participates in pointer arithmetic.
+Before materialization, the Scheduler recomputes the Definition content hash and
```
zhusy54: There is no Definition `content_hash` field or recomputation in the current implementation; `GraphDefinitionHeader` only carries framing and identity fields, and the device performs structural bounds/topology checks. Please update this paragraph and the corresponding introduction so the documented integrity model matches the code.

src/common/host_build_graph/host/orchestrator.cpp
```diff
@@ -844,132 +1012,93 @@ bool graph_recording_reset(GraphRecording &recording, const GraphInflightRecordi
...
+    const TaskId owner = tensor.owner_task_id;
+    if (owner.space() == TaskId::Space::PARAM) {
+        record->buffer.addr = 0;
+        record->start_offset -= recording.bound_boundary().params.tensor(owner.local_id()).ref().start_offset;
```
ChaoZheng109: 这是一次没有任何保护的无符号减法。main 里对应的前置条件是显式的——graph_tensor_from_boundary 只在 `tensor.start_offset >= param_tensor.start_offset` 时才接受一个 view——而这个比较被删掉了：新的 graph_tensor_from_boundary 只比较 `buffer.addr` 和 `buffer.size`。

目前不可达：所有 view op 我都过了一遍（view / transpose / permute / slice / reshape，含 `view()` 那条保持 start_offset 不变的空视图分支），没有任何一个会减小 start_offset，所以 owner 为 PARAM(i) 的 tensor 总是落在参数 i 的原点之后。但这个不变量现在完全由函数外部维持，而一旦在这里下溢，写进 image 的就是一个接近 2^64 的相对原点，下游没有任何地方会再检查它——graph_rebind_tensor 加上实参原点后直接返回。

请把前置条件收回本地：在这里加 `debug_assert`，或者把那个 `>=` 判据加回 graph_tensor_from_boundary。

src/common/host_build_graph/host/orchestrator.cpp
```diff
@@ -1919,63 +2023,109 @@ static TaskOutputTensors submit_task_common(
...
-        if (tensor.ndims > MAX_TENSOR_DIMS) {
+        const simpler::hbg::Tensor &actual = args.tensor(i).ref();
+        if (actual.ndims > MAX_TENSOR_DIMS) {
             debug_assert(
-                tensor.ndims <= MAX_TENSOR_DIMS && "Graph boundary simpler::hbg::Tensor rank is not supported"
+                actual.ndims <= MAX_TENSOR_DIMS && "Graph boundary simpler::hbg::Tensor rank is not supported"
             );
             LOG_WARN(
-                "[GraphExecution] simpler::hbg::Tensor rank %u exceeds the fixed Graph boundary limit", tensor.ndims
+                "[GraphExecution] simpler::hbg::Tensor rank %u exceeds the fixed Graph boundary limit", actual.ndims
```
ChaoZheng109: `ndims` 现在是 `uint8_t`，这里直接把它喂给了 `%u`。2111 行对 `actual.ndims` 和 `expected.ndims` 同样如此。

靠默认实参提升，在本项目所有目标 ABI 上都能工作，但同一个改动在 `Tensor::to_string()`（tensor.h:486）里专门为这件事加了 `static_cast<unsigned>`，于是同一个转换在树里出现了两种写法。请统一：三处都加转换，或者三处都不加。

src/common/host_build_graph/host/orchestrator.cpp
```diff
@@ -378,8 +402,181 @@ struct GraphBoundary {
...
+            slot = (slot + 1) & (SLOT_COUNT - 1);
+        }
+    }
+    return true;
+}
+
+// Move a captured boundary's parameters into the recording's own address space: each
+// tensor takes a recording-space address and an owner naming its own parameter index.
+// Returns false when the boundary is not one this runtime can represent, in which case
+// nothing has been rewritten.
```
ChaoZheng109: "in which case nothing has been rewritten" 不准确。在失败路径上——即 557 行 graph_alias_partition 返回 false——`boundary.param_match` 已经被上面那个循环 resize 并填好了，只有 `alias_rep` 字段还停在值初始化的 0。

因为 entry 在发布前就被丢弃，所以无害；但这句注释是写给下一个读这个函数的人的契约。真正成立的是更窄的那句：没有任何 tensor 被改写，所以边界仍然持有调用方自己的地址和 owner。请改成这个说法。

src/common/host_build_graph/device/graph_execution.cpp
```diff
@@ -261,14 +251,14 @@ bool graph_predicate_resolve(
...
     }
     if (!operator_known) return false;
-    const uint64_t element_size = get_element_size(static_cast<DataType>(operand.dtype));
+    const uint64_t element_size = get_element_size(operand.dtype);
     if (element_size != 1 && element_size != 2 && element_size != 4 && element_size != 8) return false;
-    if (predicate.elem_size != element_size || predicate.elem_offset >= operand.extent_elem) return false;
-    // graph_tensor_wire_valid bounds start_offset + extent_elem by the buffer's
-    // element count, so the scaled sum cannot leave the buffer.
+    if (predicate.elem_size != element_size || predicate.elem_offset >= operand.extent_elem_cache) return false;
+    // The recorder bounded elem_offset by the operand's own extent, and the operand's
```
ChaoZheng109: 这段论证的后半句不成立。录制侧确实用 operand 自身的 extent 约束了 elem_offset——orchestrator.cpp:2552 会拒绝 `flat_offset - operand->start_offset >= operand->extent_elem_cache`。但它没有做 "the operand's extent by its buffer"：main 的 graph_tensor_wire_valid 曾经在 replay 侧证明 `start_offset <= buffer_elements && extent_elem <= buffer_elements - start_offset`，而新的 graph_tensor_from_boundary 只比较 `buffer.addr` 和 `buffer.size`。

也就是说，这条注释引用的是一个被删除、而非被前移的 record-time 检查。请二选一：要么把这个包含性检查补回录制侧——录制器在那里同时持有参数和派生出的 tensor，成本是每个 Definition 一次——要么把注释改成陈述真实的前置条件，即调用方传入的 tensor 是良构的。

src/common/host_build_graph/tensor_create_info.h
```diff
@@ -60,7 +60,7 @@ class TensorCreateInfo {
     }
 
 public:
-    uint32_t ndims{0};
+    uint8_t ndims{0};
```
ChaoZheng109, line 63: 不阻塞本 PR，只是记一下这里有同类问题，后续单独处理即可。

这个 struct 唯一的消费者 `ChipTensorMapEntry::copy_tensor_create_info` 是死代码，有三份副本（`src/common/host_build_graph/tensormap.h:190`、`src/{a2a3,a5}/runtime/tensormap_and_ringbuffer/runtime/tensormap.h:187`），全仓没有任何调用者。

它不只是没人用，本身还有两处是错的，而这次窄化让其中一处更糟：`memcpy(this, &tensor_create_info, 64)` 从一个 TensorCreateInfo 上读 64 字节，而该对象在 main 上是 28 字节、在 ndims 窄化之后是 24 字节；并且字段偏移根本对不上——`ndims` 在 create info 里位于偏移 0，在 entry 里位于偏移 36，所以即使尺寸正确，这次拷贝也填不出任何有意义的东西。

后续起一个清理 PR 把这三份一并删掉比较合适（另外两份不在本 PR 的 diff 范围内）。

src/common/host_build_graph/tensor_create_info.h
```diff
@@ -60,7 +60,7 @@ class TensorCreateInfo {
...
     DataType dtype{DataType::FLOAT32};
     bool manual_dep{false};
     uint32_t shapes[MAX_TENSOR_DIMS]{};
```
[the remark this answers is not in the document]
poursoul: 这条不改，三点说明：

- `build_output_layout` 在本仓库不存在。真正对 create info 调用 `buffer_size_bytes()` 的函数是 `calculate_output_layout`（`src/common/host_build_graph/host/orchestrator.cpp:1558`）。
- 该路径不可达。仓库里没有任何一处直接给 `TensorCreateInfo::ndims` 赋值，所有构造都走带校验的构造函数，其中断言 `0 < ndims <= MAX_TENSOR_DIMS`。
- 把字段改回宽类型也解决不了所指的问题。用 `uint32_t` 时，`257` 会一路走到 `init_tensor_from_create_info` 并触发 `always_assert`，而不是静默变成 rank 1 —— 两种结果都不优雅。「公开字段绕过构造函数校验」的正解是把字段私有化、配一个带校验的 setter，那属于独立改动。

`MAX_TENSOR_DIMS` 是 5，`uint8_t` 仍有 50 倍余量；被替换掉的 `uint32_t` 是多给的。

src/common/host_build_graph/host/orchestrator.cpp
poursoul: 这条不改 —— 目前这个 fallthrough 是穷举的。

`TaskId::Space` 只有三个取值（`src/common/host_build_graph/task_id.h:71`：`GLOBAL = 0, IN_GRAPH = 1, PARAM = 2`）。上面的分支已拒掉 invalid 与 GLOBAL、并对 PARAM 直接返回，所以能走到 producer 路径的只剩 IN_GRAPH。

即使 raw 字有个落在枚举之外的位型，紧随其后的守卫也会在任何数组被索引之前拦住它：先是 `producer_index < 0 || producer_index > task_index`，再是后面的块范围检查。所以这个显式比较背后没有可达的情形。

同意写出来是更好的习惯，但它属于防御性补充，而不是修缺陷。

src/common/host_build_graph/host/orchestrator.cpp
poursoul: 这条不改 —— 这里的整段检查是冗余而非缺失。

本运行时的 buffer 由统一的分配器给出，人为构造的 tensor 不在受支持的契约内。一个录制输出的 `buffer` 正是 `calculate_output_layout` 从该任务自身 `total_output_size` 里切出来的那一块，而 `start_offset` 是这同一个 buffer 内的原点。所以 `addr` 落在 `[record_packed_base, record_packed_base + total_output_size)` 内，就已经意味着整个 tensor 都在其中：`addr + start_offset` 不可能越过 `addr + buffer.size`，而 `buffer.size` 由构造即在块内。

也就是说，包含性是地址产生方式带来的性质，不是在这一点上再加一次比较能够加强的东西。

src/common/host_build_graph/host/orchestrator.cpp
poursoul: 这条不改。拷贝 `ndims` 个 slot 是正确行为，不是「只拷了一部分」。

`ndims` 之后的 slot 不描述任何轴，而所有读者 —— host 侧和 device 侧一样 —— 都以 `ndims` 收界。改成拷 `MAX_TENSOR_DIMS` 个 slot 同样正确，作为定长拷贝甚至可能更快；但那是性能取舍，不是正确性问题，也不能说明按 `ndims` 拷贝是错的。

单就确定性这一点：没有任何地方比较或摘要 Definition 的 image 字节。`graph_definition_from_object`（`src/common/host_build_graph/device/graph_execution.cpp:177-181`）校验的是 `magic`、`definition_bytes`、`total_bytes` 和 `full_key`，实现里不存在对 image 字节的 hash。

src/common/task_interface/tensor.h
poursoul: 这个改动是有意保留的。

`TensorArgType` 只有五个取值，本来就该是字节宽的；`int32_t` 是一直没人回头处理的遗留。收窄它正是这个声明早该做的清理，所以放在这里，而不是用一份 Graph 局部的 tag 副本绕过去。

它也不跨任何边界：tag 在 args 越过 dispatch 边界之前就被剥离，既不出现在 Definition image 里，也不出现在 `remote_wire` 里，而 `tags_` 是 host 侧 `TaskArgsTpl` 的存储。所以对 HBG 和 TMR 的实际影响，就是两边的 `Arg` 都变小了。

src/common/host_build_graph/host/orchestrator.cpp
poursoul: 已加 `debug_assert`。把 `start_offset` 提成局部量 `param_origin`，断言写在减法前面，并把不变式作为一句陈述放在旁边：没有任何 view op 会降低 `start_offset`，所以归属参数 i 的 tensor 其原点必然在该参数原点之后。

你对当前不可达的判断我核对过，一致：main 上那个 `>=` 判据确实随着 `graph_tensor_from_boundary` 改成只比较 `buffer.addr` 和 `buffer.size` 而一并没了，而减法搬到了这里。理论上不可能为负，但这个不变量现在由本函数之外维持，所以按你说的把前置条件收回本地，而不是依赖调用方。

src/common/host_build_graph/host/orchestrator.cpp
poursoul: 已统一，三处都加了 `static_cast<unsigned>`，和 `Tensor::to_string()` 保持一致。

一处订正：这两条 `LOG_WARN` 里的 `actual` 是 `simpler::hbg::Tensor`，它的 `ndims` 仍然是 `uint32_t`，所以 `actual.ndims` 配 `%u` 本来就是相符的；真正依赖默认实参提升的只有 `expected.ndims`（来自 `GraphBoundaryParamMatch`，`uint8_t`）那一处。不过一致性这个诉求本身成立，所以三处一起加了。

src/common/host_build_graph/host/orchestrator.cpp
poursoul: 已按你给的表述改：

> Returns false when the boundary is not one this runtime can represent. No tensor is rewritten on that path, so the boundary still holds the caller's own addresses and owners.

确认过时序与你说的一致：`param_match` 在上面那个循环里已经 resize 并填好，`graph_alias_partition` 要到之后才可能返回 false，所以原来那句「nothing has been rewritten」确实不成立。

src/common/host_build_graph/device/graph_execution.cpp
poursoul: 已改注释，取你给的第二个选项——陈述真实的前置条件：

> The recorder bounded elem_offset by the operand's own extent. That the extent itself lies inside the operand's buffer is a property of the well-formed tensor the caller passed, so the scaled sum stays inside the buffer this resolves against.

你的判断是对的：录制侧 `flat_offset - operand->start_offset >= operand->extent_elem_cache` 只约束了 elem_offset 落在 operand 自身 extent 内，而「extent 落在 buffer 内」那一半正是 main 上 replay 侧 `graph_tensor_wire_valid` 做的、已被删除的检查，注释原来把它写成了前移，不准确。

没有把这个包含性检查补回录制侧，是因为本运行时的 buffer 由统一分配器给出，人为构造的 tensor 不在受支持的契约内，所以这里是前置条件而非需要复验的性质。

src/common/host_build_graph/tensor_create_info.h
poursoul: 确认存在，也同意不在本 PR 处理。

数字核对过，比描述的还要差一点：`DataType` 是 1 字节，所以 `TensorCreateInfo` 在 main 上是 28 字节、`ndims` 窄化之后是 24 字节，而 `memcpy(this, &tensor_create_info, 64)` 读 64 字节 —— 越界读 40 字节。三份副本（`src/common/host_build_graph/tensormap.h`、`src/{a2a3,a5}/runtime/tensormap_and_ringbuffer/runtime/tensormap.h`）确实零调用者。

已记在待办里，后续起一个单独的 PR 把三份一并删掉。

src/common/host_build_graph/host/orchestrator.cpp
poursoul: 已加 `static_cast<uint8_t>`，把这处收窄写成显式的。

`std::copy` 的两处范围没有改动：`MAX_TENSOR_DIMS` 是 5，而 `simpler::hbg::Tensor` 的 `ndims` 由构造时的 `always_assert(in_ndims > 0 && in_ndims <= MAX_TENSOR_DIMS)` 和边界上的 `validate_tensor` 约束，所以拷贝范围本来就在界内；人为把公开字段写成越界值不在受支持的契约内。`ndims` 字段本身是 `uint32_t` 也属遗留 —— 实际取值永远不超过 5。

src/common/host_build_graph/docs/GRAPH_EXECUTION.md
poursoul: 说得对，这个论证确实不成立，已改。

precondition 本身保留，但把理由换掉了：运行时分配的 buffer 免费满足它（同一个分配器发出互不相交的块），而 boundary storage 是 caller-owned，`ChipTensor.make(..., child_memory=True)` 这种调用方自带的地址在那个保证之外，那种情况由调用方负责满足。commit message 和 PR 描述里同一处论证也一并改了。

补充一点背景：按 buffer 地址分组这个行为 main 和本分支完全相同，所以这个重叠隐患不是本 PR 引入或加重的 —— 本 PR 引入的只是那句「holds by construction」的错误论证。

src/common/host_build_graph/docs/GRAPH_EXECUTION.md
poursoul: 确认，已改。`GraphDefinitionHeader` 从来只有 `magic` / `full_key` / `definition_bytes` 三个字段（main 上也是），实现里不存在 Definition image 的 content hash —— 代码里所有 `content_hash` 命中都属于 `ChipCallable` 的 buffer 池，是另一回事。

文档里一共四处这个假声明，全部改掉了：第 18 行（改为 shell 上填的 Definition 设备地址）、第 430 行（同上）、以及你指出的这一处及其上文的 header 描述，换成实际的 framing 与 device 侧实际复验的那几项（magic、能容纳 `GraphDefinition` 的 size、`total_bytes` 与 header 相符、`full_key` 与 header 相符）。commit message 里也加了一句说明。

zhusy54 (approved):

Issue #2256, opened 2026-09-16
[Bug] hbg: sync_start body root deadlocks against a sibling root staged by the same shell


qweasdzxcht:
### Platform

a2a3sim (Ascend 910B/C simulation)

### Runtime Variant

host_build_graph

### Description

When an early-released Graph shell stages more than one body root and one of
them carries `require_sync_start`, the run can deadlock.

`stage_graph_roots_early` stages every qualifying root from a single shell
release. An ordinary sibling root therefore takes gated cores and holds them
until its own doorbell rings. The `sync_start` root then cannot fit its cohort
and enters the global stop-the-world drain, which parks every scheduler thread
on the ack tree — including the threads that would have rung the sibling and
freed those cores. The drain finds resources insufficient, advances
`drain_attempt`, and returns without clearing `sync_start_pending`, so the
threads re-enter the drain immediately and the cycle repeats.

`SIMPLER_SCHEDULER_TIMEOUT_MS` does not fire: these spins live outside the
dispatch loop's wall-clock budget, which `handle_drain_mode`'s own comment notes.

Two controls bound the scope:

- A body whose only gated-stageable root is the `sync_start` one runs normally.
- The same task set submitted at top level, with no Graph wrapper, runs normally.

So the interaction is specific to batch staging under one shell release, not to
the task shape or to `sync_start` itself.

### Steps to Reproduce

1. Build the `a2a3sim` target.
2. Submit a flagged slow producer (`allow_early_resolve`, long spin) at top level
   so the Graph shell that consumes its output qualifies as an early-dispatch
   candidate and is released while the producer still occupies its cores.
3. Give the Graph body two roots with empty fanin rows: one MIX task with
   `require_sync_start` and `block_num = rt_available_cluster_count()`, and one
   ordinary AIC task on a separate tensor.
4. Run the scene. The scene under
   `tests/st/a2a3/host_build_graph/graph_sync_start_early_dispatch/` is this
   shape minus the sibling root; adding a second ordinary root to its body
   reproduces the hang.

### Expected Behavior

The cohort either stages and rings, or fails to place and falls back to the
ordinary `ready_sync_queues` path. A body root that cannot be placed should cost
a lost pre-stage, not forward progress.

### Actual Behavior

The run hangs until killed. State read from the hung process:

```text
active_sched_threads_ = 3
drain_attempt         = 1060894166
sync_start_pending    = 8            (the cohort's block count, still set)
drain_stage_go        = 0
[...]

PR #2258, opened 2026-09-16
Add: in-graph sync_start early-dispatch scene coverage

### LINKED ISSUE #2184: hbg: in-graph sync_start early dispatch has no scene coverage
## What

A `sync_start` task inside a Graph body can now reach the early-dispatch sync-start drain queue, and no scene test exercises that path.

PR #2167 lets an early-released Graph shell stage its body's roots. A root that carries `requires_sync_start()` therefore reaches `publish_ready_to_early_sync_drain` / `try_launch_sync_start_cohort` the same way a top-level `sync_start` candidate does — `push_ready_routed` runs one body of code for both cohorts.

The only scene test on that path, `tests/st/a2a3/host_build_graph/spmd_sync_start_early_dispatch/`, submits its tasks at top level: the directory contains no `graph_begin` / `rt_submit_graph`. So the newly reachable in-graph arm is covered only by C++ unit tests of the surrounding qualification logic, never end to end on device.

## Why it is not urgent

The two cohorts share the dispatch code. They differ only in which array carries a task's completion state and where its fanin row lives, and both of those are covered by `tests/ut/cpp/common/test_hbg_graph_ed_qualification.cpp` and `test_hbg_graph_cache.cpp`. The risk is a cohort-specific interaction inside the rendezvous, not a missing branch.

## Scope

Broader than a test: worth taking together with a pass over how `sync_start` is specified for hbg, rather than bolting one more scene test onto the existing shape. A body-local rendezvous raises questions the top-level case never had to answer — what a cohort means when its members are staged by a shell release rather than by their own producers, and whether an in-graph `sync_start` root may rendezvous with a top-level one.

Suggested deliverables:

- A scene test with a `sync_start` root inside a Graph body, on both `a2a3` and `a5`.
- A statement in the docs of what a `sync_start` cohort is scoped to under hbg.

## References

- `src/{a2a3,a5}/runtime/host_build_graph/runtime/scheduler/scheduler.h` — `push_ready_routed`, `stage_graph_roots_early`, `try_launch_sync_start_cohort`
- `tests/st/a2a3/host_build_graph/spmd_sync_start_early_dispatch/`
- Raised in review of #2167

qweasdzxcht:
## Summary

- Add a scene on both arches where a `require_sync_start` root inside a Graph
  body reaches `early_sync_start_queue` and the sync-start rendezvous. #2167
  opened that arm and left it uncovered on device; #2184 asks for the coverage.
- Document what a `sync_start` cohort is scoped to — one task's blocks, never a
  set of tasks, whoever staged them.

### Why the neighbouring scenes do not already cover it

Each misses one term:

| scene | missing term |
| --- | --- |
| `spmd_sync_start_early_dispatch/` | submits everything at top level; no `graph_begin` / `rt_submit_graph` |
| `graph_execution/` (mix_spmd) | has a body `sync_start` task, but its shells consume boundary tensors and have **no producer**, so they never qualify as early-dispatch candidates and `stage_graph_roots_early` never runs |

The new scene supplies the missing term: a flagged slow seed that makes the
shell a candidate while the seed still occupies its cores.

### Scene shape

```text
seed (AIC, allow_early_resolve, slow)  ->  Graph shell  ->  body { sync_root, tail }
```

`tail` writes the tensor `sync_root` writes, so it is ordered behind the cohort:
a cohort that never rings surfaces as a scheduler timeout rather than as a
silently wrong pass.

The body carries exactly one gated-stageable root on purpose — a sibling
ordinary root under the same shell deadlocks the run, filed separately as #2256.

### Documentation

`GRAPH_EXECUTION.md` gains a "What a `sync_start` cohort is scoped to" section
answering the three questions #2184 raises: a body root and a top-level task
never rendezvous with each other; the halves of the rendezvous keep their
meaning and only change which event supplies them; and the global drain is the
only cross-task coupling, as mutual exclusion rather than as a rendezvous. Both
arches' `SUBMIT_BY_CLUSTER.md` get one line on cohort scope next to the existing
residency constraint.

## Testing

- [x] Simulation tests pass
- [x] Hardware tests pass (if applicable)

| check | a2a3 | a5 |
| --- | --- | --- |
| new scene, sim | 1 passed / 5.91 s | 1 passed / 6.38 s |
| new scene, device | 1 passed / 5.15 s | 1 passed / 6.32 s |
| `host_build_graph` sim sweep | 13 passed, 7 skipped | 14 passed |
| `host_build_graph` device sweep | 40 passed, 1 skipped | 25 passed, 1 skipped |

The a2a3 sim sweep was 12 passed / 7 skipped at #2167; the extra pass is this
scene. Every device task exited 0 and released its NPU lock.

The path is verified as actually firing rather than merely passing. Temporary
probes on `stage_graph_roots_early`, the `sync_start` branch of
`enqueue_early_dispatch_candidate`, and the cohort launch show the full chain,
with the cohort's staged and running core counts equal at the ring:

```text
stage_graph_roots_early: staging root idx=0 sync_start=1
enqueue -> early_sync_start_queue in_graph_local_id=0
RING sync_start cohort: staged_cores=24 running_cores=24
```

Without that check a `sync_start` scene can pass while silently taking the
ordinary ready path, which the output alone cannot distinguish.

## Coverage this PR does not close

An audit of `sync_start` under hbg, run alongside this change, found the
surrounding coverage much thinner than `tensormap_and_ringbuffer`'s. Recording
it so the remaining gap is explicit rather than implied.

| scene | tmr (a2a3 / a5) | hbg a2a3 | hbg a5 |
| --- | --- | --- | --- |
| `spmd_sync_start` (basic) | both | — | — |
| `spmd_sync_start_aiv` | both | — | — |
| `spmd_sync_start_early_dispatch` | both | yes | — |
| `spmd_sync_start_edge` | both | — | — |
| `spmd_sync_start_mix_spill` | both | — | — |
| `spmd_sync_start_stress` | both | — | — |
| in-graph root (this PR) | — | yes | yes |

tmr carries six scenes on each arch. Before this PR hbg had one, on a2a3 only —
even though hbg is the runtime with the extra `TaskKind::GRAPH`, with
materialization, and with `stage_graph_roots_early`.

Two reachable combinations were exercised as throwaway experiments during the
audit. Neither is covered by a committed scene and neither is addressed here:

- **An in-graph non-root `sync_start` task.** It qualifies through the publish
  list rather than through `stage_graph_roots_early`, so it is a different
  registration path from the one this PR covers. It **hangs** under a body whose
  root is staged by the shell release, which widened #2256 — the trigger there
  is not "two roots staged together" but any in-graph task holding gated cores
  alongside a `sync_start` task that needs more of the device than is left.
- **A Graph shell carrying `require_sync_start`.** `launch_spec` is a public
  member of the `Arg` base `GraphTaskArgs` inherits, so it compiles;
  `graph_submit_outer` then drops it via `slot.task_attrs = TaskAttrs{}`.
  Confirmed inert by experiment — the runtime behaviour is correct, since a
  shell places no block, and only the absence of a guard is questionable. Not
  filed: no caller has hit it and it is not a natural mistake to make. Worth
  knowing because the asymmetry looks like a defect until checked —
  `push_ready_routed` tests `requires_sync_start()` before it tests
  `task_kind == TaskKind::GRAPH`, while `enqueue_early_dispatch_candidate`
  returns early for `GRAPH`; that path is unreachable, not buggy.

Closing these is a separate piece of work — porting tmr's basic scenes to hbg
and adding a non-root scene — rather than something to fold into this PR.

Related: #2256.

## Review follow-up

CodeRabbit's two cohort-oracle findings are addressed in the latest push:

- The cohort width is now reported through a `layout` tensor from the **same**
  `rt_available_cluster_count()` call the body launches with — passed into the
  Graph as a scalar — instead of being recovered as `max() + 1` of what the
  cohort wrote. A cohort that runs short now fails rather than narrowing the
  expectation to match itself.
- `output` starts at a sentinel no kernel writes, so an unwritten cache line is
  distinguishable from a block that legitimately wrote `0.0`.

This also brings the scene onto the same pattern the top-level twin
`spmd_sync_start_early_dispatch` already uses.

The third finding — that the scene cannot assert it took the early-dispatch
path rather than the ordinary ready path — is **not** addressed here, and it is
correct. The gating in this scene is a producer spin, and nothing in the
assertions distinguishes the two routes; the path was confirmed manually with
temporary probes that are not part of the commit. A permanent assertion is
possible without touching the runtime: the scheduler already emits an
`EarlyDispatch` swimlane bar, and `tests/st/a2a3/host_build_graph/dfx/chip_swimlane/`
shows the pattern for reading those records back and asserting on a phase. It
is left out of this PR because it is a self-contained piece of DFX work, and
because such an assertion only runs under `--enable-chip-swimlane`, which
per-PR CI does not enable — so it would not guard this scene on the default
path anyway.

Fixes #2184


qweasdzxcht:
Thanks — all three are valid. Two are fixed in `9b856b4`, the third is not, and
here is why.

**Cohort-width oracle (the two width findings).** Fixed, and slightly stronger
than suggested. Rather than recording the cluster count separately, the width
now comes from a **single** `rt_available_cluster_count()` call in the
orchestration entry, which feeds both the Graph body (as a scalar, so
`set_block_num` launches with it) and a `layout` tensor the test reads as its
oracle. The launched width and the checked width are therefore the same number
rather than two independent queries. `output` also starts at a sentinel
(`-1.0`) that no kernel writes, so an unwritten cache line is distinguishable
from a block that legitimately wrote `0.0`. A cohort that runs short now fails
instead of narrowing the expectation to match itself.

This also puts the scene on the same pattern the top-level twin
`spmd_sync_start_early_dispatch` already used — it reports its width through a
`layout` tensor too. The `max() + 1` form came from `graph_execution_mix_spmd`;
the repo had both precedents and this scene had picked the weaker one.

**Deterministic / observable early-dispatch path.** Correct, and not fixed
here. The gating really is a producer spin and the assertions really cannot
distinguish the early-dispatch route from the ordinary ready route — the path
was confirmed with temporary probes on `stage_graph_roots_early`,
`enqueue_early_dispatch_candidate` and the cohort launch, which are not part of
the commit.

A permanent assertion is possible **without touching the runtime**: the
scheduler already emits an `EarlyDispatch` swimlane bar
(`scheduler_dispatch.cpp`), and
`tests/st/a2a3/host_build_graph/dfx/chip_swimlane/test_scheduler_phases.py`
shows the pattern — `read_perf_data` over the run's artifact, then assert on
`record["phase"]`.

It is left out of this PR for two reasons. It is a self-contained piece of DFX
work rather than part of adding the scene, and this PR is otherwise zero
runtime-code change plus one scene per arch. And such an assertion only runs
under `--enable-chip-swimlane`, which per-PR CI does not enable (DFX smokes are
gated behind `include_dfx_smokes`, daily only) — so it would not actually guard
this scene on the default path, which is where a silent regression to the
ordinary route would happen.

Noted in the PR description so it is not lost.

Verification after the change, unchanged from before it:

| | a2a3 | a5 |
| --- | --- | --- |
| new scene, sim | 1 passed / 6.69 s | 1 passed / 5.89 s |
| new scene, device | 1 passed / 5.15 s | (a5 device rerun pending) |
| `host_build_graph` sim sweep | 13 passed, 7 skipped | 14 passed |
| `host_build_graph` device sweep | 40 passed, 1 skipped | (pending) |

`clang-format`, `cpplint`, `ruff-check`, `ruff-format` and `markdownlint-cli2`
all pass on the changed files.

qweasdzxcht:
a5 device re-run after `9b856b4`, completing the table in the previous comment:

| | a2a3 | a5 |
| --- | --- | --- |
| new scene, sim | 1 passed / 6.69 s | 1 passed / 5.89 s |
| new scene, device | 1 passed / 5.15 s | 1 passed / 6.32 s |
| `host_build_graph` sim sweep | 13 passed, 7 skipped | 14 passed |
| `host_build_graph` device sweep | 40 passed, 1 skipped | 25 passed, 1 skipped |

Every device task exited 0 and released its NPU lock. All four sweep counts are
identical to the run before the oracle change, so reporting the width through
`layout` and starting `output` at a sentinel did not disturb any neighbouring
scene.

The a5 device runs are on 238:10135, which is the DT build — used here for
functional verification only, and it produces no timing that pairs with the PR
build.

qweasdzxcht:
## Mutation verification

CI turning green on a *newly added* test says it compiles, runs and does not
disturb its neighbours — it cannot say the test would catch the failure it
exists for, because the behaviour under test is healthy in CI. Verified that
separately by breaking the behaviour on purpose, in the spirit of #2167's
`mutation-verified` note.

Two mutations of the a2a3 scene, both on `a2a3sim`:

| mutation | expectation | result |
| --- | --- | --- |
| **M1** — launch the cohort one block short of the width reported through `layout` | should fail | **FAILED** (6.51 s) |
| **M2** — `PRODUCER_SPIN_ITERS = 0`, so the seed completes before the shell stages roots and the root takes the ordinary ready path | unknown | **PASSED** (5.95 s) |

**M1 confirms the oracle fix works.** A cohort that runs short now fails instead
of narrowing the expectation to match itself, which is what the previous
`max() + 1` form did. The queue probe in that run still shows
`ess pushes=1 maxocc=1`, so the cohort was on the early-dispatch queue even in
the mutated build — the failure is the width check, not a path change.

**M2 measures the remaining gap rather than arguing about it.** With the
producer spin removed the scene still passes, so it genuinely cannot distinguish
the early-dispatch route from the ordinary ready route. That is the finding
already flagged above and left for separate work; it is now a measured fact
rather than a stated risk.

What the scene does and does not catch, stated plainly:

| failure | caught? |
| --- | --- |
| cohort partially launched / width mismatch | yes (M1) |
| cohort staged but never rung | yes — surfaces as a scheduler timeout |
| silent regression to the ordinary ready path | **no** (M2) |

The first two are the reasons the scene exists. The third is the open item.

ChaoZheng109 (approved):

Issue #1853, opened 2026-08-17
[Feature] Let consecutive runs queue on the same stream instead of round-tripping through a host sync


ChaoWao:
### Summary

Consecutive a2a3 runs are gated by a host-side blocking sync: the executor waits for run N's streams to drain before run N+1's kernels are submitted. But a stream **is** an ordered queue — if run N+1's AICore and AICPU kernels were enqueued on the same pair while N is still executing, the driver would serialize them for free, and the host would stop paying per-run launch latency between runs.

This asks whether we can let consecutive runs queue naturally on the stream instead of round-tripping through a host sync.

### Motivation / Use Case

Today device execution is strictly serial, and the serialization is enforced on the host rather than by the queue:

- The execution claim from `try_acquire_native_run` is exclusive, and `simpler_finalize_run` releases it (`c_api_shared.cpp:965`) only **after** `drain_execution` (`:919`).
- `drain_execution` → `reap_run()` → `sync_stream_pair()` is a blocking `aclrtSynchronizeStreamWithTimeout` on both streams.
- Launch requires the claim, so run N+1 cannot submit until run N has drained *and* finalized.

Depth-two pipelining does not change this. What overlaps today is only run N+1's **preparation** — arena build, host orchestration, args H2D — against run N's device execution. That is what the two arena banks and two host `Runtime` buffers exist for (`HOST_PER_RUN`). The submission itself does not overlap, so between every pair of runs the device is idle for a host round trip: sync return → validate → finalize → claim release → next launch.

Ordering correctness would not depend on the host at all if both runs sat on the same queue, which makes the host sync a latency cost rather than a safety mechanism.

Note this is *not* what #1717 proposes. #1717 overlaps the device launch with the host-orchestration build **inside one run** (its main beneficiary is cold start), and it states that "run-to-run pipelining (prepared-successor) already hides O + launch for steady-state runs". Only preparation is pipelined; the launch is still behind the claim and the sync. This issue is the run-to-run *device* overlap that assumption implies but that does not exist yet.

### Proposed API / Behavior

Enqueue run N+1's kernel pair while run N is still in flight, and let the stream order them:

- Admit two *launched* runs, not one launched plus one prepared, in the execution-claim contract.
- Keep the AICPU and AICore streams distinct. The AICPU Run kernel spins in the handshake waiting for the AICore workers, so collapsing them onto one queue would leave the AICore submission behind a spin that can never end.
[...]

PR #2262, opened 2026-09-16
Add: publish RunWall's raw device-clock bounds on the device_wall marker

ChaoWao:
## Summary

A device span's `ts` is rebased on its own invocation's origin, so it orders spans *within* one run and cannot express the interval *between* two runs. The AICPU already stamps every phase from `get_sys_cnt_aicpu()` — the free-running `CNTVCT_EL0`, never reset per run — but the host subtracted that origin away before emitting: RunWall's start was dropped entirely and the sub-phases kept only relative offsets. The device-side interval between one run's end and the next run's start was therefore not derivable from the light timing channel at all.

- Keep RunWall's `min_start` / `max_end` as raw sys-counter ticks in `read_device_wall_ns()` and expose them through two accessors.
- Carry them on the `device_wall` marker as `dev_start_cycle` / `dev_end_cycle`, plus `dev_cnt_hz` so a consumer need not hardcode the per-arch tick rate (50 MHz a2a3, 1 GHz a5) and `dev_id` so ticks from different devices are never differenced. New `STRACE_DEV_SPAN_AT_A` for the attribute-carrying form.
- Document the interval and its limits in `docs/dfx/host-trace.md`.

The gap is `(dev_start_cycle(N+1) - dev_end_cycle(N)) / dev_cnt_hz`, **differenced as integers before converting** — converting each bound to ns first rounds both ends and can consume a sub-microsecond gap.

### Why `ts` stays 0

Deliberate, not an omission: `ts` is the origin the depth-3 sub-phases are positioned against, so moving this bracket to an absolute instant would invert their containment and break `--trace-out`'s nesting. The absolute bounds ride as attributes, which an existing parser ignores.

### Limits stated in the doc

- Comparability holds within **one device and one counter epoch**. It says nothing across a device reset, and **nothing detects such a reset today**, so a consumer must not treat these ticks as valid across one.
- These ticks are never comparable to the host clock.
- The metric is the **AICPU RunWall interval**. It does not prove the AICore kernel's exit instant, so it is not a full two-stream execution boundary.
- The onboard platform emits the attributes; the sim platform derives `device_wall` from `steady_clock` and continues to emit `ts`/`dur` only.

## Testing

Measured on a2a3 through `task-submit`. **Positive control:** the emitted bounds reproduce the span exactly — `(4055254392754 - 4055254360695)` ticks at 50 MHz is 641,180 ns, matching that same marker's `dur=641180`; a second invocation likewise gives 230,200 ns against `dur=230200`. No negative gaps across any run.

- [x] `tests/st/a2a3/host_build_graph` — 39 passed, 1 skipped
- [x] `tests/st/a2a3/tensormap_and_ringbuffer` — 46 passed
- [x] cpput (clean reconfigure + build) — 147/147
- [x] pyut — 2387 passed, 18 skipped
- [x] `examples/workers/l2/vector_add/test_run_timing.py` (marker contract, needs `--manual include`) — passed
- [x] `clang-format` and `markdownlint-cli2` clean

One sweep before the final one hit a `SCHEDULER_TIMEOUT` (`sched_error_code=100`) forward-progress flake in `worker_async_endpoint` and `run_stream_reuse`; both passed on re-run and the full sweep is green. That failure surfaces inside `sync_stream_pair`, which is strictly earlier in a run than anything this change adds — every line here runs after drain — and the box was holding 9/16 devices for other users with several all-device jobs queued.

Relates to #1853 (measuring the device-side inter-run gap is the first step there; this change is what makes that interval derivable).


ChaoWao:
CI triage, since three self-hosted lanes were red on the first run.

**Two of the three cleared on re-run** and are green now: `st-onboard-a2a3` (collectives / L3 dispatch / L3 dependency / `worker_async_fifo`) and `st-onboard-a5` (`sdma_async_completion_demo`). The a5 one failed identically on #2251, and #1489 already tracks the intermittent allreduce L3 family.

**`st-network1-onboard-a2a3` persists, and is not this change.** Filed as #2264 with the logs. The mechanism is remote-node device initialization:

```text
chip worker 0 (pid 336220) failed during init:
  chip_process dev=14 init: RuntimeError: simpler_init failed with code 507033
```

The local side came up fine (`[chip_process pid=3406204 dev=12] ready`); what failed is the *remote* machine's chip worker on `dev=14`, before any test workload ran. The failing test set also rotates between runs, and `global_tload_mpirun_l3` fails identically on #2252 — an unrelated change. #2263 passed the same lane, so it is a rate, not a constant.

Why this diff is not upstream of any of it: every line it adds runs in `read_device_wall_ns()` and `emit_device_phase_markers()`, both **after** drain, and it touches no `simpler_init`, L4, or network path. `profiling-flags-smoke`, `st-deepseek-onboard-a2a3`, `ut-a2a3`, `ut-a5` and both `st-sim` lanes pass — those exercise the emission path and would fall first if it were broken.

Local verification on a2a3 before pushing: `tests/st/a2a3/host_build_graph` 39 passed / 1 skipped, `tests/st/a2a3/tensormap_and_ringbuffer` 46 passed, cpput 147/147 from a clean reconfigure, pyut 2387 passed / 18 skipped.

**One substantive change since the first push**, found while re-reading my own wording rather than from review: `get_sys_cnt_aicpu()` does not read `CNTVCT_EL0` raw — it returns `cntvct * PLATFORM_PROF_SYS_CNT_FREQ / cntfrq_el0`, i.e. rescaled into the PROF unit. The arithmetic in the change was already right (`dev_cnt_hz` is that unit), but the doc said "reads the free-running `CNTVCT_EL0`", and a reader taking that literally would divide by `cntfrq_el0` and be wrong wherever the two disagree. The doc, the header comment and the commit body now state the rescaling and point at `sys_cnt_elapsed_ns()`.

Still pending: `ut (macos)` and `st-sim-a5 (macos)` are queued for macOS runners. Not green-and-done until those land.

PR #2265, opened 2026-09-16
CI: source CANN environment in A5 jobs

doraemonmj:
A5 CI can fail SDMA operator initialization when the runner environment lacks CANN operator paths. Source `/usr/local/Ascend/cann/set_env.sh` during setup, compilation, hardware tests, and DFX smokes, matching A2A3.

Validation: YAML parsing and Bash syntax checks passed for all 9 run steps in the 3 changed files; `git diff --check` passed. Hardware validation is pending CI.


.github/actions/run-onboard-dfx-smokes/action.yml
```diff
@@ -18,10 +18,9 @@ runs:
         set -euo pipefail
 
         case "$DFX_PLATFORM" in
-          a2a3)
+          a2a3|a5)
```
ChaoZheng109, line 26: **This breaks `ut` on both OSes.** `tests/ut/py/test_run_onboard_dfx_smokes_action.py` extracts this `run:` block verbatim and executes it under `bash -c` on a plain GitHub-hosted runner, with `DFX_PLATFORM=a5`. The explicit `a5) ;;` arm was what made that possible — `a2a3` has never been hermetically covered precisely because it sources a file that does not exist off-silicon.

All four parametrisations now fail:

```text
E   AssertionError: bash: line 5: /usr/local/Ascend/cann/set_env.sh: No such file or directory
E   assert 1 == 0
tests/ut/py/test_run_onboard_dfx_smokes_action.py:126: AssertionError
```

(`test_a5_dfx_smokes_adapt_to_device_count_without_overlap[1..4]`, run 35084704218.)

Worth noting that this is also the one hunk in the PR with no observed benefit: the a5 DFX smokes have never needed CANN — `st-onboard-a5` is green on `main` without it, and the SDMA demo passes either way (`TestSdmaAsyncCompletionDemo [PASS 11.0s]` on #2259 vs `[PASS 10.8s]` here).

Two ways out:

1. **(preferred) Parameterise the script path, reusing the precedent already in `setup-venv`.** That action carries a `cann-env` input plus `env: CANN_ENV: ${{ inputs.cann-env }}` for the network1 runners. This action already has an `env:` block, so adding the same input and using `source "$CANN_ENV"` keeps the `run:` body pure bash (no `${{ }}` for the test to choke on) — the test then just points `CANN_ENV` at a stub file. The test update belongs in the same commit.
2. **(smaller) Keep `a5) ;;`.** The other two files still deliver the stated goal on their own.

Please don't reach for `[ -f ... ] && source ...`: that would silently degrade on a runner that *should* have CANN but doesn't, which is the exact failure class this PR is trying to remove.

ChaoZheng109 (commented):
Folding `a5` into the `a2a3` case arm here breaks the only hermetic coverage this action has: `ut (ubuntu-latest, 3.10)` and `ut (macos-latest, 3.10)` are both red on this PR, and both are green on #2259. See the inline comment for the failure and two ways out.

The other two files (`_st-npu-a5.yml`, `_ut-npu-a5.yml`) look right to me — in particular they close a real hole, since the three `cann-examples` steps in `_ut-npu-a5.yml` build `${ASCEND_HOME_PATH}/tools/hcc/bin/...` and today only work because the a5 runners happen to carry CANN in their ambient environment.

.github/actions/run-onboard-dfx-smokes/action.yml
doraemonmj: Fixed — parameterized the DFX CANN setup path through a `cann-env` input, kept the real hardware path as the default, and updated the hermetic scheduler test to inject a temporary stub script. The action still fails fast when its configured CANN script is missing.

ChaoZheng109 (approved):

Issue #1455, opened 2026-07-24
[Bug] sync_start drain can deadlock when generation-less ack/election state is reused across retry attempts


Leaf-Salix:
### Platform

a2a3 (Ascend 910B/C hardware)

### Runtime Variant

tensormap_and_ringbuffer

### Description

`sync_start` drain appears to use reusable shared coordination fields without an attempt/generation tag:

```text
sync_start_pending
drain_ack_mask
drain_worker_elected
drain_stage_go
drain_stage_done_mask
```

When the elected drain owner observes insufficient resources, it resets `drain_ack_mask` and `drain_worker_elected` so scheduler threads can resume completion polling and retry the same pending `sync_start` task.

However, a scheduler thread from the old attempt can already have passed the ack barrier and then be delayed before election. After the owner resets the shared fields, that stale thread can continue and interpret the reset `drain_worker_elected == 0` / `drain_ack_mask` values as if they belonged to its own attempt.

This is a generation-less reusable-barrier ABA issue: every individual shared field value is valid, but the fields can come from different drain attempts.

I used a deterministic test-only harness to force this interleaving. The harness does not directly return a synthetic error code. It lets the stale thread continue into the original `handle_drain_mode()` election/stage protocol. The original protocol then enters a split barrier and eventually fails with an AICPU timeout.

One detail in the harness is intentionally deterministic: after proving that the stale thread crossed from attempt 1 to attempt 2, it lets the stale owner pass the availability check. This models the real timing where resources become available before the stale owner performs that check. The hook does not replace the ack/election/stage protocol; the deadlock still comes from the original protocol accepting a stale owner and publishing `stage_go` with a mixed-generation `ack_mask`.

### Steps to Reproduce

```markdown
1. Apply the env-gated runtime hook patch from Additional Context. The patch adds `SIMPLER_DRAIN_ABA_TEST`, tags drain attempts in test mode, and pauses one scheduler thread after the old attempt's ack barrier but before election.

2. Add the ST from Additional Context. The orchestration submits one holder MIX task and then one `sync_start` MIX task:


static constexpr int16_t HOLDER_BLOCKS = 1;
static constexpr int16_t SYNC_BLOCKS = 24;
static constexpr int64_t HOLDER_SPIN_ITERS = 50000000;

submit_mix(holder, scratch, HOLDER_BLOCKS, 0, HOLDER_SPIN_ITERS, false);
submit_mix(output, scratch, SYNC_BLOCKS, 0, 0, true);


[...]

PR #2266, opened 2026-09-16
Update: reduce A5 HBG scheduler hot-path traffic

zhusy54:
> #2259 is merged. This branch is rebased onto the current `main` and
> contains only the scheduler hot-path follow-up in one commit.

## Problem

After removing the largest redundant completion and dependency handoffs, the
A5 `host_build_graph` Scheduler still performed avoidable shared-GM traffic
and retained compatibility state that is not part of the production model:

- locally filled slots published ownership state to GM and then reread it;
- helpers retained `nullptr` paths that reconstructed Scheduler state from GM;
- `SchedulerDispatchSlot` carried fields written by the Scheduler but never
  consumed by the Executor;
- Ready-owner maintenance scanned inactive queues, while invariant worker
  trace fields and fallback timing metadata were reread repeatedly;
- Completion processing reread a worker's two-slot inbox during catch-up and
  cleared each producer-owned token after consumption;
- the dispatch-slot alignment left an unused fourth 64-byte cache line after
  the Executor trace had already been compressed to one line.

These costs are serialized on short dependency chains.

### Scheduler-local ownership state

- Use `SchedulerLocalState` as the sole source of Scheduler-owned slot state;
  remove the legacy `nullptr`/GM fallback paths and the old cluster-local
  compatibility terminology.
- Keep shared `READY` publication as the Scheduler-to-Executor handoff, while
  local `FILLING` and `FREE` transitions remain private to the Scheduler.
- Track active Ready-owner queues by core type so inactive queues are skipped.
- Directly refill compatible single-fanin consumers without redundant Ready
  state and wait-metadata traffic.
- Cache invariant worker entry, handshake, and release timing fields after
  their first observation.

### Dispatch and completion contract

- Retain only Executor-consumed dispatch fields: task/timing identity,
  generation, pending-slot identity, publication, and Executor trace data.
- Remove write-only kernel, subtask, fan-in, block, cohort, and gang metadata.
- Use the generation-tagged Completion Inbox token as the release/acquire
  boundary for the one-cache-line Executor trace.
- Treat Completion Inbox tokens as producer-owned publication. The single
  Scheduler consumer records each slot's last consumed generation locally;
  it does not clear the token in GM.
- Load both Completion Inbox slots in one packed 64-bit access per active
  worker and Scheduler pass. Both slots are decoded from that snapshot; a
  later publication is consumed on the next pass instead of by a catch-up
  reread.
- Share the same consumed-generation state between normal completion and the
[...]

PR #2284, opened 2026-09-17
Fix: run graph control work before the sync_start drain check

### LINKED ISSUE #2256: [Bug] hbg: sync_start body root deadlocks against a sibling root staged by the same shell

qweasdzxcht:
## Summary

- Move the scheduler loop's graph control pass ahead of the `sync_start` drain
  check, so a cohort waiting on cores held by gated in-graph tasks can be
  satisfied. Fixes the hang in #2256.
- Add a scene that puts a `sync_start` root and an ordinary root in one Graph
  body, which is the shape that triggers it.

### The cycle

`stage_graph_roots_early` stages every qualifying body root from a single shell
release. An ordinary root among them takes gated cores and holds them until its
own doorbell rings. A `require_sync_start` root that needs the rest of the
device then cannot fit, so it enters the global drain — and the drain waits for
cores that only `activate_graph_task -> graph_route_ready_roots ->
push_ready_routed` can release. That call sits **after** the drain check in
`resolve_and_dispatch`, behind a `continue`, so the loop never reaches it while
a drain is pending.

State read from a hung run (`a2a3sim`):

```text
drain_attempt         = 1060894166     (still advancing)
sync_start_pending    = 8              (the cohort's block count, still set)
drain_stage_go        = 0
drain_ack_tokens_     = [N-1, N, N]    (token is 2^63 | attempt)
```

It is a livelock rather than a stalled barrier: the drain keeps retrying the one
thing it is itself preventing. `SIMPLER_SCHEDULER_TIMEOUT_MS` does not fire
because those spins live outside the dispatch loop's wall-clock budget, which
`handle_drain_mode`'s own comment notes.

### The change

The graph control block moves ahead of the drain check, **unchanged** — no
statement added or removed (verified by comparing the added and removed lines as
sets). Its own comment already carried the argument: *"Graph control work never
consumes an AICore."* A drain that exists to free cores gains nothing by
skipping a pass that takes none, and that pass is the only route by which the
cohort's cores can come free.

Both registration paths reach the same deadlock, and both are fixed:

- a body root staged by `stage_graph_roots_early`
- an in-graph **non-root** that qualifies through the publish list
  (`graph_incremental_publish -> register_on_ed_publish_list`)

## Testing

- [x] Simulation tests pass
- [x] Hardware tests pass (if applicable)

**Before / after on the same baseline** (`31f348e`), same scene, same host:

| runtime | result |
| --- | --- |
| without the fix | **hangs**, killed at 300 s |
| with the fix | **1 passed in 2.92 s** |

Each arm re-checks the built source for the fix marker before running, so the
two arms cannot silently build the same binary.

| check | result |
| --- | --- |
| new scene (`a2a3sim`) | 1 passed |
| in-graph non-root `sync_start` variant | 1 passed (hangs without the fix) |
| `host_build_graph` sweep, `a2a3sim` | 15 passed, 7 skipped |
| `host_build_graph` sweep, `a5sim` | 14 passed |
| `host_build_graph` sweep, a2a3 device | 41 passed, 1 skipped |
| repeat runs of both repro scenes | 40 iterations, 0 failures, 0 hangs |
| pre-commit (clang-format, cpplint, ruff, …) | all pass |

The sweep counts include the throwaway repro scenes that were present in that
worktree; the committed addition is one scene.

## Review note on concurrency

The graph control pass now runs before the drain's ack barrier, and it can call
`push_ready_routed`, which may ring a staged cohort. Two properties say that is
safe, and both are in the code rather than inferred:

- `ring_staged_doorbell_bits` writes doorbell registers only. It does not touch
  the `CoreTracker`, so the coordinator's availability count is unaffected by
  whether anyone rang. The cores were already marked occupied at gated-staging
  time.
- `try_launch_sync_start_cohort` claims `early_dispatch_launch_state` with a CAS
  before ringing, so the ring happens exactly once regardless of who gets there.

Ringing also moves in the direction the drain wants: a rung cohort runs, FINs,
and releases its cores.

**That is a code argument, not a proof.** Evidence for "no new race" is the
reasoning above plus 40 repeat runs — weaker than the rest of the table, and
this protocol has had a subtle ABA bug before (#1455). Reviewers who know the
drain protocol should look hardest at the ack barrier.

Fixes #2256


qweasdzxcht:
## CI triage

Three jobs are red. None of them is caused by this change, but the two classes
differ in how firmly that can be shown, so both are spelled out.

### `ut-a5` and `st-onboard-a5` — pre-existing, confirmed on other branches

Both fail in **`Set up environment`**, before any test runs. The same two jobs
fail at the same step on unrelated branches in the same window:

| branch | failing job | failing step |
| --- | --- | --- |
| `rdma-backend-pr` | `ut-a5`, `st-onboard-a5` | `Set up environment` |
| `fix/a2a3sim-tensor-wait-budget` | `ut-a5` | `Set up environment` |
| this PR | `ut-a5`, `st-onboard-a5` | `Set up environment` |

Three unrelated changes cannot share a cause through the code, and the step that
fails runs before the runtime this PR touches is ever built or loaded. There is
also a branch named `support/source-cann-in-a5-ci` in flight, which suggests the
a5 runner environment is already being worked on.

### `st-onboard-a2a3` — not caused by this change, by exclusion

This one fails in `Run pytest scene tests (a2a3)`, a different step from the
a2a3 failure on `fix/a2a3sim-tensor-wait-budget` (`SDMA pytest`), so it is not
the same pre-existing failure and is not claimed as one.

What the log shows:

```text
149 deselected, 1 warning in 430.76s
```

The pytest body selected **zero** tests. Every failure comes from the parallel
`standalone` group:

```text
runtime_fatal_codes ...[heap_ring_deadlock]          rt=tensormap_and_ringbuffer  rc=1  5.0s
runtime_fatal_codes ...[tensor_wait_timeout]         rt=tensormap_and_ringbuffer  rc=1  6.0s
task_timing ...test_hbg_distinct_slots_emit_markers  rt=host_build_graph          rc=1  8.1s
task_timing ...test_spmd_task_aggregates_across_threads  rt=tensormap_and_ringbuffer  rc=1  7.9s
vis_isolation ...test_comm_domain_under_visible_devices  rt=tensormap_and_ringbuffer  rc=1  7.7s
```

Four of the five are `tensormap_and_ringbuffer`, which this PR does not touch at
all — the change is confined to `host_build_graph`'s scheduler, a separate
runtime tree. The one `host_build_graph` case,
`test_hbg_distinct_slots_emit_markers`, asserts on task-timing markers and does
not involve `sync_start`, the drain, or graph activation. All five fail in 5–8 s
with `rc=1`, which is a fast failure rather than the hang or timeout a
scheduler-ordering regression would produce.

The scene added by this PR lives in the pytest body, which selected nothing, so
it did not run in this job at all.

### What is not yet ruled out

The a2a3 failure has no direct same-step comparison on another branch, so the
argument above is exclusion rather than a reproduction on `main`. A re-run of
that job would settle it: if it goes green, the group is flaky; if the same five
cases fail again, it deserves its own issue. I do not have permission to trigger
a re-run.

### Verification this change did receive

Run locally on an a2a3 host, on this PR's baseline (`31f348e`):

| check | result |
| --- | --- |
| before/after on the repro scene | hangs unfixed (killed at 300 s) → **1 passed / 2.92 s** fixed |
| in-graph non-root `sync_start` variant | hangs unfixed → **1 passed / 3.92 s** fixed |
| `host_build_graph` sweep, `a2a3sim` | 15 passed, 7 skipped |
| `host_build_graph` sweep, `a5sim` | 14 passed |
| `host_build_graph` sweep, a2a3 device | 41 passed, 1 skipped |
| repeat runs of both repro scenes | 40 iterations, 0 failures, 0 hangs |

Each arm of the before/after re-checks the built source for the fix marker
before running, so the two arms cannot silently build the same binary.

qweasdzxcht:
## Rebased onto `44dccbef`, and a correction to the triage above

The `st-onboard-a2a3` failure has a more specific cause than the exclusion
argument I gave. This branch was based on `31f348eb`, one commit behind main,
and the commit it was missing is **#2283, "preserve each run's error result in
storage that run owns"** — which touches exactly what was failing:

```text
tests/st/runtime_fatal_codes/test_runtime_fatal_codes.py      <- the failing test file itself
src/common/platform/include/common/device_run_result.h        <- new
src/common/platform/onboard/host/device_runner_base.cpp
src/common/platform/onboard/host/c_api_shared.cpp
src/common/platform/include/common/host_api.h
```

A `pull_request` build runs the merge of the branch with the base, so CI took
**#2283's updated test file against this branch's older runtime**, which has
none of the `device_run_result` plumbing those tests now expect. The two
`runtime_fatal_codes` cases were the first two failures in that group, and
`task_timing` / `vis_isolation` reach the same `host_api.h` surface.

So the previous comment's conclusion (not caused by this change) stands, but its
stated mechanism was incomplete: it was a stale baseline, not just an absence of
overlap. Rebased onto `44dccbef` and re-verified there before pushing.

Re-verification on the new baseline, same protocol as before — each arm
re-checks the built source for the fix marker, so the two arms cannot silently
build the same binary:

| check | result |
| --- | --- |
| repro scene, runtime **without** the fix | hangs, killed at 300 s |
| repro scene, runtime **with** the fix | **1 passed / 2.99 s** |
| `host_build_graph` sweep, `a2a3sim` | 14 passed, 7 skipped |

`ut-a5` and `st-onboard-a5` are expected to stay red: they fail in
`Set up environment` before any test runs, and the same two jobs fail at the
same step on unrelated branches in this window.

qweasdzxcht:
## The rebase did not fix `st-onboard-a2a3`, and the baseline theory was wrong

Rebasing onto `44dccbef` (which includes #2283) did **not** turn that job green,
so the stale-baseline explanation in my previous comment is **withdrawn**. It was
plausible — #2283 touches `test_runtime_fatal_codes.py` and the error-result
plumbing — but it is not what is happening.

What the rebase did change: `ut-a5` now passes. That one really was the a5
runner environment, and it has been fixed in the meantime.

### `st-onboard-a2a3` is flaky on the runner, not branch-specific

Recent runs of this job across branches:

| branch | `st-onboard-a2a3` |
| --- | --- |
| `rdma-backend-pr` | failure |
| `rdma-backend-pr` | **success** |
| `hbg_swim` | failure |
| `fix/cacheline-alignment` | failure |
| `run-result-per-run` | **success** |
| this PR | failure (twice) |

The same branch (`rdma-backend-pr`) both passes and fails, which is the
definition of a flake rather than a defect in any one change.

The failure set also grew between my two runs — from 5 cases to ~15 — and the
new ones are all L2/L3 worker-level:

```text
examples/workers/l2/per_task_runtime_env         rc=1  devices=[3]
examples/workers/l3/device_copy_offset           rc=1  devices=[3]
examples/workers/l3/domain_rank_map              rc=1  devices=[3, 0, 4]
examples/workers/l3/dual_domain_overlap          rc=1  devices=[5, 3, 0]
examples/workers/l3/ffn_tp_parallel              rc=1  devices=[3, 0]
examples/workers/l3/multi_chip_dispatch          rc=1  devices=[3, 0]
tests/st/vis_isolation                           rc=1  devices=[3, 4]
tests/st/worker/comm_region/...                  rc=1  devices=[3]
```

Two things stand out. **Device 3 appears in nearly every failure**, and all of
them fail in 2–8 s. And the areas — multi-card communication, domain rank
mapping, device visibility, worker lifecycle — have no relationship to a
reordering of two blocks inside `host_build_graph`'s single-card scheduler loop.

`149 deselected` again: the pytest body selected nothing, so the scene this PR
adds did not run in that job either time.

### What I am not claiming

I cannot re-run the job to confirm the flake directly (no permission), and I have
not proven the device-3 correlation is causal — it is a pattern across the
failure list, not a diagnosis. If a maintainer can re-run `st-onboard-a2a3`,
that would settle whether this PR is affected at all. If it keeps failing with
the same device-3 set across branches, the runner is worth looking at
separately.

### This change's own verification, on the rebased baseline `44dccbef`

| check | result |
| --- | --- |
| repro scene, runtime **without** the fix | hangs, killed at 300 s |
| repro scene, runtime **with** the fix | **1 passed / 2.99 s** |
| `host_build_graph` sweep, `a2a3sim` | 14 passed, 7 skipped |

Each arm re-checks the built source for the fix marker before running.

qweasdzxcht:
## When `st-onboard-a2a3` started failing, and what landed around then

Tracing the job across the last 24 CI runs (times UTC) shows a clear turning
point, and this PR is not at it:

```text
09-17T01:59  success   refactor/own-the-aicore-register-tables-per-device
09-17T02:16  success   refactor/issue-2267-run-owned-completion-fence
09-17T02:21  success   feat/capture-carries-level-identity
09-17T03:17  success   rdma-backend-pr
--------------------- failures become widespread ---------------------
09-17T04:36  failure   support/source-cann-in-a5-ci      <- earliest failure seen
09-17T06:53  failure   rdma-backend-pr
09-17T07:16  failure   fix/a2a3sim-tensor-wait-budget
09-17T07:30  failure   run-result-per-run
09-17T08:08  success   rdma-backend-pr
09-17T08:09  failure   this PR                            <- first run of this PR
09-17T08:21  success   run-result-per-run
09-17T08:45  failure   fix/cacheline-alignment
09-17T09:11  failure   rdma-backend-pr
09-17T09:33  failure   this PR (rebased)
09-17T09:42  failure   hbg_swim
09-17T09:43  success   rdma-backend-pr
09-17T10:21  failure   rdma-backend-pr
```

Three properties of that window:

- **This PR is not the origin.** The earliest failure is at 04:36 on
  `support/source-cann-in-a5-ci`, about four hours before this PR first ran CI.
- **At least seven unrelated branches are affected.**
- **It is intermittent, not deterministic.** `rdma-backend-pr` alternates
  success/failure across four runs of the same branch, which rules out a commit
  that simply breaks the job.

Commits that merged to main inside the 02:00–04:36 UTC window:

```text
#2263  Add: namespace nested worker diagnostics by topology
#2258  Add: in-graph sync_start early-dispatch scene coverage   (tests + docs only)
#2272  Add: give each onboard run its own dual-stream completion boundary
#2275  Refactor: require the arena compatibility probe only when the bank is shared
#2271  Fix: retire the AICPU loader when a program close's reset also fails
#2277  Refactor: own the AICore register tables per device context
```

Several of those touch onboard device lifecycle and per-run state isolation,
which is the same area the failures cluster in — worker lifecycle, multi-card
communication, device visibility — and nearly every failing case carries
`devices=[3, ...]`.

**This is a correlation, not a diagnosis.** The intermittency argues against any
single commit breaking the job outright; it is equally consistent with a runner
device left in a bad state, or with one of those changes making a pre-existing
race easier to hit. Confirming either way needs repeated runs on `main`, which I
cannot trigger.

Posting it because the window is narrow enough to be worth looking at, and
because it makes the "not caused by this PR" claim checkable rather than just
asserted.

qweasdzxcht:
Follow-up with harder evidence than the timeline above. The timeline argued from
ordering; this argues from a control group and from the failure sets themselves.

## A branch that carries none of this work fails the same way

`rdma-backend-pr` is branched from before #2258 landed, so it contains neither
the merged scene coverage nor the scheduler change in this PR. Cross-tabulating
the `st-onboard-a2a3` job result against whether the tested commit contains
#2258:

| | job passed | job failed |
|---|---|---|
| **contains #2258** | 1 | 8 |
| **does not contain #2258** | 9 | 3 |

Both off-diagonal cells are populated, so #2258 is neither necessary nor
sufficient for the failure. The apparent association is time confounding:
#2258 merged at 03:15Z, the job went red around 04:36Z, and after that almost
every freshly-based branch contains it by construction.

`rdma-backend-pr` is the clean control — a fixed baseline without #2258 that
spans the turning point. It is green at 01:27Z–03:17Z, then alternates
failure/success from 06:53Z onward (fail 06:53, pass 08:08, fail 09:11, pass
09:43, fail 10:21). Its behaviour changes at the same moment as everyone
else's, without ever containing the change.

## The failure sets shuffle, and never touch the changed path

Comparing the genuinely-FAILED tests in this PR's run `8d1d1871` against
`rdma-backend-pr`'s run `ea8fd954`:

- this PR: 12 failed; `rdma-backend-pr`: 10 failed
- the sets overlap but differ in both directions — each run has failures the
  other does not

A failure set that reshuffles between runs is the signature of a flaky
environment, not of a deterministic defect.

Every failure on both sides sits in `examples/workers/l2|l3|l4` (worker
topology, multi-chip dispatch, comm streams), `tests/st/task_timing`,
`tests/st/vis_isolation` or `tests/st/runtime_fatal_codes`. Zero failures in
`host_build_graph`, and zero in anything named `sync_start`.

## The changed path is green

On this PR's own failing run, the a2a3 `host_build_graph` suite is **19 passed,
6 skipped, 0 failed**, including both scenes this work added:

```text
graph_sync_start_early_dispatch ... PASSED
graph_sync_start_sibling_root   ... PASSED
```

`graph_sync_start_sibling_root` is the regression test for #2256 and it passes
on device, which is the point of this PR.

## What this does not establish

None of the above identifies the actual cause. It rules this work out; it does
not rule anything else in. Confirming a cause needs repeated runs on `main`.

qweasdzxcht:
**Correction to my previous comment.** Its control-group argument is wrong.
`st-onboard-a2a3` is called without a `ref`, so on `pull_request` it checks out
`refs/pull/N/merge` — the PR merged into `main` *at trigger time*, not the
branch head. The `rdma-backend-pr` run at 10:21Z shows
`HEAD is now at ... Merge ... into 44dccbef`, and
`graph_sync_start_early_dispatch` ran (and passed) in it. So every run after
03:15Z contained #2258, and the 2×2 table I posted is meaningless. Please
disregard it.

The conclusion holds, on different evidence:

**1. Every failure is pinned to one physical device.** In both this PR's run
and `rdma-backend-pr`'s 10:21Z run, every `[FAIL ...]` entry lists device 3
among its devices (28/28 and 27/27; the single exception is an L4 two-host
case), and zero entries that include device 3 pass. The failures are
`chip_process dev=3 init: simpler_init failed with code 507018` and
`rtStreamCreate` timeouts — the device fails at initialisation, before any test
logic runs. A code defect fails on whatever device a test lands on; this fails
on one fixed device.

**2. The runners share that device.** `infra021-npu-1..4` all failed the same
way, i.e. four runner instances on one host sharing its physical device 3.

**3. Device 3 was already failing before this work's scenes ran.** In both
runs the first device-3 failure precedes `graph_sync_start_*` by about three
and a half minutes (09:44:11 vs 09:47:39; 10:30:29 vs 10:34:07).

**4. It recovered with #2258 still in place.** From 09-17 11:17Z on, the job is
green on all four runners (12 of 13 finished runs through 09-18 01:00Z), every
one of them testing a `main` that contains #2258. The one exception,
`hbg_swim`, fails its own `concurrent_prepare_stress` /
`native_run_lifecycle` cases, with no device-3 or 507018 signature.

#2284's own scheduler change was never in anyone else's merge ref, since it is
unmerged, so it cannot account for other PRs' failures.

This PR's `st-onboard-a2a3` failure was device 3 on `infra021`, which appears
to have recovered. A re-run should confirm it.

qweasdzxcht:
Rebased onto `main` @ `b776cd27` (which includes #2266's A5 scheduler header
rework); no conflicts, and the four changed files are byte-identical to before.
All 20 checks now pass, including the three onboard jobs that were red earlier
for infrastructure reasons.

This is the first CI run in which the A5 half of the fix is actually compiled
and run on device; the previous `st-onboard-a5` attempt failed in
`Set up environment` before building anything.

- `st-onboard-a2a3`: `host_build_graph` 45 passed / 1 skipped, including
  `graph_sync_start_early_dispatch` and `graph_sync_start_sibling_root` (the
  #2256 regression test)
- `st-onboard-a5`: `host_build_graph` 27 passed / 1 skipped, including
  `graph_sync_start_early_dispatch`

qweasdzxcht:
## Approach replaced: the fix now moves the release, not the block

This PR previously moved the scheduler loop's graph control block ahead of the
Phase 2 drain check. That is force-pushed away; the old head `03ed8d25` stays
reachable from the force-push entry in the timeline above.

The reorder cleared the hang and its added drain latency was bounded and small.
What it left behind is the invariant *"every path that can free a core another
task waits on must sit before the drain check"*, held by nothing but the order of
two blocks plus a comment — no compile-time signal, no test signal. That
equivalence has already been broken once without anyone noticing: the check's
position is inherited from a loop where the code after it was only dispatch, so
"skip the rest" was exactly what `SyncStartDrainState` promises, and #1444
inserted a control pass into that gap while deleting the sentence that had stated
it.

Underneath sits an asymmetry, which is what this version removes: a shell's body
roots are **staged** through the early-dispatch publish chain like any other
candidate, but were **released** through a scheduler-loop queue, while every other
early-dispatch release happens inline on the completion path — on the resolution
thread, which never takes part in a drain. `push_ready_routed`'s
`TaskKind::GRAPH` branch now calls `activate_graph_task` inline, so the scheduler
loop stops being a release path at all and the drain's critical path gains
nothing.

Thanks to @ChaoZheng109, whose analysis on #2256 is what this is built on.

The reasoning that ruled out the reorder is recorded in
`docs/investigations/2026-09-hbg-graph-release-vs-drain-reorder.md` so the next
person meeting this deadlock does not re-derive it.

### What carried over unchanged

The scene from the previous revision — a `sync_start` root and an ordinary root
in one body — is the same test, with only the prose describing the mechanism
updated.

### Re-verified on today's `main`

The branch is rebased from `b776cd27` onto `7ad20a66`; upstream had touched three
of the files this changes, so the earlier run no longer described what is pushed.
Everything below was re-run on the new base, and each arm re-checks the built
source so the two arms cannot share a binary.

| check | result |
| --- | --- |
| new scene (`a2a3sim`) | 1 passed in 3.0 s |
| same scene, `src/` reverted to the base, rebuilt | hangs, killed at 400 s (a 180 s `--pto-session-timeout` does not cut it short) |
| `tests/st/a2a3/host_build_graph` sweep (`a2a3sim`) | 14 passed, 7 skipped |
| `tests/st/a5/host_build_graph` sweep (`a5sim`) | 14 passed, 13 deselected |
| pre-commit (headers, english-only, clang-format, cpplint, ruff, whitespace) | pass |

`markdownlint-cli2` could not be installed in the environment used here (its npm
hook is blocked), so the two doc files were checked with the other hooks only.

ChaoZheng109 (approved):

Issue #2286, opened 2026-09-17
[Performance] A5 AICore retirement: porting more of the a2a3 protocol yields no net gain


qweasdzxcht:
### Platform

a5 (Ascend 950 hardware)

### Runtime Variant

tensormap_and_ringbuffer

### Summary

A5's AICore retirement implements a subset of the a2a3 protocol. This issue quantifies what completing that port actually costs.

Porting a2a3's round-robin ACK sweep, deferred IDLE reset and post-broadcast `wmb` into A5 (call it ExitSync — the closest A5 can get, since FAST_PATH CLOSE / CLOSE readback / the GM return gate have no A5 counterpart) does make `SchedulerContext::shutdown()` faster: **+1.945 µs vs the current implementation, 27/27 paired blocks**. But the saving does not transfer. The segment from `shutdown` returning to `graph_build` returning gets **slower in 27 of 27 blocks** (−2.355 µs), cancelling it out.

**ExitSync does not remove work; it moves the wait out of the instrumented segment.**

The right criterion here is the **transfer ratio**, not how fast a segment looks in isolation. Injection calibration established that this tail transfers to the whole graph with slope **1.000** — a microsecond saved in the tail is a microsecond off `device_wall`:

| | `shutdown` gain | `window_to_gb_end` gain | transfer |
|---|---:|---:|---:|
| current (batched EXIT + shared deadline) | +3.907 | +3.790 | **97%** |
| ExitSync (fuller port) | +5.908 | +3.497 | **59%** |

On the net metrics ExitSync measures −0.306 µs on `window_to_gb_end` (8/27) and −0.979 µs on `device_wall` (9/27). Both have weak directional agreement and sit below that segment's 1 µs visibility threshold (a follow-up diagnostic round measured −0.108, 10/27). The accurate statement is therefore **no measurable net gain**, not "slower" — but either way, the added protocol complexity buys nothing.

**Conclusion: stopping at batched EXIT + shared deadline is the right trade-off. The remaining steps should not be ported wholesale.**

### Git Commit ID

e63903c3027d931444d176d8278f46e04b5e4bfc

### CANN Version

9.2.0 (V100R001C11B134)

### Driver Version

25.7.rc1.6 (ascendhal 7.35.23)

### Host Platform

Linux (x86_64)

### Reproduction

```bash
# Nine v4-pro graphs, A5 development card, four arms bracketed per block:
#   M0 -> candidate -> candidate -> M0End   (candidates swapped in block 2)
[...]

PR #2287, opened 2026-09-17
Add: stamp the A5 AICore exit handshake as a device phase

qweasdzxcht:
## Summary

- Adds one fixed `AicpuPhase` (`Shutdown`) around `SchedulerContext::shutdown()` on A5 `tensormap_and_ringbuffer`, plus its entry in the onboard and sim marker tables. No behaviour change.
- **Why**: `window_to_gb_end` — from the dispatch windows closing to `graph_build` returning — has no marker of its own; it is derived as `graph_build` end − `max(orch_end, sched_end)`, so its cost cannot be attributed to code. With the stamp, on nine v4-pro graphs the exit handshake is **90.4% of that segment** (13.759 of 15.245 µs), and the remainder is visible as a separate residue. #2286 depends on that split: a candidate there is faster inside `shutdown` (+1.945 µs, 27/27) yet gains nothing net, because the saving reappears in the residue.
- **Contract**: fixed cardinality — **one stamp per run per thread**, never per core or per poll; variable-cardinality timing stays in the chip swimlane ring.
- **Cost**: two `mrs cntvct_el0` reads and two 8-byte stores per thread. Measured: going from two spans to four in the same buffer perturbed `shutdown` by 0.038 µs and `window_to_gb_end` by 0.121 µs, both below the M0-vs-M0End drift of the same runs (0.129–0.208 and 0.272–0.400 µs).
- **For reviewers**: `NUM_AICPU_PHASES` changes, so the device-side buffer layout changes — **host and AICPU must be rebuilt together**. The entry is appended before `Count`, so existing slot indices are unaffected; the `static_assert` in both host tables catches a missed table.

## Testing

- [x] Built for a5 onboard; `strings` on the host library shows `device_wall.shutdown`, absent without this change
- [x] Per-call self-consistency over 1620 calls: `win_to_sd + shutdown + sd_to_gbend == window_to_gb_end` to 0.000000 µs; `shutdown ⊂ graph_build` in 540/540 calls per arm
- [x] Onboard: nine v4-pro graphs, A5 development card, 27 paired blocks, all passing
- [ ] Simulation tests — CI `st-sim-a5sim` (an earlier run hit a `chip_swimlane` fanout flake, analysed in the comments; rerunning on this push)

Relates to #2286


qweasdzxcht:
`st-sim-a5sim` is red on both runners here. I believe it is flaky rather than caused by this change, and I cannot re-run it myself (no admin rights on the repo) — so flagging it rather than leaving it silently red.

**The failure**

```
FAILED tests/st/a5/tensormap_and_ringbuffer/dfx/chip_swimlane/test_chip_swimlane.py::TestChipSwimlane::test_run
AssertionError: Fanout edges mismatch: printed=0, oracle=6
```

`printed=0` means `sched_overhead_analysis` saw no fanout edges. That happens when `ChipSwimlaneDataHeader.num_phase_cores` is still 0, which `chip_swimlane_aicpu_init_core_assignments()` leaves untouched when `s_phase_initialized` is false — i.e. the swimlane phase pools never came up. (The nanobind leak report after it is a consequence of the failing test, not a separate fault.)

**Why this change should not reach that path**

- The diff adds one RAII timing object in `SchedulerContext::shutdown()` and one enum entry with its two marker-table rows. **No control flow changes.**
- `s_phase_initialized` is set false only when `g_platform_chip_swimlane_base` is null. The swimlane buffer and the device-phase buffer are separate regions (`g_platform_chip_swimlane_base` vs `get_platform_phase_base()`), so growing `NUM_AICPU_PHASES` does not encroach on it.
- `ut-a5` and `ut-a2a3` pass on this PR.

**Sibling-job evidence**

#2288 is stacked on this branch, so it contains **every commit here plus its own**. Its `st-sim-a5sim` passes on both runners:

- ubuntu: https://github.com/hw-native-sys/simpler/actions/runs/35219711808/job/105197394313
- macos: https://github.com/hw-native-sys/simpler/actions/runs/35219711808/job/105197394305

Failing run on this PR: https://github.com/hw-native-sys/simpler/actions/runs/35219158073

A superset of this change passing the same job on both runners, while this one fails on both, points at flakiness in the swimlane phase-pool bring-up rather than at the diff.

Could someone with the rights re-run the failed jobs? If it reproduces, I will dig further rather than assume flakiness.

qweasdzxcht:
Update: #2288 is no longer stacked on this PR — it is now a single commit on `main` without the `Shutdown` phase, and the two merge cleanly in either order. The sibling-job evidence above refers to the earlier stacked #2288. This PR was rebased onto current `main` (8aa76faa, content unchanged), which also re-runs `st-sim-a5sim`; if the `chip_swimlane` fanout failure reproduces I will dig into it rather than call it flaky.

qweasdzxcht:
Follow-up: on the rebased head (8aa76faa) `st-sim-a5sim` passes on both ubuntu and macos, together with the rest of the matrix (19/19). The `chip_swimlane` fanout failure did not reproduce, consistent with a flake in the swimlane phase-pool bring-up rather than this change.

qweasdzxcht:
Closing. The replacement series in #2387 reports its numbers from the markers
already in the tree — the retirement tail is `graph_build` end minus
`max(orch end, sched end)` — so it does not depend on this.

The finer segmentation was what located the cost inside the retirement (it sat
in the claim, not in polling or in the read-back). That question is answered
and the answer does not need to ship. If per-segment retirement timing is
wanted later it is better proposed as a benchmark-tooling change, separately
from the protocol.

Issue #1710, opened 2026-08-06
Code health: three scheduler trees pay one deinit timeout per dead core on fatal shutdown


sunkaixuan2018:
## Description

`emergency_shutdown` has two shapes in the tree after #1664.

The a2a3 `tensormap_and_ringbuffer` scheduler broadcasts the AICore exit signal to every handshake'd core first, then joins them against a **single shared deadline**:

```cpp
for (...) platform_signal_aicore_exit(reg_addr);          // broadcast
const uint64_t exit_deadline = platform_aicore_exit_deadline();
for (...) platform_finish_aicore_exit(reg_addr, exit_deadline);   // join, one deadline
```

The other three trees still run the serial `platform_deinit_aicore_regs` loop — signal, wait, quiesce, next core.

**These three are not defective.** They never lost the acknowledgement wait, so every core is still quiesced and the card stays usable for the next process. The difference is cost, and it is worth measuring rather than assuming:

- The serial shape pays **one full deinit timeout per unresponsive core**. Onboard that is `inner_get_deinit_timeout_ticks() == PLATFORM_PROF_SYS_CNT_FREQ`, i.e. 1 s per core.
- A fatal run where every core is dead is the common case, not the rare one — the AICore fault is what triggered the shutdown.
- So an N-core fatal shutdown burns ~N seconds inside the AICPU op. That is the same order as the STARS op-execute timeout that kills `aicpu-sd`, which means a slow fatal teardown can be reaped as a hang and masked as a different failure.
- The shared-deadline shape caps the whole group at one timeout regardless of core count. On the sim variants the per-core timeout is 10 s, so the gap is larger there.

The three trees also lack the `publish_fatal_shutdown` fence, which keeps a healthy per-thread `shutdown()` from racing the emergency broadcast over the same cores.

This was deliberately left out of #1664's scope: it would mean changing three schedulers that work correctly today, and a5 hardware was not available to validate them. @ChaoZheng109 reviewed that scoping and agreed, asking that the N-second cost be measured before deciding whether to touch these trees.

## Location

- `src/a2a3/runtime/host_build_graph/runtime/scheduler/scheduler_cold_path.cpp`
- `src/a5/runtime/host_build_graph/runtime/scheduler/scheduler_cold_path.cpp`
- `src/a5/runtime/tensormap_and_ringbuffer/runtime/scheduler/scheduler_cold_path.cpp`
[...]

PR #2288, opened 2026-09-17
Add: broadcast AICore EXIT to the owned set before waiting

qweasdzxcht:
## Summary

- `SchedulerContext::shutdown()` (A5 `tensormap_and_ringbuffer`) signals EXIT to every owned core first, then collects the ACKs, so the cores' exit latencies overlap instead of queueing behind each other.
- Splits `platform_deinit_aicore_regs` into three exported steps — `platform_signal_aicore_exit`, `platform_aicore_exit_deadline`, `platform_finish_aicore_exit` — matching the a2a3 helpers of the same names. `platform_deinit_aicore_regs` keeps its signature and semantics, so its other callers (HBG, emergency paths) are unchanged.
- Updates `RUNTIME_LOGIC.md` and the idempotence comment in `aicpu_executor.cpp` to the new sequence.

**Measured** (nine v4-pro graphs, A5 development card, 27 blocks bracketed `M0 → candidate → M0End`, 6 warmup + 20 measured calls per process; µs, medians):

| segment | before | after | gain | same direction |
| ------- | -----: | ----: | ---: | -------------: |
| exit handshake (`shutdown`) | 13.759 | 9.857 | **+3.907** | **27/27** |
| `window_to_gb_end` | 15.245 | 11.656 | **+3.790** | **27/27** |
| `post_orch` | 3.650 | 3.462 | +0.122 | 18/27 |

- COND reads per thread fall **41 → 20** (7 cores per thread).
- **The saving transfers**: injection calibration puts `window_to_gb_end`'s transfer slope at 1.000, and the inner and outer gains agree to **97%**. #2286 shows the opposite case — a fuller a2a3 port that is faster inside `shutdown` and gives it all back in the next segment.
- **Fixed per-call cost**: across graphs spanning 76.9–5507.4 µs of `device_wall` the gain stays 3.757–4.005 µs; two unrelated workloads (`batch_paged_attention` ≈ 6.8 ms, `qwen3_14b_decode` ≈ 35 ms) give +4.236 (4/4). Relative value is highest on short graphs (4.91% of `device_wall` on `hc_head`, 0.07% on `mtp_prefill`). Whole-graph noise is 0.58–8.56 µs against a 4 µs effect, so judge it on the segment.
- The `shutdown` segment is the `Shutdown` phase added in #2287; this PR does not depend on it and applies to `main` on its own.

**Behaviour change for reviewers to weigh.** The per-core timeout becomes **one deadline shared by the group**, taken once after the broadcast; a core that answers late is judged against the same absolute bound as its peers. On the normal path every core ACKs within ~20 polls (~7.7 µs), four orders of magnitude inside the 1 s onboard budget, so **the slow-core and unresponsive-core behaviour is not characterised by these measurements** and needs fault injection (hold a core silent; check each core's effective budget, that surviving cores are still released, the timeout upper bound, recovery on the next call). For context, #1710 argues this shape from the fatal-shutdown side: with every core dead, the per-core loop pays one 1 s timeout per core, a group deadline pays one. `emergency_shutdown` and the `publish_fatal_shutdown` fence that #1710 also asks for are not in this PR.

## Testing

- [x] Onboard (A5 development card): nine v4-pro graphs, 135 processes, all passing against the frozen golden set; artifact SHAs verified at both ends of each run
- [x] Cross-workload: `batch_paged_attention`, `qwen3_14b_decode`, 4/4
- [ ] Abnormal exit path (slow / unresponsive core) — not covered, see above
- [ ] Robustness across inputs — single frozen input
- [ ] Simulation tests — CI `st-sim-a5sim`

Relates to #2286, #1710


qweasdzxcht:
Restructured to follow the repo PR convention: this is now one commit on `main` and no longer depends on #2287. The `ExitBroadcast` phase that needed #2287's enum entry is dropped; none of the measurements in the description rely on it. `RUNTIME_LOGIC.md` and the idempotence comment in `aicpu_executor.cpp` are updated to the new `shutdown()` sequence.

qweasdzxcht:
Closing in favour of #2387.

The speedup measured here is real and reproduced across five rounds (+3.90 us
on the retirement tail, sign holding in 27/27 or 12/12 paired blocks each
round). The reason for withdrawing it as a standalone change is the fault path.

AICore exit waiting has two independent properties:

```
P1  every core is judged on its own full timeout budget
P2  the whole group's retirement is bounded by one budget
```

Today's per-core wait has P1 and not P2. This PR's shared deadline with a
blocking per-core wait has P2 and not P1: a healthy core queued behind a slow
one is judged on the budget the slow one already spent. That is the concrete
content of the "changes slow/unresponsive core behaviour" caveat in the
description above.

Only a non-blocking round-robin sweep has both, and it carries the same batched
broadcast proposed here. So the batching lands in #2387 as part of the sweep
rather than on its own.

PR #2289, opened 2026-09-17
Add: [HBG kernel] connect graph execution

TaoZQY:
## Summary

Enable HBG Kernel execution through the shared caller/AICPU/AICore binder, including eager launches and real ACLGraph capture/replay. Graphs with internally allocated intermediate tensors now survive build, packet transport and device restore. Launch packets carry live graph data instead of unused workspace capacity.

Context initialization allocates and freezes execution resources. Callable preparation registers the execution slot and resident callable before capture. Each invocation owns an immutable graph snapshot; replay restores its working state without rebuilding the graph on Host.

## Implementation and fixes

| Area | Problem / resulting behavior | Main locations |
| --- | --- | --- |
| Kernel integration | Connect context initialization, callable preparation, three-stream launch and device execution. Function IDs refer to device-side `CoreCallable` descriptors, as required by the scheduler ABI. | `src/common/host_build_graph/host/kernel_graph_owner.cpp`, `host/kernel_launch_owner.cpp`, `device/kernel_graph_execution.cpp`; common onboard runner and architecture build wiring |
| Internal tensors | Rebased internal tensors were mistaken for external tensors and rejected during restore. Validate their address and size against the registered live heap; external tensors still require DEVICE storage. Reject malformed images before destination writes. | `kernel_external_tensor_wire.h`, `device/kernel_graph_restore.cpp` |
| Packet size and restore | Separate frozen capacity from live image length. Transmit only used runtime/SM, Definition and scheduler data; clear, copy, flush and invalidate only live spans. Versioned packet/slot layouts reject incompatible inputs. | `host/kernel_graph_template.cpp`, `kernel_graph_wire.h`, `kernel_graph_slot_wire.h`, device restore and registry |
| RTS argument-memory recovery | On HostArgs allocation failure, synchronize the prelaunch `AicoreStart` event and retry once. Never wait on the current AICore completion or stream tail. Synchronization/retry errors propagate through cancellation and context poisoning. | `src/common/aicpu_loader/host/kernel_graph_launch.h` |
| Regression coverage and contracts | Add internal-intermediate, packet-capacity, recovery, queued eager and ACLGraph tests; document ownership, live lengths and capture/replay coverage. | HBG C++ tests, Kernel C API/Worker tests, `tests/st/a2a3/host_build_graph/kernel_mode_capture/`, HBG Kernel contract/slot docs |

HBG-relative paths above are under `src/common/host_build_graph/`. Based on `feat/kernel-mode-integration-test` at `4f162da0`, retaining its init-time static-resource lifecycle and TMR residency behavior.

### HBG Kernel ACLGraph capture/replay

**12 passed; 1,200 replays; 102.49 s total test-suite time.** Task: `task_20260917_185240_10239712560`.

Each scenario runs both a single-task callable and a two-task callable with an internally allocated intermediate tensor, with **100 replays per variant**:

| Scenario | What is checked | Both variants |
| --- | --- | --- |
| Fresh inputs | Change input contents at fixed addresses and poison outputs before each replay | PASS |
| Dependent calls | Capture calls with different scalars and dependent output/input buffers | PASS |
| Batched feedback | Alternate two feedback graphs without intermediate synchronization; check ordering and device execution counters | PASS |
| Eager/replay interleaving | Change eager arguments while replay retains its captured arguments; check feedback and counters | PASS |
| Graph recreation | Capture, execute and destroy 100 graphs with changing scalar specializations | PASS |
| Long chain | Capture 16 dependent calls plus a counter; enqueue 100 replays before the final synchronization | PASS |

The test uses the public `Worker(level=2, execution_mode="kernel", runtime="host_build_graph")` API and real CANN global capture. Preparation finishes before capture; no eager warmup is required. Inputs are **16,384-element FP32 vectors** with exactly representable data and scalar increments. The two-task variant adds half the graph-level increment in each child, so both variants share the same numerical oracle.

All outputs match the reference exactly. Host argument objects are cleared immediately after enqueue. Context-reported committed memory stays at **539,834,806 bytes**, with **zero growth** across each case; this is not a measurement of the whole device or RTS memory pool. Graphs are destroyed before Worker close, and caller-owned buffers, streams and device resources close successfully.

### Eager and conventional-runtime regressions

- Kernel C API + Worker: **24 passed across HBG and TMR, including 4 HBG-specific cases**. Task: `task_20260917_183624_182066115298`.
- HBG eager coverage includes 16 queued two-task invocations with different inputs/scalars and no intermediate sync; exact outputs, Host-argument lifetime and unchanged context committed memory are checked.
- Conventional HBG runtime regression, separate from Kernel-mode evidence: **47 passed, 1 skipped**. Task: `task_20260917_183638_200792610462`.

## Additional validation and scope

- HBG simulation: **20 passed, 7 skipped**.
- Python Kernel/Worker unit tests: **71 passed, 33 skipped** without hardware selection.
- HBG C++ build/packet/restore tests: **103 passed for each architecture**; native binder/HostArgs tests: **5 passed**.
- Runtime builds: `a2a3sim`, `a5sim`, `a2a3`, `a5`.
- Pre-commit checks and [documentation CI](https://github.com/hw-native-sys/simpler/actions/runs/35297169492) passed.

The new regressions failed before their fixes: restore rejected an ordinary internal intermediate; a small graph transmitted approximately 13 MB when unused capacity was included; and allocation-error fault injection showed the missing bounded retry. RTS memory recovery is validated by fault injection, not by exhausting real device memory.

Hardware evidence is limited to **A3 eager execution and global ACLGraph capture/replay with preparation before capture**. A5 hardware, prepare-inside-capture, latency benchmarking and real RTS-memory-exhaustion stress are not claimed.


nalinaly:
I checked the current PR head `47756a4c` and ran the minimal A3 hardware cases. The public HBG kernel path is now reachable, but numerical execution still fails.

### 1. Blocking: the callable table contains the wrong kind of address

In [`bind_callable`](https://github.com/hw-native-sys/simpler/blob/47756a4c362fb7352e012b8f69d563746af08885/src/common/host_build_graph/device/kernel_graph_execution.cpp#L68), the new path does:

```cpp
runtime->replay_function_bin_addr(func_id, child->resolved_addr_);
```

However, the existing [A3 HBG scheduler](https://github.com/hw-native-sys/simpler/blob/47756a4c362fb7352e012b8f69d563746af08885/src/a2a3/runtime/host_build_graph/runtime/scheduler/scheduler_dispatch.cpp#L144) interprets that table entry as a **CoreCallable descriptor address**, then reads `callable->resolved_addr()`. The A5 HBG scheduler has the same contract.

The new producer stores the **machine-code address**, so the consumer effectively reads another pointer at `code_address + offsetof(CoreCallable, resolved_addr_)`, i.e. `code_address + 136`. The add-scalar code compiled in this run is only 124 bytes long, so this read is beyond the actual code. The existing program path instead stores the child descriptor's device address via `upload_and_collect_child_addrs`.

Please preserve that contract here by registering the device address of `child`, rather than `child->resolved_addr_`.

Hardware evidence:

- `test_hbg_kernel_eager_launch_executes_graph_snapshots` fails at the output comparison in the **first invocation**.
- Init, prepare, launch, and caller-stream synchronization return success, but all 16,384 output elements remain the prefilled `-999.0` sentinel; expected values start with `1.25, 2.25, 3.25, ...`.
- The Worker `_case_hbg_eager_numerics` fails as well.
- Device logs show a one-task graph, one AIV queue push, one scheduled task, and normal AICPU completion.
- On the same device/install, the equivalent TRB kernel add-scalar test and the HBG **program** vector example pass.

The [A3 AICore executor](https://github.com/hw-native-sys/simpler/blob/47756a4c362fb7352e012b8f69d563746af08885/src/a2a3/runtime/host_build_graph/aicore/aicore_executor.cpp#L36) silently skips a zero function address while the enclosing protocol can still complete the task. This is a possible silent-failure mechanism, **not a claim that the erroneous device pointer was directly measured as zero**. The descriptor/code-address mismatch is established from the source; a fix-and-rerun A/B has not been performed.

### 2. Existing restore mismatch becomes relevant now that public HBG kernel execution is enabled

This is a static finding in the inherited packet/restore path, separate from the external-tensor-only failure above:

- `init_tensor_from_create_info` creates internal tensors with `AddressSpace::HOST`.
- `compact_live_image` rebases their virtual heap addresses to the real device heap without changing that tag.
- [`valid_kernel_graph_tensor`](https://github.com/hw-native-sys/simpler/blob/47756a4c362fb7352e012b8f69d563746af08885/src/common/host_build_graph/kernel_external_tensor_wire.h#L35), called during device restore, treats addresses below `HEAP_VIRTUAL_BASE` as external and requires `DEVICE`.

Consequently, an ordinary task chain with runtime-created intermediate tensors can be rejected after rebasing. Please distinguish tensors backed by the registered execution heap from external tensors, and add a build → packet → restore regression containing an ordinary internal intermediate. The current eager add-scalar test uses only external input/output tensors and cannot cover this.

### 3. Per-launch HostArgs is sized to frozen capacity, not the actual graph

[`prepare_kernel_graph_resources`](https://github.com/hw-native-sys/simpler/blob/47756a4c362fb7352e012b8f69d563746af08885/src/common/host_build_graph/host/kernel_graph_owner.cpp#L43) defaults both heap and Definition capacity to 256 MiB; A5 adds another 256 MiB scheduler region. [Packet construction](https://github.com/hw-native-sys/simpler/blob/47756a4c362fb7352e012b8f69d563746af08885/src/common/host_build_graph/host/kernel_graph_template.cpp#L270) copies the full non-heap destination capacities into each invocation.

Thus even a one-task graph with no Definitions carries 256 MiB of Definition capacity in default A3 HostArgs, plus runtime/SM; A5 additionally carries the scheduler capacity. Restore also copies whole capacities and clears the whole heap.

Please separate **workspace capacity** from **valid per-invocation image length**, transmit only the valid image, and initialize only the mutable state required by the execution protocol. These are source-derived sizes, not a queued-HBM stress measurement. The hardware tests above explicitly used 64 tasks and a 1 MiB heap.

Importantly, this RTS-owned per-launch graph packet is separate from both the reusable Simpler workspace and the callable-residency budget (8,192 entries / 2 GiB).

### 4. The discussed args-memory recovery policy is still absent

[`launch_graph_template`](https://github.com/hw-native-sys/simpler/blob/47756a4c362fb7352e012b8f69d563746af08885/src/common/aicpu_loader/host/kernel_graph_launch.h#L38) returns the RTS launch error directly; the binder compensates/cancels and poisons the context. It does not implement the previously discussed bounded policy: synchronize an earlier event-timeline point, retry once if synchronization succeeds, then return failure if synchronization or the retry fails.

Please account for this when claiming support for deeply queued whole-model eager launches. Any recovery synchronization must avoid waiting on the current AICore task, which may already be waiting for the not-yet-enqueued AICPU task.

Scope: these hardware observations are A3 eager only. A5 hardware, multiple in-flight launches, and ACLGraph replay were not verified. No runtime implementation was modified during this check.

nalinaly:
Follow-up on `917f5c68`: the minimal A3 recheck now passes all six selected hardware cases (eager values, 16 queued invocations with an internal intermediate, and four internal-intermediate ACLGraph scenarios with 100 replays each). The previous callable-address and internal-tensor restore failures no longer reproduce.

Two remaining integration points need to be tracked separately:

### 1. HBG callable preparation still synchronizes; clarify the first-use/capture contract

In [`prepare_hbg_kernel_callable_registration`](https://github.com/hw-native-sys/simpler/blob/917f5c68f7bd72793f81fba715f88411027f8dcf/src/common/host_build_graph/host/kernel_graph_owner.cpp#L145), `aclrtSynchronizeStreamWithTimeout` is still called:

- after the first execution-slot registration, at line 167;
- after each new callable registration, at line 191.

The [HBG capture test](https://github.com/hw-native-sys/simpler/blob/917f5c68f7bd72793f81fba715f88411027f8dcf/tests/st/a2a3/host_build_graph/kernel_mode_capture/test_hbg_kernel_capture.py#L85) completes both `worker.init()` and `kernel_prepare_callable()` **before** capture. Therefore, the passing tests establish capture/replay of an already-prepared callable, not first-time lazy preparation during capture.

The updated PR description and [HBG-specific contract](https://github.com/hw-native-sys/simpler/blob/917f5c68f7bd72793f81fba715f88411027f8dcf/docs/host-build-graph-kernel-contract.md#L114) correctly state this restriction. However, the [generic integration API table](https://github.com/hw-native-sys/simpler/blob/917f5c68f7bd72793f81fba715f88411027f8dcf/docs/zh-cn/kernel-mode-integration-test.md#L244) still says that prepare does not synchronize and can be called inside capture. These are different contracts for HBG and TRB and must not be presented as one.

Please reconcile the API documentation and define the intended behavior when a new HBG callable is first encountered during capture. If that lazy path is intended to work, registration must use capture-compatible ordering rather than these blocking synchronizations. If it remains unsupported, document the runtime-specific restriction and reject it clearly before partial registration side effects. A first-prepare-inside-capture test should check the chosen behavior; the existing prepare-before-capture tests cannot cover it. Simply deleting the waits without preserving registration-before-execution ordering would not be sufficient.

### 2. Simpler HBG support is not yet end-to-end PyPTO/PyTorch support

In the locally checked PyPTO source (`5b52a034`), `python/pypto/torch/launch.py::describe_eager_call` still accepts only:

```python
(abi.platform, abi.runtime) == ("a2a3", "tensormap_and_ringbuffer")
```

HBG is rejected at this frontend gate before reaching the now-working Simpler kernel path.

There is also a separate local build-alignment issue: the Python-side expected Simpler revision has been updated to `917f5c68`, while the installed `pypto._torch_npu` adapter still reports `6c0aff3c`. `launch.py::_load_native` correctly rejects that mismatch. Reinstalling Simpler alone does not rebuild the PyPTO adapter; bypassing the revision check is not a fix.

Please track the companion PyPTO work explicitly: enable HBG only with the appropriate supported-call contract, align/rebuild the native adapter against the selected Simpler revision, and add eager plus ACLGraph coverage through the actual PyPTO/PyTorch entry rather than only the Simpler C API/Worker.

The stale adapter is a **local integration/build issue**, not a new device-execution defect introduced by this PR. The frontend gate is also owned by PyPTO. They are recorded here as remaining end-to-end integration dependencies, not as evidence that the newly fixed HBG backend still fails.

TaoZQY:
@nalinaly Addressed the capture-preparation contract and tracked the PyPTO dependencies separately.

- HBG context init now registers and waits for the sealed execution slot outside capture. Callable prepare uploads context-owned images and stores immutable registration metadata without stream/event/device synchronization or a registration-kernel enqueue.
- Each launch enqueues idempotent callable registration immediately before graph execution on the same AICPU stream. Both nodes are captured together. A newly prepared callable can run eagerly, or in another graph, before the original graph's first replay. Registration enqueue errors use the binder's cancellation/poison path.
- Added first-prepare and second-prepare cases in global and thread-local capture, each for a single-task graph and an internal-intermediate graph. The second-prepare case first launches an existing callable to join the internal streams. An interposer rejects synchronization during prepare; after capture, the new callable is executed eagerly with different arguments before the first replay. The original implementation fails this guard; the fix passes.
- Reconciled the generic integration API table, HBG contract/slot docs and Worker docstring. Also moved the program-mode READY default into the prelaunch control declaration, so minimal Runtime constructors preserve the same default; this fixes the executor retirement regression found during broader validation.

The earlier four findings are incorporated in the same PR: descriptor addresses in the function table, internal-heap tensor restore validation, live-image packet lengths, and a bounded HostArgs allocation retry using the prelaunch event. Their regression coverage remains in place.

Validation: **49/49 Kernel onboard cases passed**, including **20 HBG capture variants × 100 = 2,000 replays** (task `task_20260919_185120_407791628910`); the general A3 sweep passed **217 cases with 1 skip** (task `task_20260919_185130_408508725280`); HBG simulation **20 passed/7 skipped**; Python units **74 passed/44 skipped**; focused C++ **44/44 targets passed**; pre-commit passed. After the initialization-only adjustment, the rebuilt final revision passed **12/12 focused onboard cases**, including all 8 capture-preparation variants (**800 replays**), HBG eager and the program vector case (task `task_20260919_190111_69924920822`). The native extension stamp matches the source revision.

The PyPTO frontend gate and native adapter are explicitly listed as pending companion work in the PR description and `docs/zh-cn/kernel-mode-integration-test.md`: admit HBG under its supported-call contract, align and rebuild `pypto._torch_npu` against the selected Simpler revision without bypassing the check, and run eager plus capture/replay through the actual PyPTO/PyTorch entry. These Simpler C API/Worker results do not claim that frontend integration is complete. A5 onboard and latency measurements remain outside the validated scope.

TaoZQY:
Resolved the merge conflict by rebasing onto the current integration base, including #2306/#2393/#2402.

The conflict was in `src/common/platform/onboard/host/kernel_launch_owner.cpp`: the base removed `kernel_result_handle_`, while the HBG dispatch guard still referenced it. The resolution keeps HostApi validation and HBG dispatch, applies the coordination-ready guard only to TMR, and preserves the base's native RTS fail-stop handling without restoring the removed result-check task. Range-diff confirms this is the only adjustment to the PR patch; the PR remains one commit.

Validation: runtime rebuild and native/source revision match; static checks passed; 6 C++ targets passed; Python 74 passed/44 skipped; HBG sim 20 passed/7 skipped. The A3 kernel sweep reported **79 passed, 1 failed** (task `task_20260920_232907_627011466`). The HBG internal-tensor eager/replay case returned `507015` on device 10, then passed once on device 12 and five consecutive times on the original device 10 without code changes (600 successful replays; tasks `task_20260920_232959_66941030071` and `task_20260920_233041_70943523126`). This intermittent failure is explicitly recorded in the updated PR description; its cause is not established.

YunjiQin (approved):

Issue #2264, opened 2026-09-16
[CI] st-network1-onboard-a2a3 fails from remote-node device init (507033), not from the change under test


ChaoWao:
### Platform

a2a3 (Ascend 910B/C hardware)

### Runtime Variant

tensormap_and_ringbuffer — but the job fails before any runtime code executes.

### Description

`st-network1-onboard-a2a3` is failing across **unrelated PRs**, and the one
failure whose log I read fails in **remote-node device initialization**, before
any test workload runs:

```text
RuntimeError: remote L3 session open failed for worker 1:
  remote L3 session startup failed for worker 1: remote session startup:
    File ".../simpler/remote_l3_session.py", line 1299, in run_session
      inner_worker.init(_startup_deadline=startup_deadline)
    File ".../simpler/worker.py", line 8331, in _await_children_ready
      raise RuntimeError(f"{kind} worker {i} (pid {pids[i]}) failed during init: ...")
  RuntimeError: chip worker 0 (pid 336220) failed during init:
    chip_process dev=14 init: RuntimeError: simpler_init failed with code 507033
```

The local side came up fine (`[chip_process pid=3406204 dev=12] ready`). What
failed is the **remote** machine's chip worker on `dev=14` — a device the local
job never allocated, so the local `--device 12,13` reservation says nothing
about whether it was free.

Filing rather than re-running to green because this lane is red on several PRs
at once, the failing test set **rotates between runs**, and one test
(`global_tload_mpirun_l3`) fails identically on two unrelated changes. That
combination is the signature of the environment rather than of any one change —
and per [`discipline.md`](../blob/main/.claude/rules/discipline.md) §5, a lane
re-run to green without recording why is indistinguishable from one hiding a
real regression.

### Observations

| # | PR | Job | Failing test(s) | Mechanism |
| - | -- | --- | --------------- | --------- |
| 1 | #2262 | [104729208186](https://github.com/hw-native-sys/simpler/actions/runs/35075949632/job/104729208186) | `test_global_tload_mixed_l3_network1`, `global_tload_mixed_l3`, `vector_add_mixed_l3`, `compute_then_tload_mixed_l3` (4) | remote `chip_process dev=14 init: simpler_init failed with code 507033` (log read) |
| 2 | #2262 (re-run) | [104745555245](https://github.com/hw-native-sys/simpler/actions/runs/35080818071/job/104745555245) | `global_tload_mpirun_l3` | **not read** — run still in progress, so only the annotation is available |
[...]

PR #2293, opened 2026-09-17
Add: attribute diagnostic records to the run that produced them

ChaoWao:
## Summary

Part of #1853 Track A §2.4 (continuous DFX collection), first step: **give diagnostic records a run identity**, so attribution stops depending on the host clearing every collector between runs.

Five collectors write into pooled device buffers and **no record says which run it belongs to**. Attribution works today only because `begin_run()` clears everything — whatever arrives afterwards must be this run's. `device_runner_base.cpp` states the dependency outright: the clear *"is only safe while this run holds the execution claim"*.

That makes the clearing load-bearing, which is what blocks reusing a collector across runs. Removing it first wouldn't fail — it would **silently mis-attribute**. So identity has to land first.

## Where the stamp goes

A buffer carries the run that acquired it; the device stamps once at acquisition and the host copies it out with the records.

- **scope_stats + the three AICPU-side swimlane pools** stamp in their module's `on_pop_success`. The generic engine calls that hook after advancing the buffer generation and **before its own `wmb()`**, so the stamp is published by that fence — ahead of the first record, and ahead of the `enqueue_ready` that makes the buffer reachable by the host. The engine stays unaware of epochs.
- **The AICore swimlane pool can't do that.** `chip_swimlane_aicore_record_task` is `__aicore__` code in a separate program that does not link the AICPU SO, so it can never know the current run. AICPU stamps that buffer for it at both publish points (prime in `chip_swimlane_aicpu_init`, and `aicore_rotate`), and AICore's records inherit the run **by residency** — a record belongs to the buffer it was reserved from, even when the head rotates between reserve and commit.

One ordering fix fell out of this: the prime block claimed *"same publish pattern as aicore_rotate"* but wrote the buffer's contents **after** publishing ptr and seq. `count` tolerates that because AICore resolves the head lazily on its first task; identity does not — a buffer published before it is stamped attributes its records to whatever the storage last held. Prime now matches rotate.

## Cost

**No device memory.** Both `ScopeStatsBuffer` and `TypedBuffer` had alignment tail behind their trailing `count`, and the fields fit inside it. Static assertions pin that — and since `chip_swimlane_profiling.h` is compiled by **ccec for the AICore program** as well as by the host compiler, those assertions also hold the two toolchains to the same layout.

**No extra D2H for scope_stats**: the host already copies the buffer's 64-byte header separately to learn `count`, and identity rides along in it.

## Host side: copied, not referenced

A buffer returned to the pool is re-stamped by whichever run takes it next, so a record already collected **cannot** recover its run from the buffer it came from. Identity is therefore copied into the collected record. `CollectedRecord<Record>` is shared rather than per-collector — attribution is the same question everywhere, and three more collectors still need it.

## What this deliberately does not do

- **Swimlane artifact schema unchanged.** `aicore_tasks` / `scheduler_tasks` are positional arrays and `swimlane_converter.py:512` rejects a row that isn't five or six columns, so widening them belongs with the session container + converter. Only scope_stats surfaces identity, where records are JSON-Lines objects and two fields are additive.
- **Per-run clearing untouched.** Removing it additionally needs the accounting that derives a buffer's live count from `total - seq * capacity`, which a partial tail buffer breaks.
- **args_dump / dep_gen / pmu have no identity yet.**

## Testing

Every identity assertion has a negative control — a passing new test proves nothing until I know it can fail:

- reverting the stamp fails the scope_stats identity tests and the swimlane host-copy test, and **leaves the accounting test passing**, because that one doesn't depend on identity;
- the swimlane host-copy test re-stamps the device buffer *after* collection, which catches a copy that aliased it rather than copying;
- on device, the scope_stats scene test now requires every record to carry one non-zero epoch. That is what shows the real AICPU stamping works rather than a fixture — the cpput AICore test cannot show it, since nothing there runs AICPU's prime/rotate.

Runs: cpput 151/151 · a2a3 onboard sweep 74/74 · a2a3 DFX 11/11 · a2a3sim 40/40 · a5sim 36/36 · pyut 2405 passed · all four platforms build.

One a2a3 DFX run out of three failed with the known `fftsplus aivector error` → 507018 (#1489 / #2264); it passes in isolation and on re-run, and this change adds no AICore code.

**a5 onboard is unverified** — no a5 silicon on this box. The change is arch-shared (`src/common/platform/`), so a5 exercises the same code path.

## Follow-ups this uncovered

While trying to replace the accounting I found two coupled **pre-existing** defects in `chip_swimlane_aicpu_flush`, filed separately as a prerequisite for removing the per-run clearing:

1. at TASK_TIMING the flush stamps `count = BUFFER_SIZE` on the stated grounds that `total_record_count` "stays 0" — but that counter is bumped per *dispatch* and is accurate at every level. Measured: the accurate count takes the host's empty-slot warnings from one-per-core down to **1** per run. The enqueue-failure path also charges `dropped` by the over-stated mark.
2. the `live == 0` branch `continue`s before clearing `current_buf_ptr`, and neither the next run's init nor host reconcile returns that buffer to the pool.

(1) can't be fixed without (2), because removing the fallback moves the `live == 0` continue into the TASK_TIMING path as well.

Refs #1853


ChaoWao:
Both fixed. Review base `3f525af2` → now `3674964d`.

**1. [P1] AICPU task pool's first buffer had no stamp — confirmed, and the reuse case is the worse one.** `chip_swimlane_aicpu_init` hand-pops that buffer and only cleared `count`, so it never reached `on_pop_success`. A run that never rotates published records under whatever the storage last held: zero on a fresh allocation, and *the previous run's epoch* on a reused one — which is silent mis-attribution rather than a visibly missing value.

I checked the sibling paths rather than just the one you named: the two phase pools prime through the engine's `try_pop_free`, so they were already covered, and scope_stats acquires lazily through the engine too. The AICPU task pool was the one remaining bypass — I had fixed the AICore prime next to it and missed this one. It also now writes the buffer's contents before publishing ptr/seq, matching what I'd already done for AICore.

**2. [P2] sim published no epoch — confirmed, and my verification of it was hollow.** Both simulated runners call `aicpu_execute` directly and nothing set the epoch, so every sim record carried none. Each now resolves `set_platform_run_result` alongside the phase-base setter and publishes the run's epoch at launch with a null region, since sim allocates none.

The more useful finding is *why* I didn't catch this. The artifact assertions in that scene test are gated on `--enable-scope-stats`, and none of my runs passed it — so `test_run` returned early and my new epoch assertion never executed. **My earlier claim that an a2a3 run had verified the device-side stamping was therefore unfounded**, and the local "pass" I got here after wiring sim was equally meaningless until I re-ran with the flag. With it:

- sim, epoch suppressed → fails with `records carry no run identity: {... 'run_epoch': 0 ...}`
- sim, epoch wired → passes
- a2a3 DFX, `--enable-scope-stats` → 11/11

That is the first run in which the assertion actually executed, so it is also the first real evidence that the AICPU stamping works on device.

## Re-verified

cpput 151/151 · a2a3 onboard sweep 74/74 · a2a3 DFX 11/11 (with `--enable-scope-stats`) · a2a3sim 40/40 · a5sim 36/36 · all four platforms build.

a5 onboard still unverified locally — no a5 silicon here; the sim half of the a5 wiring is covered by the a5sim sweep.

The count/return defects stay out of this PR, filed as a prerequisite for removing the per-run clearing rather than as follow-up.

ChaoWao:
Added. `3674964d` → `3cda9573`.

You're right that the existing test couldn't catch it — `CollectedRecordsCarryTheirRunAfterTheBufferIsReused` hand-fills `buf.run_epoch`, so it passes with or without the prime stamp. It tests the host copy, not the stamp.

The new one, `InitStampsTheFirstBufferOfAShortRun`, drives the production sequence instead: real `chip_swimlane_aicpu_init` (the hand-pop that was missing the stamp) → real `chip_swimlane_aicpu_complete_task` → real `chip_swimlane_aicpu_flush` → the collector's own `on_buffer_collected`. One record, far short of a rotation boundary, so the run publishes only out of the primed buffer.

The reuse case is built in rather than being a separate test: before `init`, the buffer sitting at the free queue's head is poisoned with `run_epoch = 4242`, standing in for a pooled buffer a previous run already stamped. A prime that doesn't stamp leaves that value in place, which is the mis-attribution rather than a visibly-absent one.

To make that possible the target now links the real AICPU producer module (`chip_swimlane_collector_aicpu.cpp` + the epoch globals + `inner_platform_regs.cpp`), so both sides of the path are production code in one binary — same shape as the scope_stats target.

**Negative control against the actual bug**, not a synthetic one: restoring the P1 defect (dropping the two stamp lines from `chip_swimlane_aicpu_init`, leaving `count = 0`) fails it with `init's hand-popped first buffer kept a stale or unset run identity`, while the other two swimlane tests keep passing. So it discriminates on exactly the defect it exists for.

## Re-verified

cpput 151/151 · a2a3 DFX 11/11 with `--enable-scope-stats` · a2a3sim scope_stats with the flag · a2a3sim 40/40 · a5sim 36/36 · all four platforms build.

a2a3 onboard sweep 73/74: `TestAllreduceTwophaseP2` failed with the `fftsplus aivector error` → 507018 signature (#1489 / #2264) and passes 10/10 in isolation on a re-run with no `aivector` in the log.

Issue #2270, opened 2026-09-16
[Feature] Introduce reusable PreparedCall with production TMR program-path integration


ChaoWao:
### Summary

Introduce an immutable, reusable `PreparedCall` containing the resolved host-side result of a TMR invocation, and make the **existing program execution path actually consume it**.

This is the next bounded milestone after merged #2176, derived from the shared program/kernel execution design. Work from current main and selectively reuse existing implementations; existing PR numbering or dependency stacks do not dictate the design or merge order.

The deliverable is a production-used preparation result with demonstrable isolation and reuse. A types-only implementation or standalone codec with fake consumers does not complete this issue.

### Motivation / Use Case

The target composition is:

```text
program: device ownership + tensor bindings/transfers ─┐
                                                     ├→ prepare → PreparedCall → submit → Submission
kernel: caller-owned device arguments ────────────────┘
```

Kernel execution may own resident runtime metadata and internal workspaces while borrowing the device and tensor storage. Program mode adds device ownership and tensor parameter management around that common execution capability.

At the inspected main baseline `fd5f2d93fc55c784a002049577b0015ebfa4194e`:

- #2176 establishes cold-path resources. `prepare_kernel_callable()` registers the callable and establishes persistent arguments; it does not produce a per-invocation prepared result. Public kernel launch remains unsupported.
- `PreparedExecution` combines a borrowed `Runtime*`, configuration, launch fields, native run identity, pipeline slot, and retirement state. `ActiveExecution` takes ownership of that object rather than allowing the preparation result to be reused.
- `launch_run()` still performs DFX arming and other host work; Runtime publication is synchronous and includes handshake state.

We need the explicit preparation result before introducing a common stream-ordered publication/reset/launch closure. Removing admission guards or enabling the launch stub first would preserve the coupling that the design is intended to remove.

#### 1. Three distinct lifetimes

| Object | Responsibility |
| --- | --- |
| `PreparedCall` | Immutable resolved invocation description and references to everything needed to consume it; reusable across executions |
| `Submission` (or the existing per-run state adapted to this role) | Identity, ordering/completion state, status, pipeline lease during migration, and references for one execution |
| `WorkingSet` | Mutable device execution storage; remains under existing serial execution ownership in this milestone |

One PreparedCall may be used by multiple submissions. Submission does not consume or mutate the PreparedCall. This milestone validates sequential reuse and does not promise simultaneous device execution or multiple queued launches.

[...]

Issue #2267, opened 2026-09-16
[Code Health] Use per-run dual-stream completion fences (Track A P2)


ChaoWao:
### Category

Technical Debt (cleanup, refactor)

### Component

Host Runtime

### Description

Make completion belong to an individual program run rather than to all work submitted to the runner's AICore/AICPU stream pair. This is **Track A / P2**, an independently mergeable prerequisite of #1853.

Today `poll_execution` queries the stream pair and successful drain synchronizes streams. Once a later change queues a successor, those operations can wait for successor work as well as the requested run. A run needs its own completion boundary before admission can safely be widened.

**This issue preserves one launched run at a time**, existing prepared-successor overlap, public APIs, result/copyback semantics, and the current admission/error policy. It introduces no throughput claim. The target is onboard a2a3 and a5, for both `host_build_graph` and `tensormap_and_ringbuffer`; keep simulation interfaces and behavior compatible.

Related: #1853 (parent queueing work), #2257 (independent KernelArgs launch-payload refactor). P1 owns timing/result-buffer isolation and other shared-resource lifetimes. P4 will widen native admission; P5 will enable the production `Worker.submit` path. P2 can proceed without waiting for P1 or #2257, provided it keeps existing admission closed.

### Location

Source reference: `4860af980ca3688612435292e0a2b256324d3751` (main after #2262). Recheck current main before implementation; other resource work may land independently.

- `src/common/worker/native_run_execution.h`: `exact_launch_transaction`, `LaunchProgress`, launch receipt/identity.
- `src/common/platform/onboard/host/device_runner_base.h`: `PreparedExecution`, `ActiveExecution`, poll/drain contracts.
- `src/a2a3/platform/onboard/host/device_runner.cpp`: `launch_run`, `poll_execution`, `drain_execution`, `reap_run`, cleanup.
- `src/a5/platform/onboard/host/device_runner.cpp`: launch transaction, runner-wide poll state, stream sync and cleanup.
- `src/common/platform/include/host/run_stream_pair.h`: stream ownership, completion and retirement.
- `src/common/platform/onboard/host/c_api_shared.cpp`: native prepare/launch/poll/wait/finalize and claim release.
- `src/common/worker/chip_run_lane.cpp`: progress/finalize consumer; preserve its admission behavior.
- `src/common/platform/onboard/host/clock_correlation.cpp`: existing event API usage, not an ownership model to reuse blindly.
[...]

PR #2295, opened 2026-09-17
Separate a run's input staging, result copy-back and binding release

### LINKED ISSUE #2291: [Code Health] Separate a run's input staging, result copy-back and binding release (Track A prerequisite)
### Category

### Component

Host Runtime

### Description

`validate_runtime_impl(Runtime *, const HostApi *, int execution_rc)` is one
entry doing three things with three different lifetimes:

1. reading the run's device-side status (on the failure path);
2. copying this run's outputs back through its `TensorLease` bindings;
3. releasing those bindings.

And the inputs go the other way: `copy_in_device_args`
(`tensormap_and_ringbuffer/host/runtime_maker.cpp:514`, called from the bind at
`:844`) copies the caller's input **bytes** into the device argument slice while
resolving the arguments, so a run's input staging happens during preparation
rather than at its own submission.

Three consequences on main today:

- **A control fact is communicated by mutating data.** The two never-launched
  paths tell the status read to skip its device access by writing
  `set_gm_sm_ptr(nullptr)` into the caller's `Runtime`
  (`c_api_shared.cpp:796`, `:1155`). So a run that fails before launch mutates
  the image it was about to publish, and the read side cannot be `const`.
  "Never launched" is a property of the submission, not of the runtime image.
- **Release is welded to copy-back**, so a run cannot have its results read and
  its device bindings retired at different times. #2267's partial-submission
  cases need exactly that: *kernel submitted, event missing → retain
  device-referenced resources until an independent quiescence proof*. With one
  entry there is no seam to retain them at.
- **Input staging belongs to preparation.** Under #1853's queueing a successor
  is prepared while its predecessor executes, and "each invocation prepares its
  own inputs and submits its execution context once" is the agreed shape. The
  bind is the wrong place for the H2D of a run's input bytes; recording *which*
  buffer each argument uses is the bind's job, and staging the bytes is the
  submission's.

This is the in-scope remainder of the withdrawn reusable-`PreparedCall`
milestone (#2270), carved down to what serves Pipeline A. Nothing here retains
a prepared-call store, reuse controls, a cross-invocation preparation cache or
any interface for resubmitting a prepared context — those are withdrawn
requirements.

### Location

Source reference: `44dccbef8` (main after #2283). Recheck current main before
implementation.

- `src/{a2a3,a5}/runtime/tensormap_and_ringbuffer/host/runtime_maker.cpp`:
  `copy_in_device_args` (`:514`), the bind (`:844`), `validate_runtime_impl`
  (`:926`), `read_runtime_status`, `read_published_run_status`.
- `src/{a2a3,a5}/runtime/host_build_graph/host/runtime_maker.cpp`:
  `validate_runtime_impl` (a2a3 `:1435`, a5 `:2090`), and the
  `set_gm_sm_ptr(nullptr)` at a2a3 `:1334` / a5 `:1989`.
- `src/common/platform/onboard/host/c_api_shared.cpp`:
  `cleanup_failed_prepare` (`:790`) and `simpler_finalize_run` (`:1155`).
- `src/common/platform/sim/host/c_api_shared.cpp`: the same two call sites.
- `src/common/worker/runtime_c_api.h`: the runtime-impl symbol list.
- `src/common/utils/tensor_lease.h`: `TensorLease`, `TensorReleaseKind`.
- Existing tests: `tests/ut/cpp/common/test_trb_runtime_temp_buffer.cpp`
  (compiles the real `runtime_maker.cpp` against a fake `HostApiOps`, host-only,
  per arch).

### Proposed Fix

Three `extern "C"` entries in place of one, provided by both runtimes:

```c
int copy_in_run_inputs_impl(const Runtime *, const HostApi *);
int copy_back_run_outputs_impl(const Runtime *, const HostApi *, int execution_rc, int launched);
int release_run_bindings_impl(Runtime *, const HostApi *);
```

- `copy_in_device_args` stops copying bytes and records `needs_copy_in` on the
  lease alongside the existing `needs_copy_back`. `copy_in_run_inputs_impl`
  performs the per-run H2D and returns the first failing rc.
  `host_build_graph`'s is a documented no-op: its host orchestrator reads the
  input tensors while it builds the graph, so its inputs are staged inside its
  own bind.
- `launched` becomes an argument. No path writes `set_gm_sm_ptr(nullptr)`, and
  the read side (`read_runtime_status`, `arm_collectors_for_run`,
  `init_args_dump`, the `get_workers()` accessors) takes `const Runtime *`.
- Release stays adjacent to copy-back at today's call sites. **This issue
  changes no normal-path behavior** — its deliverable is the seam plus the
  removal of the data mutation, not a reordering. Deferred release under
  partial submission belongs to #2267.

Internal boundary, both sides in this repo, so no version field.

#### Acceptance criteria

Each names an observation, and what the wrong value would mean:

1. **A failed prepare leaves the caller's runtime image alone.** After a prepare
   that fails before launch, the caller's `Runtime` compares equal to what it
   held before — in particular `gm_sm_ptr()` is unchanged. A changed value
   means the control fact is still travelling as a data mutation.
2. **No device status read without a launch.** With `launched == 0`, the fake
   `HostApi`'s device-read count against the shared-memory region is **0**. A
   nonzero count means the skip is still inferred from a nulled pointer rather
   than from the argument.
3. **Copy-back does not retire bindings.** After `copy_back_run_outputs_impl`,
   the fake ops' free count is **0** and the lease vector is unchanged; after
   `release_run_bindings_impl` it equals the number of `Free`-kind leases. A
   nonzero count at the first check means release is still welded to copy-back.
4. **Input bytes are staged per execution, not at bind.** The bind's H2D count
   against lease destinations is **0**, and `copy_in_run_inputs_impl` issues
   exactly one H2D per `needs_copy_in` lease. A nonzero bind count means input
   staging is still preparation-time work. `packed_temp_bytes` and the
   descriptor/scalar content the bind writes are unchanged by the move.
5. **Direction semantics are preserved.** An `IN` lease is copied in and not
   back; an `OUT` lease is copied back and not in; an `INOUT` lease is both.
   Asserted per direction rather than by a total count, so a flag inverted on
   one direction cannot pass.
6. **A failed input staging keeps the run's bindings.** If
   `copy_in_run_inputs_impl` fails, the leases are still present and releasable,
   and the failure is the rc the caller sees.
7. **Production regressions.** Full onboard a2a3 and a5 scene suites for both
   runtimes, both sim corpora, cpput, pyut, and the per-channel DFX lanes
   (`--enable-chip-swimlane`, `--dump-args`, `--enable-pmu`, `--enable-dep-gen`,
   `--enable-scope-stats`) — the `--dump-args` channel specifically, because it
   reads the argument slice this change moves the bytes into.

#### Scope boundaries

Not in this issue: #2267's completion fences and deferred release under partial
submission, #1853's admission widening, #2257's KernelArgs payload change, any
retained or reusable prepared result, and any change to native tensor access
policy.

ChaoWao:
## Summary

`validate_runtime_impl` did three things with three different lifetimes — read a run's device-side status, copy its outputs back through its `TensorLease` bindings, and release those bindings — while the inputs went the other way: the bind's argument walk copied the caller's bytes H2D while it was still deciding which device buffer each tensor gets.

Three entries in place of one, provided by both runtimes:

```c
int copy_in_run_inputs_impl(const Runtime *, const HostApi *);
int copy_back_run_outputs_impl(const Runtime *, const HostApi *, int execution_rc, int launched);
int release_run_bindings_impl(Runtime *, const HostApi *);
```

- **The bind names buffers; a separate step fills them.** `copy_in_device_args` records `needs_copy_in` on the lease and moves no bytes; `copy_in_run_inputs_impl` copies them under its own `[STRACE]` span, which this adapter calls after the bind returns — still inside `simpler_prepare_run`, since moving it to launch is #1853's business, not this seam's. `host_build_graph` keeps its copy inside its bind and implements the entry as a documented no-op — its orchestrator reads the input tensors while it builds the graph, and the graph that bind materializes carries the values it read, so bytes and graph belong to the same run either way.
- **An input with no usable endpoint fails staging.** A HOST IN/INOUT tensor may carry a null address with a nonzero size — `init_external` accepts it and the bind still gives it a real device slice — so skipping such a lease would report success while the kernel read whatever that slice already held. Only a lease carrying no input or no bytes is skipped.
- **`launched` is an argument.** Nothing writes `set_gm_sm_ptr(nullptr)` into the caller's runtime to make the status read skip its device access, so a run that fails before launch no longer mutates the image it was about to publish. The read side — `read_runtime_status`, `arm_collectors_for_run`, `init_args_dump` and the `get_workers()` accessors — takes a const runtime across both backends and both architectures.
- **Release is its own entry.** Reading a run's results and retiring the memory behind them are now separately orderable, which is what a partially submitted run needs: results readable, bindings retained until quiescence is proven (#2267). Both call sites still invoke them back to back, so no normal-path behavior changes; a failed prepare now releases its bindings without also running a copy-back it has nothing to copy.

Carved from the reusable-`PreparedCall` work in #2270, whose central requirement was withdrawn. No prepared-call store, reuse control, cross-invocation preparation cache or resubmission interface is retained.

## Testing

The new cpput cases assert the seam rather than the outcome, so each has a value that would show the old shape:

- the bind's H2D count is 1 (its arena image) and staging takes it to 2 — a bind that still moved tensor bytes would read 2 before staging;
- a null-source IN lease fails staging with no H2D attempted and its bindings still releasable, while the same missing address on a pure OUT lease is ignored — confirmed to fail without its fix on both architectures;
- the copy-back's free count is 0 with the ledger intact, and the release then frees exactly the owned allocation;
- an unlaunched copy-back performs no device read and leaves `gm_sm_ptr` alone;
- a failed staging leaves the run's bindings releasable;
- IN / OUT / child-memory directions are asserted per direction, not by a total count.

- [x] Simulation tests pass — a2a3sim 40/40, a5sim 36/36, plus the 12 per-channel DFX lanes (24 cases), including `--dump-args` on both arches, which reads the argument slice these bytes now arrive in by a different route
- [x] Unit tests pass — cpput 151/151, pyut 2405 passed / 50 skipped
- [x] Hardware tests pass — full onboard a2a3 corpus for both runtimes: 67 scene cases plus the two L2 groups (42 + 60 cases), zero failures. **a5 onboard was not run**: this box's silicon is a2a3, so its arch precheck refuses; CI's a5 lane covers it
- [x] `clang-format` and `clang-tidy` clean on every changed translation unit

Fixes #2291


ChaoWao:
Reviewed all 17 changed files at **40b98b215e4b901147e0ecdc2e9178a169488cf5**, against **44dccbef83690a86be598fd2353d0c2e105b6768** and #2291. The head is unchanged at posting.

**The split is the right bounded prerequisite for Pipeline A. Please fix the P2 below before merging; the P3 items are nonblocking cleanup and do not warrant another architectural review cycle.** No request to restore PreparedCall reuse, move staging to launch, or implement #2267's deferred retirement here.

## Spec

### P2 — A nonempty null-source input now silently succeeds instead of failing staging

Location: [`src/common/utils/tensor_lease_copy_in.h:35–36`](https://github.com/hw-native-sys/simpler/blob/40b98b215e4b901147e0ecdc2e9178a169488cf5/src/common/utils/tensor_lease_copy_in.h#L35).

For a HOST IN/INOUT tensor with nonzero size and `buffer.addr == nullptr`, TMR bind allocates a valid device slice and records `needs_copy_in=true`. The new helper skips that lease because `host_ptr == nullptr`, then returns success. Execution can therefore consume uninitialized or retained bytes instead of rejecting the input.

This is reachable through the native API: `ChipTensor::init_external` accepts the null address, and the argument/admission/bind path does not reject it. Previously the tensor copy invoked the production HostApi callback, whose [null-source check](https://github.com/hw-native-sys/simpler/blob/40b98b215e4b901147e0ecdc2e9178a169488cf5/src/common/platform/onboard/host/c_api_shared.cpp#L162) returned an error. The new skip bypasses that check.

**Fix:** skip only `!needs_copy_in` or zero-size transfers. For an input-bearing nonempty lease, reject null endpoints or let the existing HostApi callback return its error. Leave bindings intact for cleanup.

**Reproduction:** I added a temporary test using the real runtime maker and existing fake HostApi: construct a 64-byte HOST IN tensor with a null source, bind it, fill the resulting device slice with `0xab`, then stage. Both A2/A3 and A5 returned `0`, performed no additional H2D, and left the sentinel unchanged. An assertion requiring a staging error failed on both. An experimental null-endpoint rejection made both probes and the affected test targets pass. These are CPU tests of the real bind/staging code, not an onboard execution claim.

### P3 — Two HBG null-pointer control mutations remain

Locations: [A2/A3 `runtime_maker.cpp:1339–1342`](https://github.com/hw-native-sys/simpler/blob/40b98b215e4b901147e0ecdc2e9178a169488cf5/src/a2a3/runtime/host_build_graph/host/runtime_maker.cpp#L1339), [A5 `runtime_maker.cpp:1994–1997`](https://github.com/hw-native-sys/simpler/blob/40b98b215e4b901147e0ecdc2e9178a169488cf5/src/a5/runtime/host_build_graph/host/runtime_maker.cpp#L1994).

Both still call `runtime->set_gm_sm_ptr(nullptr)` with a comment explaining that it prevents a later error-status read. #2291 explicitly asks to remove that control signal; status reads are now gated by `launched`. Remove these two clears and their obsolete comments. Production constructs a fresh Runtime, so this is incomplete cleanup, not a demonstrated new execution regression. No request to roll back every bind write on failure.

### P3 — The specified read-side const conversion is incomplete

#2291 names `arm_collectors_for_run` and `init_args_dump`, but both still accept mutable `Runtime &` across onboard/sim and both architectures. Exact declaration sites:

- [A2/A3 onboard `device_runner.h:340`](https://github.com/hw-native-sys/simpler/blob/40b98b215e4b901147e0ecdc2e9178a169488cf5/src/a2a3/platform/onboard/host/device_runner.h#L340), and line 357.
- [A5 onboard `device_runner.h:271`](https://github.com/hw-native-sys/simpler/blob/40b98b215e4b901147e0ecdc2e9178a169488cf5/src/a5/platform/onboard/host/device_runner.h#L271), and line 307.
- [A2/A3 sim `device_runner.h:65`](https://github.com/hw-native-sys/simpler/blob/40b98b215e4b901147e0ecdc2e9178a169488cf5/src/a2a3/platform/sim/host/device_runner.h#L65), and line 68.
- [A5 sim `device_runner.h:66`](https://github.com/hw-native-sys/simpler/blob/40b98b215e4b901147e0ecdc2e9178a169488cf5/src/a5/platform/sim/host/device_runner.h#L66), and line 69.

Use `const Runtime &` in these declarations and matching definitions to finish the specified read-only boundary. The PR description's “whole read side” claim currently overstates completion.

## Standards

### P3 — The staging comment describes a different call boundary

[`tensor_lease_copy_in.h:22–26`](https://github.com/hw-native-sys/simpler/blob/40b98b215e4b901147e0ecdc2e9178a169488cf5/src/common/utils/tensor_lease_copy_in.h#L22) calls staging “a submission's work rather than its preparation's,” but the production calls remain inside `simpler_prepare_run`: [onboard:1015](https://github.com/hw-native-sys/simpler/blob/40b98b215e4b901147e0ecdc2e9178a169488cf5/src/common/platform/onboard/host/c_api_shared.cpp#L1015), [sim:809](https://github.com/hw-native-sys/simpler/blob/40b98b215e4b901147e0ecdc2e9178a169488cf5/src/common/platform/sim/host/c_api_shared.cpp#L809).

Under the repository's doc-consistency/comment rules, state the present contract: **binding records buffers; staging copies bytes; the current adapter stages during prepare.** Fix the comment, not the call order. Keeping today's order is appropriate for this seam-only PR.

## Validation and disposition

- Independently rebuilt and passed `test_trb_runtime_temp_buffer`, `test_a5_trb_runtime_temp_buffer`, and `test_hbg_bind_ledger` on the reviewed head before adding the counterexample.
- The added null-source counterexample failed on both TMR architectures, then passed with the experimental rejection described above. All experimental source edits were restored.
- Current CI includes successful A2/A3 and A5 onboard lanes. I did not independently rerun hardware tests.
- Three entries are wired through both runtimes and backends; HBG's bind-time input copy and no-op staging conform to #2291. Copy-back and release remain adjacent, and no withdrawn reuse interface/cache was introduced.

**Summary: Spec — one P2 and two nonblocking P3s. Standards — one nonblocking P3. Fix the null-input regression before merge; keep the follow-up scope small.**

ChaoWao:
@ChaoWao All four addressed at `e39579556`, no scope added: reuse stays out, staging stays in prepare, #2267's deferred retirement is untouched.

**P2 — null-source input now fails staging.** `copy_in_tensor_leases` skips only `!needs_copy_in` or zero-size leases; an input-bearing lease with a size and a null endpoint returns `PTO_RUNTIME_ERR_INTERNAL` before any callback, leaving the ledger and its buffers standing for the caller to release. Rejecting in the helper rather than letting the callback do it, because the cpput fake `copy_to_device` memcpy's unconditionally and the two backends' callbacks are not the only callers of this path.

Your reproduction is in the suite as two cases, so the boundary is pinned from both sides:

- `StagingRejectsAnInputWithNoHostSource` — 64-byte HOST IN tensor with a null address, bound, slice filled with `0xab`, then staged: expects a nonzero rc, an unchanged H2D count, the lease still present, and a clean release afterwards.
- `StagingIgnoresAnOutputWithNoHostSource` — the same tensor as pure OUT still stages to 0, since the kernel defines every byte it writes.

Confirmed the first fails without the fix: with the old skip restored it reported `FAILED`, and `StagingIgnoresAnOutputWithNoHostSource` stayed green, so the assertion is falsified by exactly the behaviour you found. 22/22 in both `test_trb_runtime_temp_buffer` and `test_a5_trb_runtime_temp_buffer`.

**P3 — HBG control clears removed.** Both `set_gm_sm_ptr(nullptr)` calls and their comments are gone; `git grep set_gm_sm_ptr\(nullptr\)` is now empty across `src/`.

**P3 — const read side finished.** `arm_collectors_for_run` and `init_args_dump` take `const Runtime &` in all eight declarations you listed and their matching definitions. That needed `const Handshake *get_workers() const` on the common `host_build_graph` runtime too — without it the sim TUs failed to build, which is also the answer to why the original claim was overstated: I had only const-qualified the TMR accessors.

**P3 — staging comment corrected.** `tensor_lease_copy_in.h` now states the present contract — the bind records buffers, this copies bytes, and the current adapter calls it during `simpler_prepare_run` after the bind. The two runtime-maker comments that said "at the point it owns them" and the onboard declaration block carry the same correction. Call order unchanged.

The PR description's "whole read side" claim is now true rather than trimmed; I rewrote that bullet to name the four functions instead.

Re-verified on the rebased head (now on `b776cd270`): full onboard a2a3 corpus 67 scene cases + 42/60 L2 groups with zero failures, cpput 151/151, pyut 2405 passed / 50 skipped, a2a3sim 40/40, a5sim 36/36, the 12 per-channel DFX sim lanes (24 cases). The onboard sweep ran on the pre-rebase head; #2266 is a5-only, so CI's lanes cover the rebase. a5 onboard is still unrun locally — this box's silicon refuses the precheck.

PR #2306, opened 2026-09-18
Remove extra TMR result-check task and use RTS fail-stop errors (K7)

Leaf-Salix:
# Fix: use RTS fail-stop reporting for TMR kernel errors (K7)

> Target: `hw-native-sys/simpler:feat/kernel-mode-integration-test` at `6cde5929`. Source branch: `Leaf-Salix/simpler:dev/kernel-rts-error-propagation` at `a11883c5` (pushed).

## Summary

- Remove the extra `simpler_aicpu_check_tmr_result` task from the caller stream and its Host handle, launch step, device export, and context-lifetime `KernelErrorRecord`.
- Set `ACL_STOP_ON_FAILURE` on the kernel mode's hidden AICPU and AICore streams when they are created. Keep the existing caller → hidden AICPU → hidden AICore → hidden AICPU → caller event ordering; do not add a launch-time sync or a replacement polling task.
- Keep the actual execution task, `simpler_aicpu_kernel_exec`: it returns native `0` on success and maps a nonzero internal result to CANN's native error status `2`. The separate `check_tmr_result` task is removed; deleting the formal task's error return would instead hide execution failure from RTS. The per-round status, cancellation, and finalization remain because the AICPU and AICore participants need them to finish or retire the same round safely.
- Add hardware ST assertions for caller and same-Context cross-stream error visibility, plus a two-thread caller handoff case. Repair the executor-inputs UT fixture for the current callable image/context registration contract.
- Do not change the public kernel ABI, prepare/registration timing, production submission policy, or the existing program-mode stream path.

## Why STOP is required here

The event chain establishes execution order, but its successful completion is not, by itself, a reliable transfer of a hidden AICPU task's error to the caller in eager mode. A production-path A/B test changed only the hidden-stream failure mode:

| Injected failure | With STOP | Without STOP |
| --- | --- | --- |
| Invalid invocation before the core window opens, eager | Caller sync reports an error | Caller sync returns `0`; device sync reports `507018` |
| Runtime failure after the core window opens, eager | Caller sync reports an error | Caller sync returns `0`; device sync reports `507018` |
| Invalid invocation, graph replay | Caller sync reports an error | Caller sync reports an error |
| Runtime failure, graph replay | Caller sync reports an error | Caller sync reports an error |

The no-STOP eager cases still retire their AICores. The failure is not lost on the device, but a caller that synchronizes only its stream can incorrectly observe success. The replay result does not make no-STOP safe for eager and does not prove that every graph failure follows the same reporting path. Keeping STOP uses RTS's built-in failure scheduling and reporting rather than another AICPU task. [CANN's API reference](https://www.hiascend.com/document/detail/zh/CANNCommunityEdition/900beta2/API/appdevgapi/aclcppdevg_03_0074.html) defines `ACL_CONTINUE_ON_FAILURE` as the default and `ACL_STOP_ON_FAILURE` as stopping subsequent tasks after a failure.

## Operational impact and diagnostics

- This is a **fail-stop Context policy**, not a per-call recoverable error. CANN 9.0 documents that on A2/A3, setting stop-on-failure on one stream also makes other streams in the same Context stop-on-failure. A TMR task failure can therefore terminate unrelated work sharing that Context. The serving owner must not treat the Context as reusable after such a failure; rebuilding or restarting it is outside this PR.
- The caller's RTS error (for example `507018`) establishes that execution failed; it does not identify the first faulty callable or the root cause by itself. Diagnosis must use the first Host/AICPU/RTS error and the run's isolated device logs. This PR does not claim a new per-invocation diagnostic API or guaranteed root-cause attribution from the caller error code.
- `KernelErrorRecord` and the check task are removed, but the per-round control/report status remains for admission, AICore cancellation, and finalization. This is device execution coordination, not another task that polls a completed result on the caller stream.
- No in-place recovery after a STOP fault is claimed. The injected-error ST cases terminate their isolated processes after checking core retirement; they do not call `finalize_device` on the failed Context. Close normally enqueues a revoke on the hidden AICPU stream, so same-process close after a STOP fault and resource reclamation are **not validated by this PR**. The owner must treat the Context as terminal and arrange external quiescence/rebuild; this PR does not define that recovery path.

## Validation

- On an isolated A2/A3 hardware build at the target feat revision, the four production error ST cases (`device_error_eager`, `runtime_error_eager`, `device_error_replay`, `runtime_error_replay`) pass with STOP. The same four tests, without STOP and with no other source change, fail only the two eager caller-error assertions.
- An observation-only run confirms that no-STOP eager returns caller sync `0`, core-retirement check `0`, and device sync `507018`. With STOP, the eager caller sync reports `507018`. No production assertion was weakened for the A/B run.
- On the earlier feat `e9ac1be2` validation snapshot (including #2311), the four directly related C++ targets (`test_kernel_launch_binder`, `test_tmr_kernel_execution_round`, and both architecture revoke tests) pass. The four A2/A3 hardware error ST cases, the `cold_synced`, `eager_batch`, and `prepare_in_capture` controls, and all 30 cases in the A2/A3 kernel-capture matrix passed in that validation run.
- The follow-up test-only commit fixes the executor-inputs UT fixture's stale `KernelExecutionRequest` initializer and updates its callable image/context registration. The A2/A3 and A5 host-only targets both build and pass **12/12** cases. Incremental pre-commit passes on both changed test files.
- In a later A2/A3 single-device run, each of the four error ST cases first proves an unrelated stream can record and synchronize an event, then verifies that caller sync reports the hidden-task error promptly, AICore retirement completes, and a new event on that unrelated stream is rejected. Eager caller/unrelated-stream operations return `507018`; graph replay returns `507011`. The first device error points to `simpler_aicpu_kernel_exec`, not the newly recorded event. The observed invocation contains one AICPU and one AICore launch, with no added result-check task.
- A separate two-thread, two-caller-stream ST verifies the current submission policy: while A's device round is in flight, B's cross-stream Host submission is rejected with `-1002` and enqueues no second round. After A completes correctly, B is admitted and its injected AICPU error reaches B's stream sync promptly (`507018`, about 4.5 ms); the formerly successful A stream then rejects a new event (`507018`). This does **not** demonstrate that two in-flight PyPTO calls are both admitted, nor does it exercise every RTS error category.

- Final rebase onto feat `6cde5929` (including #2327 and #2329) preserves the two K7 commits; the only K7 patch change is a one-line executor-inputs fixture correction found by rebuilding the UT. The exact final source was synchronized to an isolated A2/A3 hardware snapshot, with runtime and Python binding rebuilt. Nine related host-only C++ targets pass (9/9); the A2/A3 and A5 executor-inputs targets each pass 12/12 cases. The four eager/replay fault ST cases and the two-thread cross-stream fault ST each pass in a separate single-device job (5/5). Final child logs show eager caller/other-stream errors `507018`, graph replay errors `507011`, and B's failure making A's former caller stream reject new work (`507018`). Incremental pre-commit passes. #2329 changes tools/docs, not the TMR runtime or these ST paths; no dependency or Torch installation was performed.

**Intermittent UB fault:** the new #2311 DFX case failed once with an AICore `VEC instruction error: the ub address out of bounds`, then passed on reruns. A matching first device error had already occurred in `cold_synced` on pristine feat before both this PR and #2311. The later normal `cross_stream` control also failed once with an AICore UB out-of-bounds error (`507015`) before the new fault assertions and passed on a same-card rerun. This error class was not introduced by this PR; its root cause and whether #2311 changes its incidence remain undetermined. These intermittent failures are not counted as a fully green ST matrix.

## Scope of the PR

This PR removes the K7-specific caller-side result-check mechanism and relies on the actual AICPU execution task's native return plus RTS fail-stop behavior. It does not replace K7's device execution protocol, alter program mode, add a generic error framework, or promise that every RTS error is recoverable or self-diagnosing. A5 hardware, the repository-wide ST matrix, concurrent admission of two in-flight calls, and failed-Context close remain unverified.


Leaf-Salix:
## STOP 故障处理与恢复边界：封板结论

当前方案采用 `STOP_ON_FAILURE`：hidden stream 的执行错误升级为 RTS Context 级错误。eager 路径由调用方同步 caller stream 时观察错误；ACLGraph replay 路径通常在同步时返回 `507011`。

本阶段不要求 Simpler 在已经 fail-stop 的 Context 内继续执行 revoke、正常 close 或逐项资源释放。上层观察到错误后，应将当前执行进程判定为不可恢复，停止继续提交任务并退出该进程；进程退出后，由 CANN、驱动和操作系统完成 stream、HBM、AICPU 状态及 Host 资源的终结式回收。随后由 supervisor 拉起新进程，创建新的 RTS Context、Simpler runtime 和 context generation。

```text
Simpler hidden stream 执行失败
  → STOP_ON_FAILURE 使 RTS Context fail-stop
  → caller stream sync / ACLGraph sync 返回 CANN 错误
  → Simpler 将错误返回 PyPTO
  → PyPTO 将当前 Worker 进程标记为 fatal，停止新提交
  → 不在故障 Context 内执行 Simpler 正常资源回收
  → Worker 进程退出
  → 驱动完成终结式资源回收
  → 上层 supervisor 重启 Worker 进程
  → 新进程创建新 RTS Context 和 generation
```

这里的恢复单位是**拥有该 RTS Context 的进程**，而不是逻辑 Worker 对象。PyPTO 不应在原进程中捕获异常后继续执行，也不能只重建 Simpler 对象或调用 `Worker.close()` 后复用原 Context。PyPTO 算子执行层负责可靠向上传递错误，Worker 管理层将其升级为进程级 fatal，上层 supervisor 负责重启。

真机已验证连续故障进程使用不同 context generation 后，新进程仍可重新注册并执行；没有观察到每轮约 28.2 MB 的线性 HBM 残留。因此当前设计口径为：**Simpler 负责可靠报错，PyPTO 负责触发 Worker 进程退出，上层负责进程重建，资源由进程退出及驱动 teardown 兜底回收。**

Issue #2254, opened 2026-09-16
[Code Health] worker.run hand-copies the Runtime device image per run instead of letting KernelLaunch upload it


ChaoWao:
### Category

Technical Debt (cleanup, refactor)

### Component

Host Runtime

### Description

Every `worker.run` publishes its launch state to the device with two synchronous `rtMemcpy` calls on the prepare path, rather than passing it as a KernelLaunch argument. The AICPU side already does the latter — `launch_aicpu_kernel` hands `&kernel_args.args, sizeof(KernelArgs)` to `rtsLaunchCpuKernel`, which uploads it — so the two halves of the same struct reach the device by two different mechanisms.

**Sizes, measured by compiling the headers at `22385d2b0`:**

| runtime | per-run `Runtime` H2D | copy length is |
| ------- | --------------------- | -------------- |
| a2a3 tmr | 51,456 B | `sizeof(DeviceRuntimeLaunchDesc)` |
| a2a3 hbg | 51,584 B | `sizeof(Runtime)` |
| a5 tmr | 49,152 B | `sizeof(DeviceRuntimeLaunchDesc)` |
| a5 hbg | 56,192 B | `sizeof(Runtime)` |

Plus `sizeof(KernelArgs)` (96 B on a2a3, 88 B on a5) — copied **twice** per run whenever any DFX channel is on, because `arm_collectors_for_run` has to refresh the device copy after arming.

**What the a2a3-tmr 51,456 B actually contains:**

| component | bytes | share | real nature |
| --------- | ----- | ----- | ----------- |
| `orch_args_storage_` | 33,856 | 66% | `TaskArgsTpl<Tensor, uint64_t, 256, 128>` — fixed worst-case capacity. Live content is `tensor_count x 128 + scalar_count x 8`, typically under 2 KB |
| `func_id_to_addr_[1024]` | 8,192 | 16% | constant per callable; identical on every run |
| `workers[72]` + `teardown_gates[72]` | 9,216 | 18% | bidirectional AICPU<->AICore handshake region — device-written, and must not be overwritten by an upload at all |
| per-run scalar config | ~192 | **<1%** | `worker_count`, `aicpu_thread_num`, `ready_queue_shards`, `aicpu_allowed_cpus[16]`, `serial_orch_sched`, `gm_sm_ptr_`, `prebuilt_arena_base_`, `prebuilt_runtime_offset_`, `active_callable_id_` — the only genuinely per-run one-way data |

[...]

PR #2307, opened 2026-09-18
Support: time the runner's prepare_execution as its own host span

ChaoWao:
## What

`DeviceRunner::prepare_execution` now emits one `[STRACE]` span,
`chip.run.prepare_execution`, at depth 1 as a sibling of `chip.run.bind`.

Prerequisite for #2254, which states it directly:

> **The cost is currently unmeasurable.** `prepare_execution` carries neither a
> `HostPhaseKind` record nor a `STRACE` span, so neither `hbg-bind-phases` nor
> `strace_timing` attributes it. A span should land before any optimisation, or
> there is no before/after.

Step 1 of that issue has landed (#2297); steps 2–4 change what this function
does, so the baseline has to exist first.

**This is observability only. It claims no speedup and moves no bytes** — no
`Runtime` layout change, no H2D change, no handshake, error-decision, claim,
pipeline-capacity or collector-lifetime change, no new environment variable,
collection thread or timing framework.

## Exact timing boundary

The span brackets the `runner->prepare_execution(...)` call in each platform's
`simpler_prepare_run` — entry to return on **every** path, including each of
the function's early `return rc` failures and a throw, since `StraceScope` is
RAII and there is no manual end call to miss.

| In the span | Not in the span |
| ----------- | --------------- |
| device init, the device-wall and run-result regions, the AICore register tables, the AICPU topology probe, the orchestration-SO resolve, and the synchronous H2D of the `Runtime` device image | **before it:** native-run admission, the prepared-successor compatibility probe, `attach_current_thread`, resource provisioning, `prepare_launch_shape`, `apply_call_config`, `bind`, `stage_inputs` — **after it:** collector arming, the launch, the device run |

It is **host wall clock, not CPU time**: a blocking `rtMemcpy` and driver calls
run inside it, so any wait it contains is part of `dur`.

### How this differs from the existing "prepare" window

`strace_timing`'s `_PREPARE_SPAN` is **`chip.run.bind`** — `bind` alone, used as
a deliberately conservative proxy for preparation by `--assert-native-overlap`
(`docs/dfx/host-trace.md`). The new span is **disjoint from it**, covering a
later part of the same prepare, and must not be substituted for it. The overlap
assertion therefore keeps reading `bind`: widening it to both spans would weaken
the implication rather than strengthen it. The phased C API's own "prepare"
(`simpler_prepare_run` entry to return) is wider than either and still has no
span of its own — only the `chip.run` root, which spans prepare through
finalize.

## Why a span and not a `HostPhaseKind`

The issue offered either. `HostPhaseRecord` is not usable here, for two
independent reasons:

1. **Its pool's lifetime ends before `prepare_execution` starts.**
   `host_phase_trace_begin()` is called inside host_build_graph's
   `bind_callable_to_runtime` and `host_phase_trace_end()` runs on that
   function's RAII exit guard; `end()` clears `active`, drains in-flight
   records, calls `host_phase_pool_finish()` and nulls the pool pointer. A
   record raised from `prepare_execution` — which runs strictly after `bind` —
   would be **silently dropped**.
2. **The store is host_build_graph's alone.** `tensormap_and_ringbuffer` never
   begins a pass, while `prepare_execution` is platform-layer and runs for both
   runtimes.

A span reuses the mechanism `chip.run.bind`, `chip.run.validate` and
`chip.run.claim_release` already use, and when collection is off it costs the
same existing level check in `unified_log_host_span` — no new gate.

## Arch and runtime coverage

`src/common/platform/{onboard,sim}/host/c_api_shared.cpp` are each compiled into
**both** the a2a3 and the a5 platform library, and each is the only call site of
`prepare_execution` on its platform (`simpler_run` routes through
`simpler_prepare_run`). So two edits cover a2a3 + a5 × onboard + sim and cannot
drift apart. Verified in the built artifacts: the span string is present in all
four `libhost_runtime.so` variants.

Both runtimes emit it. For `tensormap_and_ringbuffer` it is the **only**
prepare-path span beyond `bind`'s two segments.

**Sim emits it too, and it means something different**: sim's host-side prepare
(simulated register blocks, orchestration-SO resolve, sim AICPU setup), with no
H2D and no driver call in it. A sim duration is not comparable to an onboard one
and is not a baseline for #2254. The docs say so.

## Sample output

`strace_timing --tree`, a2a3 onboard, cold first run:

```text
# tensormap_and_ringbuffer — tests/st/a2a3/tensormap_and_ringbuffer/mixed_example
callable hid=b888b1635a8b4d6f (prefill) — 1 invocation
run                                15073.0 us
  bind                                9805.9 us
    args                                 220.0 us
    prebuilt                            9476.7 us
  prepare_execution                    487.1 us
  runner_run                          3970.4 us
    device_wall            [dev]         307.5 us
      ...
  validate                             608.9 us
  claim_release                          1.7 us

# host_build_graph — tests/st/a2a3/host_build_graph/vector_example
callable hid=0 (prefill) — 1 invocation
run                                 6098.8 us
  bind                                 922.2 us
  prepare_execution                    375.7 us
  runner_run                          4693.2 us
    device_wall            [dev]         589.3 us
  validate                              16.4 us
  claim_release                          0.2 us
```

Raw marker:

```text
[STRACE] ... name=chip.run.prepare_execution ts=3001488372047075 dur=487085
```

These are single cold runs from two small scene tests, shown to demonstrate the
marker — not a benchmark.

## Tests

Assertion lands in the existing `native_run_lifecycle` scene test, which already
checks an exact per-invocation depth map over 5 (sim) / 11 (onboard)
invocations, including prepares that are abandoned without launching and three
predecessor/successor overlap pairs. It now requires **exactly one**
`chip.run.prepare_execution` per invocation, at depth 1, contained in the
`chip.run` root. No new pyut: `strace_timing` builds its tree from the dotted
name, so nothing in the parser changed.

Re-run in full after rebasing onto `6fd145369` (#2295), which reworked the same
prepare path and split out `stage_inputs`:

| Check | Result |
| ----- | ------ |
| `native_run_lifecycle` a2a3sim | 1 passed |
| `native_run_lifecycle` a2a3 onboard (`task-submit`, 11 invocations) | 1 passed |
| `concurrent_prepare_stress` a2a3 onboard (overlap assertion + its negative controls) | 5 passed |
| `mixed_example` a2a3 onboard (tensormap_and_ringbuffer) | 1 passed, 1 span |
| `vector_example` a2a3 onboard (host_build_graph) | 1 passed, 1 span |
| `vector_example` a5sim | 1 passed, 1 span |
| Collection off (`--log-level warn`) | 0 `[STRACE]` lines, run still passes |
| pyut `tests/ut/py` | 2405 passed, 50 skipped |
| cpput `tests/ut/cpp` | 152/152 passed |
| clang-format, clang-tidy, cpplint, ruff, pyright, markdownlint, repo lint hooks | clean |

**Not verified: a5 onboard.** This box is a2a3 silicon, so the a5 onboard build
is compile-verified only (span string confirmed in
`build/lib/a5/onboard/*/libhost_runtime.so`) and left to CI's a5 jobs, which
pass `st-onboard-a5` on this branch. Since the edited translation unit is shared
with the a2a3 onboard build that did run locally, the a5-specific risk is the
build, not the logic.

All onboard runs went through `task-submit --device auto --device-num 1` after
`onboard-arch-precheck`, with no other DFX channel enabled.

### One transient onboard failure, disclosed

The first post-rebase onboard attempt failed both `native_run_lifecycle` and
`concurrent_prepare_stress` with a **device-side** fault — `fftsplus aivector
error, core id 33, error code 0x4000000000000000` on device 9, then
`sched_error_code=100 SCHEDULER_TIMEOUT` and `finalize_native_run ... -100`.
Both passed on an immediate re-run of the identical tree (the numbers in the
table above are that re-run), and the lifecycle failure force-resets the card
inside the same `task-submit` lease, which plausibly explains the second test
failing behind the first.

I do not believe this change can cause it — the diff adds an RAII log scope
around a host call and cannot fault an AIV core — and the pre-rebase CI run on
this branch (`35308300006`) passed `st-onboard-a2a3`, which runs both of these
tests, with the identical span change. I am flagging it rather than calling it
unrelated. CI on the rebased head then passed `st-onboard-a2a3`,
`st-deepseek-onboard-a2a3`, `st-network1-onboard-a2a3` and `st-onboard-a5`, so
it has not recurred anywhere since.

## Docs

- `docs/dfx/host-trace.md` — the span in the tree and the depth-1 row, the
  in/out boundary table, the wall-clock and sim caveats, and a note that the
  overlap assertion still reads `bind` on purpose.
- `simpler_setup/tools/README.md` — the two host-stage enumerations that had
  become incomplete lists.

Those same two lists also gain **`chip.run.stage_inputs`**, the depth-1 span
#2295 added without documenting. It is not part of this change's subject, but
leaving it out would ship an enumeration of depth-1 spans that this diff edits
and that is still wrong.

Related: #2254 (parent), #2297 (step 1).


ChaoWao:
Fixed in 52369c316 — the tree label said `launch shape`, which `prepare_launch_shape` contradicts: it runs before the span opens, as the boundary table in the same section already lists. Label is now `register tables, topology probe, Runtime H2D`.

PR #2315, opened 2026-09-18
Name host_build_graph's device Runtime image as an explicit descriptor

### LINKED ISSUE #2312: [Code Health] Give host_build_graph an explicit device Runtime descriptor (subtask of #2254)
### Category

### Component

Host Runtime

### Description

Focused subtask of #2254, step 2, for `host_build_graph` only: give it the
explicit device descriptor `tensormap_and_ringbuffer` already has, so the
per-run upload carries the device-read fields and nothing else.

`runtime_device_copy_size()` returns `sizeof(Runtime)`
(`src/common/host_build_graph/shared/runtime.cpp:137`), and both shared copy
paths take that length verbatim — the per-run `rtMemcpy`
(`src/common/platform/onboard/host/device_runner_helpers.cpp:89`) and the
persistent-args H2D (`src/common/platform/onboard/host/kernel_persistent_args.cpp:38`).
Measured against this tree's headers:

| | `sizeof(Runtime)` | device-read fields | host-only tail |
| --- | --- | --- | --- |
| a2a3 hbg | 51,584 B | ~17,536 B | ~34,048 B (66%) |
| a5 hbg | 56,192 B | ~22,144 B | ~34,048 B (61%) |

The tail is host-only by construction, and each group was checked by accessor
name as well as field name, because a device reader reaching a host-only member
through an accessor is the one failure that compiles:

| member | status |
| --- | --- |
| `orch_args_storage_` (`simpler::hbg::EntryArgsStorage`, **33,856 B**) | Host-only. HBG orchestrates on the host; the only readers are `a2a3/.../host/runtime_maker.cpp:1340,1384` and `a5/.../host/runtime_maker.cpp:1995,2039`. The device-side `get_orch_args()` readers in the repo are trb's own (`a2a3/.../tensormap_and_ringbuffer/aicpu/aicpu_executor.cpp:595,602`), which is why trb keeps it in `dev` — the asymmetry is what makes this cheap for hbg |
| `active_callable_id_` | Host-only **for hbg**. Read and written only by the shared platform host (`device_runner_base.cpp:1134,1139,1604` onboard; `:555,560,754` sim). trb's AICPU does read its copy (`aicpu_executor.cpp:560,983`), so the accessors stay — the field just leaves the device image |
| `dev_orch_so_addr_`, `dev_orch_so_size_`, `device_orch_func_name_[]`, `device_orch_config_name_[]` | **Dead.** `set_dev_orch_so`, `set_device_orch_func_name` and `set_device_orch_config_name` have zero callers anywhere in the repo, and the trb `Runtime` does not declare them. The orch-SO plumbing that is live goes through `InitArgs` / `RegisterCallableArgs` (`{a2a3,a5}/platform/include/common/kernel_args.h:216-219`, written at `device_runner_base.cpp:1200-1204`, read on device only by trb) |
| `tensor_leases_` (`std::vector<TensorLease>`, 24 B) | Host-only ledger for the run's H2D/D2H. Its **control block is memcpy'd to the device today**, so the copied type is not trivially copyable and the upload is UB-adjacent — the split is what makes it well-defined (codestyle rule 8) |

Two comments are already wrong about this and are part of the defect:
`src/common/host_build_graph/runtime.h:195-200` claims the four dead fields "are
part of the platform↔runtime ABI and must stay", and
`src/common/host_build_graph/types.h:928` claims the executor path reads
`runtime->orch_args_storage_` — a stale copy of trb's comment.

### Location

Source reference: `6fd145369` (main after #2295).

- `src/common/host_build_graph/runtime.h`, `src/common/host_build_graph/shared/runtime.cpp` — the class, the ctor, the accessors, `runtime_device_copy_size`
- The model to mirror: `src/a2a3/runtime/tensormap_and_ringbuffer/runtime/runtime.h:132-200` (`struct alignas(64) DeviceRuntimeLaunchDesc`) with its `static_assert(sizeof(...) % 64 == 0)` at `:344-346`
- Device readers, a2a3: `runtime/host_build_graph/aicore/aicore_executor.cpp`, `aicpu/aicpu_executor.cpp`, `runtime/scheduler/scheduler_cold_path.cpp`, `runtime/scheduler/scheduler_dispatch.cpp`
- Device readers, a5: the same four plus `aicore/aicore_legacy_executor.cpp`, `aicpu/aicpu_legacy_executor.cpp`, `aicpu/aicore_lifecycle.cpp`, `aicpu/aicore_scheduler_state.h`
- Whole-image cache invalidation, three sites: `a2a3/.../aicpu/aicpu_executor.cpp:452`, `a5/.../aicpu/aicpu_executor.cpp:561`, `a5/.../aicpu/aicpu_legacy_executor.cpp:411`
- Tests that encode the current size: `tests/ut/cpp/common/test_kernel_persistent_args.cpp:504-509` and `:534-538`, `tests/ut/cpp/CMakeLists.txt:720-721`
- Docs: `docs/hardware/cache-coherency.md:138`

### Proposed Fix

`struct alignas(64) DeviceRuntimeLaunchDesc dev;` as `Runtime`'s first member,
holding `workers[]`, `teardown_gates[]`, `worker_count`, `aicpu_thread_num`,
`ready_queue_shards`, `aicpu_allowed_cpus[]`, `aicpu_allowed_cpu_count`,
`aicpu_launch_count`, `func_id_to_addr_[]`, `host_total_tasks`,
`sm_image_bytes`, `gm_sm_ptr_`, `prebuilt_arena_base_`,
`prebuilt_runtime_offset_`. `runtime_device_copy_size()` returns
`sizeof(DeviceRuntimeLaunchDesc)`.

The accessors keep their signatures — they are what lets the shared platform
layer compile against either runtime, and the shared AICPU entry reaches the
affinity gate only through them — so their bodies change and their callers do
not. Device code that reads fields directly moves to `runtime->dev.X`, trb's
spelling, so the boundary is visible at each use site. The three whole-image
invalidations become `sizeof(runtime->dev)`.

Not in this subtask, and left with #2254's owner: moving the config onto the
launch payload, retiring the per-run `rtMemcpy`, the handshake region's upload
status, and the PMU register table. This narrows *what* is copied — not when, by
whom, or into which block.

#### Acceptance criteria

Each names an observation and what the wrong value means:

1. **The uploaded image cannot carry host-only bytes.** A sentinel written into
   the tail does not appear in the first `runtime_device_copy_size()` bytes from
   `&rt`. Appearing means the partition leaks. This assertion fails on today's
   code, which is what makes it a discriminator.
2. **The copied type is trivially copyable and standard-layout**, by
   `static_assert` on the descriptor. Today's copied type is neither.
3. **The image starts at offset 0** — `static_assert(offsetof(Runtime, dev) == 0)`.
   A nonzero value means both copy sites need an offset they do not pass.
4. **The descriptor is a whole number of cache lines** —
   `static_assert(sizeof(DeviceRuntimeLaunchDesc) % 64 == 0)` — so
   `cache_invalidate_range(runtime, sizeof(runtime->dev))` cannot round into a
   neighbouring line.
5. **The per-run copy shrinks by the measured amount**: a2a3 51,584 → ~17,536 B,
   a5 56,192 → ~22,144 B. An unchanged number means the descriptor still holds
   the tail.
6. **No device reader loses its bytes**, evidenced by the full onboard
   `host_build_graph` suite on a2a3 and CI's `st-onboard-a5` lane — not by the
   sizes. Simulation cannot show this: it shares the `Runtime` by pointer
   (`src/common/platform/sim/host/device_runner_base.h:126`) and its
   `cache_invalidate_range` is a stub (`src/common/platform/sim/aicpu/cache_ops.cpp:15-21`),
   so a sim pass proves nothing about either the narrowing or the invalidation.
7. **Nothing about timing or ownership changed**: same per-slot `runtime_args`
   block, same upload point in prepare, same admission, and the same handshake
   initialization — including that the constructor does not zero
   `teardown_gates`, which matches trb and the A2/A3 protocol that resets the
   gate word before window-open.

### Priority

ChaoWao:
## Summary

Rebased onto main, and reshaped as an increment on top of **#2309** rather than a competing implementation of it.

#2309 already stopped this runtime uploading its host-only tail — `HostOnlyState host_` last, image `[0, offsetof(host_))`, three narrowed `cache_invalidate_range` sites. **That change delivered the saved bytes; this PR does not repeat them.** What it changes is how the boundary is expressed.

The device half becomes a type. `DeviceRuntimeLaunchDesc` holds what the AICPU and AICore read — handshake region and teardown gates, scheduling parameters and affinity-gate inputs, kernel dispatch table, `host_total_tasks`, `sm_image_bytes`, shared-memory and prebuilt-arena pointers — and `Runtime` is that descriptor followed by #2309's group:

```text
Runtime
├─ DeviceRuntimeLaunchDesc dev
└─ HostOnlyState host_
   ├─ orchestration args
   ├─ active callable id
   └─ tensor leases
```

Three things follow:

- **The image's length is `sizeof(DeviceRuntimeLaunchDesc)`**, not `offsetof` on a class that isn't standard-layout. Both the upload (`runtime_device_copy_size`, `device_image_bytes`) and all three AICPU invalidation sites derive from it.
- **A field crosses because of the struct it is declared in**, not because of what it happens to precede.
- **The `static_assert`s for standard-layout and trivial-copyability sit on the thing that is actually memcpy'd.** `Runtime` can never satisfy them while `HostOnlyState` holds a `std::vector`; the descriptor does.

`alignas(64)` plus a whole-cache-line size assertion is what lets `cache_invalidate_range(runtime, sizeof(runtime->dev))` never round into a neighbouring line — the same pairing TMR uses. After this, both runtimes describe their device image identically, which is one fewer difference for the shared platform layer to hide.

**#2309's encapsulation is kept as-is:** `HostOnlyState`, the `tensor_leases()` const/mutable pair, its grouping comment. Its two layout assertions are replaced by the pair that matters once the descriptor exists — `dev` first, and the host-only tail starting no earlier than the descriptor ends. `test_kernel_persistent_args.cpp` is untouched, so its per-variant callable-id assertion still reads TMR's uploaded image rather than being weakened to a host-object check.

## Testing

`TheDeviceImageCarriesNoHostOnlyBytes` checks the partition **by content, not by size**: it writes recognisable values into three host-only members (callable stamp, orchestration scalar, a lease's device pointer), searches the bytes the platform actually copies for them, then confirms three device-read fields *are* present so the exclusions aren't vacuous. A size assertion passes even when a member sits on the wrong side of the line; this one doesn't.

The P3 from the last round is fixed: the descriptor is read out of the byte buffer through an aligned object, not a `reinterpret_cast` over `std::vector<uint8_t>::data()`. I could not re-run your alignment sanitizer locally — this box has no `libubsan` (`cannot find -lubsan`) — but the cast is gone, so the class of error is removed by construction rather than by tuning.

- [x] Unit tests pass — cpput 152/152, pyut 2405 passed / 50 skipped
- [x] Simulation tests pass — a2a3sim 40/40, a5sim 36/36, plus the 12 per-channel DFX lanes. Simulation cannot speak to this change: it shares the `Runtime` by pointer and stubs out cache maintenance
- [x] Hardware tests pass — onboard a2a3, both runtimes: 67 scene cases plus the two L2 groups (42 + 60), zero failures
- [x] `clang-format` and `clang-tidy` clean
- [ ] **a5 onboard not run locally** — this box's silicon is a2a3 and its arch precheck refuses; a5's device readers rest on CI's `st-onboard-a5` lane

### A pre-existing red lane on main, found while running this

The `--runtime host_build_graph` onboard command (the one #2309 used) fails 5 `concurrent_prepare_stress` cases with `chip run lane is poisoned: finalize_native_run failed with code -100 … run_epoch=9`. Bisected by rebuilding and re-running the identical command at three trees, 4 devices each:

| tree | result |
| --- | --- |
| `8726d7e13` (#2309) | 49 passed, 1 skipped |
| `14918c37c` (#2313) | 1 failure, a different one (`worker_async_endpoint::test_run`, known load-sensitive) |
| `0a68792ed` (main tip, adds #2314) | the 5 failures, reproduced twice — with and without this commit |

That points at **#2314**. Single runs per tree, so a flake isn't fully excluded, but 5 correlated failures in one directory, twice, isn't flake-shaped. Worth noting separately: the default `pytest examples tests/st` sweep passes at main tip because its scene scheduler gives each level-2 class its own subprocess and device, while `--runtime` bypasses the scheduler and runs them in one process where a poisoned lane carries into the cases that follow — so sweep-shaped CI cannot catch this class of failure at all.

Fixes #2312


ChaoWao:
@ChaoWao Rebased and reshaped at `88c0d17e7`. You were right on both counts — the integration gap was mine to have caught, and #2315 as it stood would not have compiled against #2309 (its `runtime.h` had no `tensor_leases()`).

**Now an increment on #2309, not a replacement for it.** `HostOnlyState`, the `tensor_leases()` const/mutable pair and its grouping comment are kept verbatim; the change adds `DeviceRuntimeLaunchDesc dev` in front of it, exactly the structure you sketched. Upload length and all three invalidation sites derive from `sizeof` on the descriptor. #2309's two layout assertions are replaced by the pair that matters once the descriptor exists: `dev` first, and the host-only tail starting no earlier than the descriptor ends.

**The 34,048 B/run claim is out of the description** — that is #2309's benefit, and the body now says so in the first paragraph. What this PR claims instead: the image's length becomes a type's size rather than an `offsetof` on a non-standard-layout class, a field crosses because of the struct it is declared in, and the standard-layout / trivially-copyable assertions sit on the type that is actually memcpy'd rather than on one that can never satisfy them.

**P3 fixed.** The descriptor is read out of the byte buffer through an aligned local (`memcpy` into a `DeviceRuntimeLaunchDesc`), not a cast over `std::vector<uint8_t>::data()`. I could not reproduce your sanitizer run locally — this box has no `libubsan`, `ld: cannot find -lubsan` — so I am not claiming a sanitizer-verified fix; the cast is simply gone. The sibling cast in `test_kernel_persistent_args.cpp` is left alone and is sound: that fake allocates with `aligned_alloc(64, …)`.

**#2309's TMR assertion is preserved, not weakened.** `test_kernel_persistent_args.cpp` is untouched in this PR, so its `#if defined(SIMPLER_UT_TRB_RUNTIME)` branch still reads `get_active_callable_id()` out of the uploaded image for TMR, and only the hbg branch reads the host object.

Re-ran on the rebased head: cpput 152/152, pyut 2405 passed / 50 skipped, a2a3sim 40/40, a5sim 36/36, the 12 per-channel DFX lanes, onboard a2a3 both runtimes 67 scene cases + 42/60 L2 groups with zero failures. a5 onboard still cannot run here (arch precheck refuses), so its device readers rest on CI's `st-onboard-a5`.

---

One finding from running the HBG-restricted lane you asked for. It is red on main, independently of this PR:

```text
5 failed  tests/st/a2a3/host_build_graph/concurrent_prepare_stress/
          chip run lane is poisoned: finalize_native_run failed with code -100
          (run_id=0 slot=0 generation=1 dispatch_id=0 run_epoch=9)
```

Bisected by rebuilding and re-running the identical command at three trees, 4 devices each:

| tree | result |
| --- | --- |
| `8726d7e13` (#2309) | 49 passed, 1 skipped |
| `14918c37c` (#2313) | 1 failure, and a different one — `worker_async_endpoint::test_run`, known load-sensitive |
| `0a68792ed` (main tip, adds #2314) | the 5 above, reproduced twice: with this commit and with a clean main build |

That points at **#2314**. Single runs per tree, so I would not call a flake fully excluded, but 5 correlated failures in one directory, twice, is not flake-shaped.

Worth separating out: the default `pytest examples tests/st` sweep is **green** at main tip. Its scene scheduler gives each level-2 class its own subprocess and exclusively-locked device, while `--runtime` bypasses the scheduler and runs them in one process, where a poisoned lane carries into everything after it. So sweep-shaped CI cannot catch this class of failure at all — which is an argument for the restricted lane being part of the gate, not just an ad-hoc check.

Issue #2233, opened 2026-09-15
[Bug] a2a3 onboard: predicated clobber intermittently does not dispatch when its gate is 1 (max_diff is exactly POISON-SENTINEL)


ChaoWao:
### Platform

a2a3 (Ascend 910B/C hardware)

### Runtime Variant

tensormap_and_ringbuffer

### Description

`TestPredicatedDispatch::PredicateTrueDispatches` intermittently fails its golden comparison on `st-onboard-a2a3`, and the failing value identifies the mechanism exactly:

```
AssertionError: TestPredicatedDispatch::PredicateTrueDispatches:
  Golden mismatch on 'x': max_diff=957.0, rtol=0, atol=0
```

`957.0 == 999.0 - 42.0 == POISON - SENTINEL`. In this case the gate is 1, so the
clobber task (`X = 999.0`) is supposed to dispatch on top of the producer's
`42.0`. `x[0]` held **42.0**, so the producer ran and **the gated clobber did
not dispatch** — the predicate behaved as if the gate were 0.

This is not a stale or uninitialised buffer: `INIT_VAL` is `-1.0`, and 42.0 is
the legitimate producer output, so the graph ran and only the predicated task
was missing. The comparison is exact (`rtol=0, atol=0`), so the diff is not
numerical drift either.

**It is intermittent.** Observed once, at `inv=11` in an xdist worker (`gw3`), so
the case had already passed ten invocations in that same process before failing.

**Not reproducible on demand.** Re-running the same CI job on the same commit
passed. Locally, on real a2a3 silicon with the same binaries and the same
job-level watchdog env the workflow sets
(`SIMPLER_SCHEDULER_TIMEOUT_MS=2000`, `SIMPLER_OP_EXECUTE_TIMEOUT_US=3000000`,
`SIMPLER_STREAM_SYNC_TIMEOUT_MS=4000`), three consecutive runs passed.

### Steps to Reproduce

```bash
task-submit --device auto --device-num 1 --run \
  ".venv/bin/python -m pytest \
     tests/st/a2a3/tensormap_and_ringbuffer/predicated_dispatch/test_predicated_dispatch.py \
     --platform a2a3 --device \$TASK_DEVICE --exclude-level 4 \
     --require-pto-isa --pto-session-timeout 600 --manual exclude"
```

Intermittent; one observed occurrence. It surfaced inside the full
`st-onboard-a2a3` sweep, under xdist, after ten passing invocations of the same
[...]

PR #2317, opened 2026-09-18
Support: add a standalone AICore FIN-ordering test tool

ChaoWao:
## Summary

Adds `tools/cann-examples/aicore-fin-ordering/` — three standalone programs
(host launcher, AICPU consumer, AICore producer) that test whether a consumer
which has observed an AICore task's **FIN notification** can then read data that
task published.

Built while chasing #2233, where a dispatch predicate intermittently reads the
pre-producer value. No dependency on this repo's runtime, so the ordering
question can be asked without the scheduler in the picture.

## Why not extend `aicore-notification-perf`

That tool measures notification **latency**, so its producer must run free. This
one tests ordering **correctness**, which needs the opposite: the payload has to
be frozen across the consumer's whole read sequence. With a free-running
producer, a mismatch cannot be told apart from "the producer moved on" — and
that is not hypothetical, it is exactly what made an earlier measurement of this
question uninterpretable. The perf tool's own README names copying the directory
as the way to add a mechanism; this is that copy.

## What makes a result interpretable here

- **Frozen handoff.** The producer writes the payload exactly once per round and
  holds it until the consumer bumps `round_request`, which the consumer does only
  after finishing every read and diagnostic. A first-load value *below* the round
  the FIN carried therefore has one explanation.
- **Three-way classification.** `equal` / `older` / `newer`, with `older` the only
  failure. `newer` cannot happen under the frozen protocol, so it is reported as a
  protocol / foreign-writer anomaly rather than folded into a failure count.
- **The wait returns the MMIO word that ended it,** and the round is derived from
  that value. Re-reading COND afterwards would let the producer advance in
  between, so "expected" would name a different round.
- **No 64 B maintenance block is written by both sides.** `ctl` is AICPU-written /
  AICore-read; `rpt` and `pay` are AICore-written / AICPU-read. Each block is only
  cleaned by its writer and only invalidated by its reader.
- **Delay control arms.** Two arms spend the measured cost of the operations they
  control for and do nothing else, so "this operation ordered something" can be
  separated from "this operation took time".

## Matrix

Two producer publish sequences (with and without a barrier after the output
flush) × eight consumer read sequences (bare, `dsb sy`, invalidate, both,
`dmb ld`, acquire-load on the flag, and the two delay controls) × a cache-state
axis that optionally leaves the payload line resident with pre-publish content.

## Two constraints the scaffolding must satisfy

Recorded in the README and enforced by the build, because each costs a debugging
round-trip to rediscover:

- `rtRegisterAllKernel` needs a **linked device ELF carrying both halves of the
  mix-kernel symbol pair** (`*_0_mix_aic` / `*_0_mix_aiv`); a raw `ccec -c`
  object, or one holding only the cube half, is rejected with `107000`. The
  vector half must not include `kernel_operator.h`, whose non-weak globals would
  then be defined twice and break the combined link.
- **`get_coreid()` is not a bare core index.** It must be masked with `0x0FFF`
  before indexing the AIC_CTRL window, as `get_physical_core_id()` does. The raw
  value (observed `0x8010`) addresses outside the window, and reading it faults
  the AICPU op — surfacing as `507018 / aicpu exception`, not as a failed match.

## Testing

- [x] Builds: all three CMake projects, aarch64 host + `ccec` + `ld.lld`.
- [x] Runs on a2a3 silicon under a `task-submit` device lock: 1600 rounds across
      the full matrix, **0 wait timeouts and 0 `newer` anomalies**, so the frozen
      protocol holds for every round.
- [x] `markdownlint-cli2`, `clang-format`, `check_headers.py`,
      `check_english_only.py` clean.
- [ ] No CI coverage — `tools/` is developer tooling, not built or run by any
      workflow, and this needs a2a3 silicon plus a device lock.

## Not in this PR

- The **existing `aicore-notification-perf` tool does not currently build or run
  on this CANN** (9.0.0): it hits the two constraints above plus stale `tikcfw`
  include paths and a `libascend_hal` link directory that moved. Its COND subtest
  additionally discards any sample whose timestamp looks implausible, so it
  reports "no valid samples" rather than surfacing the ordering effect. Fixing it
  is a separate change I have not verified end to end.
- The #2233 runtime fix itself.


ChaoWao:
CI is green. The one red check was `st-sim-a5 (ubuntu)` on `TestSyncStartEarlyLocalOwner::single_scheduler_idle`; it passed on `rerun --failed` of the same commit, and this PR cannot reach it — the diff adds eight files under `tools/cann-examples/aicore-fin-ordering/` and nothing in the build or test system reads that directory. The only `tools/cann-examples/` paths any workflow builds are `query`, `aicpu-device-query` and `aicpu-kernel-launch`, named explicitly in `_ut-npu-a5.yml` / `_ut-npu-a2a3.yml`, and both `ut-a5` and `ut-a2a3` passed.

Filed as #2320 rather than waved off, because the captured stderr is specific: 227 `write_range_to_device` rejections with a field 38.7 MiB past the end of a 270 KiB shm mapping, at a regular 160-byte stride — a wrong base pointer walked as an array, not a boundary slip. The macos job in that run reports `fail` but its conclusion is `cancelled` (matrix fail-fast), so only one job genuinely failed.

PR #2321, opened 2026-09-18
Add: attribute dep_gen records to their run, and keep buffers it cannot publish

ChaoWao:
## Why

`dep_gen` cannot say which run a record came from, and it loses buffers out of its pool. Both are prerequisites for continuous collection, not follow-ups to it:

- **Attribution comes from the clearing today.** `begin_run()` drops the previous run's records, so "whatever is in the collector" is by construction one run's graph. Remove that clearing and nothing in the wire format says which run a record belongs to.
- **A pool that shrinks once per run cannot outlive per-run clearing.** An idle run cost one buffer, permanently.

This is the dep_gen instance of the shape already fixed for swimlane in #2313, plus the identity half that #2293 did for scope_stats and swimlane. Two of five collectors now have it; `args_dump` and `pmu` follow, one PR each.

## Identity

`DepGenBuffer`'s header carries `run_epoch` and `local_seq` in padding it already had — `count` was followed by `_pad0[15]` to isolate its cache line — so **the buffer does not grow** and `offsetof(records) == 64` is unchanged, with a new assert pinning the two field offsets.

Identity is in the buffer header rather than the record because `DepGenRecord` is **fully packed**: I probed the layout, and `kernel_id[3]` (624) plus `block_num` (636) fill the gap before `tensors[]` at 640. A per-record epoch would have grown the record 4736 → 4800 and cost 64 B × 1024 records per buffer. The header is the right place regardless — a buffer holds exactly one run's records by construction.

AICPU stamps at both acquisition points: the engine's `on_pop_success` hook, and `init`'s reuse path. Both precede the first record. No barrier is needed between the stamp and the acquisition, and the comment says why rather than claiming a fence: the AICPU thread that acquired the buffer is its only reader until the ready queue publishes it. (This is where AICore differs and why swimlane's prime had to stamp before publishing the pointer — dep_gen has no such second reader.)

## The host copy owns the identity

`append_buffer_records` reads the epoch and groups by it. This is the load-bearing half: the device buffer goes back to the pool and a later run **re-stamps it in place**, so a host copy that pointed back at the buffer would report the wrong run. One of the negative controls below is exactly that mistake.

`records()` is replaced by two accessors:

- `runs()` — every run held, keyed by epoch, each run's records contiguous.
- `window_records(&epoch)` — the records this window should emit. One run → that run's. **No run → an empty span**, because a run that submitted nothing has an empty graph, not a missing one, and the replay writer accepts `num_records == 0` to emit exactly that; suppressing the file would make "nothing submitted" indistinguishable from "collection failed". Several runs → `nullptr`, the only case that cannot be answered, since one path names one graph.

A run present in `runs()` always has at least one record (`append_buffer_records` never creates an entry for an empty buffer), so the one-run case can never masquerade as the no-run one.

This keeps the `extern "C" dep_gen_replay_emit_deps_json` signature untouched — it still takes a contiguous `DepGenRecord*`, which matters because it is a weak, cross-layer symbol where a signature drift would not produce a compile error at the call site.

## Ownership

A buffer is released only by a **successful** enqueue. AICPU consumes the free queue and never produces into it, so a buffer it cannot hand over — nothing to publish, or the ready queue full — has to stay the instance's, and the next run's `init` reuses it in place. Re-stamping on reuse is what makes that safe.

Two paths were losing buffers, and only one of them was where I first expected:

- `init` popped a replacement **over** a still-set `current_buf_ptr`. dep_gen's flush already returned early on an empty buffer, leaving the pointer set — so the idle-run leak was entirely at `init`, not at the flush.
- the flush-failure path cleared the pointer after charging `dropped`.

`reconcile_counters()` needs no change, and I checked this rather than assuming it: it only faults a retained buffer that **still has records**, and both retaining paths leave `count` at zero. An idle run stays clean; a failed enqueue still fails the gate via `dropped_record_count > 0`, which is correct — that graph is incomplete.

## Tests

Seven cases in `tests/ut/cpp/common/test_dep_gen_run_identity.cpp`, a new target that links the real host collector **and** the real AICPU producer, so the whole chain is production code: the collector allocates and seeds the pool, the producer acquires/records/flushes, and the collector's own `on_buffer_collected` copies what was published. Nothing in the test writes `run_epoch`.

| Case | Property |
| --- | --- |
| `EachRunsRecordsCarryItsOwnEpoch` | records reach the host under their own run, not the previous one's |
| `TheHostCopyKeepsItsEpochAfterTheBufferIsRestamped` | cycles the FIFO pool until the first run's storage is reacquired and re-stamped; the host copy still says the old epoch |
| `IdleRunsDoNotConsumeABufferEach` | pool depth flat across four idle runs, same buffer retained, re-stamped each time, nothing published |
| `AFailedFlushChargesDroppedOnceAndKeepsTheBuffer` | dropped charged by the real count, buffer kept, count zeroed, not pushed to the free queue, reusable next run |
| `APublishedBufferIsNotAlsoRetainedByTheDevice` | success still releases; the pool does not re-hand a buffer the host has not returned |
| `AZeroRecordRunStillYieldsAnEmptyGraph` | a run that submitted nothing gets an empty span and epoch 0, so its deps.json is still written |
| `WindowRecordsRefusesOnlyWhenSeveralRunsShareAWindow` | two runs refuse; one run resolves |

Plus, in `test_dep_gen_replay`, `ZeroRecordsEmitsAnEmptyGraph` — the output side of the same property: `dep_gen_replay_emit_deps_json(nullptr, 0, path)` returns 0, writes the file, and the file has `tasks` and `edges` sections. That contract is what the empty-span path relies on, so it is pinned rather than assumed.

**Negative controls**, one defect reinstated at a time (`.docs/arms/nc_dep_gen.sh`, restoring from saved copies so it cannot eat uncommitted work):

| --- | --- |
| no stamp in the engine's pop hook | 4 |
| `init` drops retention | `IdleRuns…`, `AFailedFlush…` |
| `init` reuses without re-stamping | `IdleRuns…`, `AFailedFlush…` |
| failed flush clears the pointer | `AFailedFlush…` |
| successful flush retains | `TheHostCopyKeeps…`, `APublishedBuffer…` |
| host copy ignores `buf->run_epoch` | 3 |
| `window_records` guesses when several | `WindowRecordsRefusesOnly…` |
| `window_records` refuses a zero-record run (the reviewed regression, verbatim) | `AZeroRecordRunStillYieldsAnEmptyGraph` |

Each discriminates; none turns the suite green or red wholesale, and the script verifies the tree is byte-identical afterwards.

## Results

- **cpput 153/153** (clean reconfigure + rebuild); the two dep_gen targets are 7/7 and 2/2.
- **All four platforms build** — a2a3, a2a3sim, a5, a5sim, 8/8 runtime builds, no errors.
- **st a2a3sim 40 passed / 8 skipped; st a5sim 39 passed.**
- **The dep_gen DFX channel with `--enable-dep-gen`: 1/1 on a2a3sim and 1/1 on a5sim.** This is the gate that matters here — `test_dep_gen_chain` asserts `deps.json` **exists** and validates its edge count, so it proves the emit path through `window_records()` really produced the file rather than declining.
- **pyut 2404 passed / 50 skipped / 1 failed.** The failure is `TestChipStartupFailure::test_chip_init_hang_trips_deadline`, and I am naming it rather than calling pyut green. It is a wall-clock child-reap budget — `RuntimeError: child chip pid ... still alive; shm not freed` at `worker.py:3806`, a single `waitpid(WNOHANG)` against a forked child — and it **passes standalone in 0.19 s** on the same tree. The full run was at `load average: 52.16, 63.99, 75.41` with six other tenants holding all 16 devices and two sibling worktree sessions running their own pytest. This diff contains no Python and touches nothing in the Worker startup or shm-release path. Same family as the previously recorded `TestNextLevelStartupFailure` reap-budget failures, in a different class — so it is the budget, not the class.
- **No onboard verification of this commit.** The sim dep_gen case exercises the same shared AICPU and host code, but silicon is CI's to confirm.

Refs #1853


ChaoWao:
P2 fixed — you were right, and it was a real regression against baseline.

Confirmed the mechanism before fixing: `dep_gen_replay_emit_deps_json` guards with `if (num_records > 0 && records == nullptr)`, so `num_records == 0` with a null pointer is explicitly permitted and baseline wrote a valid empty graph. My `sole_run_records()` returned `nullptr` for both "no run" and "several runs", so a zero-record window took the refusal path and produced no file at all.

The accessor is now `window_records()` and distinguishes the three cases:

- one run → that run's records;
- **no run → an empty span, epoch 0** (baseline behaviour restored);
- several runs → `nullptr`, the only case that genuinely cannot be answered, since one path names one graph.

Renamed because `sole_run_records` would have been a lie for the zero-run case. Also documented why one run can never masquerade as none: `append_buffer_records` never creates a map entry for an empty buffer, so a key implies at least one record.

Two regression tests, one per side of the property:

- `AZeroRecordRunStillYieldsAnEmptyGraph` — a real run acquires a buffer, records nothing, flushes, publishes nothing; `window_records()` must return a non-null empty span with epoch 0.
- `ZeroRecordsEmitsAnEmptyGraph` in `test_dep_gen_replay` — the output side: `emit(nullptr, 0, path)` returns 0, the file exists, and it has `tasks` and `edges`. That pins the emitter contract the empty-span path depends on, rather than assuming it.

**Negative-controlled against your exact finding**: restoring `if (records_by_run_.size() != 1) return nullptr;` fails `AZeroRecordRunStillYieldsAnEmptyGraph` and nothing else.

Re-verified on the amended commit: cpput 153/153 (dep_gen targets 7/7 and 2/2) · four platforms build 8/8 · st a2a3sim 40p/8s · a5sim 39p · dep_gen DFX channel with `--enable-dep-gen` 1/1 on both sims. Still no onboard verification of this commit.

PR #2324, opened 2026-09-18
Fix: wait for a kernel's output flush before its task reports FIN

### LINKED ISSUE #2233: [Bug] a2a3 onboard: predicated clobber intermittently does not dispatch when its gate is 1 (max_diff is exactly POISON-SENTINEL)

ChaoWao:
## Summary

Fixes #2233.

A kernel's output `dcci` issues a write-back; it does not wait for one.
`execute_task()` returns and the task writes FIN, so a consumer that observes
FIN can read that output **before the flush has landed**. The predicated-dispatch
scheduler reads a gate value exactly that way, which is why the gate could read
as its pre-producer content and a task whose predicate held was retired instead
of dispatched — the reported `max_diff=957.0` is `POISON - SENTINEL`, i.e. the
clobber never ran.

`OUT_OF_ORDER_STORE_BARRIER()` already sits between the kernel's return and the
FIN write. On onboard it was a no-op, on the premise that `dcci` alone carries
coherency; it now waits for the flush. pto-isa documents the pair for this chip
family (`pto/npu/a2a3/grid_intrinsic.hpp`), and this runtime already pairs `dsb`
with its own flushes where it publishes completion data
(`runtime/async_kernel_api.h`).

## The consumer side needs no change

Worth stating because it is where I first looked. The scheduler's completion
poll already follows its COND read with `rmb()`
(`scheduler_completion.cpp`), with a comment naming the exact
Device-to-Normal reorder hazard, and that is the **only** path that observes
completion — the other COND read only builds a stall diagnostic string. So the
read side is already ordered, and more strongly than necessary.

## Testing

Reproducer: the predicated-dispatch case replicated to 33 invocations per
process (`CASES` list), which is the axis the original report sat on (`inv=11`).
That turns roughly 1-in-489 into near-certain. 20 runs per arm on a2a3 silicon
under a `task-submit` device lock:

| arm | red runs |
| --- | -------- |
| baseline (`upstream/main`) | **18 / 20**, every one `max_diff=957.0` |
| with the flush wait | **0 / 20** |
| with the flush wait, repeat | **0 / 20** |
| with the flush wait, on the exact shipped binary | **0 / 20** |

- [x] Full a2a3 `tensormap_and_ringbuffer` suite: 44 L2 cases + 4 L3 cases pass.
- [x] `examples/a2a3/tensormap_and_ringbuffer/benchmark_bgemm` passes.
- [x] a5 runtime builds.

Corroborated independently by `tools/cann-examples/aicore-fin-ordering` (added in
#2317): with a producer that does not wait for its flush, **every** consumer read
variant fails 41–50 of 50, including ones that add ordering or cache maintenance
on the read side. Read-side treatment cannot rescue data that has not been
flushed.

## Two gaps, both deliberate

- **a5 has no reproducer, though CI now exercises it.** The macro was identically
  a no-op on a5, so the same latent defect was there and rule 10 in
  `codestyle.md` asks for the sibling in the same commit. This box is a2a3 only
  (`onboard-arch-precheck` for a5 exits 2), so I could not run a5 locally — but
  `st-onboard-a5` passes on this PR (8m39s), which is real a5 silicon saying the
  change does not break anything there. What is still missing is a **failing**
  a5 case: nothing demonstrates the bug on a5, so the a5 half rests on the code
  being identical rather than on a measured before/after.
- **The per-task cost is not measured.** This adds one `dsb` per task
  completion, on the dispatch tail. `tools/benchmark_rounds.sh` cannot run in
  this environment — its corpus names cases that no longer exist at those paths
  (`examples/.../alternating_matmul_add`, `paged_attention_unroll`), so all nine
  of its cases fail before doing any work, on this branch and on `main` alike.
  I did not want to report a number I could not produce. If the cost turns out
  to matter, the narrower shape is to wait only when the completing task has a
  consumer that reads its output, rather than on every completion.

## Note for whoever rebuilds this

`build_runtimes` does **not** treat
`src/{arch}/platform/onboard/aicore/inner_kernel.h` as a dependency: editing it
and rebuilding leaves `aicore_kernel.o` byte-identical. `rm -rf
build/cache/<arch>/onboard` first. That cost me a whole A/B — the two arms ran
the same binary and reported "no effect" — so it is worth knowing before
measuring anything that touches a platform header.


ChaoWao:
Following up on the "cost unmeasured" note in this PR's description. The cost is
now measured, and the note's stated reason was wrong — see the retraction on
\#2332: `benchmark_rounds.sh` was fine, my editable install was stale.

**Result: the barrier costs on the order of 10 ns per task, and is below 0.1% of
end-to-end wall on every workload in the corpus. On 8 of 9 cases it is not
distinguishable from run-to-run drift.**

## Setup

a2a3 onboard, one card under `task-submit`, `benchmark_rounds.sh -n 100`, full
TMR corpus. Arms differ only in `src/a2a3/platform/onboard/aicore/inner_kernel.h`:

| arm | `OUT_OF_ORDER_STORE_BARRIER()` | `aicore_kernel.o` md5 |
| --- | --- | --- |
| A, A2 | `dsb(DSB_DDR)` (this PR) | `b9ba5d90` |
| B | `((void)0)` (pre-PR) | `ca0e4eae` |

Each arm wiped `build/cache/a2a3/onboard` before building, because
`build_runtimes` does not track this header (#2331) — without the wipe both arms
run the same binary and report no difference, which is how I previously measured
a fake 18/20-vs-17/20 null. The two md5s differ, and A2's reproduces A's
byte-for-byte, so the arms are genuinely distinct and the build is deterministic.

A2 is a **repeat of arm A** run after B, as a drift control. It is what makes
the rest of this readable.

## Device wall, mean of 100 rounds (us)

| case | A | A2 | B | A1−A2 *(same binary)* | mean(A)−B |
| --- | --- | --- | --- | --- | --- |
| alternating_matmul_add C1 | 814.9 | 812.4 | 814.2 | +2.5 ±7.7 | −0.5 ±6.6 |
| benchmark_bgemm C0 | 764.1 | 762.6 | 762.6 | +1.4 ±7.5 | +0.7 ±6.3 |
| paged_attention_unroll C1 | 1127.0 | 1123.3 | 1126.9 | +3.8 ±8.7 | −1.7 ±7.6 |
| paged_attention_unroll C2 | 608.4 | 604.6 | 608.8 | +3.8 ±6.9 | −2.3 ±6.4 |
| pa_unroll_manual_scope C1 | 1118.3 | 1119.6 | 1117.7 | −1.2 ±8.4 | +1.3 ±7.2 |
| pa_unroll_manual_scope C2 | 598.1 | 592.5 | 590.4 | +5.6 ±6.2 | +4.9 ±5.5 |
| batch_paged_attention C1 | 3562.0 | 3561.5 | 3570.2 | +0.5 ±19.2 | −8.5 ±18.0 |
| sliding_window_deps Dense16 | 16803.6 | 16804.1 | 16789.9 | −0.4 ±16.0 | **+14.0 ±12.3** |
| qwen3_14b_decode Stress | 34810.6 | 34841.0 | 34791.9 | **−30.4 ±11.2** | +33.9 ±10.5 |

±  is a 95% interval on the difference of means (Welch).

## Reading it

Without A2, two cases looked like real effects: qwen3 at +33.9 ±10.5 and
pa_unroll_manual_scope C2 at +4.9 ±5.5. **The same-binary control accounts for
both.** A vs A2 on qwen3 differ by −30.4 ±11.2 — the same magnitude as the
claimed effect, in the opposite direction, between two runs of an identical
`b9ba5d90`. Same for pa_unroll_manual_scope C2 (+5.6 same-binary vs +4.9
across arms). Those were drift.

`sliding_window_deps Dense16` is the one case that survives the control (+14.0
±12.3 across arms, −0.4 ±16.0 same-binary), and it is also the case built to
show this: `sliding_window_orch.cpp:76-103` submits exactly one AIV task per
step in a strict `i → i-1` chain, so `steps=1000` puts **1001 tasks on the
critical path**. That gives

```text
+14.0 us / 1001 tasks  =  ~14 ns per task     (95% CI ~2 - 26 ns)
```

and the hot path takes up to two of these per task (`aicore_executor.cpp:43`
after the kernel, and `:196` on the early-dispatch args fill), so a single
`dsb(DSB_DDR)` here is order **7-14 ns**. I had estimated 0.1-0.5 us from the
swimlane record write-back figure; that prior was off by roughly 50x, which is
why nothing showed up anywhere else.

## Caveats

- The delta is the **whole macro change**, not an isolated `dsb`. Besides the two
  hot sites it also covers the one-shot startup handshake (`:81`) and four
  `scheduler_cold_path.cpp` sites that normal runs do not reach.
- Shared box. Devices busy at each arm: A 13/16, B 13/16, A2 9/16 — so B ran
  under *more* contention than A2 and still came out fastest on
  `sliding_window_deps`, which biases against, not toward, the measured cost.
- 9 cases, one comparison each; at alpha=0.05 roughly one false positive is
  expected, and the control shows that is what the two hits were.
- a2a3 only. This box's `onboard-arch-precheck` does not pass a5, so the
  identical a5 change is unmeasured.

## Conclusion for the record

The barrier is required for correctness (#2233), so the question was never
whether to keep it but whether it needed to be narrowed to tasks with outputs.
At order 10 ns per task it does not — narrowing would add a branch on the same
path to save something this corpus cannot resolve.

PR #2325, opened 2026-09-18
Add: attribute args_dump records to their run, and keep buffers it cannot publish

ChaoWao:
## Why

`args_dump` cannot say which run an arg came from, and it loses buffers out of its pool. Both are prerequisites for continuous collection, not follow-ups: today `begin_run()` drops the previous run's records, so "whatever is in the collector" is one run's dump by construction. Remove that clearing and nothing in the wire format says which run an arg belongs to — and a pool that shrinks once per idle run cannot outlive the clearing either.

Third of five collectors. #2293 did scope_stats and swimlane identity, #2313 swimlane ownership, #2321 dep_gen. `pmu` is last.

## Identity

`DumpMetaBuffer` carries `run_epoch` and `local_seq` in the **60-byte alignment tail that already followed `count`**, so the buffer does not grow — it is 32 KB per dump thread, and `records[]` stays at offset 0. Both facts are pinned by asserts rather than left to a reader.

AICPU stamps at both acquisition points — the engine's `on_pop_success` hook and `init`'s reuse path — each before the first record lands. No barrier is needed between the stamp and the acquisition, and the comment says why instead of claiming a fence: the AICPU thread that acquired the buffer is its only reader until the ready queue publishes it.

**The arena needs no identity of its own**, which I checked rather than assumed. `arena_write_offset` is a monotonic cursor the host reads modulo `arena_size` and **deliberately never resets** — `begin_run` says so in a comment and resets only the payload counters beside it. So a record's `payload_offset` is already unique across runs, and its run is whichever run owns the record. Nothing to add.

## The host copy owns the identity

`process_dump_buffer` reads the epoch **once, before the record loop**, and copies it into every `DumpedArg`. This is the load-bearing half: the device buffer goes back to the pool and a later run re-stamps it in place, so a host copy that pointed back at the buffer would report the wrong run. One negative control below is exactly that mistake.

The manifest emits `run_epoch` per arg. Additive — `dump_viewer.py`, `core_swimlane.py` and the scene tests read keys they know, so a new one does not disturb them.

## Ownership

A buffer is released only by a **successful** enqueue. AICPU consumes the free queue and never produces into it, so a buffer it cannot hand over — an idle run, or a full ready queue — has to stay the thread's, and the next run's `init` reuses it in place. Previously `init` popped a replacement over the retained pointer and the flush-failure path cleared it, so the host never saw that buffer and nothing recycled it.

Worth stating because it reframes the whole series: **the shared engine already had this rule right.** `DeviceProfilerEngine::switch_buffer` charges dropped, zeroes the count and returns *without* clearing `current_ptr` on enqueue failure, clearing it only on success. The per-collector run-end flushes were the deviation, in all of swimlane, dep_gen and args_dump. This change makes args_dump's agree.

## Tests

Four cases in `test_args_dump_run_identity`, built for a2a3 and a5 (matching the existing `test_args_dump_collector` split, since `PLATFORM_MAX_AICPU_THREADS` differs). The target links the real host collector **and** the real AICPU producer, so the chain is production code end to end: pool seeding, acquire, `dump_arg_record`, flush, `on_buffer_collected`, and `export_dump_files`'s manifest writer. Nothing in the test writes `run_epoch`.

| Case | Property |
| --- | --- |
| `EachRunsArgsCarryItsOwnEpoch` | the manifest shows this run's epoch and not the cleared run's |
| `TheHostCopyKeepsItsEpochAfterTheBufferIsRestamped` | cycles the FIFO pool until the first run's storage is reacquired and re-stamped; every run in the window is still present under its own epoch |
| `IdleRunsDoNotConsumeABufferEach` | pool depth flat across four idle runs, same buffer retained and re-stamped each time, nothing published |
| `APublishedBufferIsNotAlsoRetainedByTheDevice` | success still releases, and the pool does not re-hand a buffer the host has not returned |

Every record is a SCALAR with `capture_payload = 0`, so the arena stays out of it — consistent with the arena needing no identity.

Each case closes its window with `export_dump_files()`. That matters for a reason worth stating: `on_buffer_collected` starts the payload writer thread **unconditionally**, and `export_dump_files()` is what stops it. `finalize()` carries a fallback join for "any path that skips export", and the first version of the last case reached `finalize()` through that fallback — which **hung the whole binary in CI on both runners** (ctest killed it at 300 s; CI's captured output shows the case entering `[ RUN ]` and never printing a result, with the other three passing) while passing here nine times over, including under CI's exact `ctest -LE requires_hardware -j4 --timeout 300`. I could not establish the mechanism: the writer loop's wait predicate does cover `writer_done_`, and `begin_run()` does not clear `writer_started_`, so neither a lost wakeup nor a skipped join explains it. Since that teardown path is untouched by this PR, the test now closes through the supported one, and the unexplained hang is logged separately rather than papered over — its production trigger (a run whose export is skipped after a device error) is real, and nothing in the tree exercises the fallback now.

**Negative controls** (`.docs/arms/nc_args_dump.sh`, restoring from saved copies so it cannot eat uncommitted work):

| --- | --- |
| no stamp in the engine's pop hook | 3 |
| `init` drops retention | `IdleRuns…` |
| `init` reuses without re-stamping | `IdleRuns…` |
| successful flush retains | `TheHostCopyKeeps…`, `APublishedBuffer…` |
| host copy ignores `buf->run_epoch` | `EachRunsArgs…`, `TheHostCopyKeeps…` |
| manifest omits `run_epoch` | `EachRunsArgs…`, `TheHostCopyKeeps…` |

Each discriminates; none turns the suite green or red wholesale, and the script verifies byte-identity afterwards.

### One branch is not covered, and I am not implying it is

The **run-end flush's enqueue-failure branch** has no test here. `dump_args_flush` validates its own thread index before building the engine context, so the out-of-range injection that reaches `enqueue_ready`'s guard in the swimlane and dep_gen suites cannot get in — the flush returns first. The only other route is a genuinely full ready queue, which means sitting in the backpressure gate for `PLATFORM_DFX_BACKPRESSURE_TIMEOUT_CYCLES`; the host stub's `get_sys_cnt_aicpu` scales a real `steady_clock`, so that is **30 seconds of wall clock**, which a unit test has no business spending.

What keeps the gap narrow: the identical rule is the engine's on the hot path, so every mid-run rotation that fails already exercises it for this collector. Only the run-end branch rests on review. I did not change `dump_args_flush`'s validation to make it testable — that would be a production change driven by a test, and it is the maintainer's call whether the four collectors should agree on where the index is checked.

## Results

- **cpput 157/157**; the two new targets are 4/4 each.
- **All four platforms build** — a2a3, a2a3sim, a5, a5sim, 8/8 runtime builds, no errors.
- **The args_dump scene test on a2a3sim with `--dump-args --manual include`: 1/1.** It is in the manual lane on sim, so a default sweep deselects it — running it explicitly is the only way to exercise this channel outside silicon. I then read the manifest it produced: **11 args, 0 missing `run_epoch`, all carrying epoch 1** — non-zero, so genuinely stamped rather than defaulted, through the whole chain from the AICPU stamp to the JSON.
- **st a2a3sim 39 passed / 8 skipped; st a5sim 39 passed.**
- **pyut 2405 passed / 50 skipped, 0 failed.** (The `test_chip_init_hang_trips_deadline` reap-budget flake noted on #2321 did not recur here; the box was quieter.)
- **No onboard verification of this commit.** Silicon is CI's to confirm.

Refs #1853


ChaoWao:
CI `ut` fix pushed — it was mine, and the diagnosis is only partial, so stating both.

**What failed:** `test_args_dump_run_identity_a5` timed out at ctest's 300 s on **both** ubuntu and macos. Not a flake — deterministic on both runners, while the a2a3 variant of the same source passed in 0.01 s.

**What CI's own output pins down.** All four cases run; the first three pass; `APublishedBufferIsNotAlsoRetainedByTheDevice` enters `[ RUN ]` and never prints a result. Of the four it is the only one that collects a buffer and then reaches `finalize()` **without** `export_dump_files()` — the other two export, and the idle one never collects so never starts a writer. `on_buffer_collected` calls `start_writer_thread_once()` unconditionally, and `finalize()` carries a fallback join for "any path that skips export", whose own comment warns an un-joined writer "trips std::terminate when the collector is destroyed or re-run". So my test was the only thing in the tree taking that path.

**What I could not establish: the mechanism.** Nine local attempts all pass and exit 0 — the a5 binary standalone (×3, ~80 ms), the single case via `--gtest_filter`, under `ctest -j8`, and the full suite in CI's exact form (`-LE requires_hardware -j4 --output-on-failure --timeout 300`, 157/157). Two candidate explanations were checked and neither holds: the writer loop's wait predicate *does* cover `writer_done_` (so not a lost wakeup), and `begin_run()` does *not* clear `writer_started_` (so the fallback join is not being skipped). I am not going to guess past that.

**Fix:** the case now closes its window with `export_dump_files()` like the other three — the supported path, and the one every real run takes. cpput 157/157 locally under CI's exact ctest invocation, both variants 4/4.

That deliberately stops exercising the fallback, so I logged the hang separately rather than treating a green PR as a resolution. The path it guards is real (a run whose export is skipped after a device error), and it is now untested — worth a look independently of this PR, and I did not widen this one to chase it.

No review threads or review bodies were outstanding; the only conversation comment is CodeRabbit's rate-limit notice.

ChaoWao:
Correcting my own claim here, since it was wrong and it is now measured.

I wrote that the hang was **not** a lost wakeup "because the writer loop's wait predicate does cover `writer_done_`". That reasoning is invalid. Covering the flag in the predicate is necessary but not sufficient — the flag has to *change* under the waiter's mutex. Both stop sites set it with a bare atomic store while holding nothing, so a stop can land between the waiter's predicate read and its registration on the condition variable, dropping the notification and blocking it on a condition that is already true.

Two consequences for what this PR claimed:

1. **Routing the test through `export_dump_files()` does not eliminate the hang.** Export has the same unsynchronized stop as finalize, so the fix changed which unsynchronized stopper runs, not whether the race exists. What it did legitimately achieve is shaping the test like production — export is the path every real run takes. The PR body overstated it beyond that.
2. **I still cannot attribute the CI timeout to this race.** Measured: with the bare store the export-skipping case hangs **4 of 300** runs; with the flag set under the mutex, **0 of 300**. So the race is real and closable — but ~1.3% locally against 2-of-2 in CI on two OSes means either the CI environment amplifies the window a great deal or there is a second factor. Plausible, not proven.

Fixed independently in #2338, which also closes an unsynchronized `write_queue_.empty()` / `.size()` read two lines away (a data race on a non-atomic container while the writer pops), routes both stops through one `request_writer_stop()`, and adds the export-skipping teardown case this PR stopped exercising.

On the coverage gap you flagged — the run-end enqueue-failure reuse branch — that remains untested and I have not closed it. The blocker is unchanged: `dump_args_flush` validates its own thread index before building the engine context, so the out-of-range injection used in the swimlane and dep_gen suites cannot reach it, and the only other route is a 30 s backpressure spin. Making it reachable means changing where that index is checked so the four collectors agree — a production change I will not make to serve a test without your call on it. Happy to do it as its own PR if you want the branch covered.

Issue #2206, opened 2026-09-12
[Bug] a5sim tensormap_and_ringbuffer chip_swimlane suite segfaults in the L2 native wait, serially, ~1 in 6


ChaoWao:
### Component

Host Runtime

### Runtime Variant

tensormap_and_ringbuffer

### Description

The five a5sim `chip_swimlane` cases, run as one **serial** session with the
swimlane and dep_gen channels on, segfault at a low rate inside the L2 native
wait. Observed once in six runs of the same command.

```text
Fatal Python error: Segmentation fault

  File "python/simpler/worker.py", line 11585 in _wait_run_handle
  File "python/simpler/worker.py", line 4354  in wait
  File "python/simpler/worker.py", line 11227 in run
  File "simpler_setup/scene_test.py", line 1925 in _run_and_validate_l2
```

`worker.py:11585` is `return chip_run.wait(...)`, so the fault is inside the
native wait beneath the binding.

**This is not #2121, though it is the closest relative.** That issue had the same
symptom and the same deepest frame, and its root cause was `AsyncWaitList`
reaching the AICPU with a residual `count` because hbg's forked `async_wait.h`
had lost `reset_for_reuse()`. The fix (#2124) restored it for hbg and noted the
tensormap_and_ringbuffer twin already had it — and a5 tmr does:
`src/a5/runtime/tensormap_and_ringbuffer/runtime/shared/runtime_init.cpp:258`
calls `sched->async_wait_list.reset_for_reuse()` from `SchedulerState::reset_for_reuse`.
So this crash has a different cause.

It also differs from #2121 in three ways that narrow the search:

| | #2121 (fixed) | this |
| --- | --- | --- |
| platform | a2a3sim | **a5sim** |
| runtime | host_build_graph | **tensormap_and_ringbuffer** |
| trigger | needed `-n 4` | **reproduces serially, `-p no:xdist`** |

Not #2182 either: that one is a2a3 **onboard**, faults during forked-worker
init rather than in the wait, and needs a full-suite run.

The two failing case names do rhyme — #2121's was
`TestSpmdSyncStartEarlyDispatchHbg` and this one is
`TestSyncStartEarlyLocalOwner` — so the sync-start early-dispatch machinery is
worth looking at first even though the known defect in it is handled on this side.

### Steps to Reproduce

```bash
python -m pytest tests/st/a5/tensormap_and_ringbuffer/dfx/chip_swimlane/ \
[...]

Issue #2320, opened 2026-09-18
[Bug] a5sim chip_swimlane: finalize_native_run returns -100 after 227 out-of-mapping write_range_to_device rejections


ChaoWao:
### Platform

a5sim (Ascend 950 simulation)

### Runtime Variant

tensormap_and_ringbuffer

### Description

`st-sim-a5 (ubuntu-latest)` failed once on
`TestSyncStartEarlyLocalOwner::test_run`, with the run's lane poisoned:

```text
RuntimeError: SceneTest case failed:
  TestSyncStartEarlyLocalOwner::single_scheduler_idle: chip run lane is poisoned:
  finalize_native_run failed with code -100
  (run_id=0 slot=0 generation=6 dispatch_id=0 run_epoch=6)
```

**It is a flake** — `gh run rerun --failed` on the same commit passed
(5m31s). Filing it rather than leaving it as an unexplained red check,
because the captured stderr carries a specific diagnostic rather than a
generic timeout.

### What the log actually shows

Before the lane is poisoned, `write_range_to_device` rejects **227** writes,
all with the same shape:

```text
[ERROR] write_range_to_device: [buffer_pool_manager.h:536]
  BufferPoolManager::write_range_to_device: field [0x55743022b34c, 0x55743022b350)
  outside shm [0x55742db3a870, 0x55742db7e0b0)
```

Three properties make this look like a wrong base pointer rather than an
overrun:

| observation | value |
| ----------- | ----- |
| shm region size | 276544 B (270 KiB) |
| first rejected field, distance past the mapping's end | **40555164 B (38.7 MiB)** |
| stride between consecutive rejected fields | **0xa0 (160 B)** |

38.7 MiB past the end is a different allocation, not a boundary slip. And a
regular 160-byte stride over 227 iterations is a loop walking a record array
— i.e. one array's base is out of the mapping and every element in it is
then rejected in turn.

### Steps to Reproduce

```bash
python -m pytest tests/st/a5/tensormap_and_ringbuffer/dfx/chip_swimlane/ \
    --platform a5sim --device 0-15 --exclude-level 4 --require-pto-isa \
    --pto-session-timeout 600 --manual exclude
```

Intermittent; one observed occurrence, and a re-run of the same commit
passed. The failing case is `single_scheduler_idle` inside
`test_sync_start_early_local_owner.py`, with chip_swimlane diagnostics
enabled.

### Expected Behavior

`finalize_native_run` completes and the lane is not poisoned.

[...]

Issue #2237, opened 2026-09-15
Bug: sparse chip-swimlane scheduler streams are reindexed and drop DAG stats


doraemonmj:
## What happened

The A5 sim chip-swimlane smoke failed intermittently in CI:

- Run/job: https://github.com/hw-native-sys/simpler/actions/runs/34858325769/job/104024656686
- Failure: `Fanout edges mismatch: printed=11, oracle=19`
- `deps.json` contained 24 unique structural edges and the independent oracle attributed 19 fanout edges.
- The perf artifact contained all 48 expected task rows, and the scheduler report observed all 48 Complete FINs.
- However, Part 5 reported only three scheduler threads and only 12 of the 20 logical tasks. Both gaps are exactly eight: `19 - 11 == 20 - 12 == 8`.

This is the same symptom family as #2191, but the new occurrence exposes a concrete sparse-stream ID handling bug.

## Root cause

A5 defaults to five AICPU threads: one orchestrator plus four scheduler threads.

PR #2127 changed scheduler phase serialization from the legacy positional array (which emitted empty arrays and therefore preserved thread-index holes) to the unified stream schema. The unified writer deliberately omits an empty stream while storing its original ID in `scheduler_id`:

https://github.com/hw-native-sys/simpler/blob/6940c8ccbc4d8ec7111e542a0ce4a4985a99a24a/src/common/platform/include/host/scheduler_profiling_json.h#L62-L69

The converter then appends the remaining streams densely and stores their metadata separately:

https://github.com/hw-native-sys/simpler/blob/6940c8ccbc4d8ec7111e542a0ce4a4985a99a24a/simpler_setup/tools/swimlane_converter.py#L364-L415

But `parse_scheduler_from_json_phases()` still uses `enumerate(phases_by_thread)` as the scheduler thread ID and ignores `scheduler_streams[*].scheduler_id`:

https://github.com/hw-native-sys/simpler/blob/e4bc6be0/simpler_setup/tools/sched_overhead_analysis.py#L229-L232

Therefore, for example, real scheduler IDs `[1, 2, 3]` are interpreted as `[0, 1, 2]`. Meanwhile, `core_to_thread` retains the real IDs. `compute_dag_stats_from_deps()` later looks up each real thread ID in the reindexed `threads` dict and silently skips a missing entry:

https://github.com/hw-native-sys/simpler/blob/e4bc6be0/simpler_setup/tools/sched_overhead_analysis.py#L151-L160

[...]

Issue #2246, opened 2026-09-15
[Bug] a5sim dep_gen intermittently collects zero of the device's records, so deps.json is dropped


ChaoWao:
### Platform

a5sim (Ascend 950 simulation)

### Runtime Variant

tensormap_and_ringbuffer

### Description

The dep_gen collector intermittently collects **none** of the records the device
wrote, so `reconcile_counters()` correctly refuses to emit `deps.json` and the
run loses its whole task DAG.

Observed once in 25 serial runs of the a5sim `chip_swimlane` suite on
`8414356a1` (case `TestChipSwimlane_aicpu_threads_2`):

```text
[WARN] reconcile_counters: [dep_gen_collector.cpp:260] dep_gen reconcile:
  record count mismatch (collected=0 + dropped=0 != device_total=5 + overflow=0,
  silent_loss=5)
```

**This is a total loss, not a partial one, and not a capacity problem:**

| counter | value | what it rules out |
| ------- | ----- | ----------------- |
| `collected` | 0 | `append_buffer_records()` never ran — `on_buffer_collected` never fired for any buffer |
| `dropped` | 0 | device did not drop (`free_queue` / `ready_queue` were fine) |
| `overflow` | 0 | no submit expanded past its slot |
| `device_total` | 5 | the device did record all 5 submits |

No `dep_gen reconcile: un-flushed buffer` error accompanied it, so
`current_buf_ptr` **was** clear — the device completed its hand-off. So the five
records were published by the device and collected by nobody.

### Hypothesis (not yet proven)

The loss sits between the device publishing a full buffer and the host's
collector shard appending it — a ready-queue hand-off or shutdown race, rather
than anything on the device side.

There is a matching asymmetry in the reconcile path: it inspects
`current_buf_ptr` for a buffer the device never flushed
(`dep_gen_collector.cpp:227-241`), but never re-checks the **ready queue** for a
buffer that was flushed and never collected. So the second case is visible only
as the arithmetic mismatch above and is never recovered, where the first one is
at least named precisely.

I have not confirmed the mechanism — `stop()` is documented as "drain queues,
join threads" (`dep_gen_collector.h:31`), so if that drain is sound the loss must
be elsewhere. Treating it as a hypothesis, not a finding.

### Steps to Reproduce

```bash
for i in $(seq 1 25); do
  python -m pytest tests/st/a5/tensormap_and_ringbuffer/dfx/chip_swimlane/ \
[...]

PR #2326, opened 2026-09-18
Fix: bind DFX memory context before publishing run configuration

### LINKED ISSUE #2220: [Bug] a5 onboard chip_swimlane: single_scheduler_idle intermittently publishes zero records, so no chip_swimlane_records.json is written
### Platform

a5 (Ascend 950 hardware)

### Runtime Variant

tensormap_and_ringbuffer

### Description

On the `st-onboard-a5` chip_swimlane DFX smoke, `TestSyncStartEarlyLocalOwner::single_scheduler_idle` intermittently produces **no swimlane records at all**, so `export_swimlane_json` takes its zero-record early return and never writes the file:

```
[WARN] export_swimlane_json: [chip_swimlane_collector.cpp:1062] Warning: No performance data to export.
[WARNING] [TestSyncStartEarlyLocalOwner_single_scheduler_idle] .../chip_swimlane_records.json not produced; skipping conversion
E  AssertionError: chip_swimlane_records.json missing under .../TestSyncStartEarlyLocalOwner_single_scheduler_idle_20260914_091139
```

That branch requires the collector to hold **nothing**: no AICore task records, no scheduler phase records, and no AICPU orchestrator phases. It is not one section being dropped — it is the whole run.

**It is a flake.** `gh run rerun --failed` on the same commit passed ([green a5 job](https://github.com/hw-native-sys/simpler/actions/runs/34794777367/job/103876842167)). Filing it rather than leaving an unexplained red check.

**The run itself was fine.** The case's own correctness assertions passed; only the artifact was missing. And **4 sibling cases in the same step, same process, same device passed** (`1 failed, 4 passed in 11.47s`), so the collector was not dead for the step — this one run published nothing while its neighbours published normally.

One lead ruled out: `single_scheduler_idle` uses `aicpu_thread_num: 2`, but so does `single_scheduler_pending_only` in the same file, and that one passed. Thread count alone does not separate them.

### Steps to Reproduce

```bash
  ".venv/bin/python -m pytest tests/st/a5/tensormap_and_ringbuffer/dfx/chip_swimlane/ \
     --platform a5 --device \$TASK_DEVICE -p no:xdist --pto-session-timeout 600 \
     --require-pto-isa --manual include --enable-chip-swimlane --enable-dep-gen"
```

Intermittent — one observed occurrence in the runs seen so far. The assertion is reached only at chip-swimlane level ≥ 3 (`test_sync_start_early_local_owner.py:129`).

### Expected Behavior

A run with `--enable-chip-swimlane` writes `chip_swimlane_records.json`, because the run executed tasks and therefore produced records.

### Actual Behavior

The collector holds zero records of every kind, `export_swimlane_json` returns `PTO_RUNTIME_ERR_INTERNAL` without creating the file, and the scene test fails on the missing artifact.

### Git Commit ID

8252775562dd9e62deb86b2edb4b32c4090aa620

### CANN Version

CI self-hosted a5 runner (`a5-npu-1`) — as installed there

### Driver Version

CI self-hosted a5 runner (`a5-npu-1`) — as installed there

### Host Platform

Linux (aarch64)

### Additional Context

**Failing run:** <https://github.com/hw-native-sys/simpler/actions/runs/34794777367/job/103825791593>
**Green rerun, same commit:** <https://github.com/hw-native-sys/simpler/actions/runs/34794777367/job/103876842167>

**Dedup — this is a fourth, distinct a5 chip_swimlane record-publication failure.** They share a theme but I could not show a shared cause, so this is filed separately rather than appended to any of them:

| | platform | runtime | symptom | mechanism identified? |
| #2206 | a5sim | tmr | segfault in the L2 native wait | no |
| #2191 | a5sim | tmr | records written, but fanout edges 0 where `deps.json` has 6 | no |
| #2181 | a5sim | host_build_graph | one partially stamped trace makes an explicit guard discard the whole Scheduler section | yes — `a5/runtime/host_build_graph/host/runtime_maker.cpp:515` |
| **this** | **a5 onboard** | tmr | **zero records of any kind**, no file written | no |

The nearest neighbour is #2191 — both are "a5 tmr chip_swimlane published less than it should" — but #2191 wrote the file with a missing dependency set, whereas here nothing was published at all, and #2191 is sim while this is hardware. #2181 is the only one of the four with a named mechanism, and it is host_build_graph, a different runtime.

Worth checking whether a5's record publication has a common ordering or visibility hazard behind all four before fixing any of them individually.

Related: #2206, #2191, #2181, #995

doraemonmj:
A collector shape rebuild can replace the A5 host shadow while BufferPoolManager still holds the previous addresses. `begin_run()` publishes configuration before `start()` refreshed that mapping, allowing a counter reset to overwrite `chip_swimlane_level`. Hardware instrumentation captured an 8-byte perf-counter write changing the level from 4 to 0; computation succeeded but no swimlane JSON was produced.

Bind the manager context in `ProfilerBase::set_memory_context()` and explicitly clear it after resource cleanup. `start()` only starts collection threads. This preserves the existing allocation/teardown order and the SVM/non-SVM copy behavior. Add regressions for both address-reuse directions, SVM reinitialization, incomplete initialization, and cleared contexts.

Validation:

- All five new regressions fail against the original header and pass with the fix.
- 155 C++ tests passed after rebasing onto current main.
- A5, a5sim, and a2a3sim DFX smokes cover chip_swimlane + dep_gen, dep_gen, PMU, args_dump, and scope_stats.
- A5 swimlane previously passed six consecutive rounds (30 test invocations) with the same binding fix.

- All applicable pre-commit checks passed, including clang-format, clang-tidy, and cpplint.

The separate sim sparse-thread-ID failure tracked in #2237 is outside this fix.

Fixes #2220

---

## Scope is wider than #2220 — the same misdirected write also explains #2206 and #2320

Added after review. The mechanism above is not specific to the `chip_swimlane_level` field: **where** the misdirected write lands is the only variable, and it produces three separately-filed symptoms.

`publish_run_config()`'s `reset_head` pushes exactly 8 bytes of zeros:

```cpp
head->total_record_count = 0;
head->dropped_record_count = 0;
(void)manager_.write_range_to_device(&head->total_record_count, 2 * sizeof(uint32_t));
```

Translated through the previous region's base, that 8-byte zero write lands at a wrong offset. `reset_head` runs once per core plus twice per AICPU thread, so a rebuild issues hundreds of them.

| where the write lands | observable outcome | filed as |
| --------------------- | ------------------ | -------- |
| outside the stale window | `write_range_to_device` refuses; log noise only | #2320 (227 rejections, then `finalize_native_run` rc `-100`) |
| on `chip_swimlane_level` | device keeps its old level, records nothing | **#2220** (the case this PR was written from) |
| on `free_queue.buffer_ptrs[]` | AICPU dereferences a half-formed pointer | #2206 |

The third row is the one worth spelling out, because it is a memory-corruption path rather than a lost-configuration path. An 8-byte zero write landing **4 bytes** into `buffer_ptrs[]` zeroes slot N's high word and slot N+1's low word. A captured a5sim fault shows exactly that:

```
buffer_ptrs[1] = 0x00000000b69f7bd0     low word only  (high word zeroed)
buffer_ptrs[2] = 0x0000aaaa00000000     high word only (low word zeroed)
     recombined: 0x0000aaaab69f7bd0     a well-formed pointer
```

`chip_swimlane_aicpu_init` then pops that slot and writes `buf->count = 0` through it (`chip_swimlane_collector_aicpu.cpp:383`) — SIGSEGV when the address is unmapped, and a later SIGABRT inside `malloc` when the damage lands on live heap instead. That also explains why #2206 presented as several different bugs: the crash site is wherever the damage is consumed, not where it was caused, which is why #2145 and #2225 were each excluded on the strength of a still-crashing probe.

### Measurement

`2e331c3fd` versus the same commit plus this PR, a5sim `chip_swimlane` suite:

| arm | runs | crashes |
| --- | ---- | ------- |
| baseline | 184 valid | 2 (`rc=134`) |
| baseline + this PR | 229 | 0 |

**Deliberately not claimed as proof.** Fisher's exact on those counts is p = 0.198 — not significant. The deductive match above is the evidence; the counts are only consistent with it. Reaching p < 0.05 needs roughly 500 clean runs on the fix arm.

### Why the issue references stay as they are

Only `Fixes #2220` uses a closing keyword. #2206 and #2320 are **referenced, not closed**, on purpose:

- #2206's second and third crash signatures (SIGSEGV inside `calloc`/`_dl_allocate_tls`, SIGABRT inside `malloc`) are consistent with an A-class wild write that landed on mapped heap, but that link is inference — the two recovered half-pointers are not themselves in this box's heap range, so those particular writes would fault rather than corrupt.
- The rate evidence is underpowered, as above.

So both should be verified against a post-merge base and closed by hand, rather than auto-closed here.

Related: #2206, #2320, #2335 (an independent drain-path accounting gap in the same files, not addressed here).


ChaoWao:
Corroborating evidence for this PR's mechanism, captured on a5sim while chasing
#2206 — and a suggestion that this PR's blast radius is wider than #2220.

The ordering this PR corrects is observable in a crash. `ProfilerBase::set_memory_context()`
sets the collector's `shm_host_` (`profiler_base.h:763`) while the manager's copy
is installed later from `start()` (`profiler_base.h:863`), which is what
`BufferPoolManager`'s own contract states (`buffer_pool_manager.h:331`). A batch
of the a5sim `chip_swimlane` suite caught a run where the two disagree, with the
consequence in the same log:

```text
[ERROR] write_range_to_device: [buffer_pool_manager.h:536]
  field [0xaaaabea808b0, 0xaaaabea808b4) outside shm [0xaaaabea808f0, 0xaaaabea80ef0)
  ... two more, 4 bytes each, consecutive, all below the manager's base

#0  chip_swimlane_aicpu_init (worker_count=24)
      at src/common/platform/shared/aicpu/chip_swimlane_collector_aicpu.cpp:383
#1  SchedulerContext::pre_handshake_init (aicpu_thread_num=5, sched_thread_num=4, ...)
#2  AicpuExecutor::init  →  #3 aicpu_execute
```

The three rejected writes are `DepGenCollector`'s counter reset, which computes
its target from the collector's `shm_host_` (`dep_gen_collector.cpp:173`) while
`write_range_to_device` bounds-checks the manager's base. `:383` is
`buf->count = 0` on a pointer read out of `free_queue.buffer_ptrs[]`.

Two things that may be worth folding into this PR, or worth a follow-up:

1. **Those three calls discard their result** —
   `(void)manager_.write_range_to_device(...)` at `dep_gen_collector.cpp:180-183`.
   With this PR's fix they should stop failing, but the `(void)` means a future
   recurrence is again invisible: the only trace is the manager's `LOG_ERROR`,
   which does not name the collector it was called for. Compare
   `profiler_base.h:621-632`, where the free-queue refill checks both writes,
   rolls back the slot and the tail, and returns false.
2. **`Fixes #2220` may be understating it.** The same disagreement plausibly
   explains #2206 (an abort inside `malloc` under
   `ChipSwimlaneCollector::export_swimlane_json`, plus a SIGSEGV in the submit
   path — heap corruption faulting wherever the next allocator touch lands),
   \#2320 (the same rejection shape on a5sim, 227 of them, in the same
   `TestSyncStartEarlyLocalOwner` class), and #2246 (a counter reset that never
   reached the device leaves dep_gen's totals wrong, and reconcile then refuses
   to emit `deps.json`). I am not asking you to claim those — only flagging that
   a batch on top of this branch would settle several reports at once.

Happy to run that batch against this branch if useful: the a5sim `chip_swimlane`
suite reproduces at **1.7%** (2 hits in 120 runs) at ~15 s per run, so ~120 runs
gives ~87% power. Two practical notes from doing it — use pytest `-q`, not `-s`
(the `-s` arms went 0/150 against 2/120 for `-q`), and read the stack from gdb
rather than a core: `systemd-coredump` on the dev box truncates at 2 GiB, which
drops the link map and leaves every frame unresolved.

Full evidence and the measured/inferred split are in
https://github.com/hw-native-sys/simpler/issues/2206#issuecomment-5734640016.

ChaoWao:
## This looks like it also fixes #2206 and #2320, not only #2220

Probed this PR against #2206 (the a5sim `chip_swimlane` segfault/abort) and the mechanism you describe appears to be that bug's root cause too. Full write-up in <https://github.com/hw-native-sys/simpler/issues/2206#issuecomment-5737842996>; the short version:

`publish_run_config`'s `reset_head` pushes exactly **8 bytes of zeros** (`2 * sizeof(uint32_t)` at `&head->total_record_count`). Landing 4 bytes into `free_queue.buffer_ptrs[]` it zeroes slot N's high word and slot N+1's low word. The fault dump I captured shows precisely that:

```
buffer_ptrs[1] = 0x00000000b69f7bd0     low word only
buffer_ptrs[2] = 0x0000aaaa00000000     high word only
      recombine: 0x0000aaaab69f7bd0
```

The AICPU then pops that half pointer at `chip_swimlane_collector_aicpu.cpp:383` and writes `buf->count = 0` through it — SIGSEGV when unmapped, and a later SIGABRT inside `malloc` when the damage lands on live heap instead. So the base-pointer disagreement has three outcomes depending only on where the misdirected write lands: refused (the 227 rejections in #2320), on `chip_swimlane_level` (#2220's zero records, which your hardware capture caught as 4 -> 0), or on `buffer_ptrs[]` (#2206).

### Measurement

Your head is 8 commits behind `main`, so I rebased it onto `2e331c3fd` to make this PR the only variable:

| arm | runs | crashes |
| --- | ---- | ------- |
| `2e331c3fd` | 200 | 4 |
| `2e331c3fd` + this PR | 200 | 0 |

**Not significant on its own** — Fisher's exact is p = 0.062 one-tailed — so I am reporting it as consistent rather than conclusive. 250 more runs on the fix arm are in flight (400 total reaches p = 0.012). The deductive match above is the stronger evidence.

### Suggestion

The PR body scopes itself to #2220 and says "Not the fix for #2206". On this evidence that scope note is too modest — worth widening it, or at least linking, so the three issues do not get re-investigated separately after this lands. I am not editing your PR; flagging it for you.

Nothing here asks for a change to the diff. The rebase-needed note stands on its own (8 commits behind), and it is the only thing I would actually action before merge.

PR #2328, opened 2026-09-18
Add: attribute PMU rows to their run, and keep buffers it cannot publish

ChaoWao:
## Why

PMU cannot say which run a counter row came from, and it loses buffers out of its pool. Both are prerequisites for continuous collection: `begin_run()` drops the previous run's state, so "whatever is in the CSV shards" is one run's data by construction. Remove that clearing and nothing says which run a row belongs to — and a pool that shrinks once per idle run cannot outlive the clearing either.

**Last of the five collectors.** #2293 did scope_stats and swimlane identity, #2313 swimlane ownership, #2321 dep_gen, #2325 args_dump.

## Identity

`PmuBuffer` carries `run_epoch` and `local_seq` in the **60-byte alignment tail that already followed `count`**, so the buffer does not grow — it is 32 KB per core. Asserts pin that and `records[]` at offset 0.

The two architectures keep separate copies of this collector and their records differ (8 vs 10 counters), so I probed both rather than assuming: `PmuRecord` lands on 64 bytes either way, so `PmuBuffer` is 32832 bytes with a 60-byte tail on both, and one edit serves both.

AICPU stamps at both acquisition points — the engine's `on_pop_success` hook and `init`'s reuse path — each before the first record lands. No barrier is needed between the stamp and the acquisition: the AICPU thread that acquired the buffer is its only reader until the ready queue publishes it.

## Identity reaches the output as a CSV column

PMU is the one collector with no host-side record container — it streams rows straight to per-shard CSV files. So the row writer reads the epoch **once before its loop**, for the usual reason: the device buffer goes back to the pool and a later run re-stamps it in place.

`run_epoch` is appended as the **last** column, after `event_type`:

```
thread_id,core_id,task_id,func_id,core_type,pmu_total_cycles,<counters…>,event_type,run_epoch
```

Appending rather than prepending, deliberately: the documented contract both scene tests assert is the **six-column leading prefix**, and the counter block already varies per event type, so nothing positional is disturbed. I checked the consumers — only the two `test_pmu.py` files parse `pmu.csv`, both by leading prefix, neither indexing from the end.

Both scene tests now also assert the trailing pair by name, that every row's column count matches the header, that no row carries `run_epoch=0` (i.e. identity was actually stamped rather than defaulted), and that one run yields one epoch.

## Ownership

A buffer is released only by a **successful** enqueue. AICPU consumes the free queue and never produces into it, so a buffer it cannot hand over — an idle core, or a full ready queue — has to stay the core's, and the next run's `init` reuses it in place. Previously `init` popped a replacement over the retained pointer and the flush-failure path cleared it, so the host never saw that buffer and nothing recycled it.

As with the previous three, this makes the run-end flush agree with the shared engine, whose `switch_buffer` already retained on failure and released only on success.

`reconcile_counters` needs no change on either arch — I checked rather than assumed: it only faults a retained buffer that **still has records**, and both retaining paths leave `count` at zero. Its comments described the old rule and now describe this one.

## Tests

Five cases in `test_pmu_run_identity`, built per architecture. The target links the real host collector **and** the real AICPU producer, so the pool protocol is production code end to end: seeding, acquire-or-reuse, flush, the collector's copy, and the CSV row writer.

| Case | Property |
| --- | --- |
| `EachRunsRowsCarryItsOwnEpoch` | every CSV row carries its run's epoch |
| `RowsKeepTheirEpochAfterTheBufferIsRestamped` | cycles the FIFO pool until the first run's storage is reacquired and re-stamped; the earlier rows still say the earlier epoch |
| `IdleRunsDoNotConsumeABufferEach` | pool depth flat across four idle runs, same buffer retained and re-stamped each time, nothing published |
| `AFailedFlushChargesDroppedOnceAndKeepsTheBuffer` | dropped charged by the real count, buffer kept, count zeroed, not pushed to the free queue, reusable next run |
| `APublishedBufferIsNotAlsoRetainedByTheDevice` | success still releases, and the pool does not re-hand a buffer the host has not returned |

The pool-cycle length is derived from the **measured** seeded depth rather than a constant, so it stays correct if the pool is resized.

**Record contents are a fixture, and that is a real boundary.** PMU records are not written by any host-callable function: a2a3's `pmu_aicpu_record_task` reads counters through MMIO from a per-core register window, and a5's arrive in an AICore-written staging ring that `pmu_aicpu_complete_record` validates. Neither belongs in a host unit test and neither is what this change touches, so the test writes `count` and the record fields and asserts on the identity and ownership around them. It sets the PMU register address to zero, which `pmu_aicpu_init` reads as "no PMU on this core" and skips the MMIO programming; the register accessors are defined in the test to **fail loudly** if anything reaches them.

**Negative controls** (`.docs/arms/nc_pmu.sh`, restoring from saved copies so it cannot eat uncommitted work):

| --- | --- |
| no stamp in the engine's pop hook | 3 |
| `init` drops retention | `IdleRuns…`, `AFailedFlush…` |
| `init` reuses without re-stamping | `IdleRuns…`, `AFailedFlush…` |
| failed flush clears the pointer | `AFailedFlush…` |
| successful flush retains | `APublishedBuffer…` |
| CSV row omits the epoch | `EachRunsRows…`, `RowsKeepTheirEpoch…` |
| CSV header omits the column | `EachRunsRows…`, `RowsKeepTheirEpoch…` |

Each discriminates; none turns the suite green or red wholesale, and the script verifies byte-identity afterwards.

## Results

- **cpput 157/157** (clean reconfigure + rebuild); the two new targets are 5/5 each.
- **All four platforms build** — 8/8 runtime builds, no errors.
- **The PMU scene test with `--enable-pmu --manual include`: 1/1 on a2a3sim and 1/1 on a5sim.** It sits in the manual lane on sim, so a default sweep deselects it. I read the CSV it produced:

  ```
  …,pmu_fix_instr_busy,event_type,run_epoch
  0,8,0x0000000100000001,1,1,0,…,2,1
  ```

  `run_epoch` is the last column and carries **1** on every row — non-zero, so genuinely stamped, through the whole chain from the AICPU stamp to the file.
- **st a2a3sim 39 passed / 8 skipped; st a5sim 39 passed; pyut 2405 passed / 50 skipped, 0 failed.**
- **No onboard verification of this commit.** Silicon is CI's to confirm.

## What is still open after this

The five wire structs now all carry identity, and no collector loses a buffer it could not publish. Two step-one items from the matrix remain before the per-run clearing can go:

- **the AICore-side device probe** — three assertions host unit tests structurally cannot make (that the epoch read at stamp time is this run's, that AICore sees flushed metadata rather than a cached copy, and the visibility order between the stamp and publication);
- **swimlane's `total_record_count − current_buf_seq × BUFFER_SIZE` derivation**, which #2313 made accurate but did not replace with the three-way split the matrix asks for.

Refs #1853


ChaoWao:
CI `pre-commit` fix pushed. It was one hook: **`ruff format`** on the two `test_pmu.py` files I edited — it wanted

```python
assert tuple(header_cols[-len(_REQUIRED_TRAILING_COLUMNS) :]) == _REQUIRED_TRAILING_COLUMNS, (
    f"header must end with {_REQUIRED_TRAILING_COLUMNS} in order: {header_cols}"
)
```

rather than my parenthesized form. Nothing else failed; every other job was gated behind pre-commit and never ran. Reformatted with the locally installed ruff, which is v0.14.8 — the exact version the hook pins.

My local gate was the gap, again: I ran clang-format and markdownlint but not the Python formatter. So I wrote `.docs/arms/run_local_precommit.sh`, which runs **every** hook that exists locally against `git diff upstream/main...HEAD` — `check_headers`, `check_english_only`, `check_retired_names`, `check_test_platform_literals`, `ruff format --check`, `ruff check`, `pyright`, `check_kernel_wire_isolation`, `clang-format`, `cpplint`, `markdownlint`. I also installed cpplint, which was the last hook with no local counterpart. All clean on this branch. Combined with running cpput as `ctest -LE requires_hardware -j4 --timeout 300` rather than bare `ctest`, that closes both gaps behind the last two red runs.

Re-verified: cpput 157/157 under CI's exact invocation, four platforms build 8/8.

**On the P2 you confirmed** — you are right and my #2325 claim was wrong. Covering `writer_done_` in the wait predicate is necessary but not sufficient; the flag has to *change* under the waiter's mutex, and both stop sites used a bare atomic store. So routing the test through export did **not** eliminate the hang, only which unsynchronized stopper runs. Measured it: bare store hangs 4/300 runs, flag-under-mutex 0/300. Fixed independently in **#2338**, which also closes an unsynchronized `write_queue_.empty()`/`.size()` read in the same block and adds the export-skipping teardown case. I still cannot attribute the original CI timeout to it — ~1.3% locally against 2-of-2 in CI is too large a gap to call it proven.

**Coverage gap** (run-end enqueue-failure reuse) is unclosed, and I said the same on #2325: reaching it needs `dump_args_flush` to stop pre-validating its thread index so the four collectors agree on where that check lives. That is a production change I will not make to serve a test without your call.

No review threads or review bodies outstanding on this PR; the only conversation comment is CodeRabbit's rate-limit notice.

PR #2342, opened 2026-09-18
Separate host_build_graph's run-image preparation from its publication

### LINKED ISSUE #2337: [Code Health] Separate host_build_graph's run-image preparation from its device publication
### Category

### Component

Host Runtime

### Description

`host_build_graph` computes a run's device execution image and writes it to the
device in one breath. The tail of `run_host_orchestration`
(`src/a2a3/runtime/host_build_graph/host/runtime_maker.cpp:626`, a5 twin at
`:1276`):

```cpp
std::vector<std::byte> storage(upload_bytes + CHIP_ALIGN_SIZE, std::byte{0});   // a2a3 :960
… memcpy the copied zone, compact the live shared-memory image into it …
api->copy_to_device(arena_dev + layout.off_copied_begin, upload_base, upload_bytes);  // :980
return total_tasks;                                                             // storage dies here
```

Two moments are conflated — the preparation of the bytes, and their submission
to the device — and the second is the one that has to move later:

- **The source cannot outlive the compute step.** `storage` is a per-bind local.
  A write that is not synchronous-and-immediate — enqueued behind a dependency,
  or replayed from a capture — needs those bytes to still exist when the DMA
  reads them. Today they are gone the instant orchestration returns.
- **A capture records device operations, not host graph building.** Replay does
  not call `run_host_orchestration` again, so the write has to be an operation
  that can be enqueued, given a dependency, and error-checked on its own. A
  synchronous copy issued from inside orchestration can be neither captured nor
  replayed: leave it outside a capture and replay has nothing that restores the
  execution data; put today's synchronous copy inside one and it is rejected.

The `Runtime` already has the right home for the answer. #2309 grouped the
host-only state into `HostOnlyState`, and #2315 named the device half
`DeviceRuntimeLaunchDesc` — so "what was prepared and where it goes" can be
recorded where it provably cannot travel to the device.

### Location

Source reference: `235353966` (main after #2325).

- `src/{a2a3,a5}/runtime/host_build_graph/host/runtime_maker.cpp` —
  `run_host_orchestration`, its staging vector and its `copy_to_device`
- `src/common/host_build_graph/runtime.h` — `HostOnlyState`
- `src/common/platform/{onboard,sim}/host/device_runner_base.{h,cpp}` —
  `acquire_sm_mirror` is the shape to mirror for the retained staging
  (onboard `device_runner_base.cpp:343`)
- `src/common/platform/include/common/host_api.h`,
  `src/common/platform/{onboard,sim}/host/c_api_shared.cpp` — the op table and
  the `simpler_prepare_run` call site
- `tests/ut/cpp/common/test_hbg_bind_ledger.cpp` — compiles the real bind
  against a fake `HostApiOps`, per arch

### Proposed Fix

1. **Retained staging.** A new `HostApi` op beside the existing retained-host-buffer
   ones (`acquire_sm_mirror`, `acquire_graph_definition_block` /
   `get_graph_definition_staging`):
   `acquire_run_image_staging(runner_ctx, pipeline_slot, bytes, alignment, addr_out)`,
   implemented on both runner bases as a per-slot, grow-only block in the exact
   shape of `acquire_sm_mirror` — which exists for this same reason, that a
   per-bind buffer costs an mmap and an munmap per bind.
2. **A publication record in the host-only tail.** `HostOnlyState` gains
   `{device_target, source, bytes}`. Host-only by placement, so #2315's boundary
   assertions already keep a pointer into host staging off the device.
3. **Orchestration prepares; it no longer writes.** It builds into the staging,
   records the publication, and returns. The `copy_to_device` and its
   `record_bind_phase(HostPhaseKind::BindArenaH2d, …)` move together — the phase
   record measures that copy, and `host_phase_kind_is_device_upload()`
   classifies the kind as a real host→device handover on the timeline.
4. **A runtime-impl publication entry**, beside the ones #2295 added:
   `int publish_run_image_impl(Runtime *, const HostApi *)`. `host_build_graph`
   performs the recorded copy and clears the record; a record already consumed,
   or never set, is an error rather than a silent success. TMR implements a
   documented no-op, as it does for `copy_in_run_inputs_impl`. The platform calls
   it from `simpler_prepare_run` immediately after `bind_callable_to_runtime`
   returns, under its own `STRACE` span — the ordering of device writes is
   unchanged, because between the end of orchestration and that point the bind
   only closes the host tensor view, latches `host_total_tasks` and reads back
   the pooled arena base.

**Eager behaviour does not change**: one graph build and one publication per run,
same order, same synchronous copy at the same point. No admission change, no
retained or reusable prepared result, no cross-invocation cache, and no new
public API — the `simpler_*` surface is untouched. Out of scope and left with
their owners: making the write asynchronous or captured, the platform-side
descriptor upload in `KernelArgsHelper::init_runtime_args` (its source is the
host `Runtime`, which already outlives the run), and anything in Pipeline A's
scheduling and completion work.

#### Acceptance criteria

Each names an observation and what a wrong value means:

1. **The bind moves no image bytes.** After `bind_callable_to_runtime_impl`
   returns, the count of `copy_to_device` calls targeting the arena image is
   **0**. Nonzero means the copy is still inside orchestration. This fails on
   today's code, which is what makes it a discriminator rather than a
   restatement of the diff.
2. **The publication moves exactly the recorded bytes, once.** After
   `publish_run_image_impl`, one copy to the recorded target of the recorded
   length, and the destination bytes equal the staging contents. A second call
   returns an error rather than copying again.
3. **An unpublished bind leaves nothing half-done.** If the publication never
   runs, the record is still set and the staging still holds the bytes, and no
   device byte was written — so the caller may publish or abandon.
4. **The source outlives the preparation.** The staging address the record names
   is still readable after `bind_callable_to_runtime_impl` returns, which is the
   property a later asynchronous or captured write depends on and the current
   per-bind vector cannot provide.
5. **The timeline still shows the handover.** A `--enable-chip-swimlane` run
   carries a `BindArenaH2d` segment whose `detail` is the same byte count as
   before. A missing segment, or one with no bytes, means the phase record moved
   without the copy it measures.
6. **No behaviour change on the eager path**, evidenced by the full onboard
   `host_build_graph` suite on a2a3 plus CI's `st-onboard-a5` lane, not by the
   structure of the diff.

### Priority

ChaoWao:
## Summary

S2 of the runtime-design work, and an internal boundary rather than a new API: this run's **host preparation** and its **device publication** become two steps with an owner, a lifetime and a submission point.

`run_host_orchestration` used to do both in one breath — assemble the device execution image into a per-bind `std::vector<std::byte>`, copy it to the device, and let the buffer die with the frame that made it. Two things follow from that, and the second is what has to move:

- **The source cannot outlive the compute step.** A write that is not synchronous-and-immediate — enqueued behind a dependency, or replayed from a capture — needs those bytes to still exist when the DMA reads them.
- **A capture records device operations, not host graph building.** Replay does not call the orchestrator again, so the write has to be an operation that can be enqueued, ordered and error-checked on its own. A copy issued from inside orchestration can be neither captured nor replayed.

| piece | what it does |
| --- | --- |
| `acquire_run_image_staging` | Per-slot, grow-only host block, released at finalize — the shape `acquire_sm_mirror` already uses, applied to the buffer one line below it |
| `HostOnlyState` | Gains `{device_target, source, bytes}` plus the three pool counters, so "what was prepared and where it goes" lives where #2315's boundary assertions prove it cannot travel to the device |
| `publish_run_image_impl` | Performs the recorded copy, records the `BindArenaH2d` segment that measures it, and consumes the record |

The platform calls the entry from `simpler_prepare_run` immediately after the bind returns, under its own `chip.run.publish_image` span. **Device writes keep today's order**: between the end of orchestration and that point the bind only closes the host tensor view, latches `host_total_tasks` and reads back the pooled arena base. `tensormap_and_ringbuffer` implements a documented no-op — its bind writes its own prebuilt arena image where it builds it — so the prepare path has one shape, as it does for `copy_in_run_inputs_impl`.

`BindArenaH2d` stays recorded at **exactly one site**, which is now the publication: the kind measures the copy, and `host_phase_kind_is_device_upload()` classifies it as a real host→device handover on the timeline. Its attribute string is unchanged — the three pool counters travel in the record because only the bind knows them, the task count and image size come off `dev`, and the copied zone is what the length has left over.

A record already consumed, or one a bind never set, **fails** rather than succeeding quietly: a run whose image never reached the device must not launch.

**Eager behaviour is unchanged** — one graph build and one publication per run, same order, same synchronous copy at the same point. No admission change, no retained or reusable prepared result, no cross-invocation cache, no new public API. Making the write asynchronous or captured, and the platform-side descriptor upload in `init_runtime_args` (its source is the host `Runtime`, which already outlives the run), stay out.

## Testing

One trace spans preparation and publication: the bind's guard ends it only on an exit that leaves nothing to publish, the publication ends it after recording the `BindArenaH2d` segment, and releasing a run's bindings closes a trace no publication consumed. Without that the segment would be recorded into an ended trace and dropped in silence — caught in review, and now guarded by a test that drives a real `HostPhaseRecordPool`.

The new cases assert **which bytes moved and when, and what was recorded**, not the shape of the diff:

- after the bind, **zero** writes to the image's target;
- after the publication, **exactly one** of the recorded length, carrying a stamp written into the staging *after the bind returned* — which is the property the per-bind buffer could not offer;
- a second publication is an error, not a second write;
- an unpublished bind left the image target unwritten and is still publishable, so abandoning is a choice rather than a loss;
- the publication's `BindArenaH2d` segment is present with the published byte count, while other segments prove the recorder was armed;
- an abandoned preparation closes its trace exactly once, on release.

Each fails against the shape it guards — the copy back inside the bind, or the trace ended at the bind's return. Confirmed by doing both, then removing them.

- [x] Unit tests pass — cpput 160/160 (8 in the HBG ledger suite), pyut 2405 passed / 50 skipped
- [x] Simulation tests pass — a2a3sim 40/40, a5sim 36/36, plus the 12 per-channel DFX lanes, including both `chip_swimlane` channels that read the `BindArenaH2d` segment
- [x] Hardware tests — the `--runtime host_build_graph` lane, the one this change touches: 49 passed, 1 skipped. The **full** a2a3 sweep currently reports two failures on this box (`TestWorkerAsyncEndpoint::two_frame_sequential_execution` plus one varying `tensormap_and_ringbuffer` case, each `507018 ACL_ERROR_RT_AICPU_EXCEPTION` beside a `recover_device_or_mark_unusable` device fault). **The same sweep on this branch's base reproduces the identical two**, and the tensormap victim moves between runs, so they are the box's rather than this change's; earlier sweeps on this branch were 67/67
- [x] `clang-format` clean
- [ ] **a5 onboard not run locally** — this box's silicon is a2a3 and its arch precheck refuses; a5's device readers rest on CI's `st-onboard-a5` lane

Fixes #2337


ChaoWao:
Reviewed head `b3a6c8385ae85616198ce586254ede8166935478` against base `2e331c3fd4f72b5fa5ee3b7e52dab3d41cc6d917` and issue #2337. The split is appropriately scoped. Please fix the P2 below before merging; the P3 items are non-blocking cleanup and do not require expanding this PR.

## Spec

### P2 — `BindArenaH2d` is recorded after its trace has already ended

Locations: [A2/A3 runtime_maker.cpp:1466](https://github.com/hw-native-sys/simpler/blob/b3a6c8385ae85616198ce586254ede8166935478/src/a2a3/runtime/host_build_graph/host/runtime_maker.cpp#L1466), [A5 runtime_maker.cpp:2121](https://github.com/hw-native-sys/simpler/blob/b3a6c8385ae85616198ce586254ede8166935478/src/a5/runtime/host_build_graph/host/runtime_maker.cpp#L2121).

`bind_callable_to_runtime_impl` still owns the scope guard that calls `host_phase_trace_end()` on return. The platform calls `publish_run_image_impl` only after that return. Consequently:

```text
bind: begin trace → prepare image → guard ends trace
publish: copy_to_device → record_bind_phase(BindArenaH2d)
                            → host_phase_record_bind sees !active and returns
```

`host_phase_trace_end()` disables recording and finishes/releases the record pool. The moved recording call therefore silently drops the actual image H2D duration and byte count on both architectures. This violates #2337 acceptance item 5, which requires the `--enable-chip-swimlane` output to retain the `BindArenaH2d` segment and its byte-count detail. The generic `chip.run.publish_image` STRACE span does not replace that typed record.

**Locally reproduced using the real HBG bind/publish functions and a real `HostPhaseRecordPool`, through the existing fake HostApi fixture:** all six existing `test_hbg_bind_ledger` tests pass; an additional test enabling the pool, calling bind then publish, and checking the recorded kind/payload fails:

```text
[  PASSED  ] 6 tests.
[  FAILED  ] HbgBindLedgerTest.ReviewPublicationRetainsArenaH2dPhase
Expected: arena_h2d_count == 1
Actual:   arena_h2d_count == 0
```

The probe first verifies that other phase records exist, so this is not a disabled recorder. Its expected `BindArenaH2d` payload is the saved pending-publication byte count.

Please keep one trace lifetime across preparation, publication, and their failure exits, and add this bind→publish recording regression test. Simply calling `host_phase_trace_begin()` again in publish would reset the preceding bind records/counters and is not sufficient.

No other definite violation of the scoped requirements was found: slot retention covers the current synchronous copy; production publication failure cleans up and destroys this run's Runtime; TMR's no-op and leaving descriptor publication/async/capture/Pipeline A outside this PR match #2337.

## Standards

### P3 — Describe the execution-image publication boundary accurately

Per `.claude/rules/doc-consistency.md`, please update these exact descriptions:

- [test_hbg_bind_ledger.cpp:321–324](https://github.com/hw-native-sys/simpler/blob/b3a6c8385ae85616198ce586254ede8166935478/tests/ut/cpp/common/test_hbg_bind_ledger.cpp#L321): the comment and `AnUnpublishedBindWroteNoDeviceByte` test name claim no device byte was written. This INOUT case already copies tensor input during bind; the assertions only establish that the execution-image target was not written. Rename/reword around **the unpublished execution image**, not all device writes.
- [A2/A3 runtime_maker.cpp:1359–1364](https://github.com/hw-native-sys/simpler/blob/b3a6c8385ae85616198ce586254ede8166935478/src/a2a3/runtime/host_build_graph/host/runtime_maker.cpp#L1359), and [A5:2014–2019](https://github.com/hw-native-sys/simpler/blob/b3a6c8385ae85616198ce586254ede8166935478/src/a5/runtime/host_build_graph/host/runtime_maker.cpp#L2014): “orchestration is what performs that copy” is stale after this split. Orchestration assembles the image; publication uploads it.
- [A2/A3:964–965](https://github.com/hw-native-sys/simpler/blob/b3a6c8385ae85616198ce586254ede8166935478/src/a2a3/runtime/host_build_graph/host/runtime_maker.cpp#L964), and [A5:1614–1615](https://github.com/hw-native-sys/simpler/blob/b3a6c8385ae85616198ce586254ede8166935478/src/a5/runtime/host_build_graph/host/runtime_maker.cpp#L1614): “every byte published … this bind writes” overstates initialization. `compact_live_image` copies live segments but leaves alignment gaps, while the new retained buffer is uninitialized. Limit the statement to live fields, or explicitly initialize padding if full-byte initialization is intended. I found no device reader of these gaps and am **not** claiming a functional failure from padding.

### P3 — Replace change-history/forward-plan comments with current contracts

Per `.claude/rules/comments.md`:

- [runtime.h:298–302](https://github.com/hw-native-sys/simpler/blob/b3a6c8385ae85616198ce586254ede8166935478/src/common/host_build_graph/runtime.h#L298): “after the split”.
- [onboard c_api_shared.cpp:1079–1081](https://github.com/hw-native-sys/simpler/blob/b3a6c8385ae85616198ce586254ede8166935478/src/common/platform/onboard/host/c_api_shared.cpp#L1079), and [sim:831–833](https://github.com/hw-native-sys/simpler/blob/b3a6c8385ae85616198ce586254ede8166935478/src/common/platform/sim/host/c_api_shared.cpp#L831): “a later change”.
- [test_hbg_bind_ledger.cpp:296–298](https://github.com/hw-native-sys/simpler/blob/b3a6c8385ae85616198ce586254ede8166935478/tests/ut/cpp/common/test_hbg_bind_ledger.cpp#L296): “the per-bind buffer this replaced”.

State the current invariants instead: slot-owned staging survives bind; publication synchronously consumes it; success clears the pending record. Keep edit history and later capture plans in the PR description.

Validation scope: CPU reproduction above; existing reported PR CI checks were green. I did not rerun hardware tests during this review.

**Summary: Spec — 1 finding, P2. Standards — 2 findings, P3.**

ChaoWao:
@ChaoWao P2 fixed, both P3 groups done, at `549dfac3a`.

**P2 — you were right, and my verification was the thing at fault.** The bind's `RAIIScopeGuard` called `host_phase_trace_end()` on *every* exit including the successful one, so the moved `record_bind_phase(BindArenaH2d, …)` ran against an ended trace and `RecordAdmission` withdrew it. The segment was dropped on both architectures. I claimed the DFX lanes covered this; they exercise the channel but assert nothing about that segment's presence, so they were never going to catch it.

One trace now spans preparation and publication:

- the bind's guard ends the trace **only on an exit that leaves nothing to publish** — every failure path, where the breakdown is worth having;
- a bind that recorded a publication hands the end to `publish_run_image_impl`, which closes it after recording the segment and on its own failure exits;
- `release_run_bindings_impl` closes a trace no publication consumed, so an abandoned preparation still reports its breakdown.

`host_phase_trace_end()` already ignores a second call without an intervening `begin()`, which is what lets two owners share the duty without either needing to know who got there first — so no `host_phase_trace_begin()` restart, and the preceding bind records and counters are untouched.

**The regression test you asked for is in**, driving a real `HostPhaseRecordPool` armed through the fake's `host_phase_pool_arm`:

- `ThePublicationRecordsTheArenaH2dSegment` — after the bind, other segments exist (so the recorder is provably armed) and `BindArenaH2d` is absent; after the publication, exactly one, with the published byte count as its payload, and the trace finished exactly once.
- `AnAbandonedPreparationClosesItsTraceOnRelease` — a preparation nobody publishes closes its trace once, on release, and drops its record.

Both fail against the shape they guard. With the guard restored to ending the trace unconditionally: `the publication's host-to-device segment was dropped` and `the bind closed a trace the publication still owns`. Confirmed by doing it, then removing it. 8/8 in that suite, cpput 160/160 on the rebased head.

**P3 — descriptions.** `AnUnpublishedBindWroteNoDeviceByte` → `AnUnpublishedBindLeavesTheImageTargetUnwritten`, with the comment saying so and noting explicitly that the INOUT input copy is not in question. "orchestration is what performs that copy" → orchestration assembles the image, the publication uploads it. The initialization claim now states what is actually written — the copied zone and the live segments `compact_live_image` emits — and that the alignment gaps between segments carry whatever the block last held, with the reason that makes clearing unnecessary: every device reader indexes a segment through the layout.

On that last point, one consequence worth naming rather than leaving implicit: the old per-bind vector was zero-initialized, so the gaps were zeros; the retained block is not, so they now carry the previous bind's bytes for that slot. No device reader touches them, so this is hygiene rather than correctness. If you would rather the image never carry unrelated host bytes, the fix is one `memset` of the gaps — but it would touch every page of the block per bind, which is the cost the retention was for. I left it uncleared and documented; say the word and I will invert that.

**P3 — comments.** "after the split", "a later change", "the per-bind buffer this replaced" all replaced with present-tense invariants: slot-owned staging survives the bind; the publication synchronously consumes it; success clears the pending record. The forward-looking reasoning stays in the PR description.

---

One correction to my own verification claim, since you did not rerun hardware and the PR description said 67/67 with zero failures. The full onboard sweep on this box now reports two failures:

```text
TestWorkerAsyncEndpoint::two_frame_sequential_execution   507018 ACL_ERROR_RT_AICPU_EXCEPTION
one varying tensormap_and_ringbuffer case                 same, beside recover_device_or_mark_unusable
```

I ran the same sweep on this branch's base to settle it, and **main reproduces the identical two** — same cases, same count, same signature. The tensormap victim moves between runs (`TestL3LaunchAcceptance`, then `TestL3TensorDispatch`), which is what a device fault taking its neighbours looks like rather than a deterministic break. The `--runtime host_build_graph` lane — the one this change touches — is green at 49 passed, 1 skipped, and earlier sweeps on this branch were 67/67. The commit message and PR description now say this instead of claiming a clean sweep.

PR #2343, opened 2026-09-18
Refactor: track the AICore record count instead of deriving it from capacity

ChaoWao:
## Why

Last of the two items blocking step one of the continuous-collection plan (#1853). The AICore pool decided how many records a buffer held by deriving it:

```cpp
live = total_record_count - current_buf_seq * PLATFORM_AICORE_BUFFER_SIZE;
```

and stamped a rotated-out buffer with the capacity outright (`publish_aicore_pending_buffer`). Both rest on premises that only hold while collection is bracketed per run:

- **The derivation needs `total_record_count` and `current_buf_seq` reset together.** The host's per-run clear and `init` guarantee that today. Remove the clear — which *is* continuous collection — and `total` carries earlier runs while `seq` restarts at 0, so the derivation reports a whole window as live in the current buffer.
- **Both need every rotation to have been full**, which a rotation that found no free buffer breaks.

#2313 made the derivation accurate at every level; it did not remove the derivation. This does.

## What changed

`ChipSwimlaneActiveHead` gains `live_record_count` and `published_record_count` in padding it already had — the head stays **one cache line**, which matters because AICore `dcci`s it as `SINGLE_CACHE_LINE`. `live` bumps per dispatch and resets when a buffer is handed over; `published` accumulates the marks actually given to the host. The rotated-out buffer's mark is captured as it is stashed, while it still describes that buffer rather than its replacement.

`PLATFORM_AICORE_BUFFER_SIZE` now appears only where it means capacity: the rotation boundary, and the clamp below.

### The clamp now settles what it used to hide

A rotation that finds no free buffer returns with the old buffer still active, so AICore keeps dispatching into a full buffer and its slot guard refuses the overflow. `aicore_rotate` cannot charge those when it fails — it does not yet know how many will follow, and its comment says so — so the excess accumulates in `live` and is charged to `dropped` at the point the count is taken. Previously that was silent loss, visible only as a broken reconcile.

**This path is not covered by a test, and I am not implying it is.** Reaching it needs the AICore free queue empty at a rotation, which means `wait_for_free_queue_entry` spins the full `PLATFORM_DFX_BACKPRESSURE_TIMEOUT_CYCLES` — 30 s of wall clock against the host stub's `steady_clock`. Same gate as the args_dump run-end flush branch.

### Host per-run clear

The four counters are contiguous, so the existing narrow write-back still covers them in one call. `live` and `published` must be cleared with the other two: the identity is per run, so clearing part of it would compare a fresh total against carried-over published records. Contiguity is asserted at the struct, next to the field order it constrains, rather than at the write-back.

## Tests

Three cases assert the accounting. One is the cross-run scenario the derivation could not survive; one is the starvation-recovery path from review:

| Case | What it pins |
| --- | --- |
| `TheTailCountSurvivesAWindowSpanningTwoRuns` | run 1 publishes 3; run 2 starts **without** the host's clear and dispatches 2. The derivation marks run 2's tail **5** — more records than were ever written into that buffer. The tracked counter says **2**. |
| `DispatchesWithNoBufferAreDroppedNotCarriedIntoTheNextOne` | a starved pool, capacity dispatches with nowhere to land, then the host returns one buffer. The recovered buffer must be marked **1**, not capacity. Asserts the mark and `dropped` **separately**, because their sum is identical under both behaviours. |
| `AnIdleRunLeavesNothingLive` | a retained buffer contributes no phantom count |

**Negative controls** (`.docs/arms/nc_aicore_accounting.sh`, restoring from a saved copy):

| --- | --- |
| restore the old derivation verbatim | `TheTailCountSurvivesAWindowSpanningTwoRuns` |
| dispatch stops bumping `live` | 4 |
| successful publish does not add to `published` | `TheTailCount…` |
| `init` does not reset `live` | **0** |
| a dispatch with no active buffer is credited to `live` (the reviewed P2) | `DispatchesWithNoBuffer…` |

That last row is worth stating rather than hiding: **the `init` reset is unreachable from any completed run**, because every path that reaches flush already zeroes `live` in `take_aicore_live_count`. It is kept for the one path that does not reach flush — a kernel the op-execute watchdog reaped mid-run leaves `live` describing records in a buffer the next run reuses — and the comment now says exactly that instead of implying a general reset. If you would rather not carry an untestable line, say so and I will drop it.

Also worth flagging, since it changes what #2313 claimed: **there were two capacity-as-count sites, not one.** #2313 fixed the flush's tail mark and left `publish_aicore_pending_buffer`'s at `:487`. That one is correct today — rotation fires exactly at a capacity boundary — so replacing it with the tracked count is **not observable in any cheaply-testable scenario**; its only divergence is the failed-rotation path above. Correctness by construction, not by test.

## Two corrections from review

**[P2, fixed] A dispatch made while no buffer is active was credited to `live`.** AICore resolves the head lazily and gets a null buffer, so such a dispatch lands nowhere — but `live` counted it, and the next buffer the pool handed over inherited the tally. With capacity B: `init` finds the free queue empty, B dispatches accumulate in `live`, the host returns a buffer, the B+1'th dispatch rotates into it with an outgoing pointer of 0 (so nothing is stashed and nothing settled), and the flush marks a buffer holding **1** record with **B**. Those dispatches are now charged to `dropped`.

This is invisible to an identity check — the sum balances either way — so the new case asserts the buffer's own mark and `dropped` separately.

It is also **not hypothetical in CI**: the `st-sim-a5` log on the previous push is full of `AICore free_queue is empty during init!` across many cores at generation 8, which is exactly this path's precondition. I am not claiming that as the proven cause of that failure (see below), but the precondition is demonstrably reached there.

**[non-blocking, documented] The identity is a checkpoint, not a running invariant.** A rotated-out buffer's count leaves `live` when the rotation stashes it and joins `published` only when the ACK gate releases it, so in between it is in neither — held in the AICPU-private pending slot. The header now says that, and scopes the identity to "after a flush whose publishes all succeeded", which is where the host reads these. The two original cases only covered tail publication, never that hand-off; that remains uncovered, and a test cannot read the pending slot because it is file-local to the AICPU translation unit.

## The `st-sim-a5` failure on the previous push

Worth recording rather than quietly force-pushing over. `st-sim-a5` failed on both runners; on re-run **ubuntu passed and macos failed again**, so it is not deterministic. Locally the same directory with CI's exact command — `-p no:xdist --pto-session-timeout 600 --require-pto-isa --manual include --enable-chip-swimlane --enable-dep-gen` — passes 5/5 on a5sim and 5/5 on a2a3sim, before and after this fix.

So: the P2 fix plausibly addresses it, since the failing log shows its precondition, but I have not reproduced the failure and cannot attribute it. If it recurs on this push, it needs its own investigation rather than another guess.

I also should have run that channel before opening this PR at all: it is gated behind `--manual include` in a separate DFX smoke step, so the plain a5sim sweep I did run deselects every case in it — on a PR whose whole subject is swimlane.

## Results

- **cpput 159/159** under CI's exact invocation (`-LE requires_hardware -j4 --output-on-failure --timeout 300`); the swimlane target is 12/12.
- **chip_swimlane DFX channel, CI's exact command: 5/5 on a5sim and 5/5 on a2a3sim.** This is the gate that caught nothing before because I had not run it.
- **All four platforms build** — 8/8 runtime builds, no errors. The header is compiled by ccec too, so the one-cache-line assert is checked by both toolchains.
- Local pre-commit hooks clean on the changed files, cpplint included with the hook's exact args.
- **No onboard verification.** Silicon is CI's to confirm.

## What step one still needs after this

Only the **device probe**: that the epoch read at stamp time is this run's, that AICore sees flushed metadata rather than a cached copy, and that the stamp is visible before publication. Three assertions host unit tests structurally cannot make. Everything else in step one is now in code.

Refs #1853


ChaoWao:
P2 fixed, and your non-blocking point is now documented in the header rather than left implicit. Pushed.

## [P2] Dispatches with no active buffer

You're right, and the path is exactly as you traced it. `init` leaves `current_buf_ptr` at 0, AICore resolves the head lazily and gets a null buffer so its reserve refuses every record, but `on_aicore_dispatch` credited `live` anyway. At the B+1'th dispatch the rotation runs with an outgoing pointer of 0, so it stashes nothing and calls nothing that would settle the stale tally — and the flush marks a buffer holding 1 record with B.

Those dispatches now go to `dropped`. Your point about the acceptance criterion is the part I'd have got wrong on my own: **the sum balances under both behaviours**, so the new case asserts the buffer's own mark and `dropped` separately, not the identity.

`DispatchesWithNoBufferAreDroppedNotCarriedIntoTheNextOne` — deterministic, via the host stub as you suggested: present an empty free queue to `init`, dispatch capacity times, return one buffer, dispatch once more to hit the rotation boundary, flush. Asserts mark == 1, `dropped` == capacity, `total` == capacity+1. No 30 s wait. Reinstating the defect fails exactly that case and nothing else.

I did not need the extra "live starts at zero on first acquire" guard: with the charge redirected, `live` is only ever incremented while a buffer is active, so it is already 0 whenever the pointer is. Rather than add an unreachable line I left it out — same reasoning as the `init` reset row in the negative-control table, which comes back 0 and which I flagged in the body.

## [non-blocking] The identity needs a scoped observation point

Also correct, and I had written it as if it were a running invariant. Between the rotation stashing a buffer's count and the ACK gate releasing it, that count is in neither `live` nor `published` — it sits in the AICPU-private pending slot. The header now states that and scopes the identity to *after a flush whose publishes all succeeded*, which is where the host reads these.

I did not extend the formula to include pending, because a test cannot observe it: `s_aicore_pending_enqueue` is file-local to the AICPU translation unit. So that hand-off window stays uncovered, and I say so in the body rather than implying the three cases cover it.

## On the `st-sim-a5` failure you saw still running

It finished, and it is worth reporting because it bears on the P2. It failed on both runners; on re-run **ubuntu passed, macos failed again** — so not deterministic. Locally the same directory with CI's exact command passes 5/5 on a5sim and 5/5 on a2a3sim, before and after this fix.

The connection: that job's log is full of `AICore free_queue is empty during init!` across many cores at generation 8 — **exactly this P2's precondition**. So the fix plausibly addresses it. I have not reproduced the failure and am not claiming it as the cause; if it recurs on this push it needs its own investigation rather than another guess.

Separately, my own process failure: that channel is gated behind `--manual include` in a dedicated DFX smoke step, so the plain a5sim sweep I ran deselects every case in it — on a PR whose entire subject is swimlane. I now run it explicitly with CI's exact flags, both arches.

## Re-verified

cpput 159/159 under CI's exact ctest invocation (swimlane target 12/12) · four platforms build 8/8 · chip_swimlane channel 5/5 on both sims · all local pre-commit hooks clean including cpplint.

PR #2368, opened 2026-09-19
Docs: separate what the descriptor's visibility chain proves from what it assumes

ChaoWao:
## Summary

#2254's field inventory observed that three onboard variants invalidate the whole device descriptor at end of run while a5 `tensormap_and_ringbuffer` does not, and read that as a missing operation. This entry traces the chain — host snapshot → slot-owned destination → synchronous H2D → launch ordering → first device read → last access → reuse — at a pinned SHA, and finds the original reading unsupported. It does **not** support the opposite reading either.

Docs only. No cache operation is changed, no contract is created, nothing was built, run or measured.

## What the trace found

**The asymmetry inverts once the documented rule is applied.** `docs/hardware/cache-coherency.md:130-145` states that a5 host-DMA writes are coherent with AICPU reads and that a5 therefore does *not* invalidate. Against that rule a5 `tensormap_and_ringbuffer` is the **conforming** variant, and a5 `host_build_graph` is the departure — it calls the operation at two sites.

**The rule also disagrees with the code in two further ways.** It is written as *"before reading host-written Runtime"*, and `chip-architecture.md:218-219` places it at AICPU entry; all four call sites are in `deinit()`, at the end of the *previous* run. And its snippet's `sizeof(Runtime)` extent predates #2309/#2315's narrowing to `sizeof(dev)`.

**The invariant neither document states:** the earliest descriptor reader is the platform AICPU kernel's affinity gate (`kernel.cpp:93` / `:121`), not the runtime executor. That gate performs no cache operation, so on **every** variant the first read of host-published bytes happens with no maintenance earlier in the same invocation — and because the gate decides how many threads proceed, a stale read changes the thread population rather than a value.

**Deliverable matrix** covers a2a3/a5 × TRB/HBG, with a5 HBG resident and legacy listed separately (the resident/legacy branch is itself taken *by reading* the host-published mode word) and program versus kernel `prepare_once`, where the single upload per context makes the visibility question different in kind rather than a variant of the program-mode one.

## Four assumptions recorded as unavailable

Not argued against — recorded as things the code does not establish:

- An end-of-run invalidate creates no ordering against a *later* host DMA; `dsb sy; isb` completes the maintenance and nothing more.
- A slot's **first** run has no previous `deinit`, so no invalidate has ever executed for those lines.
- Invalidation is not an acquire.
- `cache_invalidate_range` is **`dc civac`** — clean *and* invalidate — so at `deinit` it also **writes AICPU-dirtied descriptor bytes back**, the opposite direction from the one its comment implies. It is also line-granular and rounds outward.

## Separately established, and previously unwritten

a5 HBG's 64-byte `Handshake` line has **three** whole-line writers: the host (`aicpu_ready` mode + `task` context address), the AICore (`dcci … CACHELINE_OUT`) and the AICPU (`dc cvac`). Whole-line write-backs cannot merge at word granularity, so ordering is the only thing that makes this safe — and what provides it is **`aicore_done`**, published by the very flush that would otherwise overwrite the host's words. The AICPU does not reach `publish_context_partition` until it has observed every core's `aicore_done`.

That ordering is load-bearing and documented nowhere. It is what a future change to this line would silently break, and it is documentable today because it is a property of the current code rather than of the hardware.

## Why no fix is proposed

The candidates are mutually exclusive, and which is right depends on three unknowns this repository cannot settle:

1. Whether a5's host-DMA→AICPU coherency claim holds — it is asserted with no citation to an SDK document, driver source or measurement.
2. Whether AICPU kernel launch leaves the cache cold or performs maintenance.
3. Whether GM is cacheable for the AICPU at all — the documented `Device-nGnRE` attribute covers the MMIO window only.

Any one being true would make some or all four call sites redundant. Shipping a change now would encode a guess as an invariant.

A probe specification for each unknown is included, with its **positive control**, the version/binary provenance to record, and an explicit statement of what measurement **cannot** do — a probe can demonstrate a violation but cannot produce an architectural guarantee. **None was run**, and sim cannot exercise any of it, since both cache primitives are no-ops there.

## Baseline and evidence limits

Every anchor read at **`31da0560e202b7bb378773f7dfc1d0074fb28e89`** (`main`, the merge of #2363), pinned in the document. No hardware number from any other commit is carried forward. All cited line anchors were verified against that tree.

Three documentation corrections are **identified but deliberately not made**, because each depends on an unknown — correcting them now would replace one unsourced claim with another. Three decisions are returned to the maintainer rather than taken.

## Scope

Does not touch the descriptor gate-tail relocation or the full-allocation/prefix-upload separation in flight elsewhere, nor E1 health or DFX export. Read-only analysis of published `main`.

## Testing

- [x] `markdownlint-cli2` with the repo config — 0 errors on both files
- [x] All code anchors verified against the pinned SHA
- [n/a] No build, test, simulator, onboard or benchmark run — docs-only change, and by design for this entry

Relates to #2254.


ChaoWao:
Static review of this PR landed after it merged. Several conclusions here promoted source control flow, or observations that do not discriminate, into hardware guarantees, and are withdrawn in #2369 — the marker-serialization claim, "the affinity gate is the earliest device read", the kernel-mode repeated-run contract, write-through as an answer to input freshness, and the probe plan's stated classifications.

The source did not change: same baseline `31da0560e`, only the claims. #2369 marks each retraction in place rather than deleting it, so this PR's body stays readable as the history it is.

PR #2369, opened 2026-09-19
Docs: withdraw the visibility investigation's over-stated conclusions

ChaoWao:
## Summary

Follow-up to #2368, which merged before its static review could be applied. That review found several conclusions in the entry promoted **source control flow**, or observations that **do not discriminate**, into hardware guarantees. This PR corrects them.

**The source did not change between the two versions** — the baseline SHA is the same `31da0560e`. Only the claims do. Each retraction is marked **in place** as *"Withdrawn from the first version"* rather than deleted, so a reader arriving from #2368 sees which claim was retracted and why; the withdrawals are also carried into the README summary so nothing stale is left in the index.

Docs only. No runtime change, no test, build or probe.

### F1 — `aicore_done` is a dependency the code expresses, not proven serialization

#2368 said the marker "serializes" a5 HBG resident's three handshake writers and that this is "what keeps it correct".

Observing the marker is **not** an observation that the producer's whole-line write-back retired — the post-`dcci` `dsb` orders that producer's own subsequent accesses and nothing more — and it does not establish that the AICPU's clean-and-invalidate cannot write back an older copy. Those are the same publication and ownership assumptions the entry leaves open elsewhere, so the dependency cannot be promoted using the document's own unknowns. **No currently executing race is established either**; the point is that the intent is legible while its hardware preconditions are unwritten.

Also scoped: the discussion is now explicitly about **resident**, whose function ends at `write_reg(COND, AICORE_EXITED_VALUE)` (`aicore_executor.cpp:695`) with **no final handshake `dcci`**. Legacy has one, as do a2a3 TRB/HBG and a5 TRB — and the a2a3 paths' last descriptor access is a *read* of a teardown gate after their own write-back. Each path's final access is now listed separately instead of importing one path's exit into another.

### F2 — the first AICPU read is not the first device read

#2368 claimed the platform affinity gate is the earliest device reader on all four variants. It is the earliest traced **AICPU** reader. Both launch paths submit **AICore first**, on a distinct stream (`device_runner.cpp:664-687`), and nothing orders the two streams against each other.

First reads are now given per processor and mode. The missing case is concrete: a5 HBG's AICore reads the host-published mode word at `aicore_executor.cpp:533-535` — that read selects resident vs legacy — and the resident path consumes the host-published context address and its pointee before READY (`:553-561`), all ahead of any AICPU handshake permission. On the other paths the AICore's first descriptor access is its own report *write*, with the task pointer read after window-open; that shape is recorded rather than collapsed into the affinity gate.

**Kernel mode is withdrawn as a traced row.** `prepare_once` does provision storage, but `simpler_kernel_mode_supported` returns 0 (`c_api_shared.cpp:1561`) and `simpler_kernel_mode_launch` validates its arguments then unconditionally returns `INVALID_STATE` (`:1700`) — the section header says *"Launch remains a rejecting stub."* Allocation/upload is marked implemented scaffolding; first/last device reads and the end-of-run row are marked **not delivered**, and the inferred repeated-run cache contract is retracted.

### F3 — cache-policy alternatives do not imply the proposed removals

- **Write-through removed as a sufficient answer** to input freshness: it governs writes, while a write-through cache still holds readable lines that external DMA can leave stale. #2368 equated it with an uncached mapping.
- Because `cache_invalidate_range` is **clean-and-invalidate**, it may serve *both* an inbound freshness role and an outbound publication role for device-authored fields. So no single unknown licenses deleting it: host-DMA coherence answers one input edge only, a cold next cache says nothing about the previous run's outbound obligations, and a5 *not* being coherent would still not make a5 TRB defective while launch maintenance and mapping remain open.
- The **"mutually exclusive candidate fixes"** list is replaced by a table of the **individual obligations** any change must discharge, leaving removal or relocation undecided until every role of the existing call is accounted for. A docs-only PR should not hand a later implementer a deletion criterion that proves half the operation's role.
- Same-slot reuse is now described by **what the code proves**, with two predicates kept apart: same-slot exclusion is **unconditional** (`try_reserve_native_run` rejects any live reservation naming the same slot at `:2931`, before any successor logic), while the prepared-successor rule at `:2940` requires the **existing** reservation to be the active run — it says nothing about the caller, and combined with the first check a successor necessarily takes a **distinct** slot. So same-slot exclusion is not argued from the successor rule. Reuse then follows **completed prior device execution**: `reap_run` → `wait_run_fence` (`:2539`) waits **both** run-completion events and additionally `sync_stream_pair`s both streams, with the no-fence fallback synchronizing both streams directly, before any teardown; a non-zero result returns early into recovery, so no completion is claimed for a failed synchronization. That is more than host call ordering — but execution completion is **not** cache retention or input freshness, and it does not bear on the separate, still-unproved startup marker / whole-line hand-off.

### F4 — the probe specification is demoted to open questions

Each is restated with its non-unique interpretations and the verified initial state it would need:

- **Q1** — a cross-arch run is a *comparison*, not a positive control, until a stale readable line is independently established; the same launch-maintenance, residency and mapping uncertainties apply to a2a3.
- **Q2** — a dirty-sentinel run cannot separate "the cache survived" from "an old dirty line was written back over the host value and then read fresh", nor either from plain eviction before the read. It adds a write-back race to a read-freshness question.
- **Q3** — marked **a question, not an actionable probe**: page attributes are not classifiable by timing against an MMIO reference even with a second known-cached reference, since access path, ordering attributes and latency all differ. The "flat ratio indicates uncached" rule is withdrawn.

### Non-blocking precision notes, also applied

- "A fresh allocation per run would make most of this moot" — withdrawn; new allocation is new *ownership*, not proof of new physical backing or a cold cache.
- "The AICPU does dirty descriptor lines" — narrowed to observed **stores**, since a dirty write-back policy is exactly what unknown 3 says is unestablished.
- CI passing is recorded as an observation of the exercised cases, not a frequency bound.

## Hardware-doc corrections: verified location and extent only

`docs/hardware/cache-coherency.md` and `docs/hardware/chip-architecture.md` are corrected **only** where directly verified against source, which needs no hardware guarantee:

- the call's extent is `sizeof(runtime->dev)`, not `sizeof(Runtime)` (#2309/#2315);
- the call sits in `deinit()` at the **end** of a run, where both docs place it before the read;
- a note that a5 `host_build_graph` calls it at two sites while a5 `tensormap_and_ringbuffer` does not.

**The a5 coherency claim itself is left exactly as it stands.** Settling it needs evidence this work does not have, and replacing one unsourced claim with another is not an improvement. The discrepancy is recorded and linked, not resolved.

## Scope

Modifications only — the investigation is already in `main` via #2368, so this diff is correction work and does not re-add it. Does not touch the descriptor gate-tail relocation or allocation/prefix-upload separation in flight elsewhere, nor E1 health or DFX export.

## Testing

- [x] `markdownlint-cli2` with the repo config — 0 errors on all four files
- [x] Every code anchor re-verified against `31da0560e`, including the five per-path exits, the launch order, and the kernel-mode stub
- [n/a] No build, test, simulator, onboard or benchmark — docs-only, and by design for this entry

Relates to #2254. Corrects #2368.


ChaoWao:
Current-head static review (`7ce1ed3e8`): most of the earlier corrections are now in place. One P2 remains in “What the code does establish about reuse ordering”:

- `try_reserve_native_run` rejects matching `pipeline_slot` unconditionally at `device_runner_base.cpp:2931`. The later `active != existing->owner` check refers to the existing predecessor, not to the caller. A prepared successor uses a different slot. Please separate same-slot exclusion from successor admission.
- The following “only host-side phases / no host-side phase observes” wording omits the actual drain contract: `wait_run_fence` waits the run’s two event boundaries and then `sync_stream_pair`; its unfenced fallback also synchronizes both streams. Successful same-slot reuse follows completed prior device execution under these completion and reservation contracts. That still does **not** prove cross-run cache invalidation or input freshness. Preserve that distinction rather than demoting the boundary to host call order. Failed synchronization needs its separate recovery/mark-unusable account if discussed.

Small factual correction in the three-writer section: step 5 is `cache_flush_range` / `dc cvac` (clean), not “clean-and-invalidate”; step 4 is the `dc civac` call. The distinction matters to this document’s argument.

No runtime change, new hardware claim or measurement is needed. Keep this follow-up docs-only.

ChaoWao:
Re-reviewed the updated head `2e3da6ef5`: the remaining findings are addressed. The document now separates unconditional same-slot exclusion from predecessor-based successor admission, states the successful dual-event plus dual-stream completion boundary, and distinguishes step 5 `dc cvac` from step 4 `dc civac`.

No remaining blocking finding for this docs-only correction. This is not approval of a new hardware ordering contract or permission to remove cache maintenance. Exact-head docs build and pre-commit checks passed; code/test lanes correctly skipped. Ready for the user’s final decision; no merge performed.

PR #2370, opened 2026-09-19
Stop uploading the device-initialized teardown gates

ChaoWao:
## Summary

`teardown_gates` travelled in the per-run `Runtime` H2D and **no host value for it was ever consumed**. Its meaningful initial state is produced on the device, so the descriptor now ends in the array and the upload stops before it. The array is still allocated, still cache-line isolated, and still covered by the existing cache-maintenance extent.

This is `#2254` step 2a scoped to the gates only. **No epoch, no handshake redesign, no admission or sync change.**

## Why the gates need no host value

**On a2a3** the AICPU zeroes every active entry in `pre_handshake_init` and executes `wmb()` before publishing `hs_setup_done_`, and no register window opens before that publication. The AICore's single gate load is the post-close wait *after* its startup wait and its dispatch/exit loops, so every path that can reach it is dominated by that reset — including a pre-init abort, where unopened cores never reach the load at all.

**On a5** there is no gate reader, no gate writer and no gate initialization: the shared hbg descriptor declares the array and no a5 code touches it — **unused reserved storage**, not padding. a5 trb has no such member.

So no host-supplied gate value is consumed on either arch. The justification is init-before-read on a2a3 and non-use on a5 — *not* "the old host bytes were uninitialized".

## Allocation and upload were one number

Separating them is what the change actually required:

- `runtime_device_extent_size` — sizes every block backing a device `Runtime`
- `runtime_device_copy_size` — sizes the snapshot and the H2D

Both host paths take both. The program path's `prepare_runtime_args` allocates the extent and keeps its per-slot mismatch guard on it, while `RuntimeLaunchImage::prepare` snapshots the prefix; `PersistentKernelArgs::prepare_once` does the same within its prepare-once and rollback semantics. Sizing **either** allocation to the prefix would leave the device addressing gates past the end of its own block. A5 trb has no gate tail, so its two extents coincide; it gains the second entry point so the shared host paths need no per-runtime branch.

## Measured layouts

The arrays are **relocated** from their original position between `workers[]` and the host-published fields to the descriptor tail. At 4,608 and 6,912 bytes they are far larger than any alignment padding, so this is a reordering, not an absorption. Reordering leaves each descriptor's measured `sizeof` unchanged — so no allocation moves — while the prefix excludes the relocated array. Measured, not assumed:

| descriptor | device extent | uploaded prefix | off every H2D |
| --- | --- | --- | --- |
| a2a3 `host_build_graph` | 17,536 B | 12,928 B | **4,608 B** |
| a5 `host_build_graph` | 22,144 B | 15,232 B | **6,912 B** |
| a2a3 `tensormap_and_ringbuffer` | 51,456 B | 46,848 B | **4,608 B** |
| a5 `tensormap_and_ringbuffer` | 49,152 B | 49,152 B | none |

Per-run call count is unchanged. **No latency claim** — bytes and call counts only.

## Guards

Three assertions per affected descriptor: the gate array starts on a cache line, it **ends exactly at `sizeof`**, and the descriptor stays standard-layout and trivially copyable at offset zero of `Runtime`. The second is what fails the build if a host-published field is ever appended behind the gates, where it would silently never receive its host value.

## Unchanged

Handshake protocol, worker upload, window-open and claim ordering, gate reset and release, the AICore exit sequence, the cache-maintenance extent (deliberately still the full descriptor), recovery and admission.

## New cases and negative controls

Four new cases cover both extents in **both** host paths, each run on **both** a2a3 runtimes — the full 2x2, eight runs:

- `PersistentKernelArgs.LeavesTheDeviceInitializedTailUntouched`
- `PersistentKernelArgs.RepeatedPrepareNeitherReallocatesNorRecopies`
- `KernelArgsPublication.PublicationLeavesTheDeviceInitializedTailUntouched`
- `KernelArgsPublication.RepeatedPublicationPreservesTheDeviceInitializedTail`

Each seeds its sentinel **through the allocator**, so the mark predates the real copy — written afterwards it would paint over the evidence it exists to read — and asserts the allocation covers the tail *before* reading it, so a shrunk-allocation regression reports bounds instead of running past the block. The fake `copy_h2d` also clamps to the destination block and reports, so a control cannot turn into UB.

Two negative controls confirm they discriminate, both reverted byte-for-byte (md5-verified):

| control | arm | result |
| --- | --- | --- |
| copy widened back to the full extent | a2a3 trb (`runtime_device_copy_size`) | persistent + program cases fail on the tail-stop assertion |
| copy widened back to the full extent | a2a3 hbg (`Runtime::device_image_bytes`) | all four hbg cases fail on the tail-stop assertion |
| allocation shrunk to the uploaded prefix | `PersistentKernelArgs::prepare_once` | fails on the bounds assertion, no out-of-range read |

An earlier draft used `GTEST_SKIP` when copy == extent; the first control showed that made a regression **skip instead of fail**, so the expectation is now asserted from the descriptor type.

## Existing expectations migrated, not run

Three files encoded "the copy is the whole descriptor": `test_kernel_args_helper.cpp`'s rtMemcpy stub expected `sizeof(DeviceRuntimeLaunchDesc)` on every copy, and `test_hbg_bind_ledger.cpp` plus the `RuntimeLaunchImage` cases reconstructed a full descriptor from a byte vector that now holds only the prefix. Updated statically; **not executed here** — CI covers them.

## Verification performed

Coverage matrix — four targets built incrementally, only the new cases run by exact `--gtest_filter`:

| | a2a3 trb | a2a3 hbg | a5 trb | a5 hbg |
| persistent path | 2/2 pass | 2/2 pass | no target | no target |
| program path | 2/2 pass | 2/2 pass | no target | no target |

**8/8 pass.** a5 has no cpput target for either helper, so **a5 coverage here is compile-only** (layout probes on all four arch/runtime descriptors) — not runtime, and not claimed as such. a5 trb additionally has no gate tail at all, so its two extents coincide by construction.

No existing case was rerun, no suite, no `ctest`, no sim, no hardware. One earlier full clean/build/ctest attempt was started and stopped mid-compile before any test ran — it is not evidence of anything and is not counted.

Relates to #2254


ChaoWao:
Current-head static review (6984242cf): allocation/copy split and gate-only scope look consistent, but this PR is not ready yet.

1. GitHub reports `CONFLICTING` / `DIRTY`; there are no Actions runs for this head. Please rebase onto current main and resolve conflicts while preserving merged changes and this PR’s gate-only scope.
2. The layout explanation says the gates move into existing end padding. They are relocated 4,608/6,912-byte arrays, not padding occupants. Reordering preserves measured `sizeof`; the upload now excludes the relocated array. Please correct that explanation.
3. Coverage says both host paths on both runtimes, while the listed six executions cover three targets (persistent TMR/HBG, program TMR). Either validate only the two newly added cases in `test_hbg_kernel_args_helper`, or explicitly report that local coverage gap. Do not rerun existing cases/suites.
4. Scope common HBG comments about device gate initialization to A2/A3; A5’s array is unused reserved storage.

Please return the resolved head and actual CI run IDs; an empty Actions list must not be treated as green. No protocol/admission changes or hardware reruns are requested.

ChaoWao:
Rebased and all four points applied. Head **`d88c54d1f`**, base `eeedcaa4a`, one commit, `mergeable=MERGEABLE` — and Actions runs now exist at this head, so the conflict was indeed the blocker.

**1. Rebase.** Onto `eeedcaa4a` (was `31da0560e`; #2354/#2355/#2366/#2367/#2368/#2369 landed between). One conflict, in `test_kernel_args_helper.cpp` — append-vs-append: #2367 added three `runtime_args_published` cases at the file end, this branch added two. **Both sides kept**, all 13 cases present, no merged work reverted (three-dot diff is the same 17 files). The extent wiring in `prepare_runtime_args` and `prepare_once` survived intact.

Resolving it also surfaced a real defect: my first resolution consumed the upstream case's closing brace along with the conflict marker, which nested my two cases inside its body. It failed to compile with `local class ... shall not have static data member` — caught by building `test_hbg_kernel_args_helper`, which is the target point 3 asked for. Fixed and folded into the single commit.

**2. Relocation, not padding.** Corrected in both the commit message and the PR body. The arrays are relocated from their original position between `workers[]` and the host-published fields to the descriptor tail; at 4,608 / 6,912 bytes they are far larger than any alignment padding. The accurate statement is that *reordering leaves each measured `sizeof` unchanged* — so no allocation moves — *while the prefix excludes the relocated array*. Numbers re-measured on the rebased tree, unchanged. No latency claim.

**3. Coverage matrix, corrected.** The previous "both paths on both runtimes" was wrong: it was 6 runs over 3 targets, missing the hbg program helper. Now 8 runs over 4:

| | a2a3 trb | a2a3 hbg | a5 trb | a5 hbg |
| persistent path | 2/2 | 2/2 | no target | no target |
| program path | 2/2 | **2/2** | no target | no target |

a5 has no cpput target for either helper, so **a5 is compile-only coverage here** (layout probes on all four descriptors) — stated as such, not claimed as runtime. Negative controls re-run after the conflict edits, now on both arms:

| control | arm | result |
| --- | --- | --- |
| copy widened to the extent | a2a3 trb | tail-stop assertion fails |
| copy widened to the extent | a2a3 hbg | all four hbg cases fail |
| allocation shrunk to the prefix | `prepare_once` | bounds assertion fails, no out-of-range read |

All control edits reverted byte-for-byte, md5-verified.

**4. A5 scoping.** The shared hbg descriptor comment, the shared `runtime_device_copy_size` comment and the PR body now scope the `pre_handshake_init` zero + `wmb` + `hs_setup_done_` chain to **A2/A3** explicitly, and say a5 neither reads, writes nor initializes the gates. The a5 cache-maintenance comments no longer imply the gate range is device-written on a5 — they say the extent deliberately spans the whole allocated descriptor. Mechanism unchanged.

**Honest test accounting:** no existing case was rerun, no suite, no `ctest`, no sim, no hardware. An earlier full clean/build/ctest script was started and stopped mid-compile **before any test ran** — it is not evidence and is not counted anywhere. Host harness cases establish allocation/copy extents only; they say nothing about device cache visibility or future P4 ordering.

PR #2374, opened 2026-09-19
Refuse admission after a matched device-fault notice

ChaoWao:
## Summary

#2354 gave the driver's fault channel a consumer and a per-generation record. Nothing read
it: there was no path from a `DeviceFaultNotice` to `can_accept_run()`. **Admission now
reads it.**

A notice refuses future runs when *both* hold — its logical `device_id` is this runner's,
and its `stream_id` is one this runner's run boundaries were recorded on
(`run_stream_ids_.attribute(...) == Mine`). The refusal stands until a **confirmed** reset
retires the generation.

Node E1b of #2267. It is an availability policy and nothing else.

## What refuses, and what does not

| Evidence | Admission |
| --- | --- |
| notice matched to one of this runner's run streams | **refused** until a confirmed reset |
| matched, while the identity history is **incomplete** | **refused** — a missing id is what a stale notice carries, so incompleteness weakens the *negative* answer, not the positive one |
| notice naming another device | open |
| notice on a stream no run of this runner's used (`NotMine`) | open |
| `Undecided` | open |
| ring loss / reporter drop | open — it names no resource to refuse for |
| **no notice at all** | open — silence is not evidence of health |

## Wiring

`device_admits_new_run(arch_quarantined, health)` composes the fault suspicion with each
arch's own post-failure quarantine; `DeviceRunnerBase::accepts_new_run()` is what the shared
c_api asks, at `simpler_prepare_run` and at both `simpler_launch_run` checks.

The two refusals are independent and clear by different routes — the arch flag by its
finalize after a confirmed reset, the suspicion by the retirement that reset performs. **A
reset that was not confirmed leaves the refusal standing.**

The two *diagnostic* readers of `can_accept_run()` (the clock-correlation session's abandon
flag) keep asking the arch flag alone: a matched notice says nothing about whether this run's
own DFX resources are safe to release normally.

## What deliberately does not change

- **The run's own verdict.** `execution_rc` is settled before the consumer runs and is
  neither read nor written by it. A notice carries no run identity and may arrive late — the
  longest lag measured so far is ~16 s, which is a sample, not a bound the SDK promises — so
  it can name a fault from an earlier run than the one being finalized.
- **No drain, reset or recovery starts from a notice**, and nothing is reclaimed.
  Stop-admission is stop-admission.
- **The teardown consumer records only** — the admission it could refuse is already over, and
  `finalize_common_impl` is destroying the resources a drain would wait on.
- `sync_stream_pair`, the unfenced fallback, `poll_run_fence`'s sticky read, retirement and
  reset ownership: untouched.
- No stream split, no classifier, no consumer relocation, no a5 terminal producer, no E2/P4.

## The availability cost, stated

**The match is membership, not provenance.** It says a fault landed on a stream this runner's
runs use — not that a run caused it, and not that any run was impaired. On a5 that same
`stream_aicpu_` also carries binary load, AICPU init and callable registration. The notice
carries no timestamp, so **a late or recycled identity can refuse work the device would have
served**, at a rate nobody has measured.

**Recovery is conditional.** Admission comes back only through the existing
confirmed-generation-retirement path — the force reset that confirms, then
`retire_after_confirmed_device_reset`. A reset that was not confirmed, or a re-notification
delivered into the new generation, leaves the runner refusing; an ordinary close and re-init
is not a guarantee that admission is restored.

This has now been observed **twice on a5 CI, on device 4, with identical notice fields** —
`device_id=4 stream_id=61 task_id=82 error_code=507018` — in
`tests/st/a5/.../prepared_callable::test_run`, once at an earlier head and again on this PR's
run (`st-onboard-a5`, run `35441219825`). Both times: the consuming run returned success (its
`device_wall` markers are emitted, which is gated on `execution_rc == 0`), the notice was
matched to a recorded run stream, and the **next** `simpler_prepare_run` was refused with
`-1000` at the following generation. `1 failed, 47 passed`.

That is this policy doing exactly what it says. What it is **not** is evidence that the fault
is spurious — a repeat with identical ids on the same die is at least as consistent with a
real, repeatable fault on that sequence as with a stale identity. It is a red required check
and is reported as such, not waived.

## This does not license E2's synchronize removal

The rows that keep synchronizing include **the ordinary successful run, which produces no
notice at all**. A fallback that covers every ordinary run cannot also be the sync-free fast
path. E2 still owes an attributable positive terminal record from every producer (a5 HBG's
legacy path publishes none) and its promotion from the shadow validation it is today,
boundary/retirement proofs, and treatment of every fault path the synchronize currently
catches.

## Fixture isolation for the intentional-failure case

The consequence above reaches `tests/st/**/prepared_callable`, where a case deliberately
registers a missing AICPU orchestration entry and the class shares one `Worker` with cases
that expect a healthy runner. That case now lives in its own class,
`TestPreparedCallableRegistrationFailure`, on a2a3 and a5. Every assertion is unchanged; the
argument construction it shares with the positive cases moved to a plain mixin rather than
being copied; `host_build_graph` has no device-side registration and is untouched.

Two details are load-bearing and were both found by review rather than by me:

- **Class scope is the boundary, not a finer-scoped fixture.** `DevicePool.allocate` refuses
  rather than queues, so a per-test worker requested while the class's worker is still cached
  would need a second device and fail outright on the single-device default.
- **The new class keeps real `CASES`.** The root conftest deselects *every* item of a
  `SceneTestCase` class whose cases match no platform — an empty list would have silently
  dropped the moved test from CI rather than quieting the runner it inherits. The inherited
  `test_run` is quieted directly, by a shadow carrying `__test__ = False`.

Isolation confines shared mutable runner state between an intentional-failure case and the
positive cases around it. **It is not a claim that the device is restored or that a fault is
contained** — a notification the driver delivers later can still name that device, and only a
confirmed reset retires a generation.

### What was run

New suite `AdmissionUnderDeviceHealth` in `tests/ut/cpp/common/test_device_health_state.cpp`,
8 cases:

```bash
cmake --build tests/ut/cpp/build --target test_device_health_state
./tests/ut/cpp/build/test_device_health_state --gtest_filter='AdmissionUnderDeviceHealth.*'
```

**8/8 pass.** Three negative controls, each failing exactly its own cases and nothing else:

| Inversion of `device_admits_new_run` | Failures |
| --- | --- |
| ignore `health.suspect()` | the 2 matched-notice cases + the 2 reset cases (4) |
| also refuse on unattributed / undelivered totals | `AnUnattributedNoticeAdmits`, `LostAndDroppedNoticesAdmit` (2) |
| ignore `arch_quarantined` | `TheArchQuarantineRefusesOnItsOwnAndIsNotRetiredHere` (1) |

Two new CPU-only pyut files for the fixture work, run with explicit paths:

```bash
python -m pytest tests/ut/py/test_scene_case_collection_gate.py \
                 tests/ut/py/test_device_pool_fixture_lifetimes.py -q
```

**8/8 pass.** `test_scene_case_collection_gate.py` drives the **real** root-conftest hook and
a **real** pytest collection over a synthetic module — it asserts that a class whose cases
match no platform loses *every* item including hand-written ones (the defect an earlier draft
of this PR shipped), that `__test__ = False` stops the inherited runner being collected, and
that both real `TestPreparedCallableRegistrationFailure` classes use that shape.
`test_device_pool_fixture_lifetimes.py` pins the pool property the class boundary rests on.
Negative controls: dropping `__test__ = False` from the synthetic class fails only the
suppression case; restoring `CASES = []` on the real a5 class fails only the tie-back case;
a partial-allocating pool fails only the all-or-nothing case; a no-op `release` fails all four
pool cases.

Static hooks, at the repo's pinned versions, over **all 8 changed Python files** (the four
`prepared_callable` files included, which earlier rounds of this PR had not type- or
format-checked): `ruff check --fix` and `ruff format --check` (0.14.8) and `pyright` (1.1.407)
— all clean. Also `-fsyntax-only` on the four touched C++ TUs (`device_runner_base.cpp` and
`c_api_shared.cpp` × a2a3/a5, from the onboard compile DBs), `clang-format`,
`tests/lint/check_retired_names.py`, and `markdownlint-cli2` on the changed doc — all clean.

**Not run, and stated rather than implied:** the real `tests/st/**/prepared_callable` modules
were never collected locally. The build-stamp guard refuses to import `simpler.task_interface`
against a tree newer than the built extension, and rebuilding was outside this round. The
synthetic collection test plus the AST tie-back is what stands in for it; **CI's own
collection of those files is the check that they are selected.**

### An unauthorized rerun, corrected

An earlier revision of this section said only new cases were exercised. **That was wrong.**
After the negative controls I ran `./tests/ut/cpp/build/test_device_health_state` once with
no `--gtest_filter`, which executed all 40 cases in that binary — the 8 new ones **and 32
pre-existing ones**. Re-running existing cases was outside what had been authorized. Those 32
passing is not evidence of anything here and is not claimed as such. Test invocations from
here on are explicit new-case filters only.

### What the 8 cases actually cover, and what they do not

They exercise **`device_admits_new_run` only** — the predicate — driving the real
`DeviceHealthState`, `RunStreamIdentities`, `DeviceFaultMonitor` and
`retire_after_confirmed_device_reset`. They set health state through the production `note_*`
calls and inspect stream-id attribution separately; they do **not** execute
`DeviceRunnerBase::accepts_new_run()` and they do **not** execute the shared c_api's
prepare/launch admission checks. Two further limits worth naming:

- `AResetThatWasNotConfirmedLeavesAdmissionRefused` asserts that the refusal stands when **no
  retirement is performed**. It does not drive a reset that failed; the caller-side contract
  that only a confirmed reset reaches the retirement is not exercised by it.
- No case injects a `DeviceFaultNotice` through `consume_device_fault_notices()`. The
  notice → classification → record step is unchanged production code covered by the
  pre-existing cases in the same file.

### Evidence for the parts no test covers

The c_api wiring and rc preservation rest on **static source review at this head**, not on a
test and not on CI:

- the three admission sites now call `accepts_new_run()` (`c_api_shared.cpp:959`, `:1139`,
  `:1149`); the two diagnostic readers still call `can_accept_run()` (`:894`, `:1437`);
- `execution_rc` is assigned once, at `c_api_shared.cpp:1353`, before
  `consume_device_fault_notices()` at `:1378`, and is not reassigned after it;
  `consume_device_fault_notices()` returns a `uint64_t` count whose only consumer is a
  `LOG_WARN`;
- `copy_back_run_outputs_impl` at `:1389` is passed that same `execution_rc`.

This is reviewable but it is not execution. A focused harness — a seam that drives the
consumer with an injected notice and then asks the real admission path — could test it
without hardware; **that seam is not delivered here.** The ordinary onboard CI lanes exercise
this code but inject no notice, so a green run there is not targeted coverage of the policy
either.

Refs #2267


ChaoWao:
Static feedback for the in-progress fixture-isolation follow-up: avoid adding a function-scoped `isolated_st_worker` inside the existing class while keeping its positive tests on a class-scoped `st_worker`.

The earlier positive method instantiates the class-scoped fixture, which retains its device until the class finishes. The proposed negative method then calls `device_pool.allocate(1)` again. With a single device, `DevicePool.allocate` (`conftest.py:104-109`) returns `[]`, and `_l2_worker_for` fails with “no devices available”. This requires two simultaneous devices for a previously single-device test; changing which fixture the negative method requests does not finalize the cached class fixture.

Please use non-overlapping fixture lifetimes (for example, a dedicated negative-test class under the existing class-scoped fixture), preserve all existing assertions, and verify the decorator/collection path does not duplicate or drop tests. Do not solve this by changing test-name ordering, increasing the requested device count, skipping, or aliasing both Workers to the same live device. A new focused fake-pool lifecycle regression can demonstrate acquire → release → acquire with a capacity of one; no existing test rerun is authorized. This is a source/fixture-lifetime finding, not an executed test result.

ChaoWao:
Static blocker in the current dedicated-class draft: `TestPreparedCallableRegistrationFailure.CASES = []` silently deselects the negative test, not merely inherited `test_run`.

Root conftest.py:650–663 applies `if not any(platform ... for case in cls.CASES): deselected.append(item)` to EVERY item of a SceneTestCase class. Therefore an empty CASES list removes `test_prewarm_failure_rolls_back_registration` from ordinary platform CI. A Python AST inventory of method names cannot detect this. Keep the existing negative case actually selected under both matching platform modes.

Use a structure that retains platform-selection metadata and avoids introducing an unrelated inherited positive run into the poisoned worker. For example, an appropriate dedicated class can retain platform CASES metadata and explicitly suppress only the NEW inherited generic `test_run` on that new class; the existing positive class/run must remain unchanged. Verify against actual collection hooks rather than just counting methods. A new focused CPU regression using collection logic is authorized; do not run existing device cases, skip/drop the original negative case, or weaken the policy. Current draft is not ready for publication until this is handled.

ChaoWao:
@ChaoWao Both findings applied at `3f5d66430`.

**Fixture lifetime (12:32).** Correct and my approach was unusable: `DevicePool.allocate` returns `[]` rather than queueing (root `conftest.py:104-109`), and the class-scoped `st_worker` stays cached for the class's whole life, so a function-scoped worker would have needed a second device and failed outright on `--device 0`. The function-scoped fixture is withdrawn — all four `prepared_callable` conftests are back to a single `st_worker`, plus a comment recording why class scope is the boundary. The negative case moved to its own class instead.

**Collection gate (13:01).** Also correct, and it had already shipped in my draft: with `CASES = []` the root conftest's platform gate (`conftest.py:650-663`) deselects **every** item of the class, so `test_prewarm_failure_rolls_back_registration` would have stopped running in CI entirely — invisible to the AST inventory I was relying on. Fixed by keeping real platform metadata (`CASES = TestPreparedCallable.CASES`) and suppressing only the newly inherited runner:

```python
def test_run(self, *args, **kwargs):
    raise AssertionError("TestPreparedCallableRegistrationFailure runs no scene case")

test_run.__test__ = False
```

`__test__ = False` rather than a non-callable shadow: `_pytest/python.py:243` skips collection silently for the former and raises `PytestCollectionWarning` for the latter. No skip/xfail, no reordering, no weakening of the positive class or of the moved assertions.

**Verified against the collection path, not method names.** New `tests/ut/py/test_scene_case_collection_gate.py` imports the real hook by path into a synthetic rootdir and runs a real `pytest --collect-only --platform a5`, asserting: a matching class keeps its hand-written method; a `CASES = []` class loses *all* of them (the defect, now pinned); `__test__ = False` removes only the runner; and both real registration-failure classes carry that shape. Negative controls: removing `__test__ = False` from the synthetic class fails only the suppression case, and restoring `CASES = []` on the real a5 class fails only the tie-back case.

**Limit, stated:** the real ST modules were not collected locally — the build-stamp guard blocks importing `simpler.task_interface` against a newer tree and a rebuild was outside this round. CI's collection is the check.

Also in this push: a bad squash was caught and repaired before publication. I had soft-reset onto a moved `upstream/main`, which recorded reverts of #2373 and others as mine; the branch was rebased properly instead, and the three-dot diff is now 14 files, all this change.

